PR #9, opened 2020-05-04
Add Yahoo Finance adapter

boxhock:


yahoo-finance/adapter.js
```diff
@@ -0,0 +1,46 @@
...
+    symbol = commonKeys[symbol]
+  }
+
+  yahooFinance.quote({
+    symbol: symbol,
+    modules: ['price']
+  }, (err, body) => {
+    if (err) {
+      console.log(err)
+      callback(500, Requester.errored(jobRunID, undefined))
```
thodges-gh, line 33: Send in the `err` object here instead of `undefined`.

PR #23, opened 2020-05-14
Create helpers

thodges-gh:
I'm not entirely sure if `make new adapter=someadapter` will work the same on Mac since `sed` has a slightly different implementation.


boxhock:
You are right that it doesn't work on macos:

```bash
% make new adapter=myadapter
mkdir myadapter
cp -r example/* myadapter
sed -i 's/example/myadapter/' myadapter/package.json
sed: 1: "myadapter/package.json": invalid command code m
make: *** [new] Error 1
```

thodges-gh:
Added instructions for MacOS.
womm

boxhock:
Can confirm it now works on macos using gnu-sed with the path replaced for sed.

PR #24, opened 2020-05-14
All External Adapters should register as a service with Consul

j16r:
Spike to experiment with ExternalAdapters registering themselves as consul services for easier discovery by PFM.


j16r:
This spike has proven useful - it seems using consul from within a lambda function doesn't really make sense because of the execution model.

PR #39, opened 2020-06-19
Add Binance DEX adapter

boxhock:


binance-dex/adapter.js
```diff
@@ -0,0 +1,58 @@
+const { Requester, Validator } = require('@chainlink/external-adapter')
+const apiEndpoint = process.env.API_ENDPOINT || 'dex-asiapacific'
```
krebernisak, line 2: It would be more readable if the default options were specified at the start:

```suggestion
const DEFAULT_API_ENDPOINT = 'dex-asiapacific'
const DEFAULT_DATA_ENDPOINT = 'v1/ticker/24hr'
const apiEndpoint = process.env.API_ENDPOINT || DEFAULT_API_ENDPOINT
```

binance-dex/adapter.js
```diff
@@ -0,0 +1,58 @@
...
+
+const customError = (data) => {
```
krebernisak, line 4:
```suggestion
const isCustomError = (data) => {
```

binance-dex/adapter.js
```diff
@@ -0,0 +1,58 @@
...
+const customParams = {
+  base: ['base', 'from', 'coin'],
+  quote: ['quote', 'to', 'market'],
+  endpoint: false
+}
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || 'v1/ticker/24hr'
```
krebernisak, line 18:
```suggestion
  const endpoint = validator.validated.data.endpoint || DEFAULT_DATA_ENDPOINT
```

binance-dex/adapter.js
```diff
@@ -0,0 +1,58 @@
...
+        return callback(500, Requester.errored(jobRunID, 'data is too old'))
+      }
+
+      response.data.result = Requester.validateResultNumber(response.data, ['lastPrice'])
+      callback(response.status, Requester.success(jobRunID, response))
+    })
+    .catch(error => {
+      callback(500, Requester.errored(jobRunID, error))
+    })
+}
```
krebernisak, line 56: I would prefer to separate the functions from the Promise execution to make it more readable:

```suggestion

  const _handleResponse = response => {
    if (response.data.length < 1) {
      return callback(500, Requester.errored(jobRunID, 'no result for query'))
    }

    // Replace array by the first object in array
    // to avoid unexpected behavior when returning arrays.
    response.data = response.data[0]

    const lastUpdate = response.data.closeTime
    const curTime = new Date()
    // If data is older than 10 minutes, discard it
    if (lastUpdate < curTime.setMinutes(curTime.getMinutes() - 10)) {
      return callback(500, Requester.errored(jobRunID, 'data is too old'))
    }

    response.data.result = Requester.validateResultNumber(response.data, ['lastPrice'])
    callback(response.status, Requester.success(jobRunID, response))
  }
    
  const _handleError = error => callback(500, Requester.errored(jobRunID, error))

  const params = {
    symbol
  }

  const config = {
    url,
    params
  }

  Requester.request(config, isCustomError)
    .then(_handleResponse)
    .catch(_handleError)
}
```

binance-dex/test/adapter_test.js
```diff
@@ -0,0 +1,122 @@
...
+      {
+        name: 'unknown quote',
+        testData: {
+          id: jobID,
+          data: {
+            base: 'ETH',
+            quote: 'not_real'
+          }
+        }
+      }
```
krebernisak, line 107: Let's add a test for dummy data endpoint:

```suggestion
      },
      {
        name: 'unknown dummy endpoint,
        testData: {
          id: jobID,
          data: {
            coin: 'BNB',
            market: 'BUSD-BD1',
            endpoint: 'dummy'
          }
        }
      }
```

binance-dex/adapter.js
boxhock: This differs a lot from our usual implementations, but it avoids the "nesting hell" we get by nesting all the functions inside each other. This works for me 👍

binance-dex/adapter.js
boxhock: I think we can get rid of this function all together actually, since we have no good point of checking for errors with this.

PR #41, opened 2020-06-22
Add Nikkei adapter

boxhock:


nikkei/README.md
```diff
@@ -0,0 +1,33 @@
...
+        "opentime":"(09:00)",
+        "high_price":"22,575.74",
+        "hightime":"(12:33)",
+        "low_price":"22,311.94",
+        "lowtime":"(09:02)",
+        "divisor":"27.760",
+        "divisor_date":"(Jun/23/2020)",
+        "result":22437.27
+    },
+    "result":22437.27,
```
krebernisak, line 30: I could not figure out where this `result`(outside `response.data`) is set?

nikkei/test/adapter_test.js
```diff
@@ -0,0 +1,49 @@
...
+    ]
+
+    requests.forEach(req => {
+      it(`${req.name}`, (done) => {
+        createRequest(req.testData, (statusCode, data) => {
+          assert.equal(statusCode, 200)
+          assert.equal(data.jobRunID, jobID)
+          assert.isNotEmpty(data.data)
+          assert.isAbove(data.result, 0)
+          assert.isAbove(data.data.result, 0)
```
krebernisak, line 20:
```suggestion
          assert.isAbove(data.data.result, 0)
          assert.equal(data.result, data.data.result)
```

nikkei/adapter.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || 'get_real_data'
+  const url = `https://indexes.nikkei.co.jp/en/nkave/${endpoint}`
+  let idx = validator.validated.data.base.toUpperCase()
+
+  if (idx in commonKeys) {
```
krebernisak, line 19: What happens if key is not found? For example, if `base: nk225` is sent, will the result be the same as if `base: n225` is sent as a parameter? 

Former case will send `param.idx: NK225`, and the latter `param.idx: nk225`. I am not sure if API is case sensitive.

nikkei/README.md
boxhock: This happens in `Requester.success(jobRunID, response)`

nikkei/adapter.js
boxhock: If key is not found it should just use the input.

The common keys is there so that we can standardize the job spec (Use "N225" on all job specs instead of "nk225" on some and "n225" on others).

PR #44, opened 2020-06-24
Add dxFeed adapter

boxhock:


dxfeed/adapter.js
```diff
@@ -0,0 +1,52 @@
+const { Requester, Validator } = require('@chainlink/external-adapter')
+const username = process.env.API_USERNAME
+const password = process.env.API_PASSWORD
+const DEMO_ENDPOINT = 'https://tools.dxfeed.com/webservice/rest'
```
krebernisak, line 4: Any special reason why you named it `DEMO_ENDPOINT`?

```suggestion
const DEFAULT_ENDPOINT = 'https://tools.dxfeed.com/webservice/rest'
```

dxfeed/adapter.js
```diff
@@ -0,0 +1,52 @@
...
+const apiEndpoint = process.env.API_ENDPOINT || DEMO_ENDPOINT
```
krebernisak, line 5:
```suggestion
const apiEndpoint = process.env.API_ENDPOINT || DEFAULT_ENDPOINT
```

dxfeed/adapter.js
```diff
@@ -0,0 +1,52 @@
...
+
+const customError = (data) => {
+  return (data.status !== 'OK')
+}
```
krebernisak, line 9: Can be simplified:

```suggestion
const customError = (data) => (data.status !== 'OK')
```

dxfeed/adapter.js
```diff
@@ -0,0 +1,52 @@
...
+
+const commonSymbols = {
+  N225: 'N225:JP',
+  FTSE: 'UKX:FTSE'
+}
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || 'events.json'
```
krebernisak, line 24: Maybe extract `'events.json'` as `DEFAULT_DATA_ENDPOINT` constant.

dxfeed/adapter.js
boxhock: Because it's actually just a demo endpoint, and I don't think we should trust any data that comes from that endpoint as far as I understand.

The node operators should be provided an endpoint along with credentials when they get set up with an account with dxFeed.

dxfeed/adapter.js
boxhock: good point

dxfeed/adapter.js
```diff
@@ -0,0 +1,51 @@
+const { Requester, Validator } = require('@chainlink/external-adapter')
+const username = process.env.API_USERNAME
+const password = process.env.API_PASSWORD
+const DEFAULT_DATA_ENDPOINT = 'events.json'
+const DEMO_ENDPOINT = 'https://tools.dxfeed.com/webservice/rest'
+const apiEndpoint = process.env.API_ENDPOINT || DEMO_ENDPOINT
```
krebernisak, line 6: How about we add a warning for operators?

```suggestion
const apiEndpoint = process.env.API_ENDPOINT || DEMO_ENDPOINT
if (apiEndpoint === DEMO_ENDPOINT) console.warn(`Using demo endpoint: ${DEMO_ENDPOINT} (Please do not use in production!)`
```

PR #45, opened 2020-06-24
Add Trading Economics adapter

boxhock:


tradingeconomics/adapter.js
```diff
@@ -0,0 +1,99 @@
...
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  let symbol = validator.validated.data.base.toUpperCase()
+  if (symbol in commonSymbols) {
+    symbol = commonSymbols[symbol]
+  }
+
+  const client = new TeClient({
```
krebernisak, line 25: We probably do not need to recreate this client on every request, so we can move it outside of `createRequest` function.

tradingeconomics/adapter.js
```diff
@@ -0,0 +1,99 @@
...
+    if (tries++ > maxTries) {
+      client.ws.close()
+      _getDataFromUrl()
+      return
+    }
+
+    setTimeout(_checkTimeout, timeout)
+  }
+
+  setTimeout(_checkTimeout, timeout)
```
krebernisak, line 96: We should really consider using RxJS for dealing with WebSockets and streams of any kind. This retry logic would be more elegant for example using [retryWhen](https://rxjs-dev.firebaseapp.com/api/operators/retryWhen) operator.

Here is some random example:

```javascript
loadSomething(): Observable<SomeInterface> {
  return this.http.get(this.someEndpoint, commonHttpHeaders())
    .retryWhen(errors => 
      // Time shift the retry
      errors.delay(500)
            // Only take 10 items
            .take(10)
            // Throw an exception to signal that the error needs to be propagated
            .concat(Rx.Observable.throw(new Error('Retry limit exceeded!'))
    );
```

tradingeconomics/adapter.js
```diff
@@ -0,0 +1,99 @@
...
+      c: `${API_CLIENT_KEY}:${API_CLIENT_SECRET}`
+    }
+
+    const config = {
+      url,
+      params
+    }
+
+    const _handleResponse = response => {
+      if (response.data.length < 1) {
```
krebernisak, line 60:
```suggestion
      if (!response.data || response.data.length < 1) {
```

tradingeconomics/adapter.js
thodges-gh: Since external adapters are commonly ran on serverless frameworks, we have to assume that we don't retain state between requests.

krebernisak: I do not think there should be a problem because of serverless deployment. The server is either active between function calls, which means it retains state. Or it's teared-down and built back up (cold run) on next function call - that means the state is re-initialized before the function is run.

But this is nothing critical so let's leave it for now.

PR #47, opened 2020-07-09
Add adapter for BNC 24 hour VWAP

thodges-gh:


bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,96 @@
...
+      },
+      data: {
+        audience: 'https://api.bravenewcoin.com',
+        client_id: process.env.CLIENT_ID,
+        grant_type: 'client_credentials'
+      }
+    }).then(response => resolve(response.data.access_token))
+      .catch(error => reject(error))
+  })
+}
```
krebernisak, line 31: Using Promises directly becomes complicated fairly quickly, that's why `async/await` was introduced and IMHO it makes code a lot easier to reason about.

We could rewrite this snippet using `async/await`:
```suggestion
const authenticate = async () => {
  const response = await Requester.request({
    method: 'POST',
    url: `https://${host}/oauth/token`,
    headers: {
      'content-type': 'application/json',
      'x-rapidapi-host': host,
      'x-rapidapi-key': process.env.API_KEY,
      accept: 'application/json',
      useQueryString: true
    },
    data: {
      audience: 'https://api.bravenewcoin.com',
      client_id: process.env.CLIENT_ID,
      grant_type: 'client_credentials'
    }
  })
  return response.data.access_token
}
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,96 @@
...
+        useQueryString: true
+      },
+      params: {
+        status: 'ACTIVE',
+        symbol
+      }
+    }).then(response => resolve(response.data.content[0].id))
+      .catch(error => reject(error))
+  })
+}
```
krebernisak, line 50:
```suggestion
const getAssetId = async (token, symbol) => {
  const response = await Requester.request({
    url: `https://${host}/asset`,
    headers: {
      'content-type': 'application/octet-stream',
      'x-rapidapi-host': host,
      'x-rapidapi-key': process.env.API_KEY,
      useQueryString: true
    },
    params: {
      status: 'ACTIVE',
      symbol
    }
  })
  return response.data.content[0].id
}
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,96 @@
...
+
+const createRequest = (input, callback) => {
```
krebernisak, line 52: As our `@chainlink/ea-bootstrap` calls the `createRequest` function directly and expects a callback, to bridge the API, we could do something like the following.

Create an async `_createRequest` function that has all the logic, simplified using `async/await`:

```javascript
const _createRequest = async (input) => {
  ...
  return {
    status: response.status, 
    data: Requester.success(jobRunID, response)
  }
}
```

And then the `createRequest` would just bridge the async/callback API:

```javascript
const createRequest = (input, callback) => {
  _createRequest(input)
    .then(result => callback(result.status, result.data))
    .catch(error => callback(500, Requester.errored(jobRunID, error)))
}

File: bravenewcoin-vwap/adapter.js [hunk 4]
```diff
@@ -0,0 +1,95 @@
...
+    data: {
+      audience: 'https://api.bravenewcoin.com',
+      client_id: process.env.CLIENT_ID,
+      grant_type: 'client_credentials'
+    }
+  })
+  return response.data.access_token
+}
+
+const getAssetId = async (token, symbol) => {
```
krebernisak, line 31: Token param is not used and can be removed from function input.

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+  const response = await Requester.request({
+    method: 'POST',
+    url: `https://${host}/oauth/token`,
+    headers: {
+      'content-type': 'application/json',
+      'x-rapidapi-host': host,
+      'x-rapidapi-key': process.env.API_KEY,
+      accept: 'application/json',
+      useQueryString: true
+    },
```
krebernisak, line 21: Let's reuse these repeated API headers. We can define an `apiHeaders` top-level constant:

```javascript
const apiHeaders = {
  'x-rapidapi-host': host,
  'x-rapidapi-key': process.env.API_KEY
}
```

And then add them here:

```suggestion
    headers: {
      'content-type': 'application/json',
      accept: 'application/json',
      useQueryString: true,
      ...apiHeaders
    },
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+
+const getAssetId = async (token, symbol) => {
+  const response = await Requester.request({
+    url: `https://${host}/asset`,
+    headers: {
+      'content-type': 'application/octet-stream',
+      'x-rapidapi-host': host,
+      'x-rapidapi-key': process.env.API_KEY,
+      useQueryString: true
+    },
```
krebernisak, line 39: We can also reuse the same API headers here:
```suggestion
    headers: {
      'content-type': 'application/octet-stream',
      useQueryString: true,
      ...apiHeaders
    },
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  const yesterday = new Date()
+  yesterday.setDate(yesterday.getDate() - 1)
+
+  _createRequest({
+    jobRunID,
```
krebernisak, line 55: We can remove `jobRunID` from the input, let the `createRequest` function handle the Chainlink API part, and `_createRequest` would just execute the request and return the response.
```suggestion
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+    })
+      .then(response => {
+        response.data.result = Requester.validateResultNumber(response.data, ['content', 0, 'vwap'])
+        resolve({
+          status: response.status,
+          data: Requester.success(input.jobRunID, response)
+        })
+      })
+      .catch(error => reject(error))
+  })
```
krebernisak, line 92: We do not need to return an explicit promise here, an async function is an implicit promise, so we can just use it as such.

Let's make this function just execute the request and return a response:

```suggestion
  const assetId = await getAssetId(input.symbol)
  return await Requester.request({
    url: input.url,
    headers: {
      authorization: `Bearer ${token}`,
      useQueryString: true,
      ...apiHeaders
    },
    params: {
      indexId: assetId,
      indexType: input.indexType,
      timestamp: input.timestamp,
      size: 1
    }
  })
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+  return response.data.content[0].id
+}
+
+const createRequest = (input, callback) => {
+  const validator = new Validator(callback, input, customParams)
+  const jobRunID = validator.validated.id
+  const yesterday = new Date()
+  yesterday.setDate(yesterday.getDate() - 1)
+
```
krebernisak, line 53: Let's make this function handle all the API details of talking to a Chainlink node. We can define two small functions to handle `_createRequest` result:

```javascript
  const _onResponse = (resp) => {
    const path = ['content', 0, 'vwap']
    resp.data.result = Requester.validateResultNumber(resp.data, path)
    callback(resp.status, Requester.success(jobRunID, resp))
  }

  const _onError = (error) =>
    callback(500, Requester.errored(jobRunID, error))
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,95 @@
...
+  _createRequest({
+    jobRunID,
+    url: `https://${host}/ohlcv`,
+    symbol: validator.validated.data.symbol,
+    indexType: 'GWA',
+    timestamp: yesterday
+  })
+    .then(result => callback(result.status, result.data))
+    .catch(error => callback(500, Requester.errored(jobRunID, error)))
```
krebernisak, line 62: And just use the earlier defined handlers here:

```suggestion
    .then(_onResponse)
    .catch(_onError)
```

bravenewcoin-vwap/adapter.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const _createRequest = async (input) => {
+  const token = await authenticate()
+  const assetId = await getAssetId(input.symbol)
+  return await Requester.request({
+    url: input.url,
+    headers: {
+      'x-rapidapi-host': host,
+      'x-rapidapi-key': process.env.API_KEY,
```
krebernisak, line 81: You can also reuse the API headers here:
```suggestion
     ... apiHeaders,
```

thodges-gh: Missed that, thanks. Made my own commit so that it doesn't have that space.

PR #54, opened 2020-07-29
Expose port 8080 and allow custom docker tags

saschagoebel:
Exposing post 8080 by defaults makes it transparent which port the server listens on and also allows way simpler configuration with a Traefik proxy as LoadBalancer.

Custom docker tag allows to push directly to own repository


saschagoebel:
@thodges-gh Updated from latest master

saschagoebel:
@thodges-gh @tyrion70 Updated for latest master. Our upcoming offshift adapter will also have these changes, so it'd be nice if we could get these into master :-D

krebernisak (approved):
Thanks, @saschagoebel!

PR #87, opened 2020-09-18
[Feature] Uniswap Offshift (XFT) Price feed

jondoe1337:
As discussed with Johann (CL) and Freddy (Anyblock) - we built the adapter for retrieving Uniswap events from our anyblock database and generate the VWAP for a given time window.

Please note that this PR depends on https://github.com/smartcontractkit/external-adapters-js/pull/54


anyblock-offshift/adapter.js
```diff
@@ -0,0 +1,139 @@
...
+  } else if (start === end) {
+    start = end - 60 * 60 * 24
+  }
+
+  const body = {
+    query: {
+      bool: {
+        filter: [
+          { term: { 'address.raw': '0x2B9e92A5B6e69Db9fEdC47a4C656C9395e8a26d2' } },
+          { term: { 'event.raw': 'Sync' } },
```
krebernisak, line 104: It would be awesome if we could make this more flexible, so we can get VWAP from other AMM protocols like [Balancer](https://github.com/balancer-labs/balancer-core/blob/f4ed5d65362a8d6cec21662fb6eae233b0babc1f/contracts/BPool.sol#L28) for example.

To do this we would need to parameterize:

- network
- address
- event + params

Please share your thoughts on this.

krebernisak (changes requested):

anyblock-offshift/adapter.js
```diff
@@ -0,0 +1,139 @@
...
+
+  try {
+    end = parseInt(end)
+    if (roundDay) {
+      const date = new Date(end)
+      date.setUTCHours(0, 0, 0, 0)
+      end = date.getTime()
+    }
+  } catch (err) {
+  }
```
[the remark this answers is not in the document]
jondoe1337: Makes sense. :)

anyblock-offshift/adapter.js
jondoe1337: Mhhh, good point. It looks like we need some small abstraction of our API - what about an own module for this?

I don't even think that the parameters must be parameterized. Event, address and network should be sufficient. 
The caller of the module should/will know which parameters to be expected.
I'd say a timestamp window would be needed as well.

The current result is a list of events like this:
```
                    [
                        {
                            "value.hex": "0x1b287a18a87a333775bc",
                            "pos": 0,
                            "name": "reserve0",
                            "value.type": "uint112"
                        },
                        {
                            "value.hex": "0x16ad6bac17b7bad757",
                            "pos": 1,
                            "name": "reserve1",
                            "value.type": "uint112"
                        }
                    ]
```

krebernisak:
Thanks, @jondoe1337 for this!

I'll take over just to make it quicker and include it in the new release.
I'll just parameterize the address of the pool so this essentially becomes `anyblock-uniswap-vwap` adapter that can be used for Offshift (XFT) or any other asset available on Uniswap.

jondoe1337:
Great @krebernisak - sounds awesome! Thanks for that we appreciate the adapter.

krebernisak:
@jondoe1337 take a look here: https://github.com/smartcontractkit/external-adapters-js/pull/116

If you could contribute some Anyblock TypeScript types to it, so the code is more safe and understandable (documented) that would be great!

saschagoebel:
> If you could contribute some Anyblock TypeScript types to it, so the code is more safe and understandable (documented) that would be great!

Hi @krebernisak,

as we're mostly interacting with ElasticSearch directly, the result structure is highly dependent on what's requested. That being said, there are (generic) TS definitions available for the official ElasticSearch client library: https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/typescript.html
Using these would require to swap the current axios calls with the ElasticSearch lib and probably add some more dependencies.

Or maybe I'm misunderstanding the question altogether 🤷  We're both not very experienced in using TypeScript.

Cheers,
Sascha

PR #93, opened 2020-09-28
Add request coalescing support

krebernisak:


bootstrap/index.js
```diff
@@ -17,11 +17,19 @@ const withStatusCode = (execute) => (data, callback) => {
 const cacheOptions = envOptions()
 if (cacheOptions.enabled) logger.info('Cache enabled: ', cacheOptions)
 
+// Execution helper async => sync
+const _executeSync = (execute) => {
```
HenryNguyen5, line 21: This function prevents us from easily adding multiple middlewares in the future by dropping the resolved value, I would recommend instead reworking the `initHandlerXXX` functions and `withStatusCode` so that they are all async.

bootstrap/lib/cache.js
```diff
@@ -14,6 +15,14 @@ const parseBool = (value) => {
   return (_val === 'true' || _val === 'false') && _val === 'true'
 }
 
+// We generate an UUID per instance
+const uuid = () => {
```
HenryNguyen5, line 19: This is already used in the logging module, we should extract this function out and share the implementation

bootstrap/lib/cache.js
```diff
@@ -23,6 +32,10 @@ const envOptions = () => ({
     ...DEFAULT_IGNORED_KEYS,
     ...(process.env.CACHE_IGNORED_KEYS || '').split(',').filter((k) => k), // no empty keys
   ],
+  // Request coalescing
+  requestCoalescingEnabled:
```
HenryNguyen5, line 36: Would `remoteCaching` be a more simple terminology? `requestCoalescingEnabled` does not really allude distributed caching support

bootstrap/lib/cache.js
```diff
@@ -37,31 +50,58 @@ const withCache = (execute, options) => {
   // Algorithm we use to derive entry key
   const hashOptions = {
     algorithm: 'sha1',
+
     encoding: 'hex',
     excludeKeys: (key) => options.ignoredKeys.includes(key),
   }
+  const _getKey = (data) => hash(data, hashOptions)
+  const _getCoalescingKey = (key) => hash(`${options.requestCoalescingGroup}--${key}`, hashOptions)
```
HenryNguyen5, line 58: hashOptions seems like its not being used here, but i would question our usage of hashing the key here at all since its a simple string. Redis would handle this fine by itself and result in more debuggable logging if we ever have to look into Redis

bootstrap/lib/cache.js
```diff
@@ -37,31 +50,58 @@ const withCache = (execute, options) => {
...
+    if (retries === 0) return null
+    const entry = cache.get(key)
+    if (entry) {
+      logger.debug(`Request coalescing: Success`)
+      return entry
+    }
+    const inFlight = cache.get(coalescingKey)
+    if (!inFlight) return null
+    // Capped linear back-off: 100, 200, 400, 800, 1000..
+    ms = Math.min(ms * 2, 1000)
```
HenryNguyen5, line 76: Any reason for a retry? Imo retrying could result in delayed answers when we can just fallback to making a real request to the API instead.

HenryNguyen5 (changes requested):

bootstrap/lib/cache.js
```diff
@@ -37,31 +50,58 @@ const withCache = (execute, options) => {
...
-    const key = hash(data, hashOptions)
+  const _getWithCoalescing = async (key, coalescingKey, retries = 5, ms = 100) => {
+    if (retries === 0) return null
+    const entry = cache.get(key)
+    if (entry) {
+      logger.debug(`Request coalescing: Success`)
+      return entry
+    }
+    const inFlight = cache.get(coalescingKey)
+    if (!inFlight) return null
```
HenryNguyen5, line 74: Is there a possible scenario where the key won't be set before multiple other instances in a burst request try to access the coalescing key? This can happen if the backing cache mechanism is not strongly consistent AFAIK

HenryNguyen5 (changes requested):

bootstrap/index.js
krebernisak: Yes, we definitely need to rework `initHandlerXXX` functions and using async handlers from the start. Unfortunately, this is out of scope for this feature and will require a lot of changes through the repo. To do that we first need to bring in some changes from the private repo that are currently waiting.

bootstrap/lib/cache.js
krebernisak: As per our offline discussion, request coalescing is a separate enhancement from the remote caching support.

bootstrap/lib/cache.js
krebernisak: There are some extra bytes of memory that we are going to use if we go with the string key directly, but that is negligible for our use case. This is more to keep it consistent as the data key is also a hash.

I am going to leave this question open for now.

bootstrap/lib/cache.js
krebernisak: As per our offline discussion, request coalescing is a separate enhancement from the remote caching support.

bootstrap/lib/cache.js
krebernisak: Thanks for a great comment! Was thinking of adding a random delay here, just to scramble the timing of the request to avoid issues like this.

HenryNguyen5: That's definitely an option, another option would be to have the cache be local, and have the remote cache be used only for storing the computation results.

bootstrap/lib/cache.js
```diff
@@ -37,31 +50,58 @@ const withCache = (execute, options) => {
...
     const maxAge = _getMaxAge(data)
     // Add successful result to cache on callback
     const _cacheAndCallback = (statusCode, data) => {
       if (statusCode === 200) {
         const entry = { statusCode, data, maxAge }
         cache.set(key, entry, maxAge)
         logger.debug(`Cache set: ${key} => `, entry)
+
+        // Trigger request coalescing by removing in-flight mark
+        cache.del(coalescingKey)
```
HenryNguyen5: I would make sure the coalescingKey has a TTL so resource starvation doesn't occur from the following scenario:
- Process A grabs lock and then crashes
- Processes B, C, D, .... etc all have to wait the timeout length of the linear retry, until it finally returns null anyway

HenryNguyen5 (changes requested):

bootstrap/lib/util.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!process.env.UUID) process.env.UUID = uuidv4()
+  return process.env.UUID
+}
+
+const toAsync = (execute, data) =>
+  new Promise((resolve) => execute(data, (statusCode, data) => resolve({ statusCode, data })))
+
+const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+
+const interval = (retryCount = 1, min = 100, max = 1000, coefficient = 2) => {
```
HenryNguyen5, line 20: I think the following change would make this more clear to read, and is less computationally expensive
```ts
/**
 * Return a value used for exponential backoff in milliseconds.
 * @example
 * exponentialBackoffMs(1,100,1000,2) === 100
 * exponentialBackoffMs(2,100,1000,2) === 200
 * exponentialBackoffMs(3,100,1000,2) === 400
 *
 * @param retryCount The amount of retries that have passed
 * @param min The interval in ms
 * @param max The maximum backoff in ms
 * @param coefficient The base multiplier
 */
function exponentialBackoffMs(retryCount = 1, min = 100, max = 1000, coefficient = 2) {
  return Math.min(max, min * coefficient ** (retryCount - 1))
}
```

HenryNguyen5 (changes requested):

bootstrap/lib/util.js
krebernisak: Love it👌 Thanks!

bootstrap/lib/cache.js
krebernisak: The local cache for in-flight marks would suffer from the same issue as we have now because of AWS/GCP Lambda concurrency model. They basically start new micro containers for every concurrent request, so the local cache is empty most of the time.

krebernisak: Added random delay in [1bcc985](https://github.com/smartcontractkit/external-adapters-js/pull/93/commits/1bcc98500e3963cb2858fc6cd7e77c8884d701e5).

bootstrap/lib/cache.js
```diff
@@ -26,6 +33,8 @@ const envOptions = () => ({
     intervalMin: process.env.REQUEST_COALESCING_INTERVAL_MIN || 100,
     intervalMax: process.env.REQUEST_COALESCING_INTERVAL_MAX || 1000,
     intervalCoefficient: process.env.REQUEST_COALESCING_INTERVAL_COEFFICIENT || 2,
+    // Add entropy to absorb bursts
+    entropyMax: process.env.REQUEST_COALESCING_ENTROPY_MAX || 0,
```
HenryNguyen5: I would definitely add example configurations for the new values added

HenryNguyen5 (changes requested):

bootstrap/lib/cache.js
krebernisak: I've added some docs in [c4a3db4](https://github.com/smartcontractkit/external-adapters-js/pull/93/commits/c4a3db4c2ff3eae5c1533648ee4fd810648a1088). Would that be enough or did you have something else in mind?

HenryNguyen5: No these look great, thanks!

bootstrap/lib/cache.js
krebernisak: @HenryNguyen5 improved this in [586e93](https://github.com/smartcontractkit/external-adapters-js/pull/94/commits/586e9377b7069d59ebd2fdacac2fc2f9c50b137a).

HenryNguyen5 (dismissed):

boxhock (approved):
🙌

thodges-gh (approved):

PR #94, opened 2020-09-30
Add Redis support

krebernisak:


bootstrap/index.js
```diff
@@ -17,20 +17,32 @@ const withStatusCode = (execute) => (data, callback) => {
...
+const withMiddleware = async (execute) => {
+  return await withCache(withStatusCode(execute))
+}
+
 // Execution helper async => sync
-const _executeSync = (execute) => {
-  // Add middleware
-  const _execute = withCache(withStatusCode(execute))
+const executeSync = (execute) => {
+  let _execute = undefined
```
HenryNguyen5, line 26: Variables are undefined by default

bootstrap/index.js
```diff
@@ -17,20 +17,32 @@ const withStatusCode = (execute) => (data, callback) => {
 const cacheOptions = envOptions()
 if (cacheOptions.enabled) logger.info('Cache enabled: ', cacheOptions)
 
+const withMiddleware = async (execute) => {
+  return await withCache(withStatusCode(execute))
+}
```
HenryNguyen5, line 22: Unneeded async usage
```suggestion
const withMiddleware = (execute) => withCache(withStatusCode(execute))
```

bootstrap/index.js
```diff
@@ -17,20 +17,32 @@ const withStatusCode = (execute) => (data, callback) => {
...
-  // Add middleware
-  const _execute = withCache(withStatusCode(execute))
+const executeSync = (execute) => {
+  let _execute = undefined
   // Return sync function
-  return (data, callback) =>
-    _execute(data).then((result) => callback(result.statusCode, result.data))
+  return (data, callback) => {
+    // Init middleware only once
+    const init = _execute
```
HenryNguyen5, line 30: Won't this result in execute never having middleware wrapped around it?

bootstrap/index.js
```diff
@@ -17,20 +17,32 @@ const withStatusCode = (execute) => (data, callback) => {
...
+      ? Promise.resolve(_execute)
+      : withMiddleware(execute).then((fn) => {
+          _execute = fn
+          return _execute
+        })
+
+    return init.then((fn) => fn(data)).then((result) => callback(result.statusCode, result.data))
```
HenryNguyen5, line 37: Can we use a more descriptive name than `fn` here?

bootstrap/lib/cache/index.js
```diff
@@ -47,13 +46,23 @@ const envOptions = () => ({
   },
 })
 
-const withCache = (execute, options) => {
+const getCacheImpl = async (options) => {
+  switch (options.type) {
+    case 'redis':
+      return await redis.RedisCache.build(options.redis)
```
HenryNguyen5, line 52: Unneeded async usage
```suggestion
const getCacheImpl = (options) => {
  switch (options.type) {
    case 'redis':
      return redis.RedisCache.build(options.redis)
```

bootstrap/lib/cache/index.js
```diff
@@ -108,7 +117,8 @@ const withCache = (execute, options) => {
             // Add some entropy here because of possible scenario where the key won't be set before multiple
             // other instances in a burst request try to access the coalescing key.
             const randomMs = Math.random() * options.requestCoalescing.entropyMax
-            await delay(randomMs)
+            // Adding this next condition because of mocha timeout issues
+            if (randomMs) await delay(randomMs)
```
HenryNguyen5, line 121: I would probably change the condition to skip if `!options.requestCoalescing.entropyMax`, it reads a little better than checking if the value of `randomMs` exists or not

bootstrap/lib/cache/redis.js
```diff
@@ -0,0 +1,64 @@
...
+    this._auth = promisify(client.auth).bind(client)
+    this._get = promisify(client.get).bind(client)
+    this._set = promisify(client.set).bind(client)
+    this._del = promisify(client.del).bind(client)
+    this.client = client
+  }
+
+  async connect() {
+    if (!this.options.password) return
+    await this._auth(this.options.password)
```
HenryNguyen5, line 37:
```suggestion
    this._auth(this.options.password)
```

bootstrap/lib/cache/redis.js
```diff
@@ -0,0 +1,64 @@
...
+
+  static async build(options) {
+    const cache = new RedisCache(options)
+    await cache.connect()
+    return cache
+  }
+
+  async set(key, value, maxAge) {
+    const entry = JSON.stringify(value)
+    return await this._set(key, entry, 'PX', maxAge)
```
HenryNguyen5, line 48:
```suggestion
    return this._set(key, entry, 'PX', maxAge)
```

bootstrap/lib/cache/redis.js
```diff
@@ -0,0 +1,64 @@
...
+// TODO: reconnections and error handling
+class RedisCache {
+  constructor(options) {
+    this.options = options
+    const client = redis.createClient(options)
+    this._auth = promisify(client.auth).bind(client)
+    this._get = promisify(client.get).bind(client)
+    this._set = promisify(client.set).bind(client)
+    this._del = promisify(client.del).bind(client)
+    this.client = client
```
HenryNguyen5, line 32: I would probably just choose a redis library that supports promises out of the box (and typescript types too!)

HenryNguyen5 (changes requested):

bootstrap/lib/cache/index.js
krebernisak: I agree. Did just quickly just to make the test pass. I would like to get rid of this condition all together.

bootstrap/index.js
krebernisak: Yeah, wanted to be explicit here. But I agree, there is no need.

bootstrap/lib/cache/redis.js
krebernisak: Redis Node.js has two primary clients available `node_redis` (now just `redis`) and `ioredis`. This is the official recommendation, so I choose the more popular one of the two. Both do not support Promises out of the box.

bootstrap/index.js
krebernisak: Init will check if `_execute` exists in the closure. If it does, it means it was already wrapped with middleware and it uses it, if not it wraps it with middleware.

HenryNguyen5 (approved):

HenryNguyen5 (changes requested):

bootstrap/index.js
```diff
@@ -1,36 +1,83 @@
...
-    data && typeof data === 'object' && data.statusCode
-      ? callback(statusCode, { ...data, statusCode })
-      : callback(statusCode, data)
-  return execute(data, _callback)
+const skipOnError = (middleware) => async (execute) => {
+  // Try to execute, pass through on error
+  const _safe = (executeWithMiddleware) => async (data) => {
+    try {
+      return await executeWithMiddleware(data)
+    } catch (error) {
```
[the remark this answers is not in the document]
krebernisak: Passing the error to the next middleware sounds like functional error handling, e.g. Monads & Functors, Rx :D 

In any case, the first implementation did not skip errors but just returned a 500 to the user. When thinking about this, I realized that it is probably wrong to error the request just because it's not optimized/cached. Part of the pipeline failed, but we can still recover and fetch the data directly from the data provider.

bootstrap/index.js
```diff
@@ -1,36 +1,83 @@
...
-  const _callback = (statusCode, data) =>
-    data && typeof data === 'object' && data.statusCode
-      ? callback(statusCode, { ...data, statusCode })
-      : callback(statusCode, data)
-  return execute(data, _callback)
+const skipOnError = (middleware) => async (execute) => {
+  // Try to execute, pass through on error
+  const _safe = (executeWithMiddleware) => async (data) => {
+    try {
+      return await executeWithMiddleware(data)
```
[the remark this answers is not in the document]
krebernisak: Except in `try {} catch()` cases.

If we return a promise with `.catch` handler then we broke the async flow on error, and we can not do much in that handler except log the error as the context is gone and there is no callback available.

bootstrap/index.js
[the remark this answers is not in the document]
HenryNguyen5: Makes sense for caching in particular, I'm just thinking about when we add more middlewares than just caching.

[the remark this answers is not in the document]
HenryNguyen5: For example, if we added an auth middleware then this would leave a pretty big bug as they would still be able to access the data provider

HenryNguyen5 (approved):

thodges-gh (approved):

boxhock (approved):

itmustbejj (approved):

PR #97, opened 2020-10-06
Add adapters that supports ARS

boxhock:


bitex/adapter.js
```diff
@@ -0,0 +1,31 @@
...
+  const quote = validator.validated.data.quote.toLowerCase()
+  const url = `https://bitex.la/api/${endpoint}/${base}_${quote}`
+
+  const config = {
+    url
+  }
+
+  Requester.request(config)
+    .then(response => {
+      response.data.result = Requester.validateResultNumber(response.data, ['data', 'attributes', 'last'])
```
thodges-gh, line 23: We should probably use 'vwap' instead of 'last'. Last is usually the last traded value and can be extremely volatile.

bitso/adapter.js
```diff
@@ -0,0 +1,31 @@
...
+  const base = validator.validated.data.base.toLowerCase()
+  const quote = validator.validated.data.quote.toLowerCase()
+  const book = `${base}_${quote}`
+
+  const params = { book }
+  const config = { url, params }
+
+  Requester.request(config)
+    .then(response => {
+      response.data.result = Requester.validateResultNumber(response.data, ['payload', 'last'])
```
thodges-gh, line 23: Use 'vwap' here too.

satoshitango/adapter.js
```diff
@@ -0,0 +1,31 @@
...
+  const quote = validator.validated.data.quote.toUpperCase()
+  const url = `https://api.satoshitango.com/v3/${endpoint}/${quote}`
+
+  const config = {
+    url
+  }
+
+  Requester.request(config)
+    .then(response => {
+      response.data.result = Requester.validateResultNumber(response.data, ['data', 'ticker', base, 'bid'])
```
thodges-gh, line 23: This isn't great, but the API doesn't really offer an actual price.

thodges-gh (changes requested):

satoshitango/adapter.js
boxhock: I agree. I've informed Johann about this

thodges-gh (approved):

PR #105, opened 2020-10-10
Proof of Reserves feed (combined adapter/build  v2) 

krebernisak:


feeds/proof-of-reserves/README.md
```diff
@@ -0,0 +1,8 @@
+# Chainlink Proof of Reserves feed
```
thodges-gh, line 1: Can we add the steps to build this here?

blockchain.com/src/adapter.ts
```diff
@@ -6,51 +6,50 @@ export type JobSpecRequest = {
...
     }
     default: {
-      fn = Promise.reject(Error(`Endpoint ${endpoint} not supported.`))
-      break
+      throw Error(`Endpoint ${endpoint} not supported.`)
     }
   }
 
-  fn.then(_handleResponse).catch(_handleError)
+  const result = await fn
```
HenryNguyen5, line 42: Can we await when `balance.execute` is first called? Then store the result in the variable instead of the promise?

blockcypher/src/adapter.ts
```diff
@@ -6,51 +6,52 @@ export type JobSpecRequest = {
   id: string
   data: Record<string, unknown>
 }
-type Callback = (statusCode: number, data: Record<string, unknown>) => void
+type JobSpecResponse = { statusCode: number; data: Record<string, unknown> }
```
HenryNguyen5, line 9: Looks like we're using this across multiple packages, would be nice if we can store it in one place

bootstrap/index.js
```diff
@@ -49,8 +50,15 @@ const withLogger = (execute) => async (data) => {
   return result
 }
 
+const { toAsync } = util
+const { isAsyncFunction } = types
```
HenryNguyen5, line 54: I would say this is excessive destructuring, its hard to know what the parent module is when this is done too much

composite/proof-of-reserves/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('../../.eslintrc.ts.js'),
+}
```
HenryNguyen5, line 3:
```suggestion
module.exports = require('../../.eslintrc.ts.js')
```

composite/proof-of-reserves/.prettierrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('../../.prettierrc.js'),
+}
```
HenryNguyen5, line 3:
```suggestion
module.exports = require('../../.prettierrc.js')
```

composite/proof-of-reserves/.eslintrc.js
HenryNguyen5: Doesnt eslint resolve by crawling up the directory tree anyway?

composite/proof-of-reserves/.prettierrc.js
HenryNguyen5: Doesnt prettier resolve by crawling up the directory tree anyway?

composite/proof-of-reserves/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "start": "yarn server:dist"
+  },
+  "devDependencies": {
+    "@types/chai": "^4.2.11",
+    "@types/express": "^4.17.6",
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^14.0.13",
+    "@typescript-eslint/eslint-plugin": "^3.9.0",
+    "@typescript-eslint/parser": "^3.9.0",
+    "ts-node": "^8.10.2",
```
HenryNguyen5, line 40: Doesn't look like we're using ts-node here

composite/proof-of-reserves/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const execute = async (input: any) => {
+  // Get address set for protocol
+  const executeProtocol = getProtocolImpl({ type: process.env.PROTOCOL_ADAPTER_TYPE })
+  let output: any = throwOnError(await executeProtocol(input))
+  logger.debug('_onProtocol', { output })
+
+  // Get balances for address set
+  const executeBalance = getBalanceImpl({ type: process.env.BTC_BALANCE_ADAPTER_TYPE })
+  input = {
```
HenryNguyen5, line 20: Can we use new const here instead of mutating the argument?

composite/proof-of-reserves/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+      dataPath: 'result',
+      endpoint: 'balance',
+      confirmations: 6,
+    },
+  }
+  output = throwOnError(await executeBalance(input))
+  logger.debug('_onBalance', { output })
+
+  // Get reduce balances as total balance
+  input = {
```
HenryNguyen5, line 33: Can we use new const here instead of mutating the argument?

composite/proof-of-reserves/src/balance.ts
```diff
@@ -0,0 +1,16 @@
+import blockchainCom from '@chainlink/blockchain.com'
+import blockcypher from '@chainlink/blockcypher'
+
+const getImpl = (options: any) => {
```
HenryNguyen5, line 4: Can the options be typed?

composite/proof-of-reserves/src/protocol.ts
```diff
@@ -0,0 +1,16 @@
+import renVM from '@chainlink/renvm-address-set'
+import wBTC from '@chainlink/wbtc-address-set'
+
+const getImpl = (options: any) => {
```
HenryNguyen5, line 4: Can the options be typed?

composite/proof-of-reserves/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src",
+    "typeRoots": ["../../node_modules/@types", "../../typings", "./typings"]
```
HenryNguyen5, line 6: Do we need this?

composite/proof-of-reserves/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+  },
+  "include": ["src/**/*"],
```
HenryNguyen5, line 8:
```suggestion
  "include": ["src"],
```

composite/proof-of-reserves/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+  "exclude": ["dist", "**/*.spec.ts", "**/*.test.ts"]
```
HenryNguyen5, line 9: We should probably include the tests so they get type-checked, or create another tsconfig.json just for tests

package.json
```diff
@@ -61,7 +61,8 @@
     "blockchain.com",
     "blockcypher",
     "reduce",
-    "json-rpc"
+    "json-rpc",
+    "composite/**"
```
HenryNguyen5, line 65: composite/proof-of-reserves has a lockfile checked in, should remove that if its part of the monorepo

wbtc-address-set/src/adapter.ts
```diff
@@ -31,39 +31,41 @@ type AddressType = 'custodial' | 'merchant' | 'deposit'
...
-    callback(
-      200,
-      Requester.success(jobRunID, {
-        data: { response: out.data, result: addresses },
-        result: addresses,
-        status: 200,
-      }),
-    )
-  }
+  const reqConfig = { ...config.api, url: '' }
```
HenryNguyen5, line 50: Should URL be empty?

HenryNguyen5 (changes requested):

composite/proof-of-reserves/tsconfig.json
krebernisak: Took this from [here](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html), we may not need it for this flat structure but some refactor in the future could introduce subfolders and break things, so I would just leave it as is.

composite/proof-of-reserves/tsconfig.json
krebernisak: Yes, as we have `typings` package with types for `'@chainlink/ea-bootstrap'` and `'@chainlink/external-adapter'` as they are still in JS-land. The other two definitions are just here as a template.

composite/proof-of-reserves/.prettierrc.js
krebernisak: I guess we don't 👍

composite/proof-of-reserves/.eslintrc.js
krebernisak: I think we do, as we have on for JS `.eslintrc.js` and the other for TS `.eslintrc.ts.js`

bootstrap/index.js
krebernisak: Agreed! 💯

blockchain.com/src/adapter.ts
krebernisak: Great catch! We can now that this is `async`, it was full of promises and callbacks just a minute ago ;)

composite/proof-of-reserves/tsconfig.json
krebernisak: Ok, will keep this in mind for future TS improvements. I do not have the resources to change/test this now.

composite/proof-of-reserves/tsconfig.json
HenryNguyen5: it actually works recursively even without the glob

HenryNguyen5: This is fine though, we can address it later

HenryNguyen5 (approved):

PR #110, opened 2020-10-14
[Feature] TUSD Adapter

justinkaseman:
### Description
A simple adapter pulling Trust Token's True USD reserves from (https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD)

Includes a bugfix to the make new adapter command which was causing it to write to `package.json`'s `scripts`.

### Ticket
https://www.pivotaltracker.com/n/projects/2424827/stories/175264096


trueusd/README.md
```diff
@@ -0,0 +1,20 @@
+# Chainlink External Adapter for trueusd
```
thodges-gh, line 1: Let's get this updated, no input params, but then take an example for the output body.

trueusd/yarn-error.log
```diff
@@ -0,0 +1,45 @@
+Arguments:
```
thodges-gh, line 1: Add this file to your global .gitignore.

thodges-gh (changes requested):
Just some small things, we forgot to update that adapter's README.

thodges-gh (approved):

PR #112, opened 2020-10-16
[Feature] BTC Dominance

justinkaseman:
### Description
Adds the ability to get BTC's market dominance across 5 API's (Coin Gecko, CoinMarketCap, Coin Paprika, Coin Lore, & Messari) by:
- Extending the following adapters: Coin Gecko, CoinMarketCap, Coin Paprika
- Adding the following new adapters: Coin Lore, Messari

Also sneaks in a quick fix to the True USD adapter's output README.

### Ticket
https://www.pivotaltracker.com/story/show/175222857

### DIFF
--- a/coingecko/adapter.js
+++ b/coingecko/adapter.js
@@ -79,14 +79,14 @@ const price = (jobRunID, input, callback) => {
 }
 
 const globalParams = {
-  quote: ['quote', 'to', 'market', 'coin'],
+  market: ['market', 'to', 'quote'],
 }
 
 const global = (jobRunID, input, path, callback) => {
   const validator = new Validator(input, globalParams)
   if (validator.error) return callback(validator.error.statusCode, validator.error)
 
-  const quote = validator.validated.data.quote
+  const market = validator.validated.data.market
   const url = 'https://api.coingecko.com/api/v3/global'
 
   const config = {
@@ -97,7 +97,7 @@ const global = (jobRunID, input, path, callback) => {
     response.data.result = Requester.validateResultNumber(response.data, [
       'data',
       path,
-      quote.toLowerCase(),
+      market.toLowerCase(),
     ])
     callback(response.status, Requester.success(jobRunID, response))
   }
--- a/coingecko/adapter.js
+++ b/coingecko/adapter.js
@@ -151,7 +151,7 @@ const execute = (input, callback) => {
       return price(jobRunID, input, callback)
     case ENDPOINT_MKTCAP:
       return globalMarketCap(jobRunID, input, callback)
-    case ENDPOINT_MKTDOM: 
+    case ENDPOINT_MKTDOM:
       return globalMarketDom(jobRunID, input, callback)
     default:
       callback(500, Requester.errored(jobRunID, 'invalid endpoint provided'))
--- a/coingecko/adapter.js
+++ b/coingecko/adapter.js
@@ -2,7 +2,7 @@ const { Requester, Validator } = require('@chainlink/external-adapter')
 
 const ENDPOINT_PRICE = 'price'
 const ENDPOINT_MKTCAP = 'globalmarketcap'
-const ENDPOINT_MKTDOM = 'globalmarketdom'
+const ENDPOINT_DOMINANCE = 'dominance'
 
 const DEFAULT_ENDPOINT = ENDPOINT_PRICE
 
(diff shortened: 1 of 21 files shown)


coingecko/adapter.js
```diff
@@ -106,6 +107,35 @@ const globalMarketCap = (jobRunID, input, callback) => {
   Requester.request(config, customError).then(_handleResponse).catch(_handleError)
 }
 
+const mktdomParams = {
+  quote: ['quote', 'to', 'market'],
+}
+
+const globalMarketDom = (jobRunID, input, callback) => {
```
boxhock, line 114: We can consolidate this function with the other function using the same endpoint.

Example from the work I started on this, but never got time to finish: https://github.com/smartcontractkit/external-adapters-js/compare/master...btc-dominance

boxhock (changes requested):
I see that a few of the adapters have different endpoint names and/or expected params.

With the way FluxMonitor currently works, we should expect all adapters to expect the same params. In other words, the adapters should expect the same "endpoint" param for market dominance, and possibly another param for the market/coin.

coinmarketcap/adapter.js
```diff
@@ -50,4 +59,47 @@ const execute = (input, callback) => {
     .catch((error) => callback(500, Requester.errored(jobRunID, error)))
 }
 
+const globalMarketDom = (jobRunID, input, callback) => {
+  const validator = new Validator(input)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const url = 'https://pro-api.coinmarketcap.com//v1/global-metrics/quotes/latest'
```
boxhock, line 66:
```suggestion
  const url = 'https://pro-api.coinmarketcap.com/v1/global-metrics/quotes/latest'
```

coingecko/adapter.js
boxhock: Looks like we can do the same on coinpaprika as well

coinpaprika/adapter.js
```diff
@@ -87,7 +88,9 @@ const execute = (input, callback) => {
     case ENDPOINT_PRICE:
       return price(jobRunID, input, callback)
     case ENDPOINT_MKTCAP:
-      return globalMarketCap(jobRunID, input, callback)
+      return global(jobRunID, input, 'market_cap_usd', callback)
+    case ENDPOINT_DOMINANCE:
+      return global(jobRunID, input, 'bitcoin_dominance_percentage', callback)
```
boxhock, line 93: We should probably take the coin as a a parameter here as well.

I know they don't track any other coin's dominance here, but it will save us work in the future if they change that and we need something other than BTC's dominance.

Since all other adapters require the coin's symbol (e.g. "BTC"), you can convert this quickly like this:

```javascript
const convert = {
  BTC: 'bitcoin'
}
```

coinmarketcap/adapter.js
```diff
@@ -50,4 +59,47 @@ const execute = (input, callback) => {
...
+
+  const config = {
+    url: url,
+    headers: {
+      'X-CMC_PRO_API_KEY': process.env.API_KEY,
+    },
+  }
+
+  const _handleResponse = (response) => {
+    response.data.result = Requester.validateResultNumber(response.data, ['data', 'btc_dominance'])
```
boxhock, line 76: Same thing with making BTC dymanic here.

coinlore/adapter.js
```diff
@@ -0,0 +1,24 @@
...
+  const validator = new Validator(input)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const jobRunID = validator.validated.id
+  const url = 'https://api.coinlore.net/api/global/'
+
+  Requester.request(url, customError)
+    .then((response) => {
+      response.data = response.data[0]
+      response.data.result = Requester.validateResultNumber(response.data, ['btc_d'])
```
boxhock, line 18: Same thing with making BTC dymanic here.

coinpaprika/README.md
```diff
@@ -5,44 +5,44 @@
 - `coinid`: The CoinPaprika id of the coin to query (required if not using `from`)
 - `base`, `from`, or `coin`: The ticker of the coin to query (required if not using `coinid`)
 - `quote`, `to`, or `market`: The currency to convert the coin to (required)
-- `endpoint`: The endpoint to use (defaults to "price", one of "price", "globalMarketCap")
+- `endpoint`: The endpoint to use (defaults to "price", one of "price", "globalMarketCap", "dominance)
```
boxhock, line 8:
```suggestion
- `endpoint`: The endpoint to use (defaults to "price", one of "price", "globalMarketCap", "dominance")
```

boxhock (changes requested):

coinpaprika/adapter.js
justinkaseman: I made these more dynamic by adding a parameter on the `global` helper for if the coin should be utilized to key into the payload. It does add the caveat of having the coin required for getting market capitalization. As seen in the test I just pass in 'USD' for that example.

Let me know if this is in line with what you were thinking. I could also make `coin/market/ect.` an optimal param there.

boxhock (approved):

PR #116, opened 2020-10-23
Feature/anyblock uniswap vwap

krebernisak:
Adapter using Anyblock service to calculate VWAP for any Uniswap asset


anyblock-uniswap-vwap/README.md
```diff
@@ -0,0 +1,51 @@
...
+        "timestamp": 1600304409,
+        "reserve0": 1.3224814613263435e23,
+        "reserve1": 406170744097414250000
+      },
+      {
+        // more raw values
+      }
+    ]
+  },
+  "result": {
```
thodges-gh, line 45: We've typically only put one value in the result key. This is so that we can have different adapter implementations (for different providers) but give the same job spec to multiple node operators. In this case, the `copyPath` would need to be `["result", "vwap"]` for the vwap value instead of just result. Is there a hard reason for this or can we just return the vwap value in result?

krebernisak:
@jondoe1337 could you please test/debug this one?

If I use any address other than XTF pool, for example wBTC pool, I get this error: 

```json
{"took":976,"timed_out":false,"_shards":{"total":40,"successful":40,"skipped":0,"failed":0},"hits":{"total":{"value":0,"relation":"eq"},"max_score":null,"hits":[]}}
```

anyblock-uniswap-vwap/README.md
krebernisak: @jondoe1337 how do you plan tu use this adapter?

jondoe1337: Uhm well, for now it might just be used in the XFT Adapter #87
However I'm pretty sure that there might be other price feeds wanted in future!?

jondoe1337:
> @jondoe1337 could you please test/debug this one?
> 
> If I use any address other than XTF pool, for example wBTC pool, I get this error:
> 
> ```json
> {"took":976,"timed_out":false,"_shards":{"total":40,"successful":40,"skipped":0,"failed":0},"hits":{"total":{"value":0,"relation":"eq"},"max_score":null,"hits":[]}}
> ```

What is the address of that token?

anyblock-uniswap-vwap/README.md
thodges-gh: If this adapter is going to be used with other price feed adapters, then only the value within vwap should go to the result key.

krebernisak:
> > @jondoe1337 could you please test/debug this one?
> > If I use any address other than XTF pool, for example wBTC pool, I get this error:
> > ```json
> > {"took":976,"timed_out":false,"_shards":{"total":40,"successful":40,"skipped":0,"failed":0},"hits":{"total":{"value":0,"relation":"eq"},"max_score":null,"hits":[]}}
> > ```
> 
> What is the address of that token?

For example wBTC: https://info.uniswap.org/pair/0xbb2b8038a1640196fbe3e38816f3e67cba72d940

But I could not find any other token that works except XTF: https://info.uniswap.org/pair/0x2B9e92A5B6e69Db9fEdC47a4C656C9395e8a26d2

jondoe1337:
All of the pairs should have been translated. The problem is that you have used the lower-case address.

If you want to search for a checksum-case address then you can use `address.raw` in that line: https://github.com/smartcontractkit/external-adapters-js/blob/0250a53bf2226256db0e5a9b7f91bf761335234b/anyblock-uniswap-vwap/src/adapter.ts#L109

In case you just have a lowercase you should just search for `address`.

(we separated full-text and keyword search for some properties because our use-cases fit better in that solution)

I tend to just use web3js and transform the lowercase address to checksum case and then use `address.raw`.

Makes sense?

-- 

Regarding your question to if I can provide Typescript classes: Not sure how I can help here :( We never used the elastic provided TS classes and to be honest, I have never had contact with TS directly. Is there any other way I can help on this?

anyblock-uniswap-vwap/README.md
krebernisak: I've updated the adapter so it now only outputs the vwap value as result.

thodges-gh:
Need to update the README based on the change from 744e517.

krebernisak:
@jondoe1337 I can confirm that it works with checksum addresses. I've added a note to the readme.

Don't worry about the TS types, we'll slowly improve on this ourselves as the repo evolves so the code is more safe and readable.

thodges-gh (approved):

thodges-gh (approved):

PR #118, opened 2020-10-23
[Feature] COVID tracker adapter

justinkaseman:
### Description
Adds a new adapter for COVID tracker's API (https://covidtracking.com/data/api)

This adapter is dynamic in that it takes in a `field` param for the data field to return and an optional `date` field for accessing historical COVID data.

### Caveats
While testing validation I noticed that a test that returns a `400` error that is put in the `'error calls @integration'` group still passes. Not sure if this is a repository wide issue - needs further testing.

### Ticket
https://www.pivotaltracker.com/story/show/175417790

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -108,6 +108,10 @@ jobs:
             blockcypher,
             wbtc-address-set,
             renvm-address-set,
+            messari,
+            coinlore,
+            trueusd,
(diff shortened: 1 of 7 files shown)


thodges-gh:
We should add this adapter to the release workflow (`.github/workflows/release.yml`).

thodges-gh (approved):

PR #119, opened 2020-10-23
Update Amberdata to new VWAP endpoint

thodges-gh:
https://medium.com/amberdata/new-global-vwap-a-sunsetting-notice-809b6f1c637a

### DIFF
--- a/amberdata/adapter.js
+++ b/amberdata/adapter.js
@@ -16,7 +16,7 @@ const execute = (input, callback) => {
   const jobRunID = validator.validated.id
   const coin = validator.validated.data.base
   const market = validator.validated.data.quote
-  const url = `https://web3api.io/api/v2/market/prices/${coin.toLowerCase()}_${market.toLowerCase()}/latest`
+  const url = `https://web3api.io/api/v2/market/spot/prices/pairs/${coin.toLowerCase()}_${market.toLowerCase()}/latest`
 
   const config = {
     url,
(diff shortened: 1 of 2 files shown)


krebernisak (approved):
It would be great if we could start to improve the per adapter documentation.
For example, we could document that this adapter returns a volume-weighted average price with some links (from the article) on supported pairs and any other relevant info.

Maybe we can make this a project in itself and then someone would need to go through all adapters and improve documentation, examples, etc.

thodges-gh:
I'll add that to the README before merging. :+1:

krebernisak (approved):

amberdata/adapter.js
```diff
@@ -16,7 +16,7 @@ const execute = (input, callback) => {
   const jobRunID = validator.validated.id
   const coin = validator.validated.data.base
   const market = validator.validated.data.quote
-  const url = `https://web3api.io/api/v2/market/prices/${coin.toLowerCase()}_${market.toLowerCase()}/latest`
+  const url = `https://web3api.io/api/v2/market/spot/prices/pairs/${coin.toLowerCase()}_${market.toLowerCase()}/latest`
```
TrevorJTClarke, line 19: @thodges-gh its come to our attention that some pairs CL is requesting actually need cross rates.
If that is needed/supported in CL ecosystem, using this endpoint you just need to add: `includeCrossRates=true`

We did this to make sure API users didnt confuse data returned. 
Example needing new flag: "bnb_usd"

See docs: https://docs.amberdata.io/reference#spot-price-pair-latest

PR #120, opened 2020-10-26
Improve error handling

krebernisak:


external-adapter/src/errors.js
```diff
@@ -0,0 +1,34 @@
...
+    this.jobRunID = jobRunID
+    this.status = status
+    this.statusCode = statusCode
+    this.name = name
+    this.message = message
+    this.cause = cause
+  }
+
+  toJSONResponse() {
+    const showDebugInfo = process.env.NODE_ENV === 'development' || process.env.DEBUG === 'true'
```
thodges-gh, line 22: If you don't mind adding this, something that's commonly used (even in the Chainlink node) is `LOG_LEVEL=debug`

krebernisak: I was thinking of reusing `LOG_LEVEL`, but this is actually going to the output as a response, so I introduced a separate env var because it changes the behavior of the adapter (input/output).

In the `anyblock-uniswal-vwap` adapter we have something similar, where a `debug` input param is used to change the behavior of the adapter and in that case, it will output extra info in the response.

So my idea is to keep the logs separated from adapter behavior. 
Then on the behavior side, consolidate the two, env `DEBUG` var and `debug` input parameter, so the adapter can output some extra data in the response when either is set.

What do you think?

thodges-gh: That sounds good to me.

thodges-gh (approved):

PR #121, opened 2020-10-26
Update TrustToken adapter

justinkaseman:
### Description
Add param to the TrustToken adapter to differentiate getting the supply of TUSD token or the USD trust supply.

### Ticket
https://www.pivotaltracker.com/story/show/175399546

### DIFF
--- a/trueusd/README.md
+++ b/trueusd/README.md
@@ -6,12 +6,12 @@ https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD
 
 ## Input Params
 
-- `base`, `from`, or `asset`: Required, the symbol of the asset to get the total supply of, one of "USD", "TUSD"
+- `field`: Required, the data field to return. Must be in camel-case style.
 
 ## Output
 
 ```json
-{ "base": "TUSD" }
+{ "field": "totalTrust" }
 ```
 
 ```json
--- a/trueusd/README.md
+++ b/trueusd/README.md
@@ -1,10 +1,18 @@
 # Chainlink External Adapter for trueusd
 
+## Total Supply API
+
+https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD
+
 ## Input Params
 
-_No params_
+- `base`, `from`, or `asset`: Required, the symbol of the asset to get the total supply of, one of "USD", "TUSD"
 
 ## Output
+
+```json
+{ "base": "TUSD" }
+```
 
 ```json
 {
@@ -23,9 +31,9 @@ _No params_
     "message": [{ "msg": "get contractSupply successfully" }],
     "success": true,
     "responseCode": 200,
-    "result": 347215038.53
+    "result": 343465335.4587567
   },
-  "result": 347215038.53,
+  "result": 343465335.4587567,
   "statusCode": 200
 }
 ```
--- a/trueusd/README.md
+++ b/trueusd/README.md
@@ -1,4 +1,4 @@
-# Chainlink External Adapter for trueusd
+# Chainlink External Adapter for TrueUSD
 
 ## Total Supply API
 
--- a/trueusd/README.md
+++ b/trueusd/README.md
@@ -6,13 +6,9 @@ https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD
 
 ## Input Params
 
-- `field`: Required, the data field to return. Must be in camel-case style.
+- `field`: The data field to return. (defaults to `totalToken`, one of `totalToken`, `totalTrust`)
 
 ## Output
-
-```json
-{ "field": "totalTrust" }
-```
 
 ```json
 {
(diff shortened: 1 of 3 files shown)


trueusd/adapter.js
```diff
@@ -5,19 +5,28 @@ const customError = (data) => {
...
 }
 
+const convert = {
+  USD: 'totalTrust',
+  TUSD: 'totalToken',
+}
+
+const supplyParams = {
+  base: ['base', 'from', 'asset'],
+}
```
thodges-gh, line 15: I think instead of this we can just supply a param for "field" and parse that from responseData.

justinkaseman: My idea was to do this to normalize params in case we start supporting a lot of 'supply' feeds, but yeah this might be cleaner just as `field`.

Added a commit to change it.

thodges-gh: Ah, then maybe I like where you were going. So if `totalTrust` is the amount of USD secured, and `totalToken` are the number of tokens minted, then we could think about input parameters that would be compatible with multiple PoR-type feeds with different assets.

trueusd/README.md
```diff
@@ -1,11 +1,19 @@
 # Chainlink External Adapter for trueusd
 
+## Total Supply API
+
+https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD
+
 ## Input Params
 
-_No params_
+- `field`: Required, the data field to return. Must be in camel-case style.
```
krebernisak, line 9: Please document the different options here. I guess they are `(totalTrust|totalToken)`?

krebernisak: Also maybe document what different options mean. I guess `totalTrust` are bank locked USD, and `totalToken` would be the total supply of blockchain issued `TUSD`?

Going forward we should invest more time in writing great documentation to make our adapters easy to understand and straightforward to use.

trueusd/README.md
```diff
@@ -1,11 +1,19 @@
 # Chainlink External Adapter for trueusd
```
krebernisak, line 1:
```suggestion
# Chainlink External Adapter for TrueUSD
```

krebernisak (dismissed):
Looks good! 

Just, it would be great to add some more documentation on the different options.

trueusd/adapter.js
thodges-gh: How about an `endpoint` param where possible values are `secured` and `supply`? Secured maps to `totalTrust` and `supply` maps to `totalToken`. Trying to think if that would be compatible with future APIs for this same use-case.

justinkaseman: I like these endpoints suggestions.

Another slightly unrelated thing to think about is that Trust Token has stable coins for other fiat currencies (e.g. TrueCAD) that can be queried by changing the end of the url string. This adapter may be better as `trust-token` that takes in a param for `base` to switch between TUSD, TCAD, TAUD, ect.

Thoughts?

krebernisak: If that is the case then definitely rename it as `trusttoken` and extend it to support different `endpoint`s  like `TrueUSD`, `TrueAUD`, etc.

tyrion70:
Just a quick note, adding a required field to determine which value to get is a breaking change. The TUSD POR job however has not been sent to the node operators yet so in this case thats not a big issue, so we'll wait with POR until this PR is merged.

trueusd/README.md
```diff
@@ -1,11 +1,19 @@
...
 
-_No params_
+- `field`: Required, the data field to return. Must be in camel-case style.
 
 ## Output
 
+```json
+{ "field": "totalTrust" }
+```
+
```
thodges-gh, line 16: I think this bit is extra.

trueusd/README.md
thodges-gh, line 9: Based on the feedback from @tyrion70, we should probably make this optional and default field to `totalToken`. Is that right?

trueusd/adapter.js
```diff
@@ -5,19 +5,21 @@ const customError = (data) => {
   return false
 }
 
+const supplyParams = {
+  field: ['field'],
```
thodges-gh, line 9:
```suggestion
  field: false,
```

trueusd/adapter.js
```diff
@@ -5,19 +5,21 @@ const customError = (data) => {
...
+}
+
 const execute = (input, callback) => {
-  const validator = new Validator(input)
+  const validator = new Validator(input, supplyParams)
   if (validator.error) return callback(validator.error.statusCode, validator.error)
 
   const jobRunID = validator.validated.id
   const url = 'https://core-api.real-time-attest.trustexplorer.io/trusttoken/TrueUSD'
+  const field = validator.validated.data.field
```
thodges-gh, line 18:
```suggestion
  const field = validator.validated.data.field || 'totalToken'
```

thodges-gh (changes requested):

thodges-gh:
I went through and resolved all the conversations, just need the title udpated in the README and I think we're good to go.

thodges-gh (approved):

PR #122, opened 2020-10-26
[Feature] Lition Adapter

justinkaseman:
### Description
Adds a new adapter for querying Lition price.

Lition currently has 3 different sources to pull price from that are denotated by an integer (e.g. source 1 is agora-energiewende.de). A source param is used to switch between these.

### Ticket
https://www.pivotaltracker.com/story/show/175421333

### DIFF
--- a/lition/README.md
+++ b/lition/README.md
@@ -2,13 +2,25 @@
 
 ## Price API
 
+Returns the price in Euros per MWh
+
 ### Endpoint
 
 https://staking.lition.io/api/v1/energy/source/1/date/2020-10-07/hour/15/
 
+### Sources
+
+| Number | Source                |
+| ------ | --------------------- |
+| 1      | agora-energiewende.de |
+| 2      | smard.de              |
+| 3      | energy-charts.info    |
+
 ### Input Params
 
 - `source`: The provider to retrieve price data from (required, e.g. "1", "2", or "3")
+- `date`: The date to query formatted by `[YEAR][MONTH][DAY]` e.g. `20201012`, if not provided defaults to the current UTC date (optional)
+- `hour`: The hour to query (0-23), if not provided defaults to the current UTC hour (optional)
 
 ### Example Usage
 
--- a/lition/README.md
+++ b/lition/README.md
@@ -19,7 +19,7 @@ https://staking.lition.io/api/v1/energy/source/1/date/2020-10-07/hour/15/
 ### Input Params
 
 - `source`: The provider to retrieve price data from (required, e.g. "1", "2", or "3")
-- `date`: The date to query formatted by `[YEAR][MONTH][DAY]` e.g. `20201012`, if not provided defaults to the current UTC date (optional)
+- `date`: The date to query formatted by `[YEAR]-[MONTH]-[DAY]` e.g. `2020-10-12`, if not provided defaults to the current UTC date (optional)
 - `hour`: The hour to query (0-23), if not provided defaults to the current UTC hour (optional)
 
 ### Example Usage
--- a/lition/README.md
+++ b/lition/README.md
@@ -1,0 +1,28 @@
+# Chainlink External Adapter for Lition
+
+## Price API
+
+### Endpoint
+
+https://staking.lition.io/api/v1/energy/source/1/date/2020-10-07/hour/15/
+
+### Input Params
+
+- `source`: The provider to retrieve price data from (required, e.g. "1", "2", or "3")
+
+### Example Usage
+
+```json
+{ "source": "1" }
+```
+
+#### Output
+
+```json
+{
+  "jobRunID": "143239",
+  "data": { "price": 34.07, "source": "agora-energiewende.de", "result": 34.07 },
+  "result": 34.07,
+  "statusCode": 200
+}
+```
--- a/lition/adapter.js
+++ b/lition/adapter.js
@@ -17,7 +17,7 @@ const execute = (input, callback) => {
   const currentTime = new Date()
   const date =
     validator.validated.data.date ||
-    `${currentTime.getUTCFullYear()}-${currentTime.getUTCMonth()}-${currentTime.getUTCDate()}`
+    `${currentTime.toISOString().slice(0, 10)}` // YYYY-MM-DD
   const hour = validator.validated.data.hour || currentTime.getUTCHours()
 
   const url = `https://staking.lition.io/api/v1/energy/source/${source}/date/${date}/hour/${hour}/`
--- a/lition/adapter.js
+++ b/lition/adapter.js
@@ -15,9 +15,7 @@ const execute = (input, callback) => {
   const jobRunID = validator.validated.id
   const source = validator.validated.data.source
   const currentTime = new Date()
-  const date =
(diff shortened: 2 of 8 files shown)


lition/README.md
```diff
@@ -0,0 +1,28 @@
+# Chainlink External Adapter for Lition
+
```
krebernisak, line 2: Please document what is this adapter used for. I can't figure out what are we pricing here. The price `"price": 34.07` is a price of what exactly?

lition/adapter.js
```diff
@@ -0,0 +1,33 @@
...
+}
+
+const execute = (input, callback) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const jobRunID = validator.validated.id
+  const source = validator.validated.data.source
+
+  const url = `https://staking.lition.io/api/v1/energy/source/${source}/date/2020-10-07/hour/15/`
```
krebernisak, line 19: Why is the date part of the endpoint`date/2020-10-07/hour/15/` fixed?
How should this be used, and would it make sense to take this as a param with sensible defaults?

package.json
```diff
@@ -4,6 +4,7 @@
   "license": "MIT",
   "private": true,
   "workspaces": [
+    "lition",
```
krebernisak, line 7: Please also add `lition` to `.github/workflows/release.yml/publish-artifacts` so it's built automatically on every release.

krebernisak (changes requested):

lition/test/adapter_test.js
```diff
@@ -0,0 +1,66 @@
...
+      },
+    ]
+
+    requests.forEach((req) => {
+      it(`${req.name}`, (done) => {
+        execute(req.testData, (statusCode, data) => {
+          assertSuccess({ expected: 200, actual: statusCode }, data, jobID)
+          assert.isAbove(data.result, 0)
+          assert.isAbove(data.data.result, 0)
+          console.log(JSON.stringify(data))
```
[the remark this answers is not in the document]
justinkaseman: Good catch

lition/adapter.js
justinkaseman: That's a good point. I'll go to Devlin and get more details on how they want this used.

lition/adapter.js
```diff
@@ -0,0 +1,33 @@
+const { Requester, Validator } = require('@chainlink/external-adapter')
+
+const customError = (data) => {
+  if (data.Response === 'Error') return true
+  return false
+}
```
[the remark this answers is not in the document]
justinkaseman: Changed this in the "example" adapter as well.

lition/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Price API
+
+### Endpoint
+
+https://staking.lition.io/api/v1/energy/source/1/date/2020-10-07/hour/15/
+
+### Input Params
+
+- `source`: The provider to retrieve price data from (required, e.g. "1", "2", or "3")
```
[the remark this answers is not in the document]
justinkaseman: Do you think it is better to leave the param as an integer or just have the param be a string of the source's name?

lition/adapter.js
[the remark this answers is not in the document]
krebernisak: Great idea to fix it at the source 👍  We actually already have an updated example written in TS as part of the `release/v0.1.3`, which will hopefully be merged soon.

lition/README.md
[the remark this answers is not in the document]
krebernisak: It may be better to leave the numbers, as their API expects it as such, so if they add additional sources we do not need to update the adapter but just use the different source id.

Ideally, we would just link to their documentation where they list all the sources, if they have something like that available.

lition/README.md
justinkaseman: Good point. Added to README.

lition/README.md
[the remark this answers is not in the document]
justinkaseman: I haven't been able to find any documentation about the sources, but added to the README for what I was able to uncover via the API.

lition/adapter.js
justinkaseman: I've modified it to be a little more flexible with these.

One thing I am unsure of is what timezone they are basing their API off of, I presume UTC. I am able to query a full 24 hours ahead. I'll talk to Johann again tomorrow before merging this.

lition/adapter.js
```diff
@@ -0,0 +1,37 @@
...
+  const validator = new Validator(input, customParams)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const jobRunID = validator.validated.id
+  const source = validator.validated.data.source
+  const currentTime = new Date()
+  const date =
+    validator.validated.data.date ||
+    `${currentTime.getUTCFullYear()}-${currentTime.getUTCMonth()}-${currentTime.getUTCDate()}`
+  const hour = validator.validated.data.hour || currentTime.getUTCHours()
```
krebernisak, line 21: Same thing here... Looks like the API does not return for hour 0 or > 23.

krebernisak (changes requested):

lition/README.md
```diff
@@ -2,13 +2,25 @@
...
+| Number | Source                |
+| ------ | --------------------- |
+| 1      | agora-energiewende.de |
+| 2      | smard.de              |
+| 3      | energy-charts.info    |
+
 ### Input Params
 
 - `source`: The provider to retrieve price data from (required, e.g. "1", "2", or "3")
+- `date`: The date to query formatted by `[YEAR][MONTH][DAY]` e.g. `20201012`, if not provided defaults to the current UTC date (optional)
```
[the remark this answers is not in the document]
justinkaseman: The reason it is formatted this way is because in the recent COVID adapter it took in a date parameter in this format. 

Not sure if there is already a convention for how we want to normalize date params, but I adopted it from there. This is a completely separate adapter use case so I could see the argument for formatting it how the API wants.

lition/adapter.js
justinkaseman: I'm able to hit this endpoint: 
https://staking.lition.io/api/v1/energy/source/3/date/2020-10-29/hour/0

lition/adapter.js
```diff
@@ -0,0 +1,37 @@
...
+const execute = (input, callback) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const jobRunID = validator.validated.id
+  const source = validator.validated.data.source
+  const currentTime = new Date()
+  const date =
+    validator.validated.data.date ||
+    `${currentTime.getUTCFullYear()}-${currentTime.getUTCMonth()}-${currentTime.getUTCDate()}`
```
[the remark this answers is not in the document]
justinkaseman: Good catch and suggestion. Will change.

lition/adapter.js
krebernisak: Yeah, works for me now. Did not work before..

thodges-gh (approved):

PR #130, opened 2020-10-30
[Feature] LCX adapter

justinkaseman:
### Description
Adds a new adapter for LCX to the following prices:
- BTC -> USD
- BTC -> EUR
- ETH -> USD
- ETH -> EUR

### Ticket
https://www.pivotaltracker.com/story/show/175505362

### Caveats
@krebernisak there is a a linter warning getting thrown from the example adapter's `CustomError` having the any type on its parameter. I'm sure we will convert the external adapter lib code to TS soon. I can help on this if needed. Just figured I'd mention it.

### DIFF
--- a/lcx/src/adapter.ts
+++ b/lcx/src/adapter.ts
@@ -12,13 +12,13 @@ export const execute: Execute = async (input) => {
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
-  const base = validator.validated.data.base.toUpperCase()
-  const quote = validator.validated.data.quote.toUpperCase()
+  const coin = validator.validated.data.base.toUpperCase()
+  const currency = validator.validated.data.quote.toUpperCase()
   const url = `https://rp.lcx.com/v1/rates/current`
 
   const params = {
-    coin: base,
-    currency: quote,
+    coin,
+    currency,
   }
 
   const config = {
--- a/lcx/src/adapter.ts
+++ b/lcx/src/adapter.ts
@@ -1,10 +1,6 @@
 import { Execute } from '@chainlink/types'
 import { Requester, Validator } from '@chainlink/external-adapter'
 
-const customError = (data: any) => {
-  return data.Response === 'Request failed with status code 400'
-}
-
 const customParams = {
   base: ['base', 'from', 'coin'],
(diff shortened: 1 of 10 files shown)


lcx/src/adapter.ts
```diff
@@ -0,0 +1,33 @@
+import { Execute } from '@chainlink/types'
+import { Requester, Validator } from '@chainlink/external-adapter'
+
+const customError = (data: any) => {
+  return data.Response === 'Request failed with status code 400'
```
krebernisak, line 5: I do not think this is necessary in this case. The Requestor will actually throw on error and this function will not be called.

This custom handler is for cases where API returns 200, but we don't get the expected data in the response so we force an error. I think we should rarely use it, as in most cases most APIs will return the correct code (non 200) for error cases, and our requester will throw an error.

krebernisak (changes requested):
Please add the adapter to the CHANGELOG.

Regarding the TS warnings, they are legitimate and we should slowly start to fix them one by one. We are currently in this transition period where part of the codebase is JS and the other part is TS, so it's acceptable to have some warnings. 

Hopefully, we'll complete this transition soon.

lcx/src/adapter.ts
justinkaseman: Good to know, thanks. Removed the `customError`.

lcx/src/adapter.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const execute: Execute = async (input) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const base = validator.validated.data.base.toUpperCase()
+  const quote = validator.validated.data.quote.toUpperCase()
+  const url = `https://rp.lcx.com/v1/rates/current/?coin=${base}&currency=${quote}`
```
boxhock, line 17: We should be able to add these params as an object in the config.

boxhock (changes requested):

boxhock (dismissed):

krebernisak (approved):

krebernisak (approved):

boxhock (approved):

PR #135, opened 2020-11-03
Clean adapter build process

krebernisak:


.github/strategy/adapters.json
```diff
@@ -0,0 +1,67 @@
+{
+  "adapter": [
```
thodges-gh, line 2: We'll probably want to add writing to this file to the `make new` command.

krebernisak: I agree, but we should probably use [`jq`](https://stedolan.github.io/jq/) instead of `sed` for transforming JSON.

PR #136, opened 2020-11-03
Add XBTO adapter

boxhock:


xbto/test/adapter.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      {
+        name: 'id supplied',
+        testData: { id: jobID, data: {} },
+      },
+    ]
+
+    requests.forEach((req) => {
+      it(`${req.name}`, async () => {
+        const data = await execute(req.testData as AdapterRequest)
+        console.log(JSON.stringify(data))
```
justinkaseman, line 24:
```suggestion
```
Stray console.log

xbto/test/adapter.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      it(`${req.name}`, async () => {
+        try {
+          await execute(req.testData as AdapterRequest)
+        } catch (error) {
+          const errorResp = Requester.errored(jobID, error)
+          assertError({ expected: 400, actual: errorResp.statusCode }, errorResp, jobID)
+        }
+      })
+    })
+  })
```
justinkaseman, line 44:
```suggestion
```
I believe what is happening is this the empty body test returns a `200` because there are no params, so when asserting that the error should be  `400` it retries 3 times and then times out returning a `500` error. That's why it hits the timeout, because to do that takes about 4-5 seconds.

Moving this test to "successful calls" makes the unit tests pass. Or the other option is increasing the timeout in `package.json` **AND** changing this to assert for a `500` error works, but would be just getting around something that should be fixed in the test suite.

justinkaseman (approved):
LGTM  :+1:

PR #137, opened 2020-11-04
[Hotfix] LCX build script & makefile

justinkaseman:
### Description
Quick fix to LCX adapter's `build` script (which was built from the example adapter before Kristijan's recent change to run `yarn` first)

Also fix `make new adapter=...` that was incorrectly copying the example adapter and not adding to `package.json` workspaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ zip: deps build
 
 new:
 	mkdir $(adapter)
-	cp -r example/ $(adapter)
+	cp -r example/* $(adapter)
 	sed -i 's/example/$(adapter)/' $(adapter)/package.json
 	sed -i 's/Example/$(adapter)/' $(adapter)/README.md
 	sed -i 's/"workspaces": \[/"workspaces": \[\n    "$(adapter)",/' package.json
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ zip: deps build
 
 new:
 	mkdir $(adapter)
-	cp -r example/* $(adapter)
+	cp -rT example/ $(adapter)
 	sed -i 's/example/$(adapter)/' $(adapter)/package.json
 	sed -i 's/Example/$(adapter)/' $(adapter)/README.md
 	sed -i 's/"workspaces": \[/"workspaces": \[\n    "$(adapter)",/' package.json
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ zip: deps build
 
 new:
 	mkdir $(adapter)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ zip: deps build
 
 new:
 	mkdir $(adapter)
-	cp -r example/ $(adapter)
+	cp -r example/* $(adapter)
```
krebernisak, line 11: What is exactly wrong with this? It works on my machine (mac os). 

If I use `/*` then the cp command will not copy the `.eslinntrc.js` (hidden) file.

lcx/package.json
```diff
@@ -20,7 +20,7 @@
   "license": "MIT",
   "scripts": {
     "prepublishOnly": "yarn build && yarn test:unit",
-    "build": "tsc",
+    "build": "yarn && tsc",
```
krebernisak, line 23: Please take a look at [Clean adapter build process #135](https://github.com/smartcontractkit/external-adapters-js/pull/135).

Makefile
justinkaseman: On my machine (Ubuntu) it is copying the folder as well like: `[adapter name]/example/...` and not adding to `package.json`.

I found a way that should work for both of us, give it a try.

lcx/package.json
justinkaseman: Very nice, reverted this change.

Makefile
```diff
@@ -8,7 +8,7 @@ zip: deps build
 
 new:
 	mkdir $(adapter)
-	cp -r example/ $(adapter)
+	cp -rT example/ $(adapter)
```
krebernisak, line 11: This is what my macOS `cp` implementation states for `-r`: 

> Historic versions of the cp utility had a -r option.  This implementation supports that option; however, its use is strongly discouraged, as it does not correctly copy special files, symbolic links, or fifo's.

And looks like I do not have a `-T` option.

Maybe just `-R` would work?

justinkaseman: If you use `cp -R example/* $(adapter)` does that bring the `.eslintrc.js` with it? Looking at the macOS man page for `cp` I would guess it won't, but worth a try.

We could also add a line to just manually copy the `.eslintrc.js`

krebernisak: - `cp -R example/* $(adapter)` does NOT copy the `.eslintrc.js`
- `cp -R example/ $(adapter)` does copy the `.eslintrc.js`

krebernisak (approved):

PR #138, opened 2020-11-08
[Feature] Crypto volatility index adapter (by COTI)

yonine:
## Description

An external adapter for calculating the Crypto volatility index (CVX)


crypto-volatility-index/src/cryptoVolatilityIndex.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+let _isScheduled = false
+const scheduleCron = async () => {
+  if (_isScheduled) return
+
+  cron.schedule(`0 */${FREQ_MINUTES} * * * *`, async () => {
+    await calculate()
+  })
+}
```
boxhock, line 32: This seems unnecessary to me. On `execute()` we are calling `cryptoVolatilityIndex.calculate()` anyways. Is there a reason for running calculate() on a schedule too?

crypto-volatility-index/adapter.js
```diff
@@ -0,0 +1,34 @@
...
+  console.log('Bootstrapping CVX calculation')
+  cryptoVolatilityIndex.scheduleCron()
+}
+
+const execute = (input, callback) => {
+  const validator = new Validator(input)
+  if (validator.error) return callback(validator.error.statusCode, validator.error)
+
+  const jobRunID = validator.validated.id
+  console.log('jobRunID:', jobRunID)
```
boxhock, line 14: This seems unnecessary to log. The internal logger should log the payload regardless.

crypto-volatility-index/src/cryptoVolatilityIndex.js
```diff
@@ -0,0 +1,100 @@
+const { logger } = require('@chainlink/external-adapter')
+const Big = require('big.js')
```
boxhock, line 2: This package isn't included in `package.json`.

crypto-volatility-index/src/derivativesDataProvider.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const getOptionsData = async (currency, exchangeRate) => {
+  const config = {
+    url: bookDataEndpoint,
+    params: { currency, kind: 'option' },
+  }
+
+  try {
+    const response = await axios(config)
```
boxhock, line 50: Can we use `Requester.request()` here too?

crypto-volatility-index/test/adapter_test.js
```diff
@@ -0,0 +1,37 @@
...
+      })
+    })
+  })
+
+  context('successful calls @integration', () => {
+    const requests = [
+      {
+        name: 'id not supplied',
+        testData: { data: {} },
+      },
```
boxhock, line 23: Can we add more test cases to this?

crypto-volatility-index/package-lock.json
```diff
@@ -0,0 +1,5 @@
+{
```
boxhock, line 1: We use Yarn for this repo, so we shouldn't need this file.

crypto-volatility-index/adapter.js
```diff
@@ -0,0 +1,34 @@
...
+      const response = {
+        status: 200,
+        data: {
+          result,
+        },
+      }
+      callback(response.status, Requester.success(jobRunID, response))
+    })
+    .catch((error) => {
+      console.error(error)
```
boxhock, line 28: Don't think we need this log either because the internal logger will log this on errored requests, or we should at least use `logger` for this.

boxhock (changes requested):
Thanks for making this PR! Please see the comments that I left. I also see that the adapter is written in JS, but follows some of the structure we use for TS adapters.

We are working on moving more and more over to TS. Would you be able to rewrite this to TS, following the structure in the example adapter? It shouldn't be that big of a difference, just adding some types and replacing callbacks with async/await.

crypto-volatility-index/src/derivativesDataProvider.js
yonine: This api call pulls options data without granularity (deribit don't seem to support a filter). 
Since Requester.request logs the entire response data, it entails a significant amount of logging in this case.
The direct call to axios can be replaced with Requester.request() in case a "silent" call mode is added to the Requester, that would skip logging of the whole response

crypto-volatility-index/src/cryptoVolatilityIndex.js
yonine: the schedule was added in order to support averaging, after review - it isn't required and was removed

crypto-volatility-index/test/adapter_test.js
yonine: added test cases

yonine:
updated the adapter to TS as requested

crypto-volatility-index/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "crypto-volatility-index",
```
boxhock, line 2:
```suggestion
  "name": "@chainlink/crypto-volatility-index-adapter",
```

crypto-volatility-index/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "version": "0.1.0",
```
boxhock, line 3:
```suggestion
  "version": "0.0.1",
```

Just to match the versioning convention of our other adapters.

boxhock (changes requested):
There seems to be a few linter checks that are failing. We'll need that sorted before merging as well.

crypto-volatility-index/src/derivativesDataProvider.js
boxhock: In addition to the logging, the Requester will do a few extra things like retrying the request if the initial query failed. We have a few other calls we make that outputs unnecessary amounts of logging, so it's likely that the log level for this will be changed in Requester. I think it would be best if we stick with Requester.request() for now, even though it creates extra logging at the moment.

crypto-volatility-index/src/cryptoVolatilityIndex.ts
```diff
@@ -0,0 +1,54 @@
...
+      now,
+    })
+    return weightedSigma.sqrt().times(100)
+  })
+
+  return vixValues
+}
+
+const calculateWeighted = async (vixData: Array<Big>) => {
+  const marketCaps = await getMarketCaps(cryptoCurrencies)
```
boxhock, line 40: I was also made aware that we should probably be supporting multiple data providers here, but now it only seems to be supporting CMC. 

If that's the case, we should probably move this into the `composite` directory and follow the general framework of those adapters. An example for getting the data provider would be something like this: https://github.com/smartcontractkit/external-adapters-js/blob/master/composite/proof-of-reserves/src/balance.ts

boxhock (changes requested):

crypto-volatility-index/src/cryptoVolatilityIndex.ts
yonine: updated to the composite adapter format

crypto-volatility-index/src/derivativesDataProvider.js
yonine: updated to use Requester instead of the direct axios call

boxhock: Thanks. I just confirmed that we have a story to reduce the logging output of the Requester calls, so this should be taken care of.

composite/crypto-volatility-index/src/config.ts
```diff
@@ -0,0 +1,11 @@
+import { util } from '@chainlink/ea-bootstrap'
+
+export const CMC_API_KEY = 'CMC_API_KEY'
```
boxhock, line 3: This ENV var name seems to differ from the one in the README.

If the adapter will only use at most one data provider that requires an API key, can we just rename this `API_KEY` (instead of CMC specifically)?

boxhock (changes requested):

composite/crypto-volatility-index/src/on-chain-index/index.ts
```diff
@@ -0,0 +1,55 @@
...
+export const getIndex = async (oracleAddress: string): Promise<IndexData> => {
+  const rpcUrl = util.getRequiredEnv('RPC_URL')
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
+  const onChainOracle = new ethers.Contract(oracleAddress, ABI, provider)
+  const { answer: latestIndex, updatedAt } = await onChainOracle.latestRoundData()
+  return {
+    latestIndex,
+    updatedAt: moment.unix(updatedAt),
+  }
+}
```
boxhock, line 55: Can we move this functionality into our package here? https://github.com/smartcontractkit/external-adapters-js/tree/develop/helpers/reference-data-reader

Seems like it would make sense to have these in the same package.

boxhock (changes requested):

composite/crypto-volatility-index/src/config.ts
yonine: the adapter is using API_KEY, removed this config as it wasn't in use anymore

composite/crypto-volatility-index/src/on-chain-index/index.ts
yonine: moved to reference-data-reader

helpers/reference-data-reader/src/index.ts
```diff
@@ -27,3 +37,22 @@ export const getRpcLatestAnswer: ReferenceDataPrice = async (
...
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
+  const contract = new ethers.Contract(contractAddress, ABI, provider)
+  const result = await contract.latestRoundData()
+  result.answer = result.answer / multiply
+  result.startedAt = Number(result.startedAt)
+  result.updatedAt = Number(result.updatedAt)
+  result.answeredInRound = Number(result.answeredInRound)
+  result.roundId = Number(result.roundId)
+  return result
+}
```
boxhock, line 58: If you import the V2V3 interface, we can make it a bit better: `import { AggregatorV2V3InterfaceFactory } from '@chainlink/contracts/ethers/v0.6/AggregatorV2V3InterfaceFactory'`

We need to use BigNumber instead of Number as well, as the numbers we're dealing with are too big. This change should return BigNumber. Changes will have to be made in the consumer to reflect this change.

```suggestion
export const getRpcLatestRound: ReferenceDataRound = async (
  contractAddress: string,
): Promise<RoundData> => {
  const rpcUrl = util.getRequiredEnv('RPC_URL')
  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
  const aggregator = AggregatorV2V3InterfaceFactory.connect(contractAddress, provider)
  return await aggregator.latestRoundData()
}
```

composite/crypto-volatility-index/src/cryptoVolatilityIndex.ts
```diff
@@ -0,0 +1,113 @@
...
+  oracleAddress: string,
+  multiply: number,
+  heartBeatMinutes: number,
+): Promise<number> => {
+  const { answer: latestIndex, updatedAt } = await getRpcLatestRound(oracleAddress, multiply)
+  const indexUpdatedAt = moment(updatedAt * 1000)
+  if (!latestIndex || latestIndex <= 0) {
+    logger.warn('No on-chain index value found - Is first run of adapter?')
+    return cvi
+  }
```
boxhock, line 83: Can we try checking if the metadata was included in the request before making an RPC call? https://github.com/smartcontractkit/external-adapters-js#input

boxhock (changes requested):

helpers/reference-data-reader/src/index.ts
yonine: changes done

composite/crypto-volatility-index/src/adapter.ts
```diff
@@ -0,0 +1,29 @@
+import { Execute } from '@chainlink/types'
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { calculate } from './cryptoVolatilityIndex'
+
+const customParams = {
+  contract: ['contractAddress', 'contract'],
+  multiply: false,
+  heartbeatMinutes: false,
+  isAdaptive: false,
```
boxhock, line 9: `multiply` and `isAdaptive` should be documented in readme

composite/crypto-volatility-index/src/cryptoVolatilityIndex.ts
```diff
@@ -0,0 +1,119 @@
...
+const MAX_INDEX = 200
+const validateIndex = function (cvi: number, multiply: number) {
+  if (cvi <= 0 || cvi > MAX_INDEX * multiply) {
+    throw new Error('Invalid calculated index value')
+  }
+}
+
+const toOnChainValue = function (cvi: number, multiply: number) {
+  const trimmed = Number(cvi.toFixed(multiply.toString().length - 1)) // Keep decimal precision in same magnitude as multiply
+  return trimmed * multiply
```
boxhock, line 118: It would actually be better if we didn't multiply the value inside the adapter. In the job spec on the node we'll have a task that will multiply to the desired amount.

While it technically works as it is, it would be great to have this adapter be consistent with our other adapters.

boxhock (changes requested):
Was able to run this adapter successfully on a Kovan feed! Just a couple comments below.

typings/reference-data-reader/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'reference-data-reader' {
+  export interface RoundData {
+    roundId: BigNumber
+    answer: BigNumber
+    startedAt: BigNumber
+    updatedAt: BigNumber
+    answeredInRound: BigNumber
+  }
+}
```
boxhock, line 9: We should declare this in `helpers/reference-data-reader/src/index.ts`. Declaration here is only useful when we're dealing with packages that aren't in Typescript, but in this case we can declare it directly in the source.

helpers/reference-data-reader/src/index.ts
```diff
@@ -1,13 +1,21 @@
...
   contractAddress: string,
   multiply: number,
   meta?: Record<string, unknown>,
 ) => Promise<number>
 
+export type ReferenceDataRound = (
+  contractAddress: string,
+  multiply: number,
+  meta?: Record<string, unknown>,
+) => Promise<RoundData>
```
boxhock, line 17:
```suggestion
export type ReferenceDataRound = (
  contractAddress: string,
) => Promise<RoundData>
```

composite/crypto-volatility-index/src/cryptoVolatilityIndex.ts
```diff
@@ -0,0 +1,118 @@
...
+  return dominanceByCurrency(dominanceData)
+}
+
+const applySmoothing = async (
+  weightedCVI: number,
+  oracleAddress: string,
+  multiply: number,
+  heartBeatMinutes: number,
+): Promise<number> => {
+  const roundData = await getRpcLatestRound(oracleAddress, multiply)
```
boxhock, line 81:
```suggestion
  const roundData = await getRpcLatestRound(oracleAddress)
```

We also should be dividing the `latestIndex` below by the multiply amount.

boxhock (changes requested):
Just a few things that came to my attention when running it again.

composite/crypto-volatility-index/src/cryptoVolatilityIndex.ts
yonine: done

typings/reference-data-reader/index.d.ts
yonine: moved to src

boxhock (approved):

PR #141, opened 2020-11-09
Update dependencies

thodges-gh:
Hand-updated `@google-cloud/logging-winston` and `winston` to their latest package versions then ran `yarn upgrade` in the `external-adapter` directory.

### DIFF
--- a/external-adapter/package.json
+++ b/external-adapter/package.json
@@ -12,9 +12,9 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@google-cloud/logging-winston": "^3.0.6",
+    "@google-cloud/logging-winston": "^4.0.1",
     "axios": "^0.19.2",
-    "winston": "^3.2.1",
+    "winston": "^3.3.3",
     "uuid": "^8.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


external-adapter/package.json
```diff
@@ -12,11 +12,15 @@
...
     "axios": "^0.19.2",
-    "winston": "^3.2.1",
     "chai": "^4.2.0",
-    "uuid": "^8.3.0"
+    "eslint": "^7.2.0",
+    "eslint-config-prettier": "^6.11.0",
+    "eslint-plugin-standard": "^4.0.1",
+    "sinon": "^9.0.3",
+    "uuid": "^8.3.0",
+    "winston": "^3.3.3"
```
krebernisak, line 23:
```suggestion
    "uuid": "^8.3.0",
    "winston": "^3.3.3"
```

We already have those defined in the root `package.json`. Did `yarn upgrade` add them automatically?

thodges-gh: Yep, yarn added them. You think we should remove them?

krebernisak: Yeah, I don't think they should be declared here if they are declared in the workspace `package.json`. We should just upgrade them there, and every workspace package should have access to them.

se3000 (dismissed):

boxhock (approved):

boxhock (approved):

PR #142, opened 2020-11-10
Add Swipe(SXP) support to coingecko

tyrion70:
https://www.coingecko.com/en/coins/swipe
https://api.coingecko.com/api/v3/coins/list:
{
id: "swipe",
symbol: "sxp",
name: "Swipe"
},

### DIFF
--- a/coingecko/adapter.js
+++ b/coingecko/adapter.js
@@ -15,6 +15,7 @@ const presetTickers = {
   COMP: 'compound-governance-token',
   FNX: 'finnexus',
   UNI: 'uniswap',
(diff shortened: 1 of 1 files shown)


boxhock:
If I search for the symbol "SXP" in the Coingecko API, I only get a single result. Is it necessary to hard code it in?

tyrion70:
Hmm maybe it was an adapter/coingecko glitch then, I didn't get a response from the coingecko adapter (on both mine and the chainlink one earlier). Now it seems to work just fine

PR #143, opened 2020-11-10
IV outlier detection

boxhock:


iv-outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const customParams = {
+  symbol: ['base', 'from', 'coin', 'symbol'],
+  days: ['days', 'period'],
+}
+
+export const execute: Execute = async (input) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+  if (!input.meta || !input.meta.latestAnswer) throw new Error('missing meta data')
```
krebernisak, line 20: Where did the `meta` come from? What's the difference vs. `data`, when was it introduced and what do we use it for?

iv-outlier-detection/src/derbit.ts
```diff
@@ -0,0 +1,28 @@
...
+  const response = await Requester.request(config)
+  const values: number[][] = response.data['result']
+
+  const date = new Date()
+  date.setMinutes(0, 0, 0)
+  date.setDate(date.getDate() - days)
+
+  const result = values.find((elem) => elem[0] === date.getTime())
+  if (!result) {
+    console.log('no result')
```
krebernisak, line 21: If the log is needed, use our `logger` here.

iv-outlier-detection/src/derbit.ts
```diff
@@ -0,0 +1,28 @@
...
+    throw new Error('no derbit value for this date')
+  }
+
+  console.log(result)
```
krebernisak, line 25: Use our `logger` here.

iv-outlier-detection/src/derbit.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  return result[1]
```
krebernisak, line 27: How do we make sure the `result` holds at least 2 values here?

iv-outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,40 @@
...
+  const derbit = await fetchDerbit(symbol, days)
+
+  if (difference(result, derbit) > 30) {
+    throw new Error('value difference between Genesis Volatility and Derbit is more than 30%')
+  }
+
+  const onChainValue = input.meta.latestAnswer as number
+  if (onChainValue !== 0 && difference(result, onChainValue) > 50) {
+    throw new Error('value difference between Genesis Volatility and on-chain is more than 30%')
+  }
```
krebernisak, line 36: We should avoid making a derbit request if the on-chain value, that we can already check, is over threshold:

```suggestion
  const onChainValue = input.meta.latestAnswer as number
  if (onChainValue !== 0 && difference(result, onChainValue) > 50) {
    throw new Error('value difference between Genesis Volatility and on-chain is more than 50%')
  }
  
  const derbit = await fetchDerbit(symbol, days)
  if (difference(result, derbit) > 30) {
    throw new Error('value difference between Genesis Volatility and Derbit is more than 30%')
  }
```

krebernisak: Also, consider making these magic numbers configurable:
- DIFF_ON_CHAIN_THRESHOLD
- DIFF_DERBIT_THRESHOLD

iv-outlier-detection/src/genesisVolatility.ts
```diff
@@ -0,0 +1,48 @@
+import { Requester } from '@chainlink/external-adapter'
```
krebernisak, line 1: This code is duplicated. We should reuse the `genesisVolatility` adapter we already have.

iv-outlier-detection/src/derbit.ts
```diff
@@ -0,0 +1,28 @@
...
+export const fetchDerbit: ExternalFetch = async (
+  currency: string,
+  days: number,
+): Promise<number> => {
+  const url = 'https://www.deribit.com/api/v2/public/get_historical_volatility'
+  const params = { currency }
+  const config = { url, params }
+
+  const response = await Requester.request(config)
+  const values: number[][] = response.data['result']
```
krebernisak, line 13:
```suggestion
  const result: number[][] = response.data['result']
```

iv-outlier-detection/src/derbit.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  const date = new Date()
+  date.setMinutes(0, 0, 0)
+  date.setDate(date.getDate() - days)
+
+  const result = values.find((elem) => elem[0] === date.getTime())
+  if (!result) {
```
krebernisak, line 20:
```suggestion
  const resultForDate = result.find((elem) => elem[0] === date.getTime())
  if (!resultForDate) {
```

iv-outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const response = { data: { result }, result, status: 200 }
```
krebernisak, line 38: Looking at this, we should change `Requester.success` to append `statusCode: 200` as default.

iv-outlier-detection/README.md
```diff
@@ -0,0 +1,19 @@
+# Chainlink External Adapter for iv-outlier-detection
+
```
krebernisak, line 2: Consider adding more documentation on what this adapter actually does.

package.json
```diff
@@ -74,7 +74,8 @@
     "reduce",
     "json-rpc",
     "composite/**",
-    "xbto"
+    "xbto",
+    "iv-outlier-detection"
```
krebernisak, line 78: Please also add it to the CHANGELOG

krebernisak (changes requested):

iv-outlier-detection/src/adapter.ts
boxhock: The meta data (https://github.com/smartcontractkit/chainlink/pull/3376) provides some information about the on-chain reference contract that we can use, instead of having to specify an RPC_URL to poll (and also specify the contract address in the job spec).

In our case, we're interested in getting the latest answer from the reference contract - which we can now get from the meta data.

I'll add a short entry to the changelog about this.

iv-outlier-detection/src/derbit.ts
boxhock: Not needed, forgot to delete

iv-outlier-detection/src/derbit.ts
boxhock: I'll add a check

iv-outlier-detection/src/adapter.ts
krebernisak: Awesome! Looks super useful to have that extra metadata around.

I missed that, unfortunately. Need to keep an eye on that Chainlink CHANGELOG.

iv-outlier-detection/src/adapter.ts
krebernisak: No need to do this now, just thinking out loud.

iv-outlier-detection/src/genesisVolatility.ts
boxhock: It seems to me that this is the only outstanding item right now. Are we good to use this as-is and use the `genesisVolatility` adapter once we move this to a composite adapter, or should we publish that package and deal with that now?

krebernisak (approved):
Looking great! 🥇

PR #146, opened 2020-11-11
Add BTC.com adapter (PoR #3)

justinkaseman:
### Description
BTC.com adapter for querying proof or reserve for BTC. Also adds to Proof of Reserve composite adapter.

NOTE: This PR extends #150 

### Open Questions
Should we actually throw on [these warnings](https://github.com/smartcontractkit/external-adapters-js/pull/146/files#diff-eadca43659104776bad92c816efaf1b031f28398327ec414129b1e018c0e3387R32)?

### Ticket


btc.com/src/endpoint/balance.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const WARNING_NO_OPERATION = 'No Operation: only btc mainnet is supported by BTC.com adapter'
+const WARNING_NO_OPERATION_MISSING_ADDRESS = 'No Operation: address param is missing'
+
+const getBalanceURI = (address: string) => `/v3/address/${address}`
+
+const toBalances = async (config: Config, addresses: Address[]): Promise<ResponseWithResult[]> =>
+  Promise.all(
+    addresses.map(async (addr: Address) => {
+      if (!addr.address) return { ...addr, warning: WARNING_NO_OPERATION_MISSING_ADDRESS }
```
krebernisak, line 34: I think we should change this to an error as `address` should be required. I see this is also an issue in `blockchain.com` & `blockcypher` adapters. Please also throw an error there.

There is a difference between adapter does not support or can't process something (warning or error depending on `strict` flag) vs. adapter could not process it even if it supported that capability (error).

btc.com/src/endpoint/balance.ts
```diff
@@ -0,0 +1,94 @@
...
+  responses.reduce(
+    (accumulator, current) => {
+      accumulator.data.responses = [...accumulator.data.responses, current.response]
+      accumulator.data.result = [...accumulator.data.result, current.result]
+      return accumulator
+    },
+    {
+      data: { responses: [], result: [] },
+      status: 200,
+    } as any,
```
krebernisak, line 66: Mentioned this in the other PR. Please remove the status from here and add the type for `{ responses: [], result: [] }`.

I am thinking about how to reuse this kind of sequence processing, any ideas?

btc.com/src/endpoint/balance.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+
+  const data: RequestData = validator.validated.data
+  const dataPath = data.dataPath || DEFAULT_DATA_PATH
+  const inputData = <Address[]>objectPath.get(request.data, dataPath)
+
+  // Check if input data is valid
+  if (!inputData || !Array.isArray(inputData) || inputData.length === 0)
+    throw Requester.errored(
```
krebernisak, line 86: Don't throw `Requester.errored`, but `Error` or `AdapterError`.

composite/proof-of-reserves/src/balance.ts
```diff
@@ -45,6 +47,12 @@ export const getImpl = (options: BitcoinIndexerOptions): Execute => {
         const config = blockchair.getConfig(prefix)
         return blockchair.execute(data, config)
       }
+
+    case BitcoinIndexer.BTCCom:
```
krebernisak, line 51: We now have 5 adapters here, all the same, and this switch/case is slowly outgrowing itself...

Let's add all adapters to a map, then find the one requested from a map.

```javascript
const adapter = ... // find it
return (data) => {
  const config = adapter.getConfig(prefix)
  return adapter.execute(data, config)
}
```

krebernisak (changes requested):

btc.com/src/endpoint/balance.ts
justinkaseman: I'm thinking of adding a method to Requester called `warnOrThrow` that will handle the strict flag. 

Any ideas on a better implementation?

PR #147, opened 2020-11-12
Refactor Blockchair adapter (PoR #1)

justinkaseman:
### Description
Refactors Blockchair adapter in TypeScript with a second `endpoint` for querying proof of reserve.

### Changes from SoChain's design
- I noticed in other PoR adapters that the request's original response data gets lost, which is inconsistent with other external adapters. This PR amends this by returning the whole response from each endpoint's helper function.

- This PR adds a new type for `ExecuteWithDefaults` to guard against using execute without a config

### Open Questions
The currently functionality of this adapter is to still return successfully if an address is given that doesn't exist, but add a warning and leave off the balance property of the `response` for that address. Is this how we want it to function? Or should it completely error out when an invalid address is hit? 

### Ticket


typings/@chainlink/index.d.ts
```diff
@@ -14,7 +14,11 @@ declare module '@chainlink/types' {
     data: AdapterResponse
   }
 
-  export type Execute = (input: AdapterRequest, config?: any) => Promise<AdapterResponse>
+  export type Execute = (input: AdapterRequest, config: any) => Promise<AdapterResponse>
+  export type ExecuteWithDefaults = (
+    input: AdapterRequest,
+    config?: any,
+  ) => Promise<AdapterResponse>
```
krebernisak, line 21: Just thinking out loud for now, but I think we should move more in this direction:

```suggestion
  export type Execute = (input: AdapterRequest) => Promise<AdapterResponse>
  export type ExecuteWithConfig = (input: AdapterRequest, config: any) => Promise<AdapterResponse>
  export type ExecuteFactory = (config?: any) => Execute
```

Then we would build:

```javascript
const getDefaults = (): Config => getConfig()
const makeExecute: ExecuteFactory = (config) => async (request) => execute(request, config)
const execute: Execute = makeExecute(getDefaults())
```

blockchair/test/balance.test.ts
```diff
@@ -0,0 +1,229 @@
+import { assert } from 'chai'
```
krebernisak, line 1: These balance tests look like they could and should be reused for any "balance supported" adapter. 
Please make a specification and let's reuse it for all PoR adapters. The only thing to think about is where to put it 🧐

krebernisak (changes requested):

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,147 @@
...
+  return `/${coin}/addresses/balances?addresses=${addresses.join(',')}`
+}
+
+const getBalances = async (
+  jobRunID: string,
+  config: Config,
+  groupedData: GroupedAddresses,
+): Promise<ResponseWithResult[]> =>
+  Promise.all(
+    Object.entries(groupedData).map(async ([network, reqData]) => {
```
krebernisak, line 87: It looks like we are using Blockchairs batch endpoint, so `getBalanceURI(jobRunID, reqData.addresses, chain, coin)` will return an endpoint that will fetch us balances for all addresses. Why do we do it N times then?

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+type AddressGroup = {
+  addresses: string[]
+  result: Address[]
+}
+
+type GroupedAddresses = {
+  [network: string]: AddressGroup
+}
```
krebernisak, line 24: It's hard for me to understand what are we grouping and why do we need these extra types at this point.

blockchair/src/endpoint/balance.ts
krebernisak: Aha, do we group addresses for different chains for example `bitcoin/testnet` & `bitcoin/mainet` and then we execute M requests, where M is the number of different chains?

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,147 @@
...
+      addr.coin = 'btc'
+    }
+    if (!addr?.chain) {
+      addr.chain = 'mainnet'
+    }
+    const { address, coin, chain } = addr
+    if (output[`${coin}-${chain}`]) {
+      output[`${coin}-${chain}`].addresses.push(address)
+      output[`${coin}-${chain}`].result.push(addr)
+    } else output[`${coin}-${chain}`] = { addresses: [address], result: [addr] }
```
krebernisak, line 66: This can be made more readable like this:
```suggestion
    const key = `${coin}-${chain}`
    output[key] = output[key] ||  { addresses: [], result: [] }
    output[key].addresses.push(address)
    output[key].result.push(addr)
```

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,147 @@
...
+        url: getBalanceURI(jobRunID, reqData.addresses, chain, coin),
+      }
+
+      const response = await Requester.request(reqConfig)
+
+      result = result.map((r) => {
+        r.balance = response.data.data[r.address]
+        if (!r.balance) r.warning = WARNING_NO_OPERATION_ADDRESS_NOT_FOUND
+        return r
+      })
```
krebernisak, line 103:
```suggestion
      const toResultWithBalance = (r) => {
        r.balance = response.data.data[r.address]
        if (!r.balance) r.warning = WARNING_NO_OPERATION_ADDRESS_NOT_FOUND
        return r
      }
      result = result.map(toResultWithBalance)
```

krebernisak (changes requested):

blockchair/src/endpoint/balance.ts
justinkaseman: Yep exactly. Blockchair has an [endpoint](https://blockchair.com/api/docs#link_390) for getting multiple addresses from a certain chain. In this case it is **extraordinarily** cheaper to use.

PR #149, opened 2020-11-13
Refactor CryptoAPIs adapter (PoR #5)

justinkaseman:
### Description
Refactors CryptoAPIs adapter in TypeScript with a third `endpoint` for querying proof of reserve.

NOTE: This PR extends #145 .

### Caveats
The tests against mainnet are failing because our API_KEY needs to be upgraded

### Ticket
https://www.pivotaltracker.com/story/show/175540076

PR #150, opened 2020-11-13
Refactor Amberdata adapter (PoR #2)

justinkaseman:
### Description
Refactors Amberdata adapter in TypeScript with a second `endpoint` for querying proof of reserve.

NOTE: This PR extends #149 

### Ticket


amberdata/README.md
```diff
@@ -11,20 +24,115 @@ Gets the [latest spot VWAP price](https://docs.amberdata.io/reference#spot-price
...
+  "id": "1",
+  "data": {
+    "addresses": [
+      {
+        "address": "3EyjZ6CtEZEKyc719NZMyWaJpJG5jsVJL1"
+      },
+      {
+        "address": "38bzm6nhQMFJe71jJw1U7CbgNrVNpkonZF"
+      },
+      {
```
krebernisak, line 71: Let's make the input example also with two addresses if the output example is for two addresses.

amberdata/src/config.ts
```diff
@@ -0,0 +1,44 @@
...
+export type Config = {
+  apiKey?: string
+  api: Record<string, unknown>
+}
+
+export const getBaseURL = (): string => ENDPOINT_MAIN
+
+export const getConfig = (prefix = ''): Config => ({
+  apiKey: util.getEnv(ENV_API_KEY, prefix),
+  api: {
```
krebernisak, line 23: Please reuse this config as I mentioned in the other PR.

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,122 @@
+import objectPath from 'object-path'
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { AdapterRequest } from '@chainlink/types'
+import { Config, DEFAULT_DATA_PATH, getBaseURL } from '../config'
+import { CoinType, ChainType, isCoinType, isChainType } from '.'
+
+export const Name = 'balance'
+
+type Address = {
```
krebernisak, line 9: Please reuse this type as I mentioned in the other PR.

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,122 @@
...
+  "No Operation: invalid chain parameter, must be one of 'mainnet' or 'testnet'"
+const WARNING_NO_OPERATION_MISSING_ADDRESS = 'No Operation: address param is missing'
+
+const getBalanceURI = (address: string) => `/api/v2/addresses/${address}/account-balances/latest`
+
+const getBlockchainHeader = (chain: string, coin: string) => {
+  const network = blockchains[coin]
+  if (chain === 'testnet') {
+    if (network === 'ethereum') return 'ethereum-rinkeby'
+    throw WARNING_NO_OPERATION_TESTNET
```
krebernisak, line 48: This is not a warning but an error. We used warnings when processing sequences by multiple adapters, so if one adapter does not know how to process an item in the sequence, it would add a warning and pass it along.

There is an idea to add a `strict` flag, so these sequence processors would error as soon as they encounter a piece of data/request they can not process.

amberdata/test/balance.test.ts
```diff
@@ -0,0 +1,230 @@
+import { assert } from 'chai'
```
krebernisak, line 1: Let's try to reuse these tests!

The idea is to define a specific behavior and reuse it in the individual tests.
For inspiration take a look at how [OpenZepplin structures its tests and behaviors on the ERC20 contract example.](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/release-v3.2.0/test/token/ERC20)

amberdata/src/endpoint/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export const COIN_KEYS = ['btc', 'eth', 'bch', 'ltc', 'btsv', 'zec'] as const
+export type CoinType = typeof COIN_KEYS[number]
+export function isCoinType(key: string): key is CoinType {
+  return COIN_KEYS.includes(key as CoinType)
+}
+export const CHAIN_KEYS = ['mainnet', 'testnet'] as const
+export type ChainType = typeof CHAIN_KEYS[number]
+export function isChainType(key: string): key is ChainType {
+  return CHAIN_KEYS.includes(key as ChainType)
+}
```
krebernisak, line 13: Let's try to reuse this and add it to `@chainlink/types`

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+
+  const data: RequestData = validator.validated.data
+  const dataPath = data.dataPath || DEFAULT_DATA_PATH
+  const inputData = <Address[]>objectPath.get(request.data, dataPath)
+
+  // Check if input data is valid
+  if (!inputData || !Array.isArray(inputData) || inputData.length === 0)
+    throw Requester.errored(
```
krebernisak, line 114: Don't throw `Requester.errored`, but `Error` or `AdapterError`.

amberdata/src/endpoint/balance.ts
krebernisak: Only Etherum has testnet support?

amberdata/src/adapter.ts
```diff
@@ -0,0 +1,40 @@
...
+    case price.Name: {
+      response = await price.execute(config, request)
+      break
+    }
+    case balance.Name: {
+      response = await balance.execute(config, request)
+      break
+    }
+    default: {
+      throw Requester.errored(jobRunID, `Endpoint ${endpoint} not supported.`, 400)
```
krebernisak, line 31: Don't throw `Requester.errored`, but `Error` or `AdapterError`.

krebernisak (changes requested):

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,122 @@
...
+      reqConfig.headers['x-amberdata-blockchain-id'] = getBlockchainHeader(addr.chain, addr.coin)
+
+      try {
+        const response = await Requester.request(reqConfig)
+        return {
+          response: response.data,
+          result: { ...addr, balance: response.data.payload.value },
+        }
+      } catch (error) {
+        return error
```
krebernisak, line 79: Why do you catch and return an error here? We need the error to propagate and stop the adapter.

amberdata/src/endpoint/index.ts
justinkaseman: I don't think that will work in the way that I am using the type guards here because each adapter might support different coins or chains. We can't generate these dynamically with TypeScript as far as I know.

amberdata/src/endpoint/balance.ts
justinkaseman: > Only Etherum has testnet support?

Yeah, weirdly.

justinkaseman: > This is not a warning but an error. We used warnings when processing sequences by multiple adapters, so if one adapter does not know how to process an item in the sequence, it would add a warning and pass it along.
> 
> There is an idea to add a `strict` flag, so these sequence processors would error as soon as they encounter a piece of data/request they can not process.

We should sync about this concept further!

amberdata/src/endpoint/index.ts
krebernisak: Yeah, I know, but my thinking was that we should take over those specific types and make more general types for our "framework". We now have a specific `CoinType` for every adapter that supports some kind of coins, but we don't have a `CoinType` on the framework level, which I think would be useful. Maybe we need to type both the per adapter specific types and the general ones. 🤔 

Let's leave it like you defined it for now, and think about it some more about how to handle it in the future.

PR #153, opened 2020-11-18
Defi Pulse Index Adapter

RodrigoAD:
The adapter combines real-time DPI allocation data from on-chain with off-chain price sources in
order to calculate an accurate price for the DPI.

[Readme Preview](https://github.com/smartcontractkit/external-adapters-js/blob/915df5d0c2017a6bab7619a6655a96d18c0eaa00/defi-pulse/README.md)


defi-pulse/src/priceAdapter.ts
```diff
@@ -0,0 +1,50 @@
...
+  for (const symbol in prices) {
+    pricesMap.set(symbol.toUpperCase(), prices[symbol])
+  }
+
+  for (const i of index) {
+    i.priceData = pricesMap.get(i.asset.toUpperCase())
+  }
+
+  return index
+}
```
thodges-gh, line 50: This file should be in `defi-pulse/cryptocompare/src/priceAdapter.ts`. And maybe we can re-use the priceAdapter functions from `synth-index/cryptocompare`.

thodges-gh (changes requested):

defi-pulse/src/index-allocations/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  'function getComponents() external view returns (address[] memory)',
+  'function getUnits() external view returns (uint256[] memory)',
+]
+
+type Allocations = {
+  components: string[]
+  units: number[]
+}
+
+export const getAllocations = async (setAddress: string): Promise<Allocations> => {
```
krebernisak, line 14: Maybe we can use the typed [setProtocol.js](https://github.com/SetProtocol/setprotocol.js/blob/a6f8a181c7229b62f602f2e78bdb3ccdaf55b1e0/src/api/SetTokenAPI.ts) lib directly for this.

defi-pulse/src/priceAdapter.ts
```diff
@@ -0,0 +1,50 @@
+import { AssetIndex } from './adapter'
+import { Requester } from '@chainlink/external-adapter'
+import Decimal from 'decimal.js'
+
+const getPriceData = async (symbols: string) => {
```
krebernisak, line 5: Use cryptocompare adapter for this!

krebernisak (changes requested):

defi-pulse/src/data-providers/coinapi.ts
```diff
@@ -0,0 +1,42 @@
...
+  let value = new Decimal(0)
+  index.forEach((i) => {
+    const price = i.priceData && i.priceData.rate
+    if (price <= 0) {
+      throw Error('invalid price')
+    }
+    value = value.plus(new Decimal(i.units).times(new Decimal(price)))
+  })
+  return value.toNumber()
+}
```
krebernisak, line 28: Rename and extract this method to make it reusable:

```suggestion
const calculateIndexValue = (index: Index, toAssetPrice: (IndexAsset) => number): number =>
  index
    .map(asset => { ...asset, price: toAssetPrice(asset) }) 
    .reduce((acc, val) => acc.plus(val.units).times(val.price)), new Decimal(0))
    .toNumber()
}
```

`toAssetPrice` mapper should get the price, and throw if invalid. But why the price is not already added to the asset at this point? I think this should be done as soon as we fetch the price data.

krebernisak (changes requested):

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,88 @@
...
+    const priceIndex = await adapter.getPriceIndex(index)
+    asset.index = priceIndex
+    const indexResult = adapter.calculateIndex(priceIndex)
+
+    const response = {
+      status: 200,
+      data: { result: indexResult, ...asset },
+    }
+    return Requester.success(jobRunID, response)
+  } catch (e) {
```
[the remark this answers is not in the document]
RodrigoAD: Yep my mistake, it was only for debugging reasons. Will fix it

defi-pulse/src/data-providers/coinapi.ts
RodrigoAD: You are completely right. I'll get into it

defi-pulse/src/symbols/index.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function memoizeDirectory() {
+  let directory = getDirectory()
+  return () => {
+    return { directory, update: (): Directory => (directory = getDirectory()) }
+  }
+}
+
+const cachedDirectory = memoizeDirectory()
```
krebernisak, line 65: There are a few issues here:

- The caching implementation should not be specific to one use case or to one adapter.
- Most of these adapters are run in a serverless environment, where the filesystem is very volatile and short-living as are the serverless functions themselves. This means that you can cache work during the request, but not cross requests as the env will probably be torn down by the underlying execution scheduler.
- We already have 2x cache implementation that sits in front of all adapters (local in-memory & Redis). They cache adapter `input => output`. If we would need additional inter-request caching to speed up some work (like this example), then we should write some code to expose that caching service as an API to the adapter for use. This would be the right path forward, but it's not easy and requires some refactoring.
- The implementation should probably take into consideration multiple networks (testnet vs mainnet vs...)

My suggestion would be to KISS here. It's ok to ship this adapter with a DB of symbols, but I would use it for reading not for writing to it because of the ephemeral environment issue. Just read it to the local in-memory cache when the adapter starts, and use that as a lookup table for symbols. You can also write to that local in-memory cache if it makes sense to avoid duplicate calls to Ethereum nodes during one request but it probably does not make sense (because a specific Index has one asset repeating only once) so just read.

krebernisak (changes requested):

defi-pulse/src/index-allocations/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+type Allocations = {
+  components: string[]
+  units: number[]
+}
+
+const ABI = [
+  {
+    inputs: [{ internalType: 'contract ISetToken', name: '_setToken', type: 'address' }],
+    name: 'getAllocations',
```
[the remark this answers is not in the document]
RodrigoAD: I had to take it directly from the contract they deployed for this: https://etherscan.io/address/0x78733fa5e70e3ab61dc49d93921b289e4b667093#code

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const { components, units } = await getAllocations(asset.address)
+
+  const index = await getIndex(components, units)
+
+  const priceIndex = await priceAdapter(index)
+  asset.index = priceIndex
+  const indexResult = calculateIndex(priceIndex)
+  const response = {
+    status: 200,
```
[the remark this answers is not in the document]
RodrigoAD: It doesn't apparently: https://github.com/smartcontractkit/external-adapters-js/blob/4b37a6d82477a5b84244fe62fd8858828fa5ede7/external-adapter/src/requester.js#L86

defi-pulse/src/data-providers/coinpaprika.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  return value.toNumber()
+}
+
+const getPriceIndex = async (index: IndexAsset[]): Promise<IndexAsset[]> => {
+  const priceDatas = await getPriceData()
+  await Promise.all(
+    index.map(async (i) => {
+      i.priceData = priceDatas
+        .sort((a: any, b: any) => (a.rank > b.rank ? 1 : -1))
```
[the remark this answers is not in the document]
RodrigoAD: Apparently the tokens are repeated, and we should care about the one with the lowest rank. I've refactored this function though, was not really efficient.

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+      units: new Decimal(new utils.BigNumber(units[i]).toString()).div(1e18),
+      weight: 0,
+      price: 0,
+    }
+    index.push(indexAsset)
+  }
+  return index
+}
+
+const calculateIndex = (index: Index): number =>
```
krebernisak, line 32: We are calculating the total value here:

```suggestion
const calculateIndexValue = (index: Index): number =>
```

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+export type Index = IndexAsset[]
+
+export type IndexAsset = {
+  asset: string
+  units: Decimal
+  weight: number
+  price: number
+}
+
+async function getIndex(components: string[], units: number[]): Promise<Index> {
```
krebernisak, line 18: We are making an index here:

```suggestion
async function makeIndex(components: string[], units: number[]): Promise<Index> {
```

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+import { getAllocations } from './index-allocations'
+import { getPriceAdapter } from './priceAdapter'
+import Decimal from 'decimal.js'
+
+export type Index = IndexAsset[]
+
+export type IndexAsset = {
+  asset: string
+  units: Decimal
+  weight: number
```
krebernisak, line 14: What is `weight` used for? It's always 0 as far as I can tell.

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+  price: number
```
krebernisak, line 15: Price should be `undefined`, not set to 0 as default.
```suggestion
  price?: number
```

defi-pulse/src/adapter.ts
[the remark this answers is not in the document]
krebernisak: Ok, leave it as is, we'll tackle this another time.

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const jobRunID = validator.validated.id
+  const asset = validator.validated.data
+
+  const { components, units } = await getAllocations(asset.adapter, asset.address)
+
+  const index = await getIndex(components, units)
+
+  const priceIndex = await adapter.getPriceIndex(index)
+  asset.index = priceIndex
```
krebernisak, line 64: Don't modify, but keep data immutable!
```suggestion
```

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+  const indexResult = calculateIndex(priceIndex)
+
+  const response = {
+    status: 200,
+    data: { result: indexResult, ...asset },
```
krebernisak, line 69: Isn't this more readable?

```suggestion
    data: { ...input.data, result: totalValue, index: pricedIndex },
```

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const asset = validator.validated.data
+
+  const { components, units } = await getAllocations(asset.adapter, asset.address)
+
+  const index = await getIndex(components, units)
+
+  const priceIndex = await adapter.getPriceIndex(index)
```
krebernisak, line 63: I don't like that we hardcode `USD` everywhere. We should support to price this in any currency, so let's make interface support that even if the underlying implementation does not. The implementation can just `throw Error('Only USD supported')`.

```suggestion
  const pricedIndex = await adapter.priceIndex(index, 'USD')
```

krebernisak: Also, we don't want anybody reading this to need to go 3 layers deep to find it's priced in USD.

defi-pulse/src/data-providers/coinpaprika.ts
```diff
@@ -0,0 +1,45 @@
...
+      priceMap.set(key, price)
+    }
+  }
+
+  for (const i of index) {
+    const data = priceMap.get(i.asset.toUpperCase())
+    i.price = toAssetPrice(data)
+  }
+
+  return index
```
krebernisak, line 42: Always try to keep data immutable, it's going to make your life easier in the long run :)

```suggestion
  return index.map((i) => {
    const data = priceMap.get(i.asset.toUpperCase())
    return { ...i, price: toAssetPrice(data) }
  }
```

defi-pulse/src/data-providers/cryptocompare.ts
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error('Invalid price')
+  }
+  return price
+}
+
+const getPriceIndex = async (index: Index): Promise<Index> => {
+  const symbols: string[] = []
+  index.forEach(({ asset }) => {
+    symbols.push(asset.toUpperCase())
+  })
```
krebernisak, line 32: Again, defining data as a `const`! this time, and them mutating that data. This is an antipattern. Always while manipulating data think about transforming data from one form into another, not mutating the source to fit the destination.

```suggestion
  const symbols: string = index.map(({asset})  => (asset)).join().toUpperCase()
```

defi-pulse/src/data-providers/cryptocompare.ts
```diff
@@ -0,0 +1,48 @@
...
+  const prices = await getPriceData(symbols.join())
```
krebernisak, line 33:
```suggestion
  const prices = await getPriceData(symbols)
```

defi-pulse/src/data-providers/cryptocompare.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const pricesMap = new Map()
+  for (const symbol in prices) {
+    pricesMap.set(symbol.toUpperCase(), prices[symbol])
+  }
```
krebernisak, line 38: This seems unnecessary, why do we do this? Looks like `prices` is already in the format we want it in.

defi-pulse/src/data-providers/kaiko.ts
```diff
@@ -0,0 +1,45 @@
...
+const getPriceIndex = async (index: Index): Promise<Index> => {
+  await Promise.all(
+    index.map(async (i) => {
+      // Particular for the Kaiko API only
+      const asset = i.asset === 'UNI' ? 'uniswap' : i.asset
+      const data = await getPriceData(asset)
+      i.price = toAssetPrice(data)
+    }),
+  )
+  return index
```
krebernisak, line 42: There is a big difference between these two snippets. Keep data immutable ;)

```suggestion
  return await Promise.all(
    index.map(async (i) => {
      // Particular for the Kaiko API only
      const asset = i.asset === 'UNI' ? 'uniswap' : i.asset
      const data = await getPriceData(asset)
      return { ...i, price:  toAssetPrice(data) }
    }),
  )
```

defi-pulse/src/symbols/directory.json
```diff
@@ -0,0 +1,13 @@
+{
```
krebernisak, line 1: We need to support multiple different networks here. For example, let's say we want to support this functionality on BSC & Matic networks. Rename it as `directory.mainnet.json` and add a hardcoded `network = 'mainnet'` param to the top-level file. Probably something like this:

```javascript
const index = await makeIndex(components, units, 'mainnet')
```

defi-pulse/src/index-allocations/index.ts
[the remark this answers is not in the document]
krebernisak: This is important information! Please add a big `Notice: ` comment here and also mention it in the README (link to this file).

The current implementation uses a utility contract https://etherscan.io/address/0x78733fa5e70e3ab61dc49d93921b289e4b667093#code to fetch data from SetToken contracts. So, if we want to go to another network (Matic, BSC, or event a test network) this adapter will not work unless we deploy a copy of that contract.

This is a big limitation while providing a negligible utility. TODO: we should use SetProtocol.js typed library instead to fetch data directly from the SetToken contract directly. The `ChainlinkAdapter.getAllocations(ISetToken _setToken)` should be reimplemented in JS.

Please don't quote me word for word, but add a substantial comment explaining where is this coming from and what are the consequences.

krebernisak (changes requested):

defi-pulse/src/priceAdapter.ts
```diff
@@ -1,45 +1,46 @@
...
+const providers: Record<string, PriceAdapter> = {
+  [DataProvider.Amberdata]: amberdata,
+  [DataProvider.Bravenewcoin]: bravenewcoin,
+  [DataProvider.Cryptocompare]: cryptocompare,
+  [DataProvider.Coinpaprika]: coinpaprika,
+  [DataProvider.Nomics]: nomics,
+  [DataProvider.Coinmarketcap]: coinmarketcap,
+  [DataProvider.Coingecko]: coingecko,
+  [DataProvider.Coinapi]: coinapi,
+  [DataProvider.Kaiko]: kaiko,
```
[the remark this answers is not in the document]
krebernisak: @thodges-gh, @boxhock & @justinkaseman just tagging you here so we can all acknowledge that we need to streamline our adapter API so they are usable in a composite sense and avoid all of this C/P. No action needed, we are slowly going in that direction, just keep it in mind while working on this repo. Hopefully, soon we can remove all this duplicated code, and make adapters more fun to use.

defi-pulse/src/adapter.ts
RodrigoAD: I know, but it was asked like that in the task specs

defi-pulse/src/priceAdapter.ts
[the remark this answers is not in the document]
boxhock: > make adapters more fun to use

I support that 😂

defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,71 @@
...
+  )
+}
+
+const calculateIndexValue = (index: Index): number =>
+  index
+    .reduce((acc, { units, price }) => {
+      if (!price) throw new Error('Invalid price')
+      return acc.plus(units.times(price))
+    }, new Decimal(0))
+    .toNumber()
```
krebernisak, line 36: Maybe we can split this into an assert and compute parts:

```suggestion
const calculateIndexValue = (index: Index): number => {
  // assert all prices are set
  const isPriceSet = (i: IndexAsset) => (i.price && i.price > 0)
  if (!index.every(isPriceSet)) throw new Error('Invalid index: price not set')
  // calculate total value
  return index
    .reduce((acc, { units, price }) => acc.plus(units.times(price))}, new Decimal(0))
    .toNumber()
}
```

defi-pulse/src/data-providers/coinpaprika.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (!price || price <= 0) {
+    throw new Error('invalid price')
+  }
+  return price
+}
+
+const getPriceIndex = async (index: Index, currency: string): Promise<Index> => {
+  const priceData = await getPriceData(currency)
+
+  const sortedData = priceData.sort((a: any, b: any) => a.rank - b.rank)
```
krebernisak, line 28: Maybe just add a quick comment here that we remove duplicates after sorting by rank.

defi-pulse/src/data-providers/nomics.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const getPriceIndex: GetPriceIndex = async (index, currency) => {
+  const symbols = index
+    .map(({ asset }) => {
+      const symbol = asset.toUpperCase()
+      if (nomicsIds[symbol]) {
+        return nomicsIds[symbol]
+      }
+      return symbol
```
krebernisak, line 39:
```suggestion
      return nomicsIds[symbol] || symbol
```

defi-pulse/src/index-allocations/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    stateMutability: 'view',
+    type: 'function',
+  },
+]
+
+export const getAllocations = async (
+  contractAddress: string,
+  setAddress: string,
+): Promise<Allocations> => {
+  const rpcUrl = util.getRequiredEnv('RPC_URL')
```
krebernisak, line 36: We should ingest this as a parameter so the config location is not coupled to this function. Then on the top lvl when we call this function we can set the `rpcUrl`:

```javascript
const rpcUrl = util.getRequiredEnv('RPC_URL')
const { components, units } = await getAllocations(asset.adapter, asset.address, rpcUrl)
```

defi-pulse/src/priceAdapter.ts
```diff
@@ -0,0 +1,44 @@
...
+export type GetPriceIndex = (index: Index, currency: string) => Promise<Index>
+
+export type PriceAdapter = {
+  getPriceIndex: GetPriceIndex
+}
+
+export const getPriceAdapter = (): PriceAdapter => {
+  const dataProvider = util.getRequiredEnv('DATA_PROVIDER')
+  return providers[dataProvider]
+}
```
krebernisak, line 44: Same here, we should inject this config as a fn param. It will make the code much more understandable top to bottom:

```suggestion
export const getPriceAdapter = (dataProvider: string): PriceAdapter => {
  return providers[dataProvider]
}
```

defi-pulse/src/symbols/index.ts
```diff
@@ -0,0 +1,57 @@
+import fs from 'fs'
+import { join } from 'path'
+import { ethers, utils } from 'ethers'
+import { util } from '@chainlink/ea-bootstrap'
+import { logger } from '@chainlink/external-adapter'
+
+const rpcUrl = util.getRequiredEnv('RPC_URL')
```
krebernisak, line 7: We have the same problem here, this config is buried too deep and leaves us little options in the future to change it for something else if the situation requires. This should be a function param passed from the top.

defi-pulse/src/data-providers/kaiko.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Index } from '../adapter'
+import { util } from '@chainlink/ea-bootstrap'
+
+const getPriceData = async (symbol: string, currency: string) => {
+  const url = `https://us.market-api.kaiko.io/v2/data/trades.v1/spot_exchange_rate/${symbol.toLowerCase()}/${currency.toLowerCase()}`
+  const params = {
+    sort: 'desc',
+  }
+  const headers = {
+    'X-Api-Key': util.getRequiredEnv('API_KEY'),
```
krebernisak, line 11: Yes, these `API_KEY`s are also too deep and give us little flexibility, but just leave this as is because we plan to refactor this and use the adapters directly in the future.

krebernisak (changes requested):
Thanks for cleaning up this code, it's looking a lot better! 💯 
Just a few more small changes and it should be ready to go.

defi-pulse/src/adapter.ts
RodrigoAD: I like that more, but he still complains that he might receive an undefined price

krebernisak: I've updated the comment. You should be able to force it with a Non-Null Assertion Operator `!`.

composite/defi-pulse/src/adapter.ts
```diff
@@ -0,0 +1,77 @@
...
+    }),
+  )
+}
+
+const calculateIndexValue = (index: Index): number => {
+  // assert all prices are set
+  const isPriceSet = (i: IndexAsset) => i.price && i.price > 0
+  if (!index.every(isPriceSet)) throw new Error('Invalid index: price not set')
+  // calculate total value
+  return index.reduce((acc, i) => acc.plus(i.units.times(i!.price)), new Decimal(0)).toNumber()
```
krebernisak, line 39:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  return index.reduce((acc, i) => acc.plus(i.units.times(i.price!)), new Decimal(0)).toNumber()
```

krebernisak (approved):

thodges-gh (approved):

PR #156, opened 2020-11-20
[Feature] @chainlink/external-adapter TS

justinkaseman:
### Description
- Migrates `@chainlink/external-adapter` to TypeScript
- Migrates `@chainlink/ea-bootstrap` to TypeScript
- Adds additional types to `@chainlink/types` for `Config`, `Execute` types, and `Account` (used in the balance endpoint)
- Moves config functionality to `@chainlink/external-adapter`'s Requester class for re-use across all adapters
- Refactors `blockcypher`, `wbtc-address-set`, `blockchain.com`, `renvm-address-set` adapter to use the above

### Caveats
Additional changes such as removing all synchronous adapters and refactoring bootstrap's middleware will be handled in a separate PR.


justinkaseman:
@krebernisak This is ready for review to see if this is on the right path with all the types. Might be some stuff I misunderstood.

Tests aren't currently passing because the following adapters need to be moved toward [this pattern](https://github.com/smartcontractkit/external-adapters-js/pull/147#discussion_r524543739)
- [ ] wbtc-address-set 
- [ ] blockchain.com
- [ ] renvm-address-set

Take a look at `blockcypher` adapter and see if this is how we want our structure to be.

bootstrap/test/cache.test.ts
```diff
@@ -1,19 +1,27 @@
...
   while (n--) {
     const { data } = await fn(0)
-    if (n === 0) expect(data).to.equal(result)
+    if (n === 0) expect(data.result).to.equal(result)
   }
 }
 
 // Helper test function: a stateful counter
-const counterFrom = (i = 0) => () => ({ statusCode: 200, data: i++ })
+const counterFrom = (i = 0): ExecuteWrappedResponse => async (request) => {
```
krebernisak, line 16: Could we revert this part? We are just testing the cache caching the fn input ->  output, there is no need to couple/mix our adapters here or their complicated types 😅

external-adapter/src/errors.ts
```diff
@@ -1,12 +1,19 @@
...
+
   constructor({
     jobRunID = '1',
     status = 'errored',
     statusCode = 500,
     name = 'AdapterError',
     message = 'An error occurred.',
     cause,
-  }) {
+  }: Partial<AdapterError>) {
```
krebernisak, line 16: Why would we need to type a class constructor as a `Partial` of that same class?

external-adapter/src/requester.ts
```diff
@@ -72,18 +79,18 @@ class Requester {
     return Number(result)
   }
 
-  static getResult(data, path) {
+  static getResult(data: { [key: string]: any }, path: string[]): any {
     return path.reduce((o, n) => o[n], data)
   }
 
-  static errored(jobRunID = '1', error, statusCode = 500) {
+  static errored(jobRunID = '1', error?: any, statusCode = 500) {
```
krebernisak, line 86:
```suggestion
  static errored(jobRunID = '1', error: AdapterError | Error | string, statusCode = 500) {
```

external-adapter/src/validator.ts
```diff
@@ -1,8 +1,14 @@
...
-const { logger } = require('./logger')
+import { AdapterError } from './errors'
+import { Requester } from './requester'
+import { logger } from './logger'
+
+export class Validator {
+  input: any
+  customParams: any
+  validated: any
+  error: any
```
krebernisak, line 9:
```suggestion
  error: AdapterError | Error
```

external-adapter/src/validator.ts
```diff
@@ -1,8 +1,14 @@
...
+  errored: any
```
krebernisak, line 10: We could actually type this as `AdapterErrorResponse`

external-adapter/src/requester.ts
```diff
@@ -1,24 +1,31 @@
...
-const { logger } = require('./logger')
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
+import { AdapterError } from './errors'
+import { logger } from './logger'
+import { Config } from '@chainlink/types'
+import { util } from '@chainlink/ea-bootstrap'
 
-class Requester {
-  static async request(config, customError, retries = 3, delay = 1000) {
+const defaultCustomError = () => false
```
krebernisak, line 7: IMHO we hide a simple fn behind an obscure name and reading through the code makes it harder to understand what's exactly going on. I would just call it:

```suggestion
const getFalse = () => false
```

bootstrap/src/lib/util.ts
```diff
@@ -1,23 +1,33 @@
...
+export const uuid = () => {
   if (!process.env.UUID) process.env.UUID = uuidv4()
   return process.env.UUID
 }
 
-const toAsync = (execute, data) =>
-  new Promise((resolve) => execute(data, (statusCode, data) => resolve({ statusCode, data })))
+export const toAsync = (execute: ExecuteSync, data: AdapterRequest) =>
+  new Promise<WrappedAdapterResponse>((resolve) =>
+    execute(data, (statusCode: number, data: AdapterResponse) => resolve(data)),
```
krebernisak, line 27: Something looks off here, shouldn't we `resolve({ statusCode, data })`?

```suggestion
    execute(data, (statusCode: number, data: AdapterResponse) => resolve({ statusCode, data })),
```

krebernisak (changes requested):

external-adapter/test/config.test.ts
```diff
@@ -1,6 +1,7 @@
 import { expect } from 'chai'
```
krebernisak, line 1: This test does not make much sense anymore at this location, we should move it to the place where `API_KEY` env is set, which is now the `external-adapter` package.

external-adapter/src/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Validator } from './validator'
+import { AdapterError } from './errors'
+import { logger } from './logger'
+
+module.exports = {
+  Requester,
+  Validator,
+  AdapterError,
+  logger,
+}
```
krebernisak, line 11:
```suggestion
export * from './requester'
export * from './validator'
export * from './errors'
export * from './logger'
```

bootstrap/src/lib/cache/redis.ts
```diff
@@ -1,36 +1,40 @@
...
+import { timeout } from 'promise-timeout'
+import { createClient, ClientOpts, RedisClient } from 'redis'
 
 // Connection
 const DEFAULT_CACHE_REDIS_HOST = '127.0.0.1' // IP address of the Redis server
 const DEFAULT_CACHE_REDIS_PORT = 6379 // Port of the Redis server
-const DEFAULT_CACHE_REDIS_PATH = null // The UNIX socket string of the Redis server
-const DEFAULT_CACHE_REDIS_URL = null // The URL of the Redis server
+const DEFAULT_CACHE_REDIS_PATH = undefined // The UNIX socket string of the Redis server
+const DEFAULT_CACHE_REDIS_URL = undefined // The URL of the Redis server
```
krebernisak, line 10: Please revert this just in case. Redis client lib is especially picky and had different behavior if you feed it `null` vs `undefined` 😒
```suggestion
const DEFAULT_CACHE_REDIS_PATH = null // The UNIX socket string of the Redis server
const DEFAULT_CACHE_REDIS_URL = null // The URL of the Redis server
```

bootstrap/src/lib/cache/local.ts
```diff
@@ -1,45 +1,44 @@
...
 const DEFAULT_CACHE_UPDATE_AGE_ON_GET = false
+const DEFAULT_CACHE_TIMEOUT = 500 // Timeout in ms
 
 const env = process.env
-const defaultOptions = () => ({
+export const defaultOptions = () => ({
   max: Number(env.CACHE_MAX_ITEMS) || DEFAULT_CACHE_MAX_ITEMS,
   maxAge: Number(env.CACHE_MAX_AGE) || DEFAULT_CACHE_MAX_AGE,
   updateAgeOnGet: parseBool(env.CACHE_UPDATE_AGE_ON_GET) || DEFAULT_CACHE_UPDATE_AGE_ON_GET,
+  timeout: DEFAULT_CACHE_TIMEOUT,
```
krebernisak, line 15: Local cache impl does not have (or need) this `timeout` that our Redis impl has.

```suggestion
```

bootstrap/src/lib/cache/index.ts
```diff
@@ -56,54 +61,57 @@ const defaultCacheBuilder = () => {
...
 })
 
-const withCache = async (execute, options) => {
+export const withCache = async (
+  execute: ExecuteWrappedResponse,
+  options: CacheOptions = defaultOptions(),
+) => {
   // If no options read the env with sensible defaults
-  if (!options) options = defaultOptions()
+  if (!options) options
```
krebernisak, line 77: Please revert this line, just so we don't mess up something around cache lifecycle, as it is not so robust. Looks harmless, I know, but still :) We have some caching work incoming, so we'll make any changes there where it's going to be properly tested.

```suggestion
  if (!options) options = defaultOptions()
```

bootstrap/src/lib/cache/index.ts
```diff
@@ -37,16 +38,20 @@ const defaultOptions = () => ({
...
   },
 })
+export type CacheOptions = ReturnType<typeof defaultOptions>
+
 const defaultCacheOptions = () => {
   const type = env.CACHE_TYPE || DEFAULT_CACHE_TYPE
   const options = type === 'redis' ? redis.defaultOptions() : local.defaultOptions()
   return { ...options, type }
 }
+export type ImplCacheOptions = ReturnType<typeof defaultCacheOptions>
```
krebernisak, line 48: Maybe switch the words:
```suggestion
export type CacheImplOptions = ReturnType<typeof defaultCacheOptions>
```

krebernisak (changes requested):

external-adapter/src/validator.ts
justinkaseman: Not sure if I exactly follow on this, but I added it.

bootstrap/src/lib/cache/redis.ts
justinkaseman: Will revert, the official Redis types are looking for `undefined` so I will add a band-aid. Weird that it would differ :confused:

external-adapter/test/config.test.ts
justinkaseman: That's already where this one is moved to :smile:

external-adapter/src/validator.ts
krebernisak: If you take a look at our `AdapterError` there we construct a error response type that we return to the client:

```javascript
toJSONResponse() {
    const showDebugInfo = process.env.NODE_ENV === 'development' || process.env.DEBUG === 'true'
    const errorBasic = { name: this.name, message: this.message }
    const errorFull = { ...errorBasic, stack: this.stack, cause: this.cause }
    return {
      jobRunID: this.jobRunID,
      status: this.status,
      statusCode: this.statusCode,
      error: showDebugInfo ? errorFull : errorBasic,
    }
  }
```

My suggestion was just to type that, and then use it as `errored: AdapterErrorResponse` instead of `errored: any`.

external-adapter/src/errors.ts
justinkaseman: All of the properties of the class are getting set here in the constructor, but are not actually required params (they have the default). The use of `Partial` here makes all of those params optional.

external-adapter/src/validator.ts
justinkaseman: Ah I see now! Will add.

Makefile
```diff
@@ -30,6 +30,8 @@ deps: clean
 	yarn
 	# Call the build script for the adapter if defined (TypeScript adapters have this extra build/compile step)
 	# We use `wsrun` to build workspace dependencies in topological order (TODO: use native `yarn workspaces foreach -pt run build` with Yarn 2)
+	if [ $(adapter) = bootstrap ]; then yarn wsrun -mre -p @chainlink/ea-bootstrap -t build; fi
```
justinkaseman, line 33: @krebernisak Can you think of a cleaner way to do this?

Creating new make commands & github strategies for `adapters-core` seemed too verbose, but maybe that is a better path.

justinkaseman:
@krebernisak Can you think of a cleaner way to do this?

We could add `adapters-core` make commands and github strategy/workflow, but that might be too verbose.

https://github.com/smartcontractkit/external-adapters-js/pull/156/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R33

typings/@chainlink/index.d.ts
```diff
@@ -1,20 +1,67 @@
 // Declare missing type definitions
 declare module '@chainlink/types' {
-  export type AdapterRequest = { id: string; data: Record<string, unknown>; meta?: Record<string, unknown> }
+  export type AdapterRequest = {
+    id: string
+    data: Record<string, unknown>
+    meta?: Record<string, unknown>
+  }
+
+  export type AdapterHealthCheck = (callback: any) => any
```
krebernisak, line 9:
```suggestion
  export type Callback = (statusCode: number, data?: any) =>  void
  export type AdapterHealthCheck = (callback: Callback) => void
```

bootstrap/src/lib/server.ts
```diff
@@ -1,21 +1,25 @@
...
 
-const HEADER_CONTENT_TYPE = 'Content-Type'
-const CONTENT_TYPE_APPLICATION_JSON = 'application/json'
-const CONTENT_TYPE_TEXT_PLAIN = 'text/plain'
+export const HEADER_CONTENT_TYPE = 'Content-Type'
+export const CONTENT_TYPE_APPLICATION_JSON = 'application/json'
+export const CONTENT_TYPE_TEXT_PLAIN = 'text/plain'
 
-const notImplementedHealthCheck = (callback) => callback(HTTP_ERROR_NOT_IMPLEMENTED)
+const notImplementedHealthCheck = (callback: any) => callback(HTTP_ERROR_NOT_IMPLEMENTED)
```
krebernisak, line 17:
```suggestion
const notImplementedHealthCheck: AdapterHealthCheck = (callback) => callback(HTTP_ERROR_NOT_IMPLEMENTED)
```

typings/@chainlink/index.d.ts
```diff
@@ -1,20 +1,67 @@
...
+  }
+
   // TODO: clean this ASAP
   export type WrappedAdapterResponse = {
     statusCode: number
     data: AdapterResponse
   }
+  export type ExecuteWrappedResponse = (input: AdapterRequest) => Promise<WrappedAdapterResponse>
+
+  export type ExecuteSync = (input: AdapterRequest, callback: (statusCode, data) => void) => void
```
krebernisak, line 40:
```suggestion
  export type ExecuteSync = (input: AdapterRequest, callback: Callback) => void
```

krebernisak (changes requested):
Looking good 👍 

Few more small suggestions, and please add an entry to the CHANGELOG noting the migration to TS for `@chainlink/ea-bootstrap` and  `@chainlink/external-adapter` packages.

Makefile
```diff
@@ -30,6 +30,8 @@ deps: clean
...
+	if [ $(adapter) = external-adapter ]; then yarn wsrun -mre -p @chainlink/external-adapter -t build; fi
```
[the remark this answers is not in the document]
justinkaseman: Yeah this could be fixed by adding `-adapter` to the package names and changing the github strategies to say `ea-bootstrap` instead of `bootstrap`

#174

bootstrap/src/index.ts
```diff
@@ -33,47 +41,54 @@ const withStatusCode = (execute) => async (data_) => {
...
 
   // Return sync function
-  return (data, callback) => {
+  return (data: AdapterRequest, callback: any) => {
     // We init on every call because of cache connection broken state issue
     return withMiddleware(execute)
-      .then((executeWithMiddleware) => executeWithMiddleware(data))
-      .then((result) => callback(result.statusCode, result.data))
+      .then((executeWithMiddleware) => executeWithMiddleware && executeWithMiddleware(data))
+      .then((result) => callback(result?.statusCode, result?.data))
```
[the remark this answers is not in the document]
justinkaseman: For some reason it is seeing the Middleware promises returning a possible undefined. I'll tell it that they won't by adding 
```  return wrappedExecute as ExecuteWrappedResponse
```
on line 70

bootstrap/src/index.ts
```diff
@@ -33,47 +41,54 @@ const withStatusCode = (execute) => async (data_) => {
...
+    return (data: AdapterRequest) => execAsync(data)
+  }
+  return async (data: AdapterRequest) => util.toAsync(execute as ExecuteSync, data)
 }
 
 const middleware = [withAsync, withLogger, skipOnError(withCache), withStatusCode]
 
 // Init all middleware, and return a wrapped execute fn
-const withMiddleware = async (execute) => {
+const withMiddleware = async (execute: ExecuteWrappedResponse | ExecuteSync) => {
```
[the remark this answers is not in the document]
justinkaseman: Agreed, will open an issue.

bootstrap/src/lib/cache/index.ts
```diff
@@ -37,73 +39,80 @@ const defaultOptions = () => ({
...
     options.cacheOptions.type === 'redis'
-      ? redis.redactOptions(options.cacheOptions)
+      ? redis.redactOptions(options.cacheOptions as RedisOptions)
       : local.redactOptions(options.cacheOptions),
 })
 
-const withCache = async (execute, options) => {
+export const withCache = async (
+  execute: ExecuteWrappedResponse,
+  options: CacheOptions = defaultOptions(),
```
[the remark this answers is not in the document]
justinkaseman: Typescript doesn't see something like 
```
74 // If no options read the env with sensible defaults
75  if (!options) options = defaultOptions()
```
as guarding against undefined. Instead we can keep the default param and remove line 74 + 75.

bootstrap/src/lib/cache/redis.ts
```diff
@@ -14,23 +14,27 @@ const DEFAULT_CACHE_REDIS_TIMEOUT = 500 // Timeout in ms
...
+export type RedisOptions = ClientOpts & { maxAge: number; timeout: number }
+
+export const defaultOptions = (): RedisOptions => ({
   host: env.CACHE_REDIS_HOST || DEFAULT_CACHE_REDIS_HOST,
-  port: env.CACHE_REDIS_PORT || DEFAULT_CACHE_REDIS_PORT,
-  path: env.CACHE_REDIS_PATH || DEFAULT_CACHE_REDIS_PATH,
-  url: env.CACHE_REDIS_URL || DEFAULT_CACHE_REDIS_URL,
+  port: Number(env.CACHE_REDIS_PORT) || DEFAULT_CACHE_REDIS_PORT,
+  path: env.CACHE_REDIS_PATH || ((DEFAULT_CACHE_REDIS_PATH as unknown) as undefined),
+  url: env.CACHE_REDIS_URL || ((DEFAULT_CACHE_REDIS_URL as unknown) as undefined),
```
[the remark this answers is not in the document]
justinkaseman: This is a bandaid over the fact that we are setting these defaults to `null` rather than what the @types/Redis expects: `undefined`. 

We could assert it as undefined, but it gives this warning:
```
Conversion of type 'null' to type 'undefined' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.ts(2352)
```

bootstrap/src/index.ts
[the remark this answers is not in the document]
krebernisak: Thanks!

bootstrap/src/lib/cache/redis.ts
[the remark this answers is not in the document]
krebernisak: Ok, let's set it as undefined then 🤞. Sounds like a non-issue but that lib spanked me few times unexpectedly.

typings/@chainlink/index.d.ts
```diff
@@ -1,14 +1,14 @@
...
   export type AdapterRequestMeta = {
-    availableFunds: number
-    eligibleToSubmit: boolean
-    latestAnswer: number
-    oracleCount: number
-    paymentAmount: number
-    reportableRoundID: number
-    startedAt: number
-    timeout: number
+    availableFunds?: number
```
krebernisak, line 4: I don't think all of this should be optional? Meta itself could be optional in the request, but if meta is set then all of this params should be set.

justinkaseman: It was throws errors from things like `iv-outlier-detection` adapter where meta gets used. 

I could go through and modify these, but might be better suited for a new 1 story point PR/issue as I didn't dig into Jonas' adapters to see how much needs to be changed.

composite/proof-of-reserves/src/balance.ts
```diff
@@ -18,19 +22,13 @@ export const getBitcoinIndexer = (): BitcoinIndexer | undefined => {
...
 export const getImpl = (options: BitcoinIndexerOptions): Execute => {
   const prefix = options.type?.toUpperCase()
-  switch (options.type) {
-    case BitcoinIndexer.BlockchainCom:
-      return (data) => {
-        const config = blockchainCom.getConfig(prefix)
-        return blockchainCom.execute(data, config)
-      }
+  const impl = options.type && implLookup[options.type]
+  if (impl)
```
krebernisak, line 26: It's preferable not to nest code if possible:

```javascript
if(!impl) throw Error(`Unknown balance adapter type: ${options.type}`)
return ...
```

composite/proof-of-reserves/src/protocol.ts
```diff
@@ -18,19 +22,13 @@ export const getProtocol = (): Protocol | undefined => {
...
 export const getImpl = (options: ProtocolOptions): Execute => {
   const prefix = options.type?.toUpperCase()
-  switch (options.type) {
-    case Protocol.RenVM:
-      return (data) => {
-        const config = renVM.getConfig(prefix)
-        return renVM.execute(data, config)
-      }
+  const impl = options.type && implLookup[options.type]
+  if (impl)
```
krebernisak, line 26: Same comment as above.

krebernisak (changes requested):

composite/proof-of-reserves/src/balance.ts
justinkaseman: Yeah guard clauses look way cleaner, good suggestion. Changed.

krebernisak (dismissed):

thodges-gh (approved):

PR #163, opened 2020-11-25
Add adapter for Tradermade

boxhock:


tradermade/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@chainlink/tradermade",
```
krebernisak, line 2: Isn't it `@chainlink/tradermade-adapter` now?

krebernisak (approved):

tradermade/package.json
boxhock: Good point

krebernisak (approved):

tradermade/package.json
boxhock: Oh actually, we can't do this until the other PR is merged in. Reverting that commit now

krebernisak: Ok, let's get that other PR rebased & merged after this then.

krebernisak (approved):

PR #165, opened 2020-11-27
Add Finage adapter

boxhock:


krebernisak (changes requested):
I see that you haven't updated any CHANGELOG files. Would it make sense to do so? 😂

krebernisak (approved):

PR #173, opened 2020-12-07
Query string params

emmick4:
Will need to be merged alongside [this PR](https://github.com/smartcontractkit/infra-modules/pull/286). This allows us to pass arbitrary query string parameters to external adapters, useful for specifying a single external adapter as multiple data sources. For [this story](https://www.pivotaltracker.com/story/show/175901421)

### DIFF
--- a/bootstrap/lib/aws.js
+++ b/bootstrap/lib/aws.js
@@ -49,7 +49,10 @@ exports.initHandlerREST = (execute) => (event, _context, callback) => {
     return callback(null, UNSUPPORTED_MEDIA_TYPE_RESPONSE)
   }
   let newBody = event.requestBody
-  newBody.data = Object.assign(event.requestBody.data === undefined ? {} : event.requestBody.data, parseQueryString(event.queryStringParameters))
+  newBody.data = Object.assign(
+    event.requestBody.data === undefined ? {} : event.requestBody.data,
+    parseQueryString(event.queryStringParameters),
+  )
   execute(newBody, (_, data) => {
     callback(null, data)
   })
@@ -62,7 +65,10 @@ exports.initHandlerHTTP = (execute) => (event, _context, callback) => {
   }
   const body = JSON.parse(event.body)
   let newBody = body.requestBody
-  newBody.data = Object.assign(event.requestBody.data === undefined ? {} : event.requestBody.data, parseQueryString(event.queryStringParameters))
+  newBody.data = Object.assign(
+    event.requestBody.data === undefined ? {} : event.requestBody.data,
+    parseQueryString(event.queryStringParameters),
+  )
   execute(newBody, (statusCode, data) => {
     callback(null, {
       statusCode: statusCode,
--- a/bootstrap/lib/aws.js
+++ b/bootstrap/lib/aws.js
@@ -8,6 +8,8 @@ const {
   HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE_MESSAGE,
 } = require('./errors')
 
+const { parseQueryString } = require('./util')
+
 const awsGetRequestHeaders = (event) => {
   if (!event || !event.headers) return {}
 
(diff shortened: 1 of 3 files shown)


bootstrap/lib/server.js
```diff
@@ -24,7 +26,10 @@ const initHandler = (execute, checkHealth = notImplementedHealthCheck) => () =>
         .status(HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE)
         .send(HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE_MESSAGE)
     }
-
+    req.body.data = Object.assign(
+      req.body.data === undefined ? {} : req.body.data,
+      parseQueryString(req.query),
+    )
```
krebernisak, line 32: `Object.assign()` modifies an object in place. So we should either do:

```javascript
req.body.data = req.body.data || {}
Object.assign(req.body.data, parseQueryString(req.query))
```

or just use the spread operator:

```javascript
req.body.data = { ...req.body.data, parseQueryString(req.query) }
```

Because of [optional spreading](https://stackoverflow.com/questions/46957194/javascript-es6-spread-operator-on-undefined) feature, I suggest going with option #2.

bootstrap/lib/util.js
```diff
@@ -8,6 +8,15 @@ const parseBool = (value) => {
   return (_val === 'true' || _val === 'false') && _val === 'true'
 }
 
+// parse incoming query string, strings into Numbers where possible
+const parseQueryString = (query) => {
```
krebernisak, line 12: As query parameters are already parsed at this point, and as this fn does something else, consider naming it something more specific. For example:

```suggestion
const toObjectWithNumbers = (query) => {
```

bootstrap/lib/util.js
```diff
@@ -8,6 +8,15 @@ const parseBool = (value) => {
...
+  let formatted = {}
+  for (const [key, val] of Object.entries(query)) {
+    formatted[key] = isNaN(val) ? val : Number(val)
+  }
+  return formatted
```
krebernisak, line 17: Perfect opportunity to use ES2019 `Object.fromEntries`:

```suggestion
  const toNumber = (v) => isNaN(v) ? v : Number(v)
  return Object.fromEntries(Object.entries(obj).map(([k, v]) => [k, toNumber(v)]))
```

bootstrap/lib/aws.js
```diff
@@ -57,8 +63,13 @@ exports.initHandlerHTTP = (execute) => (event, _context, callback) => {
   if (!isContentTypeSupported(event)) {
     return callback(null, UNSUPPORTED_MEDIA_TYPE_RESPONSE)
   }
-
-  execute(JSON.parse(event.body), (statusCode, data) => {
+  const body = JSON.parse(event.body)
+  let newBody = body.requestBody
```
krebernisak, line 67: Consider renaming this to be more specific like `bodyIncludingQuery`.

bootstrap/lib/aws.js
```diff
@@ -8,6 +8,8 @@ const {
   HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE_MESSAGE,
```
krebernisak, line 8: We should also add this feature to [bootstrap/lib/gcp.js](https://github.com/smartcontractkit/external-adapters-js/blob/master/bootstrap/lib/gcp.js)

bootstrap/lib/aws.js
krebernisak: I'll comment on `requestBody` and `queryStringParameters` in the other PR.

krebernisak (changes requested):
Please also add documentation to README and a note to CHANGELOG so we document the new behavior.

bootstrap/lib/aws.js
```diff
@@ -46,8 +48,12 @@ exports.initHandlerREST = (execute) => (event, _context, callback) => {
...
     return callback(null, UNSUPPORTED_MEDIA_TYPE_RESPONSE)
   }
-
-  execute(event, (_, data) => {
+  let newBody = event.requestBody
+  newBody.data = Object.assign(
+    event.requestBody.data === undefined ? {} : event.requestBody.data,
+    parseQueryString(event.queryStringParameters),
+  )
+  execute(newBody, (_, data) => {
```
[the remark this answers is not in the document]
krebernisak: Aha, now I get it :) You see how it confused me because now this code is coupled to some configuration that is contained in another repo. That other repo is actually not public, so this is going to be even a bigger issue for someone external to the team who would want to run this open-source software and be part of the Chainlink protocol.

emmick4:
my git-fu wasn't there to deal with the merge after typescript changes were merged, moved this PR to: https://github.com/smartcontractkit/external-adapters-js/pull/186

PR #178, opened 2020-12-07
Import directory dinamically

RodrigoAD:
The build process was not including `json` files. 

This fix changes the way the directory is read. Now is directly imported, instead of using `filesystem`.   
`tsconfig.json` is also configured to include `json` files


krebernisak (approved):
That's better :D

PR #181, opened 2020-12-08
Outlier Detection composite adapter

boxhock:


typings/@chainlink/index.d.ts
```diff
@@ -55,7 +55,8 @@ declare module '@chainlink/types' {
     apiKey?: string
     network?: string
     returnRejectedPromiseOnError?: Boolean
-    api: Partial<AxiosRequestConfig>
+    api?: Partial<AxiosRequestConfig>
+    threshold?: Record<string, number>
```
boxhock, line 59: Not sure if this is the ideal way to handle this, but I'm also not familiar with the decision to have a centralized Config type.

krebernisak: Yeah, I feel your pain.

It's not meant as a centralized Config type, but as a base Config type where every adapter has a specific Config type that extends that base Config, and needs to implement a factory method:

```javascript
export const makeConfig = <C extends Config>(args...): C => {
```

But the infrastructure is not there yet.

composite/outlier-detection/src/config.ts
```diff
@@ -0,0 +1,19 @@
+import { logger } from '@chainlink/external-adapter'
+import { util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
```
krebernisak, line 4: Here you can make a specific Config type for this adapter, one that extends our base Config type.

```suggestion
import types from '@chainlink/types'

export type Config = types.Config & {
   threshold: {
     checks: number
     onchain: number
   }
}

```

composite/outlier-detection/src/source.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const getSourceDataProviders = (): SourceDataProvider[] => {
+  const sourceDataProviders = (process.env.SOURCE_ADAPTERS || '').split(',')
+  const adapters: SourceDataProvider[] = []
+
+  sourceDataProviders.forEach((provider) => {
+    if (isSourceDataProvider(provider)) adapters.push(provider as SourceDataProvider)
+  })
+
+  return adapters
```
krebernisak, line 21: This can be just a filter of the original array:

```suggestion
  return sourceDataProviders.filter(isSourceDataProvider)
    .map(provider => provider as SourceDataProvider)
```

composite/outlier-detection/src/source.ts
```diff
@@ -0,0 +1,48 @@
...
+export enum SourceDataProvider {
+  XBTO = 'xbto',
+  GenesisVolatility = 'genesisvolatility',
+}
+
+const isSourceDataProvider = (envVar?: string): envVar is SourceDataProvider =>
+  Object.values(SourceDataProvider).includes(envVar as any)
+
+export const getSourceDataProviders = (): SourceDataProvider[] => {
+  const sourceDataProviders = (process.env.SOURCE_ADAPTERS || '').split(',')
```
krebernisak, line 14: This should also be part of the config.

We can extend the config type to be:

```javascript
// config.ts
import { SourceDataProvider } from './source'

export type Config = types.Config & {
   sourceDataProviders: SourceDataProvider[]
   threshold: {
     checks: number
     onchain: number
   }
}
```

And then we can read it from the env in the `makeConfig` factory function.

composite/outlier-detection/src/source.ts
```diff
@@ -0,0 +1,48 @@
...
+  const adapters: SourceDataProvider[] = []
+
+  sourceDataProviders.forEach((provider) => {
+    if (isSourceDataProvider(provider)) adapters.push(provider as SourceDataProvider)
+  })
+
+  return adapters
+}
+
+export const getSourceImpl = (types: SourceDataProvider[]): Execute[] => {
```
krebernisak, line 24: It's easier to use this type of mapping function if it maps just one item.

```suggestion
export const getSourceImpl = (type: SourceDataProvider): Execute => {
```

Then externally if I want to map N items:

```javascript
const adapters = types.map(getSourceImpl)
```

Or in our case, we could write it close to this:

```javascript
const sourceAdapters = config.sourceDataProviders.map(source.makeExecute)
```

composite/outlier-detection/src/source.ts
```diff
@@ -0,0 +1,48 @@
...
+        adapters.push((data) => {
+          const config = XBTO.getConfig(prefix)
+          return XBTO.execute(data, config)
+        })
+        return
+      case SourceDataProvider.GenesisVolatility:
+        adapters.push((data) => {
+          const config = GenesisVolatility.getConfig(prefix)
+          return GenesisVolatility.execute(data, config)
+        })
```
krebernisak, line 40: Looks like both of these adapters now have our `makeExecute` & `makeConfig` factory functions, so we can use them directly:

```suggestion
        const config = XBTO.makeConfig(prefix)
        adapters.push(XBTO.makeExecute(config))
        return
      case SourceDataProvider.GenesisVolatility:
        const config = GenesisVolatility.makeConfig(prefix)
        adapters.push(GenesisVolatility.makeExecute(config))
```

composite/outlier-detection/src/check.ts
```diff
@@ -0,0 +1,45 @@
+import { Execute } from '@chainlink/types'
```
krebernisak, line 1: Comments for `source.ts` also apply here.

genesis-volatility/src/adapter.ts
```diff
@@ -0,0 +1,52 @@
+import { ExecuteFactory, ExecuteWithConfig } from '@chainlink/types'
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { makeConfig } from './config'
+
+const customParams = {
+  symbol: ['base', 'from', 'coin', 'symbol'],
+  days: ['days', 'period'],
+}
+
+export const execute: ExecuteWithConfig = async (input, config) => {
```
krebernisak, line 10: No need to export this fn anymore now that we export `makeExecute`

```suggestion
const execute: ExecuteWithConfig = async (input, config) => {
```

genesis-volatility/src/config.ts
```diff
@@ -0,0 +1,15 @@
+import { logger } from '@chainlink/external-adapter'
+import { util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const ENV_API_KEY = 'API_KEY'
+
+export const makeConfig = (prefix = ''): Config => {
+  return {
+    apiKey: util.getEnv(ENV_API_KEY, prefix),
+  }
```
krebernisak, line 10: We now have `Requester.getDefaultConfig` that will do this for us:

```suggestion
  return Requester.getDefaultConfig(prefix)
```

oilpriceapi/src/config.ts
```diff
@@ -0,0 +1,15 @@
+import { logger } from '@chainlink/external-adapter'
+import { util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const ENV_API_KEY = 'API_KEY'
+
+export const makeConfig = (prefix = ''): Config => {
+  return {
+    apiKey: util.getEnv(ENV_API_KEY, prefix),
+  }
```
krebernisak, line 10: We now have `Requester.getDefaultConfig` that will do this for us:

```suggestion
  return Requester.getDefaultConfig(prefix)
```

oilpriceapi/src/config.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const logConfig = (config: Config): void => {
```
krebernisak, line 13: There is now also a `Requester.logConfig` that will do this and additionally avoid leaking the `apiKey`

genesis-volatility/src/config.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const logConfig = (config: Config): void => {
```
krebernisak, line 13: There is now also a `Requester.logConfig` that will do this and additionally avoid leaking the `apiKey`

xbto/src/config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const makeConfig = (prefix = ''): Config => {
+  return {
+    apiKey: util.getEnv(ENV_API_KEY, prefix),
+  }
+}
+
+export const logConfig = (config: Config): void => {
+  logger.debug('Adapter configuration:', { config })
+}
```
krebernisak, line 15: Same comments as above!

krebernisak (changes requested):
Looking nice!

I love how we evolved our codebase to support writing these composite adapters. But this is just a start, I think we can improve the framework step by step to enable even greater composability.

Left a few comments, but expect some more soon as I did not go through it all in this pass.

typings/@chainlink/index.d.ts
krebernisak: We should now be able to remove this line from here, as you defined a specific config for that specific adapter.

krebernisak, line 59:
```suggestion
```

composite/outlier-detection/src/source.ts
```diff
@@ -0,0 +1,35 @@
...
+export const getSourceImpl = (type: SourceDataProvider): Execute => {
+  const prefix = type?.toUpperCase()
+  switch (type) {
+    case SourceDataProvider.XBTO:
+      return XBTO.makeExecute(XBTO.makeConfig(prefix))
+    case SourceDataProvider.GenesisVolatility:
+      return GenesisVolatility.makeExecute(GenesisVolatility.makeConfig(prefix))
+    default:
+      throw Error(`Unknown source data provider adapter type: ${type}`)
+  }
```
krebernisak, line 34: This is ok for now, but it's a lot of boilerplate that we can improve on, or at least make it more general. We are currently in the process of figuring out how to best do it for the PoR composite adapter. When we figure that out, we'll probably refactor this also. Ping @justinkaseman

krebernisak (changes requested):

composite/outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  return returnValue(jobRunID, sourceMedian)
+}
+
+const getExecuteMedian = async (executes: Execute[], request: AdapterRequest): Promise<number> => {
+  const responses = await Promise.allSettled(executes.map((execute) => execute(request)))
+  const values = responses
+    .filter((result) => result.status === 'fulfilled' && 'value' in result)
```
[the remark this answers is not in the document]
boxhock: Since that's how we do things with FM, I think this is the best approach to take. We also don't want the entire thing to fail if 1 out of 2 data providers fail.

composite/outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,93 @@
...
+  const half = Math.floor(values.length / 2)
+  if (values.length % 2) return values[half]
+  return (values[half - 1] + values[half]) / 2.0
+}
+
+const difference = (a: number, b: number): number => {
+  return (Math.abs(a - b) / ((a + b) / 2)) * 100
+}
+
+const returnValue = (jobRunID: string, result: number): AdapterResponse => {
```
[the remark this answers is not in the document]
boxhock: Not a bad idea :smile:

composite/outlier-detection/src/adapter.ts
```diff
@@ -0,0 +1,93 @@
...
+export const execute = async (input: AdapterRequest, config: Config): Promise<AdapterResponse> => {
+  const sourceExecute = config.sourceDataProviders.map(getSourceImpl)
+  if (sourceExecute.length === 0) {
+    throw Error('No source adapters provided')
+  }
+  const checkExecute = config.checkDataProviders.map(getCheckImpl)
+  return await executeWithAdapters(input, sourceExecute, checkExecute, config)
+}
+
+export const executeWithAdapters = async (
```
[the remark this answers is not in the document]
boxhock: I've made some changes now that should hopefully address this.

composite/outlier-detection/test/adapter.test.ts
```diff
@@ -79,16 +86,14 @@ describe('executeWithAdapters', () => {
...
       },
     ]
 
     requests.forEach((req) => {
       it(`${req.name}`, async () => {
-        const data = await executeWithAdapters(req.input, req.adapters, req.checks, {
-          ...config,
-          ...req.config,
-        })
+        const execute = makeExecute(makeMockConfig(req.adapters, req.checks, req.threshold))
```
krebernisak, line 95: Nice! 💯

krebernisak (dismissed):
Looking great! 👍 

Just take a look at deribit tests, it's failing there:

```
> @chainlink/deribit-adapter
$ mocha --exit --grep @integration --invert -r ts-node/register 'test/**/*.test.ts'

TSError: ⨯ Unable to compile TypeScript:
Error: test/adapter.test.ts(32,28): error TS2554: Expected 2 arguments, but got 1.
Error: test/adapter.test.ts(49,17): error TS2554: Expected 2 arguments, but got 1.
Error: test/adapter.test.ts(69,17): error TS2554: Expected 2 arguments, but got 1.
```

krebernisak (approved):

Issue #175, opened 2020-12-07
Sequences processing semantics [Draft]


krebernisak:
Some of our adapters like the PoR bitcoin indexer adapters process sequenced data.

Adapters that process sequences should only change items that they know how to process. If they don’t have the ability to process them they should add a warning. Maybe the next adapter in the chain will have the know-how to process those items and possibly remove the warning.

If they have the ability to process the item, but the item is found to be invalid, the error should be thrown as there is no way to continue processing this sequence. For example, if we have an adapter that processes `bitcoin/mainnet` addresses. That adapter should add warnings for all `bitcoin/testnet` and `ethereum/mainnet` addresses. But if it finds an invalid `bitcoin/mainnet` address, then it throws an error!

We can additionally add a `strict` configuration flag, which we can use to configure the sequence processing adapter with: “No warnings, all errors.” Or in other words, you need to be able to process everything or nothing.

TODO:

- [ ] How sequence processing semantic should work
- [ ] How we should be able to chain different adapters with different processing abilities
- [ ] How error handling should work
- [ ] How can we extract this in a framework to easily add sequence processing abilities to an adapter

boxhock:
Story created

PR #184, opened 2020-12-08
Feature/proof of reserves

justinkaseman:
### Description
Adds `balance` endpoint functionality to:
- `amberdata`
- `cryptoapis`
- `blockchair`

New adapters:
- `btc.com`
- `sochain`

Updates other PoR adapters:
- `blockchain.com`
- `blockcypher`


amberdata/src/config.ts
```diff
@@ -0,0 +1,19 @@
...
+export const ENV_API_TIMEOUT = 'API_TIMEOUT'
+
+export const ENDPOINT_MAIN = 'https://web3api.io'
+
+export const DEFAULT_DATA_PATH = 'addresses'
+export const DEFAULT_ENDPOINT = 'price'
+
+export const getBaseURL = (): string => ENDPOINT_MAIN
+
+export const getConfig = (prefix = ''): Config => {
```
krebernisak, line 14:
```suggestion
export const makeConfig = (prefix = ''): Config => {
```

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      if (!addr.coin) addr.coin = 'btc'
+      if (isCoinType(addr.coin) === false)
+        return {
+          ...warnedResponse,
+          warning: WARNING_NO_OPERATION_COIN,
+        }
+
+      if (!addr.chain) addr.chain = 'mainnet'
+      if (isChainType(addr.chain) === false)
```
krebernisak, line 55:
```suggestion
      if (!isChainType(addr.chain))
```

amberdata/src/endpoint/price.ts
```diff
@@ -0,0 +1,33 @@
...
+const customError = (data: any) => {
+  return Object.keys(data.payload).length === 0
+}
+
+const customParams = {
+  base: ['base', 'from', 'coin'],
+  quote: ['quote', 'to', 'market'],
+}
+
+export const execute = async (config: Config, request: AdapterRequest) => {
```
krebernisak, line 16:
```suggestion
export const execute: ExecuteWithConfig = async (request, config) => {
```

amberdata/src/endpoint/balance.ts
```diff
@@ -0,0 +1,114 @@
...
+    } as any,
+  )
+
+export const inputParams = {
+  dataPath: false,
+  confirmations: false,
+}
+
+// Export function to integrate with Chainlink node
+export const execute = async (config: Config, request: AdapterRequest) => {
```
krebernisak, line 96:
```suggestion
export const execute: ExecuteWithConfig = async (request, config) => {
```

amberdata/test/balance.test.ts
```diff
@@ -0,0 +1,235 @@
+import { makeExecute } from '../src/adapter'
+import { shouldBehaveLikeBalanceAdapter } from '@chainlink/adapter-test-helpers'
+
+shouldBehaveLikeBalanceAdapter(makeExecute(), ['bitcoinmainnet'])
+
+// import { assert } from 'chai'
```
krebernisak, line 6: We should remove this now as we are using the test helper package

blockchair/README.md
```diff
@@ -1,83 +1,264 @@
...
+## Configuration
+
+The adapter takes the following environment variables:
+
+- `API_KEY`: Optional Blochair API key to use
+- `API_TIMEOUT`: Optional timeout param, defaults to `30000`
+
 ## Input Params
 
+- `endpoint`: Optional endpoint param, defaults to `difficulty`
```
krebernisak, line 12:
```suggestion
- `endpoint`: The endpoint to use, one of (difficulty|balance). Defaults to `difficulty`
```

blockchair/package.json
```diff
@@ -1,15 +1,39 @@
...
   "name": "@chainlink/blockchair-adapter",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "license": "MIT",
-  "main": "index.js",
+  "main": "dist/index.js",
   "scripts": {
-    "server": "node -e 'require(\"./index.js\").server()'",
+    "prepublishOnly": "yarn build && yarn test:unit",
+    "build": "tsc",
```
krebernisak, line 8:
```suggestion
    "build": "tsc -b",
```

blockchair/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ExecuteWithConfig, ExecuteFactory } from '@chainlink/types'
+import { getConfig, DEFAULT_ENDPOINT } from './config'
+import { difficulty, balance } from './endpoint'
+
+const inputParams = {
+  endpoint: false,
+}
+
+// Export function to integrate with Chainlink node
+export const execute: ExecuteWithConfig = async (request, config) => {
```
krebernisak, line 11:
```suggestion
const execute: ExecuteWithConfig = async (request, config) => {
```

amberdata/src/config.ts
krebernisak: The prefix is not being considered here

blockchair/src/config.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const ENDPOINT_MAIN = 'https://api.blockchair.com'
+
+export const DEFAULT_DATA_PATH = 'addresses'
+export const DEFAULT_CONFIRMATIONS = 6
+export const DEFAULT_ENDPOINT = 'difficulty'
+
+export const getBaseURL = (): string => ENDPOINT_MAIN
+
+export const getConfig = (): Config => Requester.getDefaultConfig()
```
krebernisak, line 12: makeConfig + prefix

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,137 @@
...
+      jobRunID,
+      message: `Unsupported coin parameter`,
+      statusCode: 500,
+    })
+  if (chain === 'testnet') {
+    if (coin === 'bitcoin') coin = 'bitcoin/testnet'
+    else
+      throw new AdapterError({
+        jobRunID,
+        message: WARNING_NO_OPERATION_TESTNET,
```
krebernisak, line 64: Should not be called a warning if we throw it as an error.

blockchair/src/endpoint/balance.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (!coin)
+    throw new AdapterError({
+      jobRunID,
+      message: `Unsupported coin parameter`,
+      statusCode: 500,
+    })
+  if (chain === 'testnet') {
+    if (coin === 'bitcoin') coin = 'bitcoin/testnet'
+    else
+      throw new AdapterError({
```
krebernisak, line 62: We should not throw here in the first place. If we can't process an item, we add a warning to it and continue.

krebernisak (changes requested):

amberdata/src/endpoint/balance.ts
justinkaseman: This seems like a style choice of clear vs. readable, but if you have a preference I'm fine with switching. I'd agree that the not operator is easier to read.

I read some opinion piece a while ago that argued being explicit is better, especially for helping the understanding of newer devs.

amberdata/src/adapter.ts
```diff
@@ -35,10 +34,8 @@ export const execute: ExecuteWithConfig = async (request, config) => {
...
     }
   }
-
-  return Requester.success(jobRunID, response)
 }
 
-export const makeExecute: ExecuteFactory = (config) => {
-  return async (request) => execute(request, config || getConfig())
+export const makeExecute: ExecuteFactory<Config> = (config) => {
+  return async (request) => await execute(request, config || getConfig())
```
krebernisak, line 40: This `await` is not needed here.

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const WARNING_NO_OPERATION = 'No Operation: unsupported'
+const ERROR_NO_OPERATION_MISSING_ADDRESS = 'No Operation: address param is missing'
+
+export type IsSupported = (coin: string, chain: string) => boolean
+export type GetBalance = (account: Account, config: BalanceConfig) => Promise<ResponseData>
+export type GetBatchBalance = (
+  accGroup: [string, AccountGroup],
+  config: Config,
+) => Promise<ResponseData>
```
krebernisak, line 22: My suggestion would be to have a generic type that will be a response to our data requests. It will carry some kind a result (an answer to our data request) and a payload or some underlying data that the result was made from:

```javascript
export type DataResponse<R, P> = {
  result: R,
  payload: P
}
```

Then we can declare these two type of requests for data:

```suggestion
export type GetBalance = (account: Account, config: BalanceConfig) => Promise<DataResponse<Account, any>>
export type GetBalances = (accounts: Account[], config: BalanceConfig) => Promise<DataResponse<Account[], any>>
```

krebernisak (changes requested):
Great job! 💯 This now looks a lot better when reusing a bunch of code 🪂 

Added a few more comments we can go through today, but overall it's a big improvement.
The only thing is this branch now has a 100+ commits because of excessive merging. You should think about rebasing next time to keep your branches cleaner.

justinkaseman:
> Great job! 100 This now looks a lot better when reusing a bunch of code parachute
> 
> Added a few more comments we can go through today, but overall it's a big improvement.
> The only thing is this branch now has a 100+ commits because of excessive merging. You should think about rebasing next time to keep your branches cleaner.

Usually I rebase, but at this point when rebasing there are ~20+ merge conflicts to go through each time. Doing that every time develop drifts ends up being a lot of time.

At least we can squash them together when we merge.

external-adapter/src/config.ts
```diff
@@ -21,6 +23,7 @@ export function getDefaultConfig(prefix = ''): Config {
   const apiKey = util.getEnv(ENV_API_KEY, prefix)
   return {
     apiKey,
+    returnVerboseResponse: !!util.getEnv(ENV_API_KEY, prefix),
```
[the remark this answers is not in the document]
justinkaseman: Oops this should use `ENV_API_VERBOSE_RESPONSE`.

This is for whether the adapter responses should include all of the response data for adapters that are making multiple requests. We talked about cases where if you have 50 addresses that you are fetching balances it might be a lot of response data. 

It is currently more of a balance specific use case, likely we will have more multiple request adapters in the future. All adapters _could_ also use this.

I'm open to moving it though, thoughts?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (!acc.address)
+      throw new AdapterError({
+        jobRunID,
+        message: ERROR_NO_OPERATION_MISSING_ADDRESS,
+        statusCode: 400,
+      })
+
+    // Defaults
+    if (!acc.chain) acc.chain = 'mainnet'
+    if (!acc.coin) acc.coin = 'btc'
```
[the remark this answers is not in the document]
justinkaseman: I've never seen nullish coalescing used like this - the compiler doesn't seem to like the syntax.

Could you explain this?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,155 @@
...
+export type BalanceImplConfig = Config & {
+  confirmations?: number
+  shouldOverwrite?: boolean
+  verbose?: boolean
+}
+export type BalanceConfig = BalanceImplConfig & {
+  isSupported: IsSupported
+  getBalance?: GetBalance
+  getBatchBalance?: GetBatchBalance
+}
```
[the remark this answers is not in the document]
justinkaseman: The idea here was that the former type would be the extended Config for the implementation factory and the latter would be the Config for the EA factory.

I didn't really end up using it that way though, so I can merge them.

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,155 @@
...
+): Promise<ResponseData[]> =>
+  Promise.all(
+    accounts.map((acc) => {
+      if (acc.warning) return { result: acc }
+      return getBalance(acc, config)
+    }),
+  )
+
+export type AccountGroup = { addresses: string[]; result: Account[] }
+const group = (accounts: Account[]) => {
```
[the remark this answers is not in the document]
justinkaseman: Where do you think this should live? `external-adapter`?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,155 @@
...
+  getBalance: GetBalance,
+): Promise<ResponseData[]> =>
+  Promise.all(
+    accounts.map((acc) => {
+      if (acc.warning) return { result: acc }
+      return getBalance(acc, config)
+    }),
+  )
+
+export type AccountGroup = { addresses: string[]; result: Account[] }
```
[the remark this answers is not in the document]
justinkaseman: The reason there was AccountGroups was in an attempt to have the ea-factory filter warning accounts out similar to how the non-batched functionality works.

With the change to groupBy to be more generic it just means the `getBalances` function in the implementation will have some repeated code to filter warnings.

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,168 @@
...
+      config,
+      Array.from(grouped.values()),
+      config.getBalances,
+    )
+    response = reduceBalancesBatch(config, responses)
+  }
+  if (config.getBalance) {
+    const responses = await getBalances(config, accounts, config.getBalance)
+    response = reduceBalances(config, responses)
+  }
```
krebernisak, line 165: We need just one workflow here, not two different ones that are almost the same.

First, define the function you require:

```javascript
const getBalances = config.getBalances || // create it from config.getBalance
```

Then group the data and execute M requests for M groups:

```javascript
const key = (acc) => `${acc.coin}-${acc.chain}`
const requests = groupBy(accounts, key)
  .values()
  .map(group => getBalances(group, config))
const responses = await Promise.all(requests)
```

Then reduce:

```javascript
const data = {
  responses: responses.map(r => r.payload),
  result: responses.flatMap(r => r.result),
}
```

Then in the end act on the `verbose` config:

```javascript
if (!config.verbose) delete data.responses;
```

krebernisak (changes requested):

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,168 @@
...
+  })
+  return map
+}
+
+export function addresses(accounts: Account[]) {
+  return accounts.reduce<string[]>((accumulator, current) => {
+    if (!current.warning) accumulator.push(current.address)
+    return accumulator
+  }, [])
+}
```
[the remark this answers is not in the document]
justinkaseman: By using reduce it only iterates over the accounts one time. 

Are you suggesting this because of readability?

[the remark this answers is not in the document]
krebernisak: Yeah, readability! It's one of the most important and hard to achieve things. Naming, readability, composability...
Always think about the next guy, and better yet think about yourself debugging the code a few months down the road, and make it easy for yourself 🥇

Optimization is not even to be considered until the end, or when performance problems arise.

And here we do `O(n)` work in both cases. It is the same if you:

- iterate N elements once, do two operations per element
- iterate N elements twice, do one operation per element

Both are `O(2*n)`, or `O(n)`.

Even if you introduce an extra fixed amount of work because of readability, that is also ok as long as you don't change the time complexity. Going from `O(n)` to `O(6*n)` because of readability is ok while going to `O(n^2)` is not ok, and readability should suffer if that is the tradeoff we need to make in that case.

ea-factories/src/factories/balance.ts
justinkaseman: This is a good suggestion. Ended up being a lot cleaner. 

I originally split these because I was having type issues, but this way is a good approach.

ea-factories/src/factories/balance.ts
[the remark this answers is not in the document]
justinkaseman: Very true. I'm still in the bad habit of doing things "fancy" sometimes. Blanked on dropping the coefficient :sweat_smile: 

Great advice that should be added to your next book! :book:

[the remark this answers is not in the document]
krebernisak: Haha, don't worry, I only record YouTube videos nowadays. Books are boring and you can't put ads in them 😇

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  // If warning, clear and continue to processing
+  const { warning, ...accNoWarning } = account
+  return accNoWarning
+}
+
+const toGetBalances = (getBalance?: GetBalance) => (accounts: Account[], config: BalanceConfig) => {
+  if (!getBalance) throw new Error('Get Balance function not supplied')
+  return accounts.map(async (acc) => {
+    if (acc.warning) return { result: [acc] }
```
krebernisak, line 78: This is too late to filter warnings. If we do it here, we don't do it in the `config.getBalances` method.

There is also the problem of not keeping the original order for the sequence, grouping is messing up the order, and we need to restore it in the end.

First, we should filter out all not supported accounts before grouping:

```javascript
const supportedAccounts =  accounts.filter(a => !a.warning)
const groups = Array.from(util.groupBy<string, Account>(supportedAccounts, key).values())
const requests = ...
const responses = ...
```

Now that we have the  results back, we need to build an index of balances:

```javascript
const balances = new Map(responses.flatMap((r: any) => r.result).map(a => [a.address, a.balance])))
```

Then we build the response, using the original order:

```javascript
const data: SequenceResponseData<Account> = {
  responses: responses.map((r: any) => r.payload),
  result: accounts.map(a => {...a, balance: a.warning ? a.balance || balances.get(a.address)})
}
```

krebernisak (changes requested):

ea-factories/src/factories/balance.ts
justinkaseman: This is a good suggestion, thanks. I had the warnings getting filtered _within_ `config.getBalances`, but that would be better abstracted away. Plus maintaining the order seems better too.

composite/proof-of-reserves/src/balance.ts
```diff
@@ -1,15 +1,30 @@
...
 import blockchainCom from '@chainlink/blockchain.com-adapter'
+import blockchair from '@chainlink/blockchair-adapter'
 import blockcypher from '@chainlink/blockcypher-adapter'
+import btc_com from '@chainlink/btc.com-adapter'
+import cryptoapis from '@chainlink/cryptoapis-adapter'
+import sochain from '@chainlink/sochain-adapter'
 
 export type BitcoinIndexerOptions = { type?: BitcoinIndexer }
 export enum BitcoinIndexer {
+  Amberdata = 'amberdata',
```
krebernisak, line 12: This mapping between the enum name and the adapter defined `NAME` could pose a problem. I think that enums are too static for our use case here. I think it would be better if we just define `BitcoinIndexer` as string type alias and compute it from the map. Something like this:

```javascript
export type BitcoinIndexerOptions = { type?: BitcoinIndexer }
export type BitcoinIndexer = string

const implLookup: Implementations = {
  [amberdata.NAME]: amberdata,
  [blockchainCom.NAME]: blockchainCom,
  [blockcypher.NAME]: blockcypher,
  [blockchair.NAME]: blockchair,
  [btc_com.NAME]: btc_com,
  [cryptoapis.NAME]: cryptoapis,
  [sochain.NAME]: sochain,
}

const isBitcoinIndexer = (envVar?: string): envVar is BitcoinIndexer =>
  Object.keys(implLookup).includes(envVar as any)

export const getBitcoinIndexer = (): BitcoinIndexer | undefined => {
  const bitcoinIndexer = process.env.BTC_INDEXER_ADAPTER
  return isBitcoinIndexer(bitcoinIndexer) ? (bitcoinIndexer as BitcoinIndexer) : undefined
}

export const getImpl = (options: BitcoinIndexerOptions): Execute => {
  const prefix = options.type?.toUpperCase()
  const impl = options.type && implLookup[options.type]
  if (!impl) throw Error(`Unknown balance adapter type: ${options.type}`)

  return (data) => {
    const config = impl.makeConfig(prefix)
    const execute = impl.makeExecute(config)
    return execute(data)
  }
}
```

krebernisak: @HenryNguyen5 would love your input on how to best type this piece of code?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,120 @@
...
+import { util } from '@chainlink/ea-bootstrap'
+
+const DEFAULT_DATA_PATH = 'addresses'
+const DEFAULT_CONFIRMATIONS = 6
+
+const WARNING_UNSUPPORTED_PARAMS = 'No Operation: this provider does not support'
+const ERROR_MISSING_ADDRESS = 'No Operation: address param is missing'
+
+export function addresses(accounts: Account[]) {
+  return accounts.reduce<string[]>((accumulator, current) => {
```
krebernisak, line 19: We are just filtering accounts without a warning and mapping them to address, no need to do a more complex sequence reduction.

I also think that this function should not be exported. I see it used in only one place and we can instead do `accounts.map(a => a.address)` directly, when needed.

external-adapter/src/config.ts
```diff
@@ -5,26 +5,30 @@ import { logger } from './logger'
...
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 const cloneNoSecrets = (config: Config): Config => (({ apiKey, ...o }) => o)(config)
 
 export function getDefaultConfig(prefix = ''): Config {
   const apiKey = util.getEnv(ENV_API_KEY, prefix)
+  const timeout = util.getEnv(ENV_API_TIMEOUT, prefix)
   return {
     apiKey,
+    returnVerboseResponse: !!util.getEnv(ENV_API_VERBOSE_RESPONSE, prefix),
```
krebernisak, line 27: What is this used for?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export type IsSupported = (coin: string, chain: string) => boolean
+export type BalancesResponse = DataResponse<Account[], any>
+export type GetBalance = (account: Account, config: BalanceConfig) => Promise<BalancesResponse>
+export type GetBalances = (accounts: Account[], config: BalanceConfig) => Promise<BalancesResponse>
+
+export type BalanceConfig = Config & {
+  confirmations?: number
+  shouldOverwrite?: boolean
+  verbose?: boolean
```
krebernisak, line 33: This functionality to trigger verbose responses on/off should be part of the framework and available to all adapters to use. Let's not expand the scope of this PR further, but please open a GitHub issue documenting this so we can improve the implementation in the future and bring it to all adapters.

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,120 @@
...
+      message: ERROR_MISSING_ADDRESS,
+      statusCode: 400,
+    })
+
+  // Defaults
+  if (!account.chain) account.chain = 'mainnet'
+  if (!account.coin) account.coin = 'btc'
+
+  // Should we process?
+  if (!config.shouldOverwrite && typeof account.balance === 'number') return account
```
krebernisak, line 67: This flag is not documented and I am having trouble figuring out how it should be used and why it's useful?

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,120 @@
...
+    return { ...account, warning: WARNING_UNSUPPORTED_PARAMS + ` ${account.chain} ${account.coin}` }
+
+  // If warning, clear and continue to processing
+  const { warning, ...accNoWarning } = account
+  return accNoWarning
+}
+
+const toGetBalances = (getBalance?: GetBalance) => (accounts: Account[], config: BalanceConfig) => {
+  if (!getBalance) throw new Error('Get Balance function not supplied')
+  return accounts.map(async (acc) => {
```
krebernisak, line 81:
```suggestion
  return accounts.map((acc) => {
```

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (acc.warning) return { result: [acc] }
```
krebernisak, line 82: Warnings are handled all over the place in a number of different functions. I don't think that `toGetBalances` is a place to handle any warnings. This function should just map N accounts to N `getBalance` calls, and the data (with all the error/warning handlings) should be prepared before.

external-adapter/src/requester.ts
```diff
@@ -86,18 +87,17 @@ export class Requester {
     return new AdapterError({ jobRunID, statusCode, message: error }).toJSONResponse()
   }
 
-  static success(jobRunID = '1', response: AxiosResponse) {
-    if (!('result' in response.data)) {
```
krebernisak, line 90: What would be the reason to remove this?

typings/@chainlink/index.d.ts
```diff
@@ -50,37 +73,37 @@ declare module '@chainlink/types' {
...
     config: Config,
   ) => Promise<AdapterResponse>
 
-  export type ExecuteFactory = (config?: Config) => Execute
+  export type ExecuteFactory<C extends Config> = (config?: C) => Execute
 
-  import { expose } from '@chainlink/ea-bootstrap'
+  export type ConfigFactory = (prefix?: string) => Config
+
+  // import type { ExecuteHandlers } from '@chainlink/ea-bootstrap/src'
```
krebernisak, line 87: Please clean up these comments.

krebernisak (changes requested):
I left a few final comments, but overall, I think we are very close to merging this.

Great work @justinkaseman 💯

external-adapter/src/requester.ts
justinkaseman: The optional chaining on line 94 is more concise but if you find this more readable or there was a reason for `null` specifically can revert.

ea-factories/src/factories/balance.ts
justinkaseman: #211

ea-factories/src/factories/balance.ts
justinkaseman: So this should probably be brought to a separate issue as well (and possibly removed here for now).

This flag _specifically_ for composite adapters knowing whether they should write on top of previous results or if they should skip when a result is already filled. Composite adapter specific configs haven't been implemented yet and might need some thought.

external-adapter/src/config.ts
justinkaseman: This is used for if the adapter should return the whole data response or not. With composite adapters these responses could end up massive. Maybe we should default to true as that is the behaviour of every other current adapter?

krebernisak: But this specific flag `returnVerboseResponse`, I see it being defined, but not used anywhere?

external-adapter/src/requester.ts
krebernisak: Hmm... :) Disregard then.

ea-factories/src/factories/balance.ts
krebernisak: Let's remove this for now, as we are not using it at the moment. You can add your thoughts and ideas on this [here](https://github.com/smartcontractkit/external-adapters-js/issues/175). Thanks!

external-adapter/src/config.ts
justinkaseman: Ahhh you are right, hadn't changed the variable name in the ea-factories/balance code to match this name.

ea-factories/src/factories/balance.ts
justinkaseman: Removed for now :+1:

composite/outlier-detection/src/config.ts
```diff
@@ -1,9 +1,9 @@
 import { util } from '@chainlink/ea-bootstrap'
-import types, { Execute } from '@chainlink/types'
+import { Execute } from '@chainlink/types'
 import { getSourceDataProviders, getSourceImpl } from './source'
 import { getCheckDataProviders, getCheckImpl } from './check'
 
-export type Config = types.Config & {
+export type Config = {
```
justinkaseman, line 6: @boxhock Changed this bit from your composite Outlier-Detection PR as it was throwing a TS error.

Just want to make sure this still works for you as expected. Opened #217 which includes investigating that this Config type here still works with the underlying adapter's Config.

I think we may want to verify that Configs are working as expected here while doing

krebernisak: Yeah, it's not needed here.

sochain/README.md
```diff
@@ -0,0 +1,164 @@
...
+        "address": "38bzm6nhQMFJe71jJw1U7CbgNrVNpkonZF",
+        "coin": "btc",
+        "chain": "mainnet",
+        "balance": 307499838499
+      },
+      {
+        "address": "3ANaBZ6odMrzdg9xifgRNxAUFUxnReesws",
+        "coin": "btc",
+        "chain": "mainnet",
+        "balance": 904070305884
```
krebernisak, line 96: Some of our BTC indexer adapters return `"balance"` as strings, while other as numbers. We need to adapt those data providers to the same API, and my vote goes that we report all balances as strings (because big numbers).

There is also an issue where some adapters return decimal value of BTC, and other int value of SATS.

Let's update this so that every adapter returns **balance as string representing total amount of SATS** on the account.

cryptoapis/README.md
```diff
@@ -1,15 +1,22 @@
 # Chainlink External Adapter for CryptoAPIs
 
+## Configuration
+
+The adapter takes the following environment variables:
+
+- `API_KEY`: Optional Blochair API key to use
```
krebernisak, line 7:
```suggestion
- `API_KEY`: Optional API key to use
```

ea-factories/src/factories/balance.ts
```diff
@@ -0,0 +1,114 @@
...
+    responses.flatMap((r) => r.result).map((a) => [`${a.address}-${a.coin}-${a.chain}`, a]),
+  )
+
+  const data: SequenceResponseData<Account> = {
+    responses: responses.map((r: any) => r.payload),
+    result: accounts.map((a, i) => {
+      const hasResp = responseLookup[`${a.address}-${a.coin}-${a.chain}`]
+      if (hasResp) return hasResp
+      return a
+    }),
```
krebernisak, line 108:
```suggestion
    result: accounts.map((a, i) => responseLookup[`${a.address}-${a.coin}-${a.chain}`] || a),
```

external-adapter/src/config.ts
```diff
@@ -5,26 +5,30 @@ import { logger } from './logger'
...
   const apiKey = util.getEnv(ENV_API_KEY, prefix)
+  const timeout = util.getEnv(ENV_API_TIMEOUT, prefix)
   return {
     apiKey,
+    verbose: !!util.getEnv(ENV_API_VERBOSE, prefix),
     api: {
       withCredentials: !!apiKey,
       baseURL: util.getEnv(ENV_API_ENDPOINT, prefix),
-      timeout: parseInt(util.getEnv(ENV_API_TIMEOUT, prefix)) || DEFAULT_API_TIMEOUT,
+      timeout: timeout ? parseInt(timeout) : DEFAULT_API_TIMEOUT,
```
krebernisak, line 31: `parseInt` will just return a `NaN` on parse error, so we can just do:

```suggestion
      timeout: parseInt(timeout) || DEFAULT_API_TIMEOUT,
```

package.json
```diff
@@ -92,7 +92,7 @@
   "scripts": {
     "lint": "yarn workspaces run lint",
     "lint:fix": "yarn workspaces run lint:fix",
-    "setup": "yarn wsrun -mre -t setup",
+    "setup": "yarn wsrun -mre -p @chainlink/ea-bootstrap @chainlink/ea-factories @chainlink/external-adapter -t setup && yarn wsrun -mre -t setup",
```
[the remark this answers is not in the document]
krebernisak: Could you please open  a separate issue to keep track of this, thx!

typings/@chainlink/index.d.ts
```diff
@@ -19,13 +20,34 @@ declare module '@chainlink/types' {
...
+    api: Partial<AxiosRequestConfig>
+  }
+
+  /* RESPONSES */
+  export type DataResponse<R, P> = {
+    result: R
+    payload?: P
+  }
+
+  export type SequenceResponseData<R> = {
```
krebernisak, line 38: If the one above is named `DataResponse` let's also rename this as `SequenceDataResponse `

krebernisak: We'll also need to improve and structure these types in the future, but this should be good for now.

krebernisak (changes requested):

typings/@chainlink/index.d.ts
justinkaseman: Agreed, will open an issue for this so we remember.

blockchair/test/balance.test.ts
```diff
@@ -0,0 +1,10 @@
+import { makeExecute } from '../src/adapter'
+import { shouldBehaveLikeBalanceAdapter } from '@chainlink/adapter-test-helpers'
+
+shouldBehaveLikeBalanceAdapter(makeExecute(), [
+  'bitcoin_mainnet',
```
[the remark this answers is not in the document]
justinkaseman: Opened #223

amberdata/README.md
```diff
@@ -11,20 +24,118 @@ Gets the [latest spot VWAP price](https://docs.amberdata.io/reference#spot-price
...
+    "result": 12.02087667
+  },
+  "result": 12.02087667,
+  "statusCode": 200
+}
+```
+
+### Balance endpoint
+
+- `dataPath`: Optional path where to find the addresses array, defaults to `addresses`
```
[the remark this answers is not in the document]
justinkaseman: Resolved in this PR.

blockchair/README.md
```diff
@@ -1,83 +1,266 @@
...
+  "result": 19298087186263,
+  "statusCode": 200
+}
+```
+
+### Balance endpoint
+
+https://blockchair.com/api/docs#link_390
+
+\*Note because blockchair does not fill 0 balances, they are also not filled by this adapter
```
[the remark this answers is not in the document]
justinkaseman: Agreed, made the change to set balance to 0.

The only thing to think about here is if an address is invalid I believe it will also be setting the balance to 0.

blockcypher/README.md
```diff
@@ -117,43 +116,43 @@ Address object:
       "address": "3D8DJLwUXFfZvE8yJRu729MZ8uLy25SuLz",
       "coin": "btc",
       "chain": "mainnet",
-      "balance": 44900000000
+      "balance"":" 44900000000
```
krebernisak, line 119:
```suggestion
      "balance": "44900000000"
```

sochain/README.md
```diff
@@ -120,43 +121,43 @@ The adapter takes the following environment variables:
       "address": "3D8DJLwUXFfZvE8yJRu729MZ8uLy25SuLz",
       "coin": "btc",
       "chain": "mainnet",
-      "balance": 44900000000
+      "balance"":" 44900000000
```
krebernisak, line 124:
```suggestion
      "balance": "44900000000"
```

krebernisak (changes requested):

krebernisak (approved):

PR #185, opened 2020-12-08
Async execution fix

krebernisak:


emmick4 (approved):
Kristijan the 🐐, confirmed this works on lition and linkpool in infra-staging

PR #186, opened 2020-12-08
Query string params (post-typescript)

emmick4:
Moving [this PR](https://github.com/smartcontractkit/external-adapters-js/pull/173) here to avoid the ugly merge after typescript changes (my git-fu could use some improvement)
This allows us to pass arbitrary query string parameters to external adapters, useful for specifying a single external adapter as multiple data sources. For [this story](https://www.pivotaltracker.com/story/show/175901421)

Querystrings won't be enabled for us until https://github.com/smartcontractkit/infra-modules/pull/286 is merged, but the two PRs are independent in that they won't break anything if one is merged before the other.

### DIFF
--- a/bootstrap/src/lib/aws.ts
+++ b/bootstrap/src/lib/aws.ts
@@ -70,12 +70,12 @@ export const initHandlerHTTP = (execute: ExecuteSync) => (
   if (!isContentTypeSupported(event)) {
     return callback(null, UNSUPPORTED_MEDIA_TYPE_RESPONSE)
   }
-  const bodyIncludingQuery = JSON.parse(event.body)
-  bodyIncludingQuery.data = {
-    ...bodyIncludingQuery.data,
-    ...toObjectWithNumbers(bodyIncludingQuery.queryStringParameters || {}),
+  const body = JSON.parse(event.body)
+  body.data = {
+    ...body.data,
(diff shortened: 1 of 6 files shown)


bootstrap/src/lib/aws.ts
```diff
@@ -66,8 +70,12 @@ export const initHandlerHTTP = (execute: ExecuteSync) => (
...
     return callback(null, UNSUPPORTED_MEDIA_TYPE_RESPONSE)
   }
-
-  execute(JSON.parse(event.body), (statusCode, data) => {
+  const bodyIncludingQuery = JSON.parse(event.body)
+  bodyIncludingQuery.data = {
+    ...bodyIncludingQuery.data,
+    ...toObjectWithNumbers(bodyIncludingQuery.queryStringParameters || {}),
+  }
+  execute(bodyIncludingQuery, (statusCode, data) => {
```
krebernisak, line 78: This piece of code changed form, now I am ok with just calling it body.

```suggestion
  const body = JSON.parse(event.body)
  body.data = {
    ...body.data,
    ...toObjectWithNumbers(body.queryStringParameters || {}),
  }
  execute(body, (statusCode, data) => {
```

krebernisak (approved):
Great work on this! 💯

PR #190, opened 2020-12-10
[Feature] EA-factories

justinkaseman:
### Description
- Makes a shell of a package to start putting higher level factories for specific adapter type implementations
- Changes `makeExecute` to be a generic so that we can use specific Configs for each adapter type

### Caveats
This is a very barebones start!


typings/@chainlink/index.d.ts
```diff
@@ -65,7 +65,9 @@ declare module '@chainlink/types' {
     config: Config,
   ) => Promise<AdapterResponse>
 
-  export type ExecuteFactory = (config?: Config) => Execute
+  export interface ExecuteFactory<C extends Config> {
```
krebernisak, line 68: Any specific reason why you used an fn interface vs an fn type? Interfaces are more powerful, but in the case of functions, fn types read so much better :)

justinkaseman: Just habit, I see what you mean :smile:

krebernisak:
Closing this PR in favor of [Feature/proof of reserves #184](https://github.com/smartcontractkit/external-adapters-js/pull/184)

PR #193, opened 2020-12-13
Fix/update dxFeed symbols

johnnymugs:
- These symbols are for the OTC data rather than the licensed FTSE and N225 data (which is quite expensive). This is what node operators are using in production (Chainlayer, AlphaVantage)
- Their demo endpoint uses the licensed FTSE and N225 feeds, so you won't be able to use the develop against it after merging this
- We have a subscription that has these feeds, so you could mock a call for integration tests if necessary


krebernisak (approved):
Thanks, @johnnymugs!

PR #194, opened 2020-12-15
multiple api keys

emmick4:
Ready for review, not a huge change but should be a nice feature where it's needed. I know there was a provider mentioned that can use _any_ characters for its API key, so would just need to document that and change the delimiter for its adapter. I don't love that the API_KEY can be undefined or a string, kinda messes with the typescript expectations and leads to an awkward `if` in the util function since I don't want to change existing behavior for the case when the key's not provided. Would love it if anyone has ideas for a cleaner way to handle it.


bootstrap/src/lib/util.ts
```diff
@@ -22,6 +22,15 @@ export const toObjectWithNumbers = (obj: any) => {
   return Object.fromEntries(Object.entries(obj).map(([k, v]) => [k, toNumber(v)]))
 }
 
+// pick a random string after splitting with the delimiter ("a&b&c" "&" -> choice(["a","b","c"]))
+export const pickRandomFromString = (str: string, delimiter: string) => {
```
krebernisak, line 26: We can define the default delimiter here:

```suggestion
export const pickRandomFromString = (str: string, delimiter = ',') => {
```

krebernisak (changes requested):
Not really a fan of this solution as it brings more complexity to the incorrect place, but there is some infrastructure missing, so I understand this is the KISS way to bring this feature in now 👍 

@boxhock could you please check that "special one" API with funky chars in API keys.

In general, we need to move away from components reading their own configuration directly from the source, but they should just require it, and it gets injected in. The components should have no idea how that configuration is being generated.

Outside, whoever is building this component, needs to find a way to set up the configuration for it. That outside actor could use different config building strategies for different use cases.
One of the strategies could be to just read the config from the ENV, but there may be others where we read it from a JSON file, a remote server, etc.

This is something we will try to define and build in the near term.

bootstrap/src/lib/util.ts
```diff
@@ -22,6 +22,15 @@ export const toObjectWithNumbers = (obj: any) => {
...
+  if (typeof str !== 'string') {
```
RodrigoAD, line 27: I think I would be a good opportunity to use the `util.getRequiredEnv` helper. Will help with the undefined Env var too. Something like:
```js
export const getRandomEnv = (name: string, delimiter = ',', prefix = '') => {
    const val = getRequiredEnv(name, prefix)
    const items = str.split(delimiter)
    return items[Math.floor(Math.random() * items.length)]
}
```
And use it like: `const apiKey = util.getRandomEnv('API_KEY')`

boxhock:
> @boxhock could you please check that "special one" API with funky chars in API keys.
The one API key I have to sample against would work. How easy is it to change the delimiter though?
Can we make it an env var? So if the node operator sees they have a "," in their API key, they can set the delimiter themselves to something that's not in the key?

emmick4:
> The one API key I have to sample against would work. How easy is it to change the delimiter though?
> Can we make it an env var? So if the node operator sees they have a "," in their API key, they can set the delimiter themselves to something that's not in the key?

Seems like it's rare enough then that it isn't worth the added complexity of specifying via an env var, and if we know a provider uses commas in their keys we should explicitly set a working delimiter in its adapter.ts (seriously though who would do such a thing?)

bootstrap/src/lib/util.ts
emmick4: Ah yep great call, included your suggestion. Don't know how I missed that function while I was working in `util.ts`

bootstrap/src/lib/util.ts
emmick4: 👍 included `,` as the default delimiter

1forge/adapter.js
```diff
@@ -1,3 +1,4 @@
+const util = require('@chainlink/ea-bootstrap/dist/lib/util')
```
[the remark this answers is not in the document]
emmick4: IntelliJ let me down on this one 😅

emmick4:
> Not really a fan of this solution as it brings more complexity to the incorrect place, but there is some infrastructure missing, so I understand this is the KISS way to bring this feature in now 👍
> 
> @boxhock could you please check that "special one" API with funky chars in API keys.
> 
> In general, we need to move away from components reading their own configuration directly from the source, but they should just require it, and it gets injected in. The components should have no idea how that configuration is being generated.
> 
> Outside, whoever is building this component, needs to find a way to set up the configuration for it. That outside actor could use different config building strategies for different use cases.
> One of the strategies could be to just read the config from the ENV, but there may be others where we read it from a JSON file, a remote server, etc.
> 
> This is something we will try to define and build in the near term.

Agreed, definitely not ideal but aside from completely changing how we import configs it's about as clean as I could get it. Could use a story to define and implement that.

synth-index/kaiko/priceAdapter.js
```diff
@@ -1,13 +1,14 @@
...
+const { util } = require('@chainlink/ea-bootstrap')
 
 const getPriceData = async (synth) => {
   const url = `https://us.market-api.kaiko.io/v2/data/trades.v1/spot_exchange_rate/${synth.asset.toLowerCase()}/usd`
   const params = {
     sort: 'desc',
   }
   const headers = {
-    'X-Api-Key': process.env.API_KEY,
+    'X-Api-Key': util.pickRandomFromString(process.env.API_KEY, ","),
```
krebernisak, line 11: Looks like `util.pickRandomFromString` fn is not available anymore, right? Should be `getRandomRequiredEnv`.

synth-index/nomics/priceAdapter.js
```diff
@@ -10,7 +11,7 @@ const getPriceData = async (synths) => {
   const params = {
     ids: synths,
     convert: 'USD',
-    key: process.env.API_KEY,
+    key: util.pickRandomFromString(process.env.API_KEY, ","),
```
krebernisak, line 14: Should be `getRandomRequiredEnv`, as above.

krebernisak (changes requested):
The problem I see now is that we charged the behavior for these adapters. While it should be required to use an API_KEY in the production environment while testing it is useful to be able to test the API without the API_KEY, as most APIs probably have some kid of a free tier 🤔

krebernisak (approved):

PR #196, opened 2020-12-15
Dns adapter factory

RodrigoAD:
- DNS: Factory adapter
- DNS-record-check: Makes use of the DNS factory adapter. Checks if a certain record exists in the specified domain


dns-record-check/README.md
```diff
@@ -0,0 +1,35 @@
...
+    - `cloudfare`
+    - `google`
+
+## Input Params
+
+- `name`: Query Name, eg. "example.com"
+- `type`: Query Type (either a numeric value or text), eg. "TXT"
+- `do` (optional): DO bit - set if client wants DNSSEC data (either boolean or numeric value), eg. "true"
+- `cd` (optional): CD bit - set to disable validation (either boolean or numeric value), eg. "false"
+- `additional` (required): Record to check, eg: "{ "record": "adex-publisher" }"
```
krebernisak, line 19: Just call it a `record`

dns-record-check/README.md
```diff
@@ -0,0 +1,35 @@
...
+- `DNS_PROVIDER`: DNS provider to use. Options available:
+    - `cloudfare`
+    - `google`
+
+## Input Params
+
+- `name`: Query Name, eg. "example.com"
+- `type`: Query Type (either a numeric value or text), eg. "TXT"
+- `do` (optional): DO bit - set if client wants DNSSEC data (either boolean or numeric value), eg. "true"
+- `cd` (optional): CD bit - set to disable validation (either boolean or numeric value), eg. "false"
```
krebernisak, line 18: This `dns-record-check` adapter uses the underlying `dns-query` adapter. We don't need to duplicate the documentation of `dns-query` here in `dns-record-check` package. We should just link to the underlying adapter README, and point to stuff that should be configured but documented there.

dns-record-check/README.md
```diff
@@ -0,0 +1,35 @@
+# Chainlink External Adapter to Check DNS Records
```
krebernisak, line 1: This `dns-record-check` is not a standalone adapter, it needs to use the underlying `dns-query` adapter to complete its work. We call these types of adapters composite adapters and currently we group them in `/composite`.

We do this just to make it explicit, for now. In the future, we may accept that almost all adapters are some kind of composite and as such may stop grouping them in this way.

For now, I would prefer if we stay explicit and group these kinds of adapters in `/composite`.

dns-record-check/src/adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const execute: Execute = async (input) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const { Answer, record } = validator.validated.data
+
+  const foundRecord = Answer.find((ans: DNSResponseAnswer) => ans.data.includes(record))
```
krebernisak, line 19: This adapter does not get an `Answer` on input. It queries (delegates most of the work to) the underlying `dns-query` adapter to obtain an `Answer` and then tries to find a record.

dns-record-check/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+
+const NAME = 'Adex'
+
+export = { NAME, execute: makeExecute, ...expose(util.wrapExecute(makeExecute())) }
```
krebernisak, line 6: Now adapters only expose `makeExecute`, an `ExecuteFactory` that you can use to build the `Execute` function for the adapter.

```suggestion
export = { NAME, makeExecute, ...expose(util.wrapExecute(makeExecute())) }
```

ea-factories/src/factories/dns.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type DNSResponse = {
+  Status: number
+  Question: DNSResponseQuestion[]
+  Answer: DNSResponseAnswer[]
+}
+
+export type DNSConfig = Config & {
+  execute: Execute
```
krebernisak, line 27: We should configure this adapter with the URL of the underlying DNS Query API endpoint it should call.,

krebernisak (changes requested):

RodrigoAD:
Closing this in favor of #199

PR #197, opened 2020-12-15
Dns and adex adapter

RodrigoAD:
Provisional version until DNS Factory version is ready.

In order to use Adex and decouple DNS from it, there are 2 adapters, DNS and Adex.

- DNS: It queries `cloudfare` with the parameters provided
- Adex: It makes use of the response from previous adapter, to check if certain information is set in the DNS

### DIFF
--- a/adex/src/adapter.ts
+++ b/adex/src/adapter.ts
@@ -20,17 +20,11 @@ export const execute: Execute = async (input) => {
   const { Answer } = validator.validated.data
 
   const record = 'adex-publisher'
-  const publisher = Answer.find((ans: DNSAnswer) => {
-    return ans.data.includes(record)
-  })
+  const publisher = Answer.find((ans: DNSAnswer) => ans.data.includes(record))
 
-  try {
-    const response = {
-      status: 200,
-      data: { result: !!publisher },
-    }
-    return Requester.success(jobRunID, response)
-  } catch (e) {
-    console.log(e)
+  const response = {
+    status: 200,
+    data: { result: !!publisher },
   }
+  return Requester.success(jobRunID, response)
 }
--- a/adex/src/adapter.ts
+++ b/adex/src/adapter.ts
@@ -1,0 +1,36 @@
+import { Execute } from '@chainlink/types'
+import { Requester, Validator } from '@chainlink/external-adapter'
+
+const customParams = {
+  Answer: true,
+}
+
+type DNSAnswer = {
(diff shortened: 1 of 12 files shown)


dns/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+const customParams = {
+  name: true,
+  type: true,
+  do: false,
+  cd: false,
+}
+
+const baseUrl = 'https://cloudflare-dns.com/dns-query'
+
+export const execute: Execute = async (input) => {
```
krebernisak, line 13: The important thing with this data provider adapter is to add appropriate types and then map to/from those types for specific underlying data providers.

We are missing an input type, lets call it `DNSDataRequest`, and also an output type `DNSDataResponse`.

Then when we define those we can have N DNS data providers answering to our `DNSDataRequest` with `DNSDataResponse`. If the underlying API uses different types, we need to map from/to those types.

Then we'll have N DNS data providers that have the same API and we can easily build composite adapters from it, which are nothing more then simple data processing pipelines that use underlying data providers.

krebernisak: The DNS adapter factory is the base package where we can define all those types, configuration and validation. The the specific implementations would then just map from `DNSDataRequest` to the actual API request, and back from the response to the `DNSDataResponse`.

adex/src/adapter.ts
```diff
@@ -0,0 +1,30 @@
+import { Execute } from '@chainlink/types'
+import { Requester, Validator } from '@chainlink/external-adapter'
+
+const customParams = {
+  Answer: true,
+}
+
+type DNSAnswer = {
```
krebernisak, line 8: This should be part of the DNS adapter.

adex/src/adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const execute: Execute = async (input) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const { Answer } = validator.validated.data
+
+  const record = 'adex-publisher'
```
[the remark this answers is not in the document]
krebernisak: We should call this adapter something like `check-dns-record` and configure it with:

- DNS data provider we are using
- record we are searching for

krebernisak (changes requested):

PR #199, opened 2020-12-16
DNS query and record check adapters

RodrigoAD:
- `dns-query`: Let query DNS over Https 
- `dns-record-check`: Composite adapter that relies on `dns-query`, and check if a certain record exists in the DNS


composite/dns-record-check/src/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const { record, ...dnsInputParams } = validator.validated.data
+
+  const dnsRequest: AdapterRequest = {
+    id: jobRunID,
+    data: dnsInputParams,
+  }
+  const dnsResponse = await DNS.execute(dnsRequest)
```
krebernisak, line 22: If all input parameters are flat, which here is the case, you can just delegate the input to the underlying adapter.

```suggestion
  const dnsResponse = await DNS.execute(input)
```

dns-query/src/adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const execute: Execute = async (input) => {
+  try {
+    const validator = new Validator(input, inputParams)
+    if (validator.error) throw validator.error
+
+    const dnsProvider = util.getRequiredEnv('DNS_PROVIDER')
+    if (!Object.values(DNSProviders).includes(dnsProvider))
+      throw new Error(`Unknown DNS Provider: ${dnsProvider}`)
```
krebernisak, line 29: Please require all configuration to be given to you instead of reading it yourself from the ENV.

```suggestion
const execute: ExecuteWithConfig = async (input, config) => {
  try {
    const validator = new Validator(input, inputParams)
    if (validator.error) throw validator.error
```

Then you can export a `makeConfig` fn that will help create this config by reading from the ENV:

```javascript
export const makeConfig = () => {
    const provider = util.getRequiredEnv('DNS_PROVIDER')
    if (!Object.values(DNSProviders).includes(provider))
      throw new Error(`Unknown DNS Provider: ${provider}`)
    return { provider }
}

Comment (krebernisak) [reply dns-query/src/adapter.ts:29]: We could also add an `endpoint/url` as configuration, so we have a few predefined providers but also anyone can set up an adapter with custom `endpoint/url` without needing to update the code.

File: dns-query/src/adapter.ts [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+      type,
+      ...(doBit && { do: doBit }),
+      ...(cdBit && { cd: cdBit }),
+    }
+    const headers = {
+      Accept: 'application/dns-json',
+    }
+
+    const result = await Requester.request({
+      url: endpoints[dnsProvider],
```
krebernisak, line 45:
```suggestion
      url: config.endpoint || endpoints[config.provider],
```

dns-query/src/adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const data: DNSResponse = { ...result.data }
+
+    return Requester.success(jobRunID, {
+      status: 200,
+      data: data,
+    })
+  } catch (e) {
+    console.error(e)
+  }
```
krebernisak, line 58: Is this leftover from debugging? No need to catch any exceptions here, just pass them through.
```suggestion
```

typings/@chainlink/index.d.ts
```diff
@@ -82,6 +82,24 @@ declare module '@chainlink/types' {
...
+    type: number
+    TTL: number
+    data: string
+  }
+
+  export type DNSResponse = {
+    Status: number
+    Question: DNSResponseQuestion[]
+    Answer: DNSResponseAnswer[]
+  }
```
krebernisak, line 102: Please move these types to the `dns-query` adapter, as they are specific for that domain.

Our DNS base adapter should be the source of all DNS domain types.

krebernisak: This `typings/@chainlink` was a bucket for all types because our core packages were not written in TS. Now that we migrated those libs, we can start moving those types there.

composite/dns-record-check/README.md
```diff
@@ -0,0 +1,30 @@
+# Chainlink External Adapter to Check DNS Records
```
krebernisak, line 1: Add this adapter to our list of adapters being built by the CI.

dns-query/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { execute, inputParams } from './adapter'
+
+const NAME = 'DNS-Query'
+
+export = { NAME, inputParams, execute, ...expose(util.wrapExecute(execute)) }
```
krebernisak, line 6: No need to export the `inputParams`, but you will need to export the `makeExecute` & `makeConfig` factory functions:

```suggestion
export = { NAME, makeExecute, makeConfig, ...expose(util.wrapExecute(makeExecute())) }
```

dns-query/src/adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+}
```
krebernisak, line 59: We just export the factory fn that anyone can use to build its function, by providing some configuration or not in which case we will build the default config for them:

```javascript
export const makeExecute: ExecuteFactory = (config?: Config) => (input) => execute(input, config || makeConfig())
```

composite/dns-record-check/src/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Validator } from '@chainlink/external-adapter'
+import DNS from '@chainlink/dns-query'
+import { Requester } from '@chainlink/external-adapter'
+
+const inputParams = {
+  record: true,
+  ...DNS.inputParams,
+}
+
+export const execute: Execute = async (input) => {
```
krebernisak, line 11: Same story here, we don't expose the raw function, but a factory to build the function.

```suggestion
const execute: ExecuteWithConfig = async (input, config) => {
```

```javascript
export const makeExecute: ExecuteFactory = (config?: Config) => (input) => execute(input, config || makeConfig())
```

The question is what is the required `Config` type? Well in this case it is the config of the underlying `dns-query` adapter, and nothing else as it looks like this adapter does not require any specific configuration. It just needs to configure the underlying adapter.

composite/dns-record-check/src/adapter.ts
krebernisak: When other changes are made to this PR, in the end, it should look like something like this:

```suggestion
  const dnsQueryExecute = DNS.makeExecute(config)
  const dnsResponse = await dnsQueryExecute(input)
```

In the future with a more fluid API definition, that could transition to something like this:

```javascript
const resp = await dnsQuery.make(config).execute(input)
```

krebernisak (changes requested):
Looking nice! Few more suggestions :)

dns-query/src/adapter.ts
RodrigoAD: Do we want that to be possible? They could then call any endpoint, even if it's not a DNS one

composite/dns-record-check/src/adapter.ts
```diff
@@ -0,0 +1,33 @@
+import { ExecuteWithConfig, ExecuteFactory, Config } from '@chainlink/types'
+import { Validator } from '@chainlink/external-adapter'
+import { Requester } from '@chainlink/external-adapter'
+import DNS from '@chainlink/dns-query'
+import { DNSResponse, DNSResponseAnswer } from '@chainlink/dns-query/dist/types'
```
RodrigoAD, line 5: Not really fan of this

composite/dns-record-check/src/adapter.ts
```diff
@@ -0,0 +1,33 @@
...
+import { makeConfig } from './config'
+
+const inputParams = {
+  record: true,
+}
+
+const execute: ExecuteWithConfig = async (input, config) => {
```
RodrigoAD, line 12: Do we want to use anything from config here?

composite/dns-record-check/src/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const makeConfig = (): Config => {
```
RodrigoAD, line 4: For now I haven't found any use case for config, I just return the default one

dns-query/src/adapter.ts
RodrigoAD: It was for debugging. But note that the adapter doesn't log any error right now

dns-query/src/adapter.ts
krebernisak: Yeah, why not enable that? If they provide a non-compatible endpoint their adapter won't work. I am ok with that.

dns-query/src/adapter.ts
krebernisak: It's returned to the caller but not logged? We can introduce error logging in the `bootstrap` module if that is the case.

composite/dns-record-check/src/adapter.ts
krebernisak: We need the config to set up the underlying `dns-query` adapter, so just pass it on to it if this adapter does not have any use for it.

composite/dns-record-check/src/config.ts
krebernisak: How does the `dns-record-check` configure the underlying adapter it uses? That underlying adapter requires some configuration, right?

composite/dns-record-check/src/config.ts
```diff
@@ -0,0 +1,6 @@
...
+  return Requester.getDefaultConfig()
```
krebernisak, line 5:
```suggestion
  return DNS.makeConfig()
```

composite/dns-record-check/src/adapter.ts
```diff
@@ -1,25 +1,23 @@
...
   const jobRunID = validator.validated.id
-  const { record, ...dnsInputParams } = validator.validated.data
+  const { record } = validator.validated.data
 
-  const dnsRequest: AdapterRequest = {
-    id: jobRunID,
-    data: dnsInputParams,
-  }
-  const dnsResponse = await DNS.execute(dnsRequest)
+  const dnsExecute = DNS.makeExecute(DNS.makeConfig())
```
krebernisak, line 19:
```suggestion
  const dnsExecute = DNS.makeExecute(config)
```

or we can namespace it:

```suggestion
  const dnsExecute = DNS.makeExecute(config.dnsQuery)
```

krebernisak (changes requested):

composite/dns-record-check/src/config.ts
RodrigoAD: Yes that makes sense

dns-query/src/types.ts
```diff
@@ -0,0 +1,17 @@
...
+  type: number
+  TTL: number
+  data: string
+}
+
+export type DNSResponse = {
+  Status: number
+  Question: DNSResponseQuestion[]
+  Answer: DNSResponseAnswer[]
+}
```
krebernisak, line 17:
```suggestion
export type DNSQuestion = {
  name: string
  type: number
}

export type DNSAnswer = {
  name: string
  type: number
  TTL: number
  data: string
}

export type DNSQueryResponse = {
  Status: number
  TC: boolean
  RD: boolean
  RA: boolean
  AD: boolean
  CD: boolean
  Question: DNSQuestion[]
  Answer: DNSAnswer[]
  Comment?: string
}
```

dns-query/src/adapter.ts
RodrigoAD: We can open an issue on it

krebernisak: Yeah, please open an issue so we can keep track of it. Thanks!

krebernisak (approved):
Nice job @RodrigoAD! 💯

dns-query/src/adapter.ts
krebernisak: [Log adapter errors #204](https://github.com/smartcontractkit/external-adapters-js/issues/204) issue is opened to track this.

Issue #204, opened 2020-12-18
Log adapter errors


RodrigoAD:
Currently if there is some internal error, it just get notified to the client with a `500` without logging anything. Logging this would be really helpful for debugging and recognize faster possible problems in production adapters

justinkaseman:
I took a glance at this while working on the log size, do you have some steps to reproduce?

For example when I run `amberdata` without an API key the client gets:
```JSON
{"jobRunID":"1","status":"errored","statusCode":500,"error":{"name":"AdapterError","message":"Invalid value \"undefined\" for header \"x-api-key\""}}
```
and the node operator sees:
```JSON
{"message":"Could not reach endpoint: \"Invalid value \\\"undefined\\\" for header \\\"x-api-key\\\"\"","level":"error","instanceId":"68b86ad1-c120-47f2-a004-5245a3dff690","timestamp":"2021-01-07T18:15:26.890Z"}
```
---

Giving an invalid `Requester.validateResultNumber` path gives the following to the client:
```JSON
{"jobRunID":"1","status":"errored","statusCode":500,"error":{"name":"AdapterError","message":"Result could not be found in path"}}
```
and to the node operator:
```JSON
{"message":"Result could not be found in path","level":"error","instanceId":"bad3f72e-ff7b-4853-896f-354e4bd03e2d","timestamp":"2021-01-07T18:26:39.447Z"}
```

which also seem fine to me.

justinkaseman:
Following up - reproduced this while acceptance testing the Proof of Reserves composite adapter.

RodrigoAD:
Those errors seem that are caught by the Requester. If you introduce any error, ie a `throw` inside the adapter execute function, the adapter will fail without notifying anything.

boxhock:
This should be covered by separate stories in Shortcut now

PR #212, opened 2021-01-06
Feature/logger

justinkaseman:
### Description
Node operator's log files have been growing very large. The Coin Paprika adapter for example had a >100Gb logfile.

This PR removes the log emitted when a response is returned to a request. Logs will still be shown on error or when LOG_LEVEL="debug".

Additionally some other logs that should probably be better served to debug level logs are changed to be so.

### Ticket
[#176137608](https://www.pivotaltracker.com/story/show/176137608)


justinkaseman:
Addresses #204 that was experienced when using the Node.js Error class:
```
throw new Error(xxxx)
```

external-adapter/src/requester.ts
```diff
@@ -82,7 +86,13 @@ export class Requester {
   static errored(jobRunID = '1', error?: AdapterError | Error | string, statusCode = 500) {
     if (error instanceof AdapterError) return error.toJSONResponse()
     if (error instanceof Error)
-      return new AdapterError({ jobRunID, statusCode, cause: error }).toJSONResponse()
+      return new AdapterError({
+        jobRunID,
+        statusCode,
+        message: error.message,
+        name: error.name,
+        cause: error.stack,
```
krebernisak, line 94: Why did you make this change?

I would keep these two errors separate. The cause and the `AdapterError` that wraps the cause. Here you basically merged it in one. On the output as a user, I would like to see an `AdapterError` with a specific error message but also the ability to investigate the underlying cause.

external-adapter/src/requester.ts
```diff
@@ -53,7 +53,11 @@ export class Requester {
       }
 
       // Success
-      logger.info(`Received response: ${JSON.stringify(response.data)}`)
+      const { request: r, ...rest } = response
+      logger.debug({
+        message: 'Received response',
+        ...rest,
```
krebernisak, line 59: We logged just the `response.data` before. Why the change?

bootstrap/src/index.ts
```diff
@@ -42,9 +43,9 @@ const withStatusCode = (execute: ExecuteWrappedResponse) => async (data_: Adapte
 
 // Log adapter input & output data
 const withLogger = (execute: ExecuteWrappedResponse) => async (data: AdapterRequest) => {
-  logger.debug('Input: ', { input: data })
+  logger.debug({ input: data })
   const result = await execute(data)
-  logger.debug(`Output: [${result.statusCode}]: `, { output: result.data })
+  logger.debug({ output: result.data })
```
krebernisak, line 48: Why remove the messages from this log outputs?

bootstrap/src/index.ts
```diff
@@ -77,7 +78,10 @@ const executeSync = (execute: ExecuteWrappedResponse): ExecuteSync => {
     return withMiddleware(execute)
       .then((executeWithMiddleware) => executeWithMiddleware(data))
       .then((result) => callback(result.statusCode, result.data))
-      .catch((error) => callback(error.statusCode || 500, Requester.errored(data.id, error)))
+      .catch((error) => {
+        logger.error(error.toString(), { stack: error.stack })
```
krebernisak, line 82: Please add this to the available `withLogger`, so we don't interrupt this specific flow. There you can `try {...} catch`  the error to log it and throw again.

external-adapter/src/requester.ts
krebernisak: This is also short enough to be a one-liner.

krebernisak (changes requested):

bootstrap/src/index.ts
justinkaseman:
```
{
  output: {
    jobRunID: '1',
    data: { result: '1316957652992' },
    result: '1316957652992',
    statusCode: 200
  },
  level: 'debug',
  message: 'Output: [200]: ',
  instanceId: '85070dc4-cdc7-4ac4-9269-5a9bdedee369',
  timestamp: '2021-01-12T19:02:36.199Z'
}
```

`output` can already been seen as the key to the data object. Doesn't seem to add any new useful information.

external-adapter/src/requester.ts
justinkaseman: In my testing when the data response was logged it includes the `request` object which is a massive (200+ lines) reference to Axios that doesn't seem add to present any useful information. This takes out the request property.

external-adapter/src/requester.ts
justinkaseman: The reason I made this change is that as a user the error cause shows the stack trace from the `AdapterError` but not the underlying Error that would be useful for debugging the underlying cause.

ie. throwing a `new Error("TESTING THROW ERROR")` from inside an EA gives

```
{"jobRunID":"1","status":"errored","statusCode":500,"error":{"name":"AdapterError","message":"TESTING THROW ERROR","stack":"AdapterError: TESTING THROW ERROR\n    at Function.errored (/home/justinkaseman/Desktop/dev/external-adapters-js/external-adapter/dist/requester.js:86:20)\n    at /home/justinkaseman/Desktop/dev/external-adapters-js/bootstrap/dist/index.js:88:94\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)","cause":{}}}
```

where as the actual useful stack trace is:

```
Error: TESTING THROW ERROR
    at getBalance (/home/justinkaseman/Desktop/dev/external-adapters-js/blockchain.com/dist/endpoint/balance.js:11:11)
    at /home/justinkaseman/Desktop/dev/external-adapters-js/ea-factories/dist/factories/balance.js:49:34
    at Array.map (<anonymous>)
    at /home/justinkaseman/Desktop/dev/external-adapters-js/ea-factories/dist/factories/balance.js:49:21
    at /home/justinkaseman/Desktop/dev/external-adapters-js/ea-factories/dist/factories/balance.js:71:48
    at Array.flatMap (<anonymous>)
    at /home/justinkaseman/Desktop/dev/external-adapters-js/ea-factories/dist/factories/balance.js:71:29
    at Object.exports.execute (/home/justinkaseman/Desktop/dev/external-adapters-js/blockchain.com/dist/adapter.js:20:58)
    at /home/justinkaseman/Desktop/dev/external-adapters-js/blockchain.com/dist/adapter.js:32:39
    at /home/justinkaseman/Desktop/dev/external-adapters-js/composite/proof-of-reserves/dist/balance.js:40:16 object

```

justinkaseman:
@krebernisak Left some reasoning on why I made certain changes

external-adapter/src/requester.ts
krebernisak: Ok, that makes a lot of sense.

This is what the [AxiosResponse](https://github.com/axios/axios/blob/fe52a611efe756328a93709bbf5265756275d70d/index.d.ts#L76) looks like. I would suggest to log only the `{ data, status, statusCode }`, and remove the `request`, `config` and `headers` as those could leak secrets out.

RodrigoAD:
I've been playing with it, this is great. Just something I noticed, when there's an internal error, this is sent to the client as well. I don't know if this is the behavior we want

justinkaseman:
> I've been playing with it, this is great. Just something I noticed, when there's an internal error, this is sent to the client as well. I don't know if this is the behavior we want

I was wondering about which behaviour we want there too. My thought process is that we are open source and want to empower people to debug on their own. @krebernisak thoughts? This is related to that first change that you requested.

krebernisak:
My thoughts are that we should handle most/all of errors, output them with correct status codes and appropriate messages. We have work to do here.

Internal errors should be caught, mapped to a HTTP 500 status code and construct an appropriate message. Then if we are in dev env (NODE_ENV=development) or some other `verbose/debug` flag is set, we dump the whole internal error on the output.

bootstrap/src/index.ts
krebernisak: It adds a `message`, you can think of it as a `label` that logs can be indexed by and then searched. How do we find this specific log otherwise?

We already have our metrics dependant on some of our log messages. For example while tracking caching hits/misses, we look for following logs by its `message/label`:

```javascript
logger.debug(`Cache: SET ${key}`, entry)
```

Think about how would we do that if we log just the `entry` here...

external-adapter/src/requester.ts
justinkaseman: Reverted this

bootstrap/src/index.ts
justinkaseman: Reverted this

justinkaseman:
> My thoughts are that we should handle most/all of errors, output them with correct status codes and appropriate messages. We have work to do here.
> 
> Internal errors should be caught, mapped to a HTTP 500 status code and construct an appropriate message. Something like `Internal Server Error: ${error.name}`. Then if we are in dev env (NODE_ENV=development) or some other `verbose/debug` flag is set, we dump the whole internal error on the output, as an underlying `cause` to be investigated.

Well said for a direction forward. Agreed that there is more work to do here. I reverted the code that had requested changes, leaving this PR to be: adding the `verbose/debug` internal error dump and setting most logs to be `verbose/debug` level to keep node operator log files small.

krebernisak (approved):
Thanks, @justinkaseman!

PR #213, opened 2021-01-07
dydx-stark adapter

krebernisak:


dydx-stark/src/config.ts
```diff
@@ -0,0 +1,37 @@
...
+  starkMessage: string
+  oracleName: string
+}
+
+export const makeConfig = (prefix?: string): Config => {
+  const defaultConfig = Requester.getDefaultConfig(prefix)
+  defaultConfig.api.baseURL = defaultConfig.api.baseURL || DEFAULT_API_ENDPOINT
+  return {
+    ...defaultConfig,
+    privateKey: util.getRequiredEnv(ENV_PRIVATE_KEY, prefix),
```
justinkaseman, line 26: Are these required environment variables or optional?

krebernisak: These are required.

But this adapter does not require an API_KEY, even in production; @emmick4 that feature where we now require an API_KEY for every adapter is posing a problem.

emmick4: yep running into that now updating adapters to typescript, thinking we could just add a `required` flag to the function where the keys are actually optional

dydx-stark/src/endpoint/send.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+  const validator = new Validator(request, customParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const { asset, ...data } = validator.validated.data
+
+  const dataPath = data.dataPath || DEFAULT_DATA_PATH
+  const price = <number | string>objectPath.get(data, dataPath)
```
boxhock, line 23: Why do we parse the dataPath here, instead of using the copy adapter in the task pipeline in the job?

krebernisak: Just so there is another option for example when using this from another composite adapter. By default, it'll look in the `data.result` so it should not need a copy adapter or a specific `dataPath` set, by default.

boxhock: Ok. Seems like it's a bit premature feature (haven't seen the need for it before as most adapters pre-set the result key, but maybe I just don't understand the usecase), but looks good otherwise.

boxhock (approved):

dydx-stark/src/config.ts
krebernisak: I've opened an issue for it and assigned it to you: [Improve support for multiple keys in API_KEY env variable #219](https://github.com/smartcontractkit/external-adapters-js/issues/219)

Please make these few changes when you get a chance. Thx!

PR #214, opened 2021-01-08
New Adapter Paxos

emmick4:
Creates a new adapter for checking Paxos asset supply attestations.


paxos/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'PAXOS'
```

paxos/src/endpoint/assetAttestations.ts
```diff
@@ -0,0 +1,52 @@
...
+  lastAttestedAt: string
+  amount: number
+  verified: boolean
+}
+
+const getAttestationURI = (asset: string) => `/asset-attestations/${asset.toUpperCase()}`
+
+const toAttestation = async (config: Config, asset: string): Promise<Attestation> => {
+  const url = getAttestationURI(asset)
+  const params = {}
```
justinkaseman, line 29: Could remove this here and the next line unless you are intending to clear out `params`

justinkaseman (changes requested):

thodges-gh (approved):

PR #215, opened 2021-01-08
Add block height in relevant adapters

ebarakos:


ebarakos:
Currently, the common interface between all 4(3 APIs + json-rpc) adapters that return bitcoin info is: `{ "blockchain": "BTC", q: "difficulty"}` & `{ "blockchain": "BTC", q: "height"}`.

I didn't add many validations on q, so it can potentially support other parameters as well.
Tried to make it backwards compatible, in case previous adapter params are already used by node operators. (in cryptoapis you can use _endpoint_ instead of _q_ and _q_ param defaults to _difficulty_ in other adapters).

We can improve the uniform interface/naming for blockchain info/stats, either now or in the future and choose a convention either a) created by us b) identical to the _getblockchainfo_ params naming c) one of external APIs, like cryptoapis

blockchair/adapter.js
```diff
@@ -13,25 +14,31 @@ const convertBlockchain = {
...
+  height: 'blocks',
+}
+
 const execute = (input, callback) => {
   const validator = new Validator(input, customParams)
   if (validator.error) return callback(validator.error.statusCode, validator.errored)
 
   const jobRunID = validator.validated.id
   let blockchain = validator.validated.data.blockchain
+  let q = validator.validated.data.q || 'difficulty'
```
krebernisak, line 27: We currently use an `endpoint` input param to differentiate between different data points we want to fetch from the external data provider.

I would suggest we don't introduce new behavior to the adapter API at this point but continue using what we have now.

Take a look at how we handle endpoint queries in different providers, and please try to mimic the interface and behavior here.

---

In the future, we are looking to streamline the adapter API. The first thing we should do is list the data points we support and codify them. Here you are adding support for a new data point, let's call it `BlockNumber`. All of our adapters that know how to answer requests for that specific data point should follow the same API (input/output).

In this context, it may make sense to wrap the `json-rpc` adapter into `bitcoin-json-rpc` adapter that will then expose this specific data point (and others).

cryptoapis/adapter.js
```diff
@@ -38,31 +39,32 @@ const price = (jobRunID, input, callback) => {
...
+const latestBlockParams = {
   blockchain: ['blockchain', 'coin'],
+  endpoint: false,
+  q: false,
   network: false,
 }
 
-const difficulty = (jobRunID, input, callback) => {
-  const validator = new Validator(input, difficultyParams)
+const latestBlock = (jobRunID, input, callback) => {
```
krebernisak, line 49: I like how you renamed this as `latestBlock`, as this actually is the `latestBlock` endpoint from the `cryptoapis` data provider. We also call our input param `endpoint`, but IMHO this is the wrong name for this input param.

What our input param should be called is a `dataPoint`, which we then map to 1 (or more) underlying API endpoints from which we fetch data.

Here the `cryptoapis` adapter supports two data points (we call them `endpoint`) and for both of those (difficulty|height) we fetch data from the same underlying endpoint (latestBlock).

krebernisak (changes requested):
Thanks, @ebarakos! 

I tried to leave some comments and ideas, but we should probably also sync offline on this.

In general:

1. Please use the existing `endpoint` input param instead of a new one `q` (take a look at existing adapters using `endpoint` on input)
2. Don't add custom logic to our general `json-rpc` adapter,  but you can make a new `bitcoin-json-rpc` adapter, that wraps/uses the underlying `json-rpc` but follows our standard API.
3. In the future we should rename our `endpoint` input param as `dataPoint` and codify all data point types that we support in TS + docs

ebarakos:
Thank you @krebernisak for your helpful review. 

I will revert q to endpoint, along with the other changes, and we can tackle this in the future, across the whole codebase, as you mentioned.

json-rpc/src/adapter.ts
```diff
@@ -5,16 +5,32 @@ const inputParams = {
   url: false,
   method: false,
   params: false,
+  blockchain: false,
```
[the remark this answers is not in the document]
ebarakos: Yes, it feels weird changing the general adapter. 

@thodges-gh any input on this? Should we create a new adapter for achieving uniformity in JobSpecs, just for this specific case? 

Or maybe tackle this with abstractions(either on the context of this task or separately), like data translations/mapping from specific adapter results to some broadly-used interfaces?

ebarakos:
Created `bitcoin-json-rpc` as composite adapter. It can be used with the same parameters as `json-rpc`, but also adheres to the `{ "blockchain": "BTC", endpoint: "difficulty"}` & `{ "blockchain": "BTC", endpoint: "height"}` convention of other relevant adapters (`blockchain` param can be omitted).

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
+import JSONRPC from '@chainlink/json-rpc-adapter'
+import { Execute, AdapterRequest } from '@chainlink/types'
```
RodrigoAD, line 2: In general I would recommend you to follow any of the recent PR's. [`dns-record-check`](https://github.com/smartcontractkit/external-adapters-js/pull/199) is very recent and similar. I think it could help

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Validator } from '@chainlink/external-adapter'
+import { Requester } from '@chainlink/external-adapter'
+
+const inputParams = {
+  url: false,
+  method: false,
+  params: false,
```
RodrigoAD, line 9: `url`, `method` and `params` will be handled by JSONRPC adapter. You can just forward the input to the json adapter an it will handle it

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Export function to integrate with Chainlink node
+export const execute: Execute = async (request: AdapterRequest) => {
+  const validator = new Validator(request, inputParams)
+  const blockchain = validator.validated.data.blockchain || validator.validated.data.coin
+
+  let endpoint = validator.validated.data.endpoint
+
+  if (validator.error) throw validator.error
```
RodrigoAD, line 26: We should throw the validation error right after validating right?
```suggestion
   const validator = new Validator(request, inputParams)
   if (validator.error) throw validator.error
  
  const blockchain = validator.validated.data.blockchain || validator.validated.data.coin
  let endpoint = validator.validated.data.endpoint
```

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (endpoint != undefined || blockchain != undefined) {
```
RodrigoAD, line 27: This block is a bit hard to understand

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (endpoint in convertEndpoint) endpoint = convertEndpoint[endpoint]
+    if (!endpoint) endpoint = 'difficulty'
+    request.data.method = 'getblockchaininfo'
```
RodrigoAD, line 30: We want method only to be 'getblockchaininfo' only if we don't have endpoint or blockchain params? Default values could come from `makeConfig` better.
Also, let's try to not modify directly the `request` object.

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  const response = await JSONRPC.execute(request)
+
+  if (endpoint) {
```
RodrigoAD, line 35: I think you are repeating the same logic as before here? `endpoint` will be defined already right?

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (endpoint in convertEndpoint) endpoint = convertEndpoint[endpoint]
+    response.result = Requester.validateResultNumber(response.data, ['result', endpoint])
+    // data are returned in result, due to the called adapter, needs to be moved to data object
+    response.data = response.data.result
+    response.data.result = response.result
+  }
+
+  return response
```
RodrigoAD, line 43: I think this could be refactored to something like, avoiding mutating response as well
```suggestion
  const result = Requester.validateResultNumber(response.data, ['result', endpoint])
  return {
    result,
    data: { ...response.data.result, result: result },
  }
```

cryptoid/src/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+  const blockchain = validator.validated.data.blockchain.toLowerCase()
+
+  const key = util.getRandomRequiredEnv('API_KEY')
+  const q = endpointToQ[endpoint]
```
RodrigoAD, line 27: What is `q`? Could it have a more explanatory name?

cryptoid/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
RodrigoAD, line 5: I think you forgot to change the `NAME`

composite/bitcoin-json-rpc/src/adapter.ts
ebarakos: True, I had already checked it.

composite/bitcoin-json-rpc/src/adapter.ts
ebarakos: I just left them here, so this composite adapter can inherit the actual functionality of the json-rpc one.

composite/bitcoin-json-rpc/src/adapter.ts
ebarakos: True, changing this.

cryptoid/src/adapter.ts
ebarakos: q is a param of the cryptoid endpoint.

composite/bitcoin-json-rpc/src/adapter.ts
ebarakos: Very good, thank you.

composite/bitcoin-json-rpc/src/adapter.ts
ebarakos: Reverted to simpler version.

cryptoid/src/adapter.ts
RodrigoAD: I would give the var an explicit name, even if it will be a param of the request, but we'll know what is it.
```js
const someExplicitName = endpointToXXX[endpoint]
const params = { key, q: someExplicitName }
```

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,36 @@
...
+// Export function to integrate with Chainlink node
+export const execute: ExecuteWithConfig<Config> = async (request: AdapterRequest) => {
+  const validator = new Validator(request, inputParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  let endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+
+  const jsonrpcRequest: AdapterRequest = { id: jobRunID, data: { method: 'getblockchaininfo' } }
+  const response = await JSONRPC.execute(jsonrpcRequest)
```
RodrigoAD, line 23: You could include the incoming `inputRequest` into the `jsonrpcRequest` so you give this adapter the option of having the input params of the `json-rpc` adapter.  Something like:
```suggestion
const response = await JSONRPC.execute({
    ...request,
    data: { ...request.data, method: 'getblockchaininfo' },
  })
```

composite/bitcoin-json-rpc/README.md
```diff
@@ -0,0 +1,87 @@
+# Bitcoin JSON-RPC Blockchain info adapter for Chainlink
+
+- Supports AWS Lambda and GCP Functions
+- Takes optional connection to RPC endpoint (set via `RPC_URL` environment variable)
+
+## Input Params
+
+Returns blockchain info stats, by calling `"method": "getblockchainfo"`
+
+- `endpoint`: The parameter to query for. Default: "difficulty"
```
RodrigoAD, line 10: Specify that it relies on `json-rpc` adapter

composite/bitcoin-json-rpc/README.md
```diff
@@ -0,0 +1,87 @@
...
+  "statusCode": 200
+}
+```
+
+## Install
+
+Install dependencies
+
+```bash
+npm install
```
RodrigoAD, line 72: Better using `yarn` so we align with the rest of the documentation

RodrigoAD (commented):
Thanks @ebarakos for the changes! Looking great, just some final notes

RodrigoAD:
Take a look at the failing tests of `bitcoin-rpc` adapter, the build is trying to fetch a local node

cryptoid/src/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+  const blockchain = validator.validated.data.blockchain.toLowerCase()
+
+  const key = util.getRandomRequiredEnv('API_KEY')
+  const apiParam = endpointToApiParam[endpoint]
```
RodrigoAD, line 27: Sorry for being picky here, as I understand `q` param is the function you want to call from cryptoId, so something like `cryptoIdFunction` or `apiFunction` would be better.

ebarakos: `q` is the query parameter of the cryptoid endpoint. Not sure if it should be called function. In essence though, it has identical name as the JSON RPC methods.

I was thinking of something to let us know that the endpoint is mapped to the actual api's parameter.

ebarakos: Changed to `apiFunctionName` after discussion.

cryptoapis/src/endpoint/stats.ts
```diff
@@ -1,24 +1,32 @@
...
 import { AdapterRequest, Config } from '@chainlink/types'
+import { DEFAULT_ENDPOINT } from '../config'
 
-export const Name = 'difficulty'
+export const Name = 'stats'
 
-const difficultyParams = {
+const statsParams = {
   blockchain: ['blockchain', 'coin'],
+  endpoint: false,
```
krebernisak, line 9: We should not reuse the name "endpoint" in our internal API. It just does not fit, just as it does not fit in the external API, but we are currently stuck with it there because of backward compatibility.

Internally, we should structure this differently.

So let's take this adapter and endpoint as an example, the "stats" endpoint.
The stats endpoint impl should have no knowledge of a concept like "height", this is a concept of our external API, it has nothing to do with "stats" endpoint. We should map a CL "height" request to a Cryptoapis "stats" request, but the impl for Cryptoapis "stats" request should have no knowledge of the outside CL "height" request.

The input to "stats" endpoint should look something like this:

```javascript
const inputParams = {
  blockchain: true,
  network: false,
  path: false  
}
```

It makes a "stats" requests, and returns the data point at "path". If the "path" is empty, it returns the whole response body.

The component that is calling this should map from the CL API request to the "stats" API request. So your adapter in the top lvl execute function gets the input:

```json
{
  "blockchain": "btc",
  "endpoint": height
}
```

We should map that to the underlying API endpoint fn call like this:

```json
{
  "blockchain": "bitcoin",
  "network": "mainnet",
  "path": "payload.blocks"
}
```

@ebarakos @RodrigoAD @justinkaseman @emmick4 please acknowledge & discuss!

As this is a high priority for today, you can ignore this for this specific PR, but keep it in mind on the next one. I am also going to open an issue for it, to improve our older implementations.

cryptoid/src/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+  const validator = new Validator(request, customParams)
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+  const blockchain = validator.validated.data.blockchain.toLowerCase()
+
+  const key = util.getRandomRequiredEnv('API_KEY')
```
krebernisak, line 26: If you need some configuration in your adapter, it should be constructed in the `makeConfig` function and injected here to be used. You already have an API key setup in your config here:

```suggestion
  const key = config.apiKey
```

krebernisak (changes requested):

composite/bitcoin-json-rpc/src/adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Config, ExecuteWithConfig, ExecuteFactory, AdapterRequest } from '@chainlink/types'
+import { Validator, Requester } from '@chainlink/external-adapter'
+import { makeConfig, DEFAULT_ENDPOINT } from './config'
+
+const inputParams = {
+  endpoint: false,
+}
+
+const convertEndpoint: { [key: string]: string } = {
+  height: 'headers',
```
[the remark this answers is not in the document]
ebarakos: I have noticed in my local BTC node that headers was better synchronized with the actual block height. For e.g if you open your node, and it is a bit asynced with the network, you will get a lesser number if you query for `blocks`.

cryptoapis/src/endpoint/stats.ts
```diff
@@ -1,24 +1,32 @@
 import { Requester, Validator } from '@chainlink/external-adapter'
 import { AdapterRequest, Config } from '@chainlink/types'
+import { DEFAULT_ENDPOINT } from '../config'
 
-export const Name = 'difficulty'
+export const Name = 'stats'
```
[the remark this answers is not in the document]
ebarakos: I was thinking of adhering to a certain name for this category of calls. Like we have `balance` and `price`, which also do not correspond to the actual endpoint naming.

composite/bitcoin-json-rpc/src/adapter.ts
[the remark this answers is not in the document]
krebernisak: Yeah, well headers are a lot faster to sync. Block validation is a lot slower as it needs to validate all transactions.

I think it should be safe to report height as a number of headers processed by a light client 👍

cryptoapis/src/endpoint/stats.ts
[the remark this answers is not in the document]
krebernisak: Well just because we have some code implemented currently does not make it the correct way to do it :)

I suggest you rename this for now, and we'll make sure to improve/refactor other adapter implementations in the future.

"balance", "price", "difficulty" & "height" are all data points for which one can query our system. There is no "stats" data point currently. Here you are mapping user request for "height" data point to a cryptoapis "bc-info" endpoint request.

composite/bitcoin-json-rpc/src/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'difficulty'
+
+export const makeConfig = (prefix?: string): Config => Requester.getDefaultConfig(prefix)
```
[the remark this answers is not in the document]
ebarakos: Added this as a TODO, since the json-rpc does not have a `makeConfig` right now.

cryptoapis/src/endpoint/stats.ts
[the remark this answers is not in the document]
ebarakos: Changed it to `bc_info` for now. 

I think we have to find a uniform convention for these adapters that are returning blockchain info, while keeping `difficulty` (and now `height`) in backwards compatibility(probably by just delegating to `stats`/`bc_info` types when seeing these types of `endpoint`).

cryptoid/src/adapter.ts
```diff
@@ -0,0 +1,46 @@
...
+  const validator = new Validator(request, customParams)
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+  const blockchain = validator.validated.data.blockchain.toLowerCase()
+
+  const key = config.apiKey
```
RodrigoAD, line 25: Make sure you are actually passing the `apiKey` value.

cryptoid/src/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'difficulty'
+
+export const makeConfig = (prefix?: string): Config => Requester.getDefaultConfig(prefix)
```
RodrigoAD, line 6: `makeConfig` needs to return the `apiKey`, something like:
```js
export const makeConfig = (): Config => {
  const apiKey = util.getRandomRequiredEnv('API_KEY')
  return { ...getDefaultConfig(), apiKey }
}
```

ebarakos: Isn't it passed from `getDefaultConfig`?

RodrigoAD: Oh right! Sorry

cryptoapis/README.md
```diff
@@ -16,10 +16,11 @@ The adapter takes the following environment variables:
 - `base`, `from`, or `coin`: The symbol or ID of the coin to query
 - `quote`, `to`, or `market`: The symbol or ID of the market to convert to
 
-### Difficulty endpoint
+### Blockchain stats
```
krebernisak, line 19: "Blockchain stats" is a non-existing concept. 

We should document input params for `difficulty` and `height` endpoints (data points). Something like this:

```markdown
## Input Params

- `endpoint` or `dataPoint`: The requested data point. One of (`price`|`difficulty`|`height`). Defaults: `price`.

### Price

- `base`, `from`, or `coin`: The symbol or ID of the coin to query (required).
- `quote`, `to`, or `market`: The symbol or ID of the market to convert to (required).

### Difficulty

- `blockchain` or `coin`: The blockchain name (required).
- `network` or `chain`: The blockchain network name. Default: `mainnet`

### Height

- `blockchain` or `coin`: The blockchain name (required).
- `network` or `chain`: The blockchain network name. Default: `mainnet`
```

I hope this makes it more clear now. I've introduced new names in this example like `dataPoint` & `chain`, just to  make it complete and more understandable. Please do not include those in the docs as we do not support them currently.

krebernisak (changes requested):

cryptoid/README.md
```diff
@@ -2,17 +2,18 @@
 
 ## Input Params
 
-- `blockchain` or `coin`: The blockchain to get difficulty from
+- `blockchain` or `coin`: The blockchain to get stats from
```
krebernisak, line 5: Please improve this docs similar to my suggestion in the other comment.

cryptoapis/README.md
ebarakos: Added Difficulty & Height separately, along with sample responses.

cryptoapis/README.md
```diff
@@ -115,3 +169,7 @@ https://docs.cryptoapis.io/rest-apis/blockchain-as-a-service-apis/btc/index#btc-
   "statusCode": 200
 }
 ```
+
+```
+
+```
```
krebernisak, line 175:
```suggestion

```

krebernisak (approved):
Should be good to go!

A lot of valuable discussions in this PR 🙌  Thanks, @ebarakos 👍

ebarakos:
Thanks for the thorough review @krebernisak and @RodrigoAD!

PR #220, opened 2021-01-12
APY Finance adapter + Token Allocation

RodrigoAD:

- `token-allocations` adapter. This adapter could be used for price data relying adapters, such as Defi Pulse or Synthx
- `apy-finance` adapter. Calculates the Total Value Locked in APY Finance.

To not make this PR really big, Defi Pulse migration to use `token-allocations` adapter will be added in a separate PR.


token-allocation/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+`units[n]` would correspond to the `components[n]` balance
+
+```json
+{
+	"jobID": "1",
+    "data": {
+        "components": [ "DAI", "USDC", "USDT" ],
+        "units": ["1000000000000000000", "10000000000000000000", "1000000000000000000"]
+    }
```
boxhock, line 36: Should update the formatting/indentation here

token-allocation/src/config.ts
```diff
@@ -0,0 +1,56 @@
...
+import { util } from '@chainlink/ea-bootstrap'
+import { Index } from './adapter'
+import cryptocompare from './data-providers/cryptocompare'
+import nomics from './data-providers/nomics'
+import coinpaprika from './data-providers/coinpaprika'
+import coinmarketcap from './data-providers/coinmarketcap'
+import coingecko from './data-providers/coingecko'
+import coinapi from './data-providers/coinapi'
+import amberdata from './data-providers/amberdata'
+import kaiko from './data-providers/kaiko'
```
boxhock, line 11: Do we have no way of using the existing adapters instead of re-writing them here?

RodrigoAD: - When we migrate the price adapters to `.ts`, `token-allocation` will be using them. Right now we can't
- Everything under `data-providers/` is exactly the same on `defi-pulse`, but this will use `token-allocation` adapter after this is merged, so all the repeated code will be removed from `defi-pulse`
- `Synthx` could use `token-allocation` as well, so the repeated code will be removed there too.

boxhock: Thanks for the explanation! Just worried about the technical debt.

Do we have a story for this migration? I feel like we should also include some work on the CVI adapter (which isn't even merged in yet, but I think it can use the same work once it's in).

RodrigoAD: I included 2 stories yes (#176484089 and #176464064). You can include the story for CVI too so we don't forget

token-allocation/README.md
```diff
@@ -0,0 +1,78 @@
...
+    - `kaiko` (K)
+    - `nomics`(K)
+- `API_KEY`: For those data providers who need an api key
+- `DEFAULT_CURRENCY` (Optional): Currency that the price will be fetched by default. `USD` used by default
+
+
+## Input Params
+
+- `components`: Array of the token symbols. 
+- `units`: Array of balances in wei of each token.
```
boxhock, line 25: Would it make sense to make this optional, and default it to an array of 1s?

token-allocation/README.md
```diff
@@ -0,0 +1,78 @@
...
+  "data": {
+    "components": [
+      "DAI",
+      "USDC",
+      "USDT"
+    ],
+    "units": [
+      "10000000000000000000",
+      "10000000000000000000",
+      "10000000000000000000"
```
boxhock, line 42: Does this match with the output in this readme?
And why does units change from wei to "Ether"?

token-allocation/README.md
```diff
@@ -0,0 +1,78 @@
...
+    - `nomics`(K)
+- `API_KEY`: For those data providers who need an api key
+- `DEFAULT_CURRENCY` (Optional): Currency that the price will be fetched by default. `USD` used by default
+
+
+## Input Params
+
+- `components`: Array of the token symbols. 
+- `units`: Array of balances in wei of each token.
+- `currency` (optional). Currency we want the price on. `DEFAULT_CURRENCY` by default
```
boxhock, line 26: More of a generic question, but would it be easy to extend this adapter with a param for "endpoint"?
Say the default currently is to get the prices, could it easily be extended to say marketcap percentage?

token-allocation/README.md
RodrigoAD: - It doesn't match you're right. Will fix it. 
- It would make sense to be the same units actually, either both 'wei' or both 'ether'. Wei I'd say makes more sense as it's more precise, but not sure.

token-allocation/README.md
RodrigoAD: I don't think it would be hard, and make sense to have that functionality in this adapter.

token-allocation/README.md
RodrigoAD: Yep ok

token-allocation/README.md
boxhock: I agree. On the last point, also consider that "ether" more sense normally, if you're not explicitly aware that this adapter takes "wei".

How much precision do we usually need with an adapter like this? I'm guessing there's a reason why you picked wei to begin with 😅

RodrigoAD: Haha there are some adapters that query the units from contracts directly, so they give it in wei. But they can be converted to "ether" of course 😄

boxhock: Alright, sounds like the easiest thing to do is just give it in wei then.

boxhock (approved):
Would like some more :eyes: on this, but looks good to me

composite/apy-finance/src/registry/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const getRegistry = (address: string, rpcUrl: string, abi: any): ethers.Contract => {
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
+  return new ethers.Contract(address, abi, provider)
+}
+
+type Allocations = {
+  components: string[]
+  units: string[]
+}
```
krebernisak, line 13: This type is a relic of `defi-pulse` underlying contract implementation and is inflexible. How about we change this type to something more usable like:

```javascript
type TokenAllocation = {
   symbol: string
   decimals: number
   balance: string | BN
}
```

This type should then be moved to the `token-allocations` adapter and used by APY & DeFi Pulse.

composite/apy-finance/src/registry/index.ts
```diff
@@ -0,0 +1,40 @@
+import { ethers } from 'ethers'
+import registryAbi from './IRegistry.json'
+import assetAllocationAbi from './IAssetAllocation.json'
+
+const getRegistry = (address: string, rpcUrl: string, abi: any): ethers.Contract => {
```
krebernisak, line 5: This function is really unnecessary :)

composite/apy-finance/src/registry/IAssetAllocation.json
```diff
@@ -0,0 +1,53 @@
+[
```
krebernisak, line 1: Kind of doesn't make sense to have `IAssetAllocation` abi in a registry folder. Let's just put all ABIs in the `/src/abi` folder.

The second thing is these static ABIs are really hard to identify. What is the origin and source code for these ABIs?  Where can we find some more documentation for these contracts? Please document answers to these questions.

token-allocation/src/adapter.ts
```diff
@@ -0,0 +1,66 @@
...
+  components: true,
+  units: false,
+  currency: false,
+}
+
+export function makeIndex(components: string[], units: any[], currency: string): Index {
+  return components.map((component, i) => {
+    return {
+      asset: component,
+      units: new Decimal(new utils.BigNumber(units[i]).toString()),
```
krebernisak, line 27: Either `Decimal` or `BigNumber`, why both?

krebernisak (commented):
@RodrigoAD here are some more suggestions you can work on in the next defi-pulse PR.

composite/apy-finance/src/registry/index.ts
RodrigoAD: Haha ok

composite/apy-finance/src/registry/index.ts
RodrigoAD: Yep agree

PR #221, opened 2021-01-12
TS migrations

ebarakos:
Migrating the following adapters to Typescript:

`google-finance`
`kaiko`
`marketstack`
`metalsapi`
`nikkei`
`nomics`
`openexchangerates`
`poa-gasprice`
`polygon`


kaiko/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'KAIKO'
```

@krebernisak Do we want this capitalized or lowercase? I've seen both ways, but let's pick a convention.

kaiko/src/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'example'
```
justinkaseman, line 4: This is currently an unused variable as there is only one endpoint.

I would instead make every EA similar to the `example` adapter's  `adapter.ts` code with the switch statement routing the logic, even if there is only one endpoint. It makes it a lot easier to extend in the future.

marketstack/package.json
```diff
@@ -1,15 +1,46 @@
...
+    "build": "tsc -b",
     "lint": "eslint --ignore-path ../.eslintignore . --ext .js,.jsx,.ts,.tsx",
     "lint:fix": "eslint --ignore-path ../.eslintignore . --ext .js,.jsx,.ts,.tsx --fix",
-    "test": "yarn _mocha --timeout 0",
-    "test:unit": "yarn _mocha --grep @integration --invert --timeout 0",
-    "test:integration": "yarn _mocha --grep @integration --timeout 0"
+    "test": "mocha --exit --timeout 0 -r ts-node/register 'test/**/*.test.ts'",
+    "test:unit": "mocha --exit --grep @integration --invert -r ts-node/register 'test/**/*.test.ts'",
+    "test:integration": "mocha --exit --timeout 0 --grep @integration -r ts-node/register 'test/**/*.test.ts'",
+    "server": "node -e 'require(\"./index.js\").server()'",
```
justinkaseman, line 31:
```suggestion
```
Take these out of every TS migrated EA as `./index.js` is now gone.

marketstack/src/adapter.ts
```diff
@@ -0,0 +1,41 @@
...
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+
+  switch (endpoint) {
+    case eod.Name: {
+      const data = await eod.execute(config, request)
+
+      return Requester.success(jobRunID, {
```
justinkaseman, line 23: This will quickly grow to be hard to read if say 10 other endpoints were added.

I think the best style would be how Jonas modified the `example` adapter as seen here: 
https://github.com/smartcontractkit/external-adapters-js/blob/develop/example/src/endpoint/example.ts#L32
Each endpoint handles its own return, which keeps `adapter.ts` a lot cleaner.

marketstack/src/endpoint/eod.ts
```diff
@@ -0,0 +1,42 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { AdapterRequest, Config } from '@chainlink/types'
+import { DEFAULT_INTERVAL, DEFAULT_LIMIT } from '../config'
+import { util } from '@chainlink/ea-bootstrap'
+
+export const Name = 'eod'
+
+const customError = (data: any) => data.Response === 'Error'
```
justinkaseman, line 8: @krebernisak Should we be adding typing for each API response shape? Seems like it might be hard to maintain, but could provide valuable debugging information.

marketstack/src/endpoint/eod.ts
```diff
@@ -0,0 +1,42 @@
...
+    symbols,
+    interval,
+    limit,
+    access_key: util.getRandomRequiredEnv('API_KEY'),
+  }
+
+  const reqConfig = {
+    ...config.api,
+    params,
+    baseURL: 'http://api.marketstack.com/v1/',
```
justinkaseman, line 35: I would expect the `baseURL` and `API_KEY` to be the same no matter what endpoint is being hit. These might be better injected in Config.ts,

marketstack/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'MARKETSTACK'
```

metalsapi/package.json
```diff
@@ -1,15 +1,46 @@
 {
   "name": "@chainlink/metalsapi-adapter",
   "version": "0.0.1",
```
justinkaseman, line 3: Should we be bumping versions for all of these TS migrations?

metalsapi/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'METALSAPI'
```

nikkei/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'NIKKEI'
```

nomics/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE'
```
justinkaseman, line 5:
```suggestion
const NAME = 'NOMICS'
```

openexchangerates/src/config.ts
```diff
@@ -0,0 +1,6 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'latest.json'
```
justinkaseman, line 4: As with the other EA I think this should be changed to be used as other `endpoint`s are.

justinkaseman (changes requested):
Great!

Added a few small tweaks.

One thing I am unsure of is what the `endpoint` would be for some of these adapters. I know we want to keep similar endpoints conformed to the same name and params. Just thinking aloud that we need them documented somewhere.

kaiko/src/index.ts
krebernisak: This was first used as an ENV prefix to avoid naming collisions when using two or more adapters in the same env. So for now it was a simple hardcoded uppercase string

What we should do now to make this more robust is this:

- Import the `name` from `package.json`, remove the `-adapter` suffix, and export that as NAME.
- Create a `toEnvName` util function that will map it to a safe string to use as ENV name

> Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit.

metalsapi/package.json
krebernisak: Yeah, I think we should!

marketstack/src/endpoint/eod.ts
krebernisak: Not in this PR to keep it fast & simple.

But in the future, we should type anything that's useful, and I think that will become useful at one point. The further we go with composite adapters, the more we use the data provider adapters programmatically, the more useful we are going to find those types.

First, we should finish this migration, then focus on improvements to our internal framework data types, and then we can make another push like this where we add a bunch of underlying API data types in one push. At that point it could start being useful :)

marketstack/package.json
ebarakos: Taking it out of the `example` as well.

metalsapi/package.json
ebarakos: Since all versions are now set to `0.0.1`, let's wait until everything is finished/merged, and batch-update all TS migrated ones, or all TS ones in general.

ebarakos:
Converted this to draft as it had been split to 3 smaller batches. To be closed after merging the migrations.

PR #228, opened 2021-01-13
Fix CMC API for multiple currencies with the same symbol

krebernisak:


tyrion70 (approved):
Tested on my staging and mainnet env, looks good!

boxhock (approved):
If we're doing fixes for the other versions in different PRs, this looks good.

RodrigoAD (approved):
LGTM

PR #240, opened 2021-01-14
add .envrc on .gitignore

ebarakos:
In order to use `direnv`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 *.zip
 dist
 .DS_Store
+.envrc


boxhock:
I'd usually prefer files like these to be in the user's global gitignore config, to keep the repo .gitignore files more clean and specific to that repo. But interested in hearing more opinions on that.

ebarakos:
I can put it the global config, but since there are too many different folders that require different environment setup, maybe is it repo-related as well? Maybe also potentially suggest the use of `direnv` to README?

krebernisak:
@ebarakos I'd agree with @boxhock here, this should go to your global `.gitignore`.

Feel free to add a README section on how to use and setup `.direnv`, which could be useful.

PR #242, opened 2021-01-15
Add CurrioInvest data provider adapters

krebernisak:


boxhock:
We have two more adapters that were included in this set, do you want to include them too?

krebernisak:
@boxhock yeah! I meant to do that today, to bring these adapters to the new codebase quickly before the deployment, but after the first one, I instead decided not to rush. We'll deploy the private version if it needs to go this week, and then update them on a normal schedule, next week. I think that's the safer/saner way to do it :)

boxhock:
@krebernisak sounds good, I agree with that approach :slightly_smiling_face:

boxhock:
I don't think this PR is needed any longer, closing it...

PR #246, opened 2021-01-18
Token allocation input params update

RodrigoAD:
Token allocation adapter accepts the units in ETH, it doesn't make any conversion. Adapters using `token-allocation` are responsible of sending the units in the correct format


token-allocation/README.md
```diff
@@ -22,7 +22,7 @@ The adapter takes the following environment variables:
 ## Input Params
 
 - `components`: Array of the token symbols. 
-- `units` (optional): Array of balances in wei of each token. `10000000000000000000` wei (1 ether) for each token by default.
+- `units` (optional): Array of balances of each token in ETH. 1 ether for each token by default.
```
boxhock, line 25: Maybe we should just make this as generic as possible? No mention of ETH, just whole units.

krebernisak: Yeah, no need to mention ETH here. These are just whole-number units that we convert to currency using 18 decimals by default if not specified differently.

composite/apy-finance/src/abi/ERC20.json
```diff
@@ -0,0 +1,222 @@
+[
```
krebernisak, line 1: We should be able to include this from a public package as this is a standard interface. Perhaps from `@openzeppelin/contracts/build/contracts/ERC20.json`

token-allocation/src/adapter.ts
```diff
@@ -24,7 +23,7 @@ export function makeIndex(components: string[], units: any[], currency: string):
   return components.map((component, i) => {
     return {
       asset: component,
-      units: new Decimal(new utils.BigNumber(units[i]).toString()),
+      units: new Decimal(units[i]),
```
krebernisak, line 26: This gives the wrong signal about `units`. Just because it's convenient to latter make calculations with decimals, that does not make `units` a decimal number. Units type should be a big number integer.

composite/apy-finance/src/registry/index.ts
```diff
@@ -1,39 +1,62 @@
...
+    ),
+    Promise.all(tokenAddresses.map(async (address: string) => getDecimals(address))),
   ])
-  return { components, units }
+
+  return {
+    components,
+    units: units.map((unit, i) => {
+      const decimal = new utils.BigNumber(10).pow(decimals[i])
+      return new utils.BigNumber(unit).div(decimal).toString()
```
krebernisak, line 39: This is not correct. It behaves differently than DeFi Pulse.

You either need to return `units` + `decimals`, or normalize `units` as the number with 18 decimals. Like they do in the DeFi Pulse `getAllocations` function.

token-allocation/src/adapter.ts
```diff
@@ -36,7 +35,7 @@ export const calculateIndexValue = (index: Index): number => {
   if (!index.every(isPriceSet)) throw new Error('Invalid index: price not set')
   // calculate total value
   return index // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    .reduce((acc, i) => acc.plus(i.units.div(1e18).times(i.price!)), new Decimal(0))
+    .reduce((acc, i) => acc.plus(i.units.times(i.price!)), new Decimal(0))
```
krebernisak, line 38: You need to divide it here, as we receive `units` on input, not a decimal number of coins.

```suggestion
    .reduce((acc, i) => acc.plus(i.units.div(i.decimal || 1e18).times(i.price!)), new Decimal(0))
```

krebernisak (changes requested):

token-allocation/src/config.ts
```diff
@@ -60,6 +60,6 @@ export const makeConfig = (defaultUnit?: string): Config => {
   return {
     priceAdapter,
     defaultCurrency,
-    makeDefaultUnits: makeDefaultUnits(defaultUnit || new Decimal(1e18).toString()),
+    makeDefaultUnits: makeDefaultUnits(defaultUnit || new Decimal(1).toString()),
```
krebernisak, line 63: Why 1? This should be 0, the default is 0. If I have no units I don't have 1 unit :)

composite/apy-finance/src/registry/index.ts
RodrigoAD: I'd say this is correct. It's true that is missing the default value for `decimal` to be `1e18`, but the calculation is correct.

token-allocation/src/adapter.ts
RodrigoAD: So `token-allocation` adapter won't be responsible of conversion anymore, either aware of decimals. It receives as input the units only, that should be formatted already.

composite/apy-finance/src/registry/index.ts
krebernisak: I don't agree.

In this PR you changed that the token-allocation adapter accepts decimal numbers on input and not base units. This is not a good API. These are big numbers and we should work with base units.

Then this apy-finance adapter returns a decimal number. In that sense it's correct.

But then defi-pulse adapter returns a int number of units (normalized to 18 decimal points), and when you pipe that to token-allocation adapter which now expects decimal numbers, you are going to get a wrong result.

Please calculate everything in units, just print the formatted decimal price on the output.

token-allocation/src/config.ts
boxhock: The idea here is that you'd give an input of symbols, but won't have to also do a list of units.

E.g. if I give it [BTC, ETH], I shouldn't also have to provide [1, 1] - that should be implied. If I want anything other than 1 unit of each, that should be specified.

boxhock: So for no units, you should just not provide that symbol. I think this makes sense for most applications of this adapter.

composite/apy-finance/src/registry/index.ts
krebernisak: Also looks like ethers `BigNumber` does not support decimal numbers, so your `.div` is going to round the result. Take a look at [this comment](https://github.com/ethers-io/ethers.js/issues/452#issuecomment-472168526) on how to do it properly.

token-allocation/src/adapter.ts
krebernisak: The token-allocation adapter should accept base units, and then calculate the price based on a number of base units and defined decimal points for that specific token. If it doesn't know the number of decimal points, because it was not provided, it uses 18 as default.

token-allocation/src/config.ts
krebernisak: Aha, I see... Well, it confused me, but I guess it can make sense. I hope this implicit 1 won't turn into an off-by-one error somewhere down the line :)

Please just document it explicitly in the readme, if it's not already.

composite/apy-finance/src/abi/ERC20.json
RodrigoAD: Is it possible that this `abi` doesn't have `decimals` included?

krebernisak: [It should](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5a58fd2de01ac4292510a9f0b30d52a9b4ef8ecd/contracts/token/ERC20/ERC20.sol#L44), why do you ask?

token-allocation/README.md
```diff
@@ -21,23 +21,27 @@ The adapter takes the following environment variables:
...
 ## Input Params
 
-- `components`: Array of the token symbols. 
-- `units` (optional): Array of balances of each token. 1 unit for each token by default.
-- `currency` (optional). Currency we want the price on. `DEFAULT_CURRENCY` by default
+- `allocations`: Array of allocations, being each allocation:
+  - `symbol`: Token symbol
+  - `balance` (optional): Token balance. `1e18` by default
+  - `decimals` (optional): Token decimals. `18` by default
+  - `currency` (optional). Currency we want the price on. `DEFAULT_CURRENCY` by default
```
krebernisak, line 28: This last one is too indented:

```suggestion
- `currency` (optional). The currency we want the price on. `DEFAULT_CURRENCY` by default
```

composite/apy-finance/src/registry/index.ts
```diff
@@ -1,39 +1,54 @@
...
   const tokenAddresses = await registry.getTokenAddresses()
-  const [components, units]: string[][] = await Promise.all([
+  const [components, balances, decimals]: any = await Promise.all([
     Promise.all(tokenAddresses.map((address: string) => registry.symbolOf(address))),
-    Promise.all(tokenAddresses.map((address: string) => registry.balanceOf(address))),
+    Promise.all(
+      tokenAddresses.map(async (address: string) =>
+        utils.bigNumberify(await registry.balanceOf(address)),
+      ),
+    ),
```
krebernisak, line 27:
```suggestion
    Promise.all(tokenAddresses.map((address: string) => registry.balanceOf(address))),
```

composite/apy-finance/src/registry/index.ts
```diff
@@ -1,39 +1,54 @@
...
+    Promise.all(tokenAddresses.map(async (address: string) => getDecimals(address))),
   ])
-  return { components, units }
+
+  return components.map((symbol: string, i: number) => {
+    return { symbol, balance: balances[i].toString(), decimals: decimals[i] }
```
krebernisak, line 32:
```suggestion
    return { symbol, balance: utils.bigNumberify(balances[i]), decimals: decimals[i] }
```

token-allocation/src/adapter.ts
```diff
@@ -1,31 +1,34 @@
...
 import { Requester, Validator } from '@chainlink/external-adapter'
-import Decimal from 'decimal.js'
-import { utils } from 'ethers'
 import { Config, makeConfig } from './config'
+import { Allocations } from './types'
+import { Decimal } from 'decimal.js'
 
 export type Index = IndexAsset[]
 
 type IndexAsset = {
```
krebernisak, line 9: Do we even need this type anymore?

I think this a relic of defi-pule and looks like a duplicate of `TokenAllocation` now.

token-allocation/src/types.ts
```diff
@@ -0,0 +1,6 @@
+export type Allocation = {
+  symbol: string
+  balance: number
+  decimals: number
+}
+export type Allocations = Allocation[]
```
krebernisak, line 6:
```suggestion
export type TokenAllocation = {
  symbol: string
  decimals: number
  balance: BigNumber
}
export type TokenAllocations = TokenAllocation[]
```

krebernisak (changes requested):

token-allocation/README.md
```diff
@@ -48,31 +50,25 @@ The adapter takes the following environment variables:
...
-                "price": "1.00139295"
-            }
-        ]
-    },
-    "result": 12.0322909,
-    "statusCode": 200
+  "jobRunID": "1",
+  "data": {
+    "result": 36221.40305171679,
+    "index": [
```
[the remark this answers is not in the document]
boxhock: What is your suggestion for output here instead?

Something like this?

```json
[...]
"allocations": [
        {
            "base": "wBTC",
            "quote": "USD",
            "price": 36221.40268850176
        }
]
```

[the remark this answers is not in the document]
krebernisak: That's why I was asking you all those questions yesterday :) I was trying to figure out do we need this output or not. Looks like we do.

My suggestion would be similar to CMC API response, as this type of schema can support multiple base tokens, with multiple quotes per token, and multiple data points per base/quote pair:

```json
{
  "jobRunID": "1",
  "data": {
    "result": 36221.40305171679,
    "DAI": {
      "quote": {
        "USD": {
          "price": 155.22087406,
          "marketCap": 17152734279.383095,
          }
        }
      }
    },
    "wBTC": {
        "USD": {
          "price": 155.22087406,
          "marketCap": 17152734279.383095,
          }
        }
      }
    }
  }
  "result": 36221.40305171679,
}
```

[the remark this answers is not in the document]
RodrigoAD: Let me ask, because the units (or whichever name 😆 ) was a business requirement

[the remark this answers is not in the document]
boxhock: If units is already provided as input, is it necessary as output too?

@krebernisak this output works for me at least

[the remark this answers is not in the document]
krebernisak: Yeah, I think that it is unnecessary.

token-allocation/src/adapter.ts
RodrigoAD: This is basically the response schema now.

token-allocation/README.md
```diff
@@ -48,31 +50,27 @@ The adapter takes the following environment variables:
...
-                "units": "10",
-                "currency": "USD",
-                "price": "1.00139295"
-            }
-        ]
+  "jobRunID": "1",
+  "data": {
+    "result": 37340.964000680404,
+    "wBTC": {
+      "balance": "1",
```
boxhock, line 57: Still not sure if balance is needed in the output here, since it's already provided in the input?

token-allocation/README.md
```diff
@@ -25,7 +25,7 @@ The adapter takes the following environment variables:
   - `symbol`: Token symbol
   - `balance` (optional): Token balance. `1e18` by default
   - `decimals` (optional): Token decimals. `18` by default
-  - `currency` (optional). Currency we want the price on. `DEFAULT_CURRENCY` by default
+- `currency` (optional). Currency we want the price on. `DEFAULT_CURRENCY` by default
```
krebernisak, line 28: I would suggest we rename this as `quote` as this is the standard name for a currency we want to price an asset in, and we already use it throughout the codebase in other adapters.

token-allocation/src/types.ts
```diff
@@ -1,6 +1,29 @@
-export type Allocation = {
+import { BigNumberish } from 'ethers/utils'
+
+export type TokenAllocation = {
   symbol: string
-  balance: number
+  balance: BigNumberish
   decimals: number
+  price?: number
+  currency?: string
```
krebernisak, line 8: TokenAllocation data type should not have a price or currency:

```suggestion
```

token-allocation/src/types.ts
```diff
@@ -1,6 +1,29 @@
...
+  [symbol: string]: QuoteSymbol
+}
+
+type ResponseAsset = {
+  quote: Quote
+  balance: string
+}
+
+export type Response = {
+  [symbol: string]: ResponseAsset
```
krebernisak, line 28: No need to name all these intermediate types. Naming is hard.

```suggestion
export type Response = {
  [symbol: string]: {
    quote: {
      [symbol: string]: {
        price?: number
        marketCap?: number
      }
    }
    balance: string
  }
```

boxhock (dismissed):
Only see a couple things to fix, but they will already be fixed in #254.

krebernisak (approved):

RodrigoAD:
Everything tested, ready to go

krebernisak (approved):

krebernisak (approved):

PR #248, opened 2021-01-18
Cryptoapis tests and README

ebarakos:
~~According to this comment: https://github.com/smartcontractkit/external-adapters-js/pull/215#discussion_r558236636~~

Adding improvements in tests and README in this PR.


justinkaseman (commented):
Some thoughts, 

1. bc_info needs README documentation

2. Instead of removing tests, could they be re-written?

3. Based on that discussion with @krebernisak the way I envisioned this working in my head goes like:
Each data provider has an External Adapter. Their API may have multiple "endpoints" such as spot price API or blockchain info, each which return a different data response. Internally we want to deprecate this param we call an `endpoint` in favor of a more flexible `path` that is used to navigate the data response. In Adapter Legos a developer will want to declaratively write one line: `Legos.cryptoapis.price()`. This "behavior" of `price` is just a common name mapped to params that are sent to this lower level External Adapter. So in the price example it might internally be something like
```
{
  "from": "BTC",
  "to": "USD",
  "path":  "'payload.weightedAveragePrice",
  "endpoint": "price",
}
```
If this is how we want to move forward then:
- `adapter.ts` should probably allow the `path` parameter to come through and endpoints changed to utilize it.
- If we want backwards compatibility, if path isn't set we look at the `endpoint` and construct it from a map. This would happen in the EA.
- New "behaviors" mappings will be handled in Adapter Legos and/or ea-behaviors.

cryptoapis/src/adapter.ts
```diff
@@ -23,20 +23,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
...
     }
-    case 'difficulty':
-    case 'height': {
-      response = await bc_info.execute(config, request)
-      break
-    }
     case balance.Name: {
       return balance.makeExecute(config)(request)
     }
     default: {
```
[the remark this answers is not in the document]
ebarakos: I understand your concern. If we don't do this though, it would be hard to pass arbitrary "endpoint" parameters to retrieve blockchain info(if we want to maintain backwards compatibility for this certain call). We might need to create a map that declares all or almost most of them. Or we can just support only these 2 declared/documented.

ebarakos:
Added some bc_info tests.

cryptoapis/src/adapter.ts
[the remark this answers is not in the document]
ebarakos: Changing these on adapter.ts duplication PR

cryptoapis/README.md
```diff
@@ -67,36 +87,6 @@ The adapter takes the following environment variables:
 }
 ```
 
-### Height
```
ebarakos, line 70: Left this as it is @justinkaseman, because probably it could be batch-updated along other balance endpoint docs.

cryptoapis/README.md
```diff
@@ -67,36 +87,6 @@ The adapter takes the following environment variables:
...
-      "verificationProgress": 0.9999935897991173
-    },
-    "result": 666185
-  },
-  "result": 666185,
-  "statusCode": 200
-}
-```
-
 ### Balance
```
ebarakos, line 90: Left this as it is @justinkaseman, because probably it could be batch-updated along other balance endpoint docs.

test-helpers/src/behaviors/balance.ts
```diff
@@ -18,22 +18,22 @@ function base(execute: Execute) {
...
-                {
-                  address: '35ULMyVnFoYaPaMxwHTRmaGdABpAThM4QR',
-                  coin: 'btc',
-                  chain: 'mainnet',
-                },
-              ],
-            },
-          },
-        },
+        // {
```
justinkaseman, line 21: Why is this test being disabled?

When I checkout to the branch and run the test it passes for me? Was this an issue when running in CI?

I'm seeing a different issue with the balance decimal place conversion that I will open an issue for.

ebarakos: It was related to a default routing to an endpoint instead of error routing. Removing it.

justinkaseman:
Looks good to me after a rebase :+1:

krebernisak (approved):

PR #251, opened 2021-01-19
Add IEX Cloud adapter

boxhock:


iex-cloud/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,39 @@
...
+  const url = `${base.toUpperCase()}${quote.toUpperCase()}/quote`
+
+  const params = {
+    token: config.apiKey,
+  }
+
+  const reqConfig = {
+    ...config.api,
+    params,
+    baseURL: 'https://cloud.iexapis.com/stable/crypto',
```
RodrigoAD, line 27: This could come from `config`, and then you add the endpoint? Same for `stock`. 
`${config.baseURL}/crypto`, `${config.baseURL}/stock`

boxhock: I was following the format of the example adapter. Is this something we should change there as well?

RodrigoAD: I'm not very familiar with the example adapter, I'm sure @krebernisak can bring some light here 😄

RodrigoAD (approved):
It LGTM

PR #256, opened 2021-01-20
Exposing adapter exports through a single package @chainlink/ea

ebarakos:


ebarakos:
New package has been created, @chainlink/adapters, that can expose all adapter exports directly through TypeScript. 

Old functionality will be retained(each adapter will have its own workspace and can be built separately).

Created 2 examples:

`json-rpc-adapter` is under the `adapters/` folder, but there is no need to move adapters, especially at this point in time, since they can be referenced through the parent folder, as the 2nd example, `dns-query`.

krebernisak (changes requested):

composite/dns-record-check/src/adapter.ts
```diff
@@ -1,8 +1,8 @@
 import { ExecuteWithConfig, ExecuteFactory, Config } from '@chainlink/types'
 import { Validator } from '@chainlink/external-adapter'
 import { Requester } from '@chainlink/external-adapter'
-import DNS from '@chainlink/dns-query-adapter'
-import { DNSQueryResponse, DNSAnswer } from '@chainlink/dns-query-adapter/dist/types'
+import { adapters } from '@chainlink/adapters'
+import { DNSQueryResponse, DNSAnswer } from '@chainlink/adapters/dist/dns-query/src/types'
```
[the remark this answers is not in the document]
ebarakos: Yes, that is the intended use, just imported it temporarily.

package.json
```diff
@@ -4,6 +4,8 @@
   "license": "MIT",
   "private": true,
   "workspaces": [
+    "adapters/**",
```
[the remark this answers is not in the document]
ebarakos: Keeping folder structuring for the end, as it changes paths of imports/settings in many places.

ebarakos:
Adapters can be run from the @ea package, using the relevant yarn scripts(still need to remove some duplication on them). 

Also can be imported through the @ea package, like in the 2 composite adapters changed + the example added under the `examples` folder.

cryptoapis/src/index.ts
```diff
@@ -3,5 +3,5 @@ import { makeExecute } from './adapter'
 import { makeConfig } from './config'
 
 const NAME = 'CRYPTOAPIS'
-
-export = { NAME, makeExecute, makeConfig, ...expose(util.wrapExecute(makeExecute())) }
+const handlers = expose(util.wrapExecute(makeExecute()))
```
ebarakos, line 6: Not sure if we are going to have any kind of problems with deployments or node operators with the exposed handlers changed this way. But I couldn't find a way to spread the exposed object to separate exports.

adapters/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "lint": "eslint --ignore-path ../.eslintignore . --ext .js,.jsx,.ts,.tsx",
+    "lint:fix": "eslint --ignore-path ../.eslintignore . --ext .js,.jsx,.ts,.tsx --fix",
+    "test": "echo \"Error: no test specified\"",
+    "test:unit": "echo \"Error: no test specified\"",
+    "test:integration": "echo \"Error: no test specified\"",
+    "start:jsonrpc": "node -e 'require(\"./dist/adapters/index.js\").jsonrpc.handlers.server()'",
+    "start:dnsquery": "node -e 'require(\"./dist/adapters/index.js\").dnsquery.handlers.server()'",
+    "start:cryptoapis": "node -e 'require(\"./dist/adapters/index.js\").cryptoapis.handlers.server()'",
+    "start:npm": "node -e 'require(\"./dist/adapters/index.js\").'$npm_config_adapter'.handlers.server()'",
+    "start": "node -e 'eval(`require(\"./dist/adapters/index.js\").${process.env.ADAPTER}.handlers.server()`)'"
```
ebarakos, line 24: `start:npm` works only with npm. e.g `npm run start:npm --adapter=cryptoapis`. 

`start` works only with yarn. e.g `ADAPTER=cryptoapis yarn start.

adapters/index.ts
```diff
@@ -0,0 +1,14 @@
+import { AdapterRequest, AdapterResponse } from '@chainlink/types'
+import * as dnsquery from '../dns-query/src'
```
ebarakos, line 2: These can also be imported from workspaces, instead of direct path, if we keep this functionality. Also could be moved to subfolders under the adapters folder.

ebarakos:
In order to enable tree-shaking (so we don't fetch code and dependencies of all adapters when importing a specific adapter through the single package) we must use ES6 modules(by setting "`type":"module"` in `package.json`). However I am experiencing `ERR_REQUIRE_ESM` errors locally when I try to import the created package under this setup. Also ts-node seems to have some issues: [](https://stackoverflow.com/questions/62096269/cant-run-my-node-js-typescript-project-typeerror-err-unknown-file-extension)

I tried not to move current folder structure, but there are some issues and different behaviors with the dist folder structure if you set the `"rootDir": ".."`(to the parent/main folder); a new adapters folder is created inside dist, and `ncc` has trouble finding the correct path of typescript dependency.

Docker images do not find the relative imports(e.g `Error: Cannot find module './dns-query/src'`) with the current setup.

In case we change imports by referring to packages (`@chainlink/dns-query-adapter` for e.g) instead of path, we need to take care of the build/setup order on the main `package.json`.

It is also hard to setup up submodule imports(e.g @chainlink/ea/dnsquery), as it is an open problem here (https://github.com/microsoft/TypeScript/issues/33079). There are some workarounds inside the discussion that could solve the problem, but I haven't found anything that could work out of the box.

justinkaseman:
Closing as recent PRs have taken over these changes.

We likely still need to re-visit tree-shaking and what `ea-legos` could become.

Thanks for the initial attempt at this @ebarakos :slightly_smiling_face:

PR #263, opened 2021-01-21
Synthx as composite using TA

RodrigoAD:
As Synthetix can now make use of the Token Allocation adapter, this PR:
- migrates old synthetix adapter to `.ts`
- uses `token-allocations` adapter


composite/token-allocation/src/adapter.ts
```diff
@@ -43,11 +43,14 @@ const inputParams = {
 const toValidAllocations = (allocations: TokenAllocations): TokenAllocations => {
   if (!allocations.every((t) => !!t.symbol))
     throw new AdapterError({ message: `Symbol not available for all tokens.`, statusCode: 400 })
-  return allocations.map((t) => ({
-    symbol: t.symbol.toUpperCase(),
-    decimals: t.decimals || DEFAULT_TOKEN_DECIMALS,
-    balance: t.balance || DEFAULT_TOKEN_BALANCE * 10 ** (t.decimals || DEFAULT_TOKEN_DECIMALS),
-  }))
+  return allocations.map((t) => {
+    const decimals = Number.isInteger(t.decimals) ? t.decimals : DEFAULT_TOKEN_DECIMALS
```
RodrigoAD, line 47: In the case of synth, they are giving us the balance already in "units", so it would make sense to have an option to send `decimal: 0`.

typings/@chainlink/index.d.ts
```diff
@@ -105,4 +105,5 @@ declare module '@chainlink/types' {
 }
 declare module '@chainlink/ea-bootstrap'
 declare module '@chainlink/external-adapter'
+declare module 'synthetix'
```
RodrigoAD, line 108: Don't know if this is correct

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+  asset: string
+  index: any[]
+  inverted: Record<string, any>
+}
+
+const getAllocations = (synth: Synth): types.TokenAllocations => {
+  return synth.index.map((index) => ({
+    symbol: index.asset,
+    balance: index.units,
+    decimals: 0,
```
RodrigoAD, line 26: The units are already formatted taking in account the token decimals

RodrigoAD:
@krebernisak Don't know what to do with the old `synthx` adapter, as he had a really specific build procedure.

krebernisak:
Make sure the new code works and feel free to delete all legacy synth-index code, this includes the Make build commands and the CI synth-index build configuration 👍

boxhock:
Thanks for this! Looking forward to getting rid of the `Dockerfile-SynthIndex` file :grin:

RodrigoAD:
@boxhock @krebernisak Please double check that I removed the old build steps properly

.github/strategy/adapters.json
```diff
@@ -110,24 +110,8 @@
...
       "bitcoin-json-rpc",
-      "token-allocation"
-    ]
-  },
-  "synth-index": {
-    "docker": "make docker-synth-index adapter=__ADAPTER__",
-    "cmd": "make zip-synth-index adapter=__ADAPTER__",
-    "asset_path": "./synth-index/__ADAPTER__/dist/synth-index-__ADAPTER__-adapter.zip",
-    "asset_name": "synth-index-__ADAPTER__-adapter.zip",
-    "adapter": [
```
krebernisak, line 121: This is going to speed up our CI run for a couple of minutes 💪

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
   - `blockchair`
   - `cryptoapis`
 - dxFeed now uses OTC feeds for FTSE and N225, rather than licensed data feeds
+- `synth-index` adapter is now a composite adapter.
```
krebernisak, line 55:
```suggestion
- `synth-index` adapter is now a composite adapter. Going forward there is only one instance of `synth-index` adapter built, one that you configure with the underlying data provider you wish to use.
```

composite/synth-index/README.md
```diff
@@ -0,0 +1,123 @@
+# Chainlink External Adapter to Synthetix Index Value
+
+The adapter calculates a Synthetix Index value in the currency selected
+
+## Configuration
+
+- `DEFAULT_NETWORK` (Optional). Network to fetch the Synth Index
+
+This adapter relies on [`token-allocation`](../../token-allocation/README.md) adapter. Required `token-allocation` input params and configuration apply to this adapter as well.
+
```
krebernisak, line 10: Please add an example start command here, it could be useful to node operators.

composite/synth-index/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Input Params
+
+- `asset`, `from`: Synthx Index asset to fetch
```
krebernisak, line 13: Add `base` here as a default param name. It's standard on our other pricing adapters. We also price an asset here, just the asset is more complex.

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+import { AdapterResponse, Execute, AdapterRequest } from '@chainlink/types'
+import {
+  types,
+  makeExecute as tokenAllocationMakeExecute,
+} from '@chainlink/token-allocation-adapter'
+import snx from 'synthetix'
+import { makeConfig, Config } from './config'
+
+const customParams = {
+  asset: ['asset', 'from'],
```
krebernisak, line 11:
```suggestion
  base: ['base', 'asset', 'from'],
```

krebernisak (changes requested):

typings/@chainlink/index.d.ts
krebernisak: Yes, this is ok. We need to declare it as it's a JS package, not TS.

composite/synth-index/src/adapter.ts
krebernisak: I think the `toValidAllocations` fn in `token-allocation` is going to give you problems with `decimals: 0`:

```javascript
decimals: t.decimals || DEFAULT_TOKEN_DECIMALS,
```

It is going to set decimals as 18, the default.

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+  const network = validator.validated.data.network || config.defaultNetwork
+
+  const synths: Synth[] = snx.getSynths({ network: network.toLowerCase() })
+  const synth = synths
+    .filter(({ index, inverted }) => index && !inverted)
+    .find((d) => d.name.toLowerCase() === base)
+
+  if (!synth) {
+    return Requester.errored(jobID, new Error('Synth not found'))
+  }
```
krebernisak, line 45: This should throw, not return an error response:


```suggestion
  if (!synth) throw new Error('Synth not found')
```

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const customParams = {
+  base: ['base', 'asset', 'from'],
+  network: false,
+}
+
+type Synth = {
+  name: string
+  asset: string
+  index: any[]
```
krebernisak, line 18: Please expand this type to add the properties we use.

composite/synth-index/test/adapter.test.ts
```diff
@@ -0,0 +1,79 @@
+import { Requester } from '@chainlink/external-adapter'
+import { assertSuccess, assertError } from '@chainlink/adapter-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../src/adapter'
+import { makeConfig } from '../src/config'
+
+describe('execute', () => {
+  const jobID = '1'
+  process.env.DATA_PROVIDER = 'coingecko'
+  const execute = makeExecute(makeConfig())
```
krebernisak, line 10: Here you are using the env to inject the specific config into the underlying adapter. We should have the ability to provide the config directly and mock it here by writing an fn like `makeMockConfig`. 

`synth-index` fn needs to create the underlying `token-allocation` fn, so it should in its `makeConfig` probably require data to configure the underlying adapter.

composite/synth-index/src/adapter.ts
krebernisak: Ok, I see now this is addressed later.

composite/token-allocation/src/adapter.ts
krebernisak: We can also add a check to avoid negative numbers. Both `decimals` and `balance` should be >= 0.

Consider creating three functions inside `toValidAllocations`:

- `_toValidSymbol`
- `_toValidDecimals`
- `_toValidBalance`

krebernisak (changes requested):

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const synths: Synth[] = snx.getSynths({ network: network.toLowerCase() })
+  const synth = synths
+    .filter(({ index, inverted }) => index && !inverted)
+    .find((d) => d.name.toLowerCase() === base.toLowerCase())
+
+  if (!synth) throw new AdapterError({ message: `Synth not found`, statusCode: 400 })
+
+  const allocations = getAllocations(synth)
+  const _execute = TAMakeExecute(TAMakeConfig(config.dataProvider))
```
RodrigoAD, line 49: Passing `TAConfig` manually to be able to select the underlying data provider

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
...
-    balance: t.balance || DEFAULT_TOKEN_BALANCE * 10 ** (t.decimals || DEFAULT_TOKEN_DECIMALS),
-  }))
+
+  const _toValidDecimals = (decimals: number) =>
+    Number.isInteger(decimals) && decimals >= 0 ? decimals : DEFAULT_TOKEN_DECIMALS
+  const _toValidBalance = (balance: BigNumberish, decimals: number) => {
+    if (!balance) {
+      return DEFAULT_TOKEN_BALANCE * 10 ** decimals
+    }
+    if (!Number(balance)) {
```
RodrigoAD, line 49: Not using `BigNumber` because small numbers give `underflow` issues

krebernisak: Could you please explain the issue here? When do you have underflows?

Also, what are you exactly checking here? When do you expect to throw `"Invalid balance"`?

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
...
+      throw new AdapterError({
+        message: `Invalid balance`,
+        statusCode: 400,
+      })
+    }
+    if (Number(balance) < 0)
```
krebernisak, line 55: I don't think this is going to work with balance being `BigNumber`?

You'll need to convert from `BigNumberish` to `number`:

```suggestion
    if (BigNumber.from(balance).toNumber() < 0)
```

composite/synth-index/src/config.ts
```diff
@@ -0,0 +1,13 @@
...
+  defaultNetwork: string
+  dataProvider: string
+}
+
+export const makeConfig = (dataProvider = ''): Config => {
+  return {
+    defaultNetwork: util.getEnv('DEFAULT_NETWORK') || 'mainnet',
+    dataProvider,
+  }
+}
```
krebernisak, line 13: You should create the required underlying config here, something like this:

```suggestion
export const makeConfig = (prefix = ''): Config => {
  return {
    defaultNetwork: util.getEnv('DEFAULT_NETWORK', prefix) || 'mainnet',
    taConfig: ta.makeConfig(prefix)
  }
}
```

composite/synth-index/src/adapter.ts
krebernisak, line 49:
```suggestion
  const _execute = ta.makeExecute(config.taConfig)
```

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,55 @@
+import { Validator, AdapterError } from '@chainlink/external-adapter'
+import { AdapterResponse, Execute, AdapterRequest } from '@chainlink/types'
+import {
+  types,
+  makeExecute as TAMakeExecute,
+  makeConfig as TAMakeConfig,
+} from '@chainlink/token-allocation-adapter'
```
krebernisak, line 7: I think the code is going to read better if you just import this as `ta`:

```suggestion
import * as ta from '@chainlink/token-allocation-adapter'
```

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
...
 const toValidAllocations = (allocations: TokenAllocations): TokenAllocations => {
   if (!allocations.every((t) => !!t.symbol))
     throw new AdapterError({ message: `Symbol not available for all tokens.`, statusCode: 400 })
-  return allocations.map((t) => ({
-    symbol: t.symbol.toUpperCase(),
-    decimals: t.decimals || DEFAULT_TOKEN_DECIMALS,
-    balance: t.balance || DEFAULT_TOKEN_BALANCE * 10 ** (t.decimals || DEFAULT_TOKEN_DECIMALS),
-  }))
+
+  const _toValidDecimals = (decimals: number) =>
```
krebernisak, line 43: This can be undefined:

```suggestion
  const _toValidDecimals = (decimals: number | undefined) =>
```

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
     .toNumber()
 }
 
-const inputParams = {
-  allocations: true,
-  quote: false,
-}
-
 const toValidAllocations = (allocations: TokenAllocations): TokenAllocations => {
```
krebernisak, line 39: We don't have `TokenAllocations` on input, we have `any` that we need to validate:

 ```suggestion
const toValidAllocations = (allocations: any): TokenAllocations => {
```

**Review by krebernisak — DISMISSED** [review]

File: composite/token-allocation/src/adapter.ts (hunk 14 above)
Comment (krebernisak) [reply composite/token-allocation/src/adapter.ts:55]: The input types are wrong, should be:

```javascript
const _toValidBalance = (balance: number | string | undefined, decimals: number) => {
```

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
...
-    decimals: t.decimals || DEFAULT_TOKEN_DECIMALS,
-    balance: t.balance || DEFAULT_TOKEN_BALANCE * 10 ** (t.decimals || DEFAULT_TOKEN_DECIMALS),
-  }))
+
+  const _toValidDecimals = (decimals: number) =>
+    Number.isInteger(decimals) && decimals >= 0 ? decimals : DEFAULT_TOKEN_DECIMALS
+  const _toValidBalance = (balance: BigNumberish, decimals: number) => {
+    if (!balance) {
+      return DEFAULT_TOKEN_BALANCE * 10 ** decimals
+    }
```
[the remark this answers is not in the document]
krebernisak: I also think we support too much here. We support `number | string | undefined`, which is ok, but then we support both integers and decimal numbers which complicates the API, and validation.

Instead of supporting data like this (coming from synth-index):

```json
{
  "symbol": "BTC",
  "balance": "4.44",
  "decimal": 0
}
```

... let's just throw (or floor the value to 4), and force the client to convert to units.

Because if we continue to support this, then there is no reason not to support something like this:

```json
{
  "symbol": "BTC",
  "balance": "4.44",
  "decimal": 1
}
```

... which would represent the number `0.444`!?

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,34 @@ export const marketCapTotalValue = (
...
+    return balance
+  }
+  return allocations.map((t) => {
+    const decimals = _toValidDecimals(t.decimals)
+    return {
+      symbol: t.symbol.toUpperCase(),
+      decimals,
+      balance: _toValidBalance(t.balance, decimals),
+    }
+  })
```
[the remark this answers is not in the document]
RodrigoAD: Oh yes right

composite/token-allocation/src/adapter.ts
RodrigoAD: Accepting only `number | string | undefined` will have some conflicts, as `TokenAllocations` (type) accepts balance as `BigNumberish`.

composite/token-allocation/src/adapter.ts
RodrigoAD: I'm not sure it can, it's a mandatory field on `TokenAllocation`

composite/token-allocation/src/adapter.ts
krebernisak:
```
type BigNumberish = BigNumber | Bytes | string | number;
```

`_toValidBalance` is happening before we have `TokenAllocations`. We are trying to map and validate the `number | string | undefined` input as `BigNumber`.

RodrigoAD: > Also, what are you exactly checking here? When do you expect to throw "Invalid balance"?

Anything that is not `BigNumberish`. This could be thrown if the `TokenAllocations` adapter is used directly, and the client gives some invalid balance.

composite/token-allocation/src/adapter.ts
krebernisak: Yes, on `TokenAllocation` but this is before we have `TokenAllocation`. We are mapping and validating an `any[]` input as `TokenAllocation[]`

composite/token-allocation/src/adapter.ts
krebernisak: The easiest way is to check for `NaN` like so: `if (isNaN(Number(balance)) || Number(balance) < 0) throw`

composite/synth-index/src/adapter.ts
```diff
@@ -0,0 +1,52 @@
...
+    category: string
+    units: number
+  }[]
+  inverted: Record<string, any>
+}
+
+const getAllocations = (synth: Synth): ta.types.TokenAllocations => {
+  return synth.index.map((index) => {
+    const decimals = 18
+    const balance = new Decimal(index.units).mul(10 ** decimals).toString()
```
RodrigoAD, line 27: The units are coming as Floating numbers, parsing them to 18 decimals balance

composite/token-allocation/src/adapter.ts
```diff
@@ -35,19 +36,36 @@ export const marketCapTotalValue = (
...
+  }
+  const _toValidDecimals = (decimals: number | undefined) => {
+    if (decimals === undefined) return DEFAULT_TOKEN_DECIMALS
+    return Number.isInteger(decimals) && decimals >= 0 ? decimals : DEFAULT_TOKEN_DECIMALS
+  }
+  const _toValidBalance = (balance: number | string | undefined, decimals: number) => {
+    if (!balance) return DEFAULT_TOKEN_BALANCE * 10 ** decimals
+    let BNbalance
+    try {
+      BNbalance = BigNumber.from(balance)
```
RodrigoAD, line 53: Converting balance into BN will make sure it doesn't receive any floating number/any other kind of value. The block looks kind of ugly, any ideas are welcome

RodrigoAD:
Ready for review. 😃

composite/synth-index/README.md
```diff
@@ -0,0 +1,139 @@
...
+  - coingecko
+  - coinmarketcap
+  - coinpaprika
+  - cryptocompare
+  - kaiko
+  - nomics
+- `API_KEY`: For those data providers who need an api key
+- `DEFAULT_QUOTE` (Optional): Currency that the price will be fetched by default. `USD` used by default
+
+E.g. if we wish to use Coincapmarket as data provider, we should run (docker):
```
boxhock, line 20:
```suggestion
E.g. if we wish to use CoinMarketCap as data provider, we should run (docker):
```

boxhock (changes requested):
Looks good otherwise

composite/synth-index/README.md
RodrigoAD: Wops

composite/token-allocation/src/adapter.ts
```diff
@@ -35,14 +36,36 @@ export const marketCapTotalValue = (
...
+    } catch (e) {
+      throw new AdapterError({ message: `Invalid balance: ${e.message}`, statusCode: 400 })
+    }
+    if (!BNbalance.gte(0))
```
boxhock, line 57: We can make this easier to read

```suggestion
    if (BNbalance.lt(0))
```

boxhock (changes requested):
One more thing

composite/token-allocation/src/adapter.ts
boxhock: Or we could do `.isNegative()` ?

composite/synth-index/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'Synth Index'
```
justinkaseman, line 5: We should keep the naming convention of no spaces. We have also been capitalizing this in a lot of places.
```suggestion
const NAME = 'SYNTH-INDEX'
```

justinkaseman (changes requested):
Nice! So so much leaner.

Just one quick change.

composite/synth-index/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^14.0.13",
+    "@typescript-eslint/eslint-plugin": "^3.9.0",
+    "@typescript-eslint/parser": "^3.9.0",
+    "ts-node": "^8.10.2",
+    "typescript": "^3.9.7"
+  },
+  "dependencies": {
+    "@chainlink/token-allocation-adapter": "^0.0.1",
+    "synthetix": "^2.35.6"
```
boxhock, line 46: We are currently going to use `2.39.1` on mainnet, so this needs to be updated.

boxhock (dismissed):

composite/synth-index/package.json
RodrigoAD: I'm updating the the presetSlugs on TA-CMC as well

composite/synth-index/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "@types/mocha": "^7.0.2",
+    "@types/node": "^14.0.13",
+    "@typescript-eslint/eslint-plugin": "^3.9.0",
+    "@typescript-eslint/parser": "^3.9.0",
+    "ts-node": "^8.10.2",
+    "typescript": "^3.9.7"
+  },
+  "dependencies": {
+    "@chainlink/token-allocation-adapter": "^0.0.1",
+    "synthetix": "^2.39.1"
```
RodrigoAD, line 46: Is this the correct version @boxhock ?

boxhock: I think we'll want to remove the caret so we get version "2.39.1" and nothing else.

boxhock: @RodrigoAD sorry but it seems we are going with version `"2.38.0-sccp-77-defi-rebalance"` instead :sweat_smile:

justinkaseman (approved):
Great work :1st_place_medal:

PR #264, opened 2021-01-22
Update Token Allocation and related Readme's

RodrigoAD:
Updated readme's related to `token-allocation` adapter: 
- [TA Readme](https://github.com/smartcontractkit/external-adapters-js/blob/2ce0bc7d0f58a6ad6d8ab00a9d4c644377c33754/composite/token-allocation/README.md)

### DIFF
--- a/composite/apy-finance/README.md
+++ b/composite/apy-finance/README.md
@@ -9,7 +9,7 @@ The adapter takes the following environment variables:
 - `RPC_URL`: Blockchain RPC endpoint to get the needed on-chain data
 - `REGISTRY_ADDRESS`: Address Registry contract address used to query the Chainlink Registry
 
-This adapter relies on [`token-allocation`](../../token-allocation/README.md) adapter. Required `token-allocation` input params and configuration apply to this adapter as well.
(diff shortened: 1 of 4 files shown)


boxhock:
Should we add something similar to the CVI adapter too? The only caveat there is that not all of the adapters on token-allocation supports marketcap.

RodrigoAD:
They should all work right? Even if they don't use it, is an extra functionality that they should know. Same for CVI and the price, even if there's no use case now.
On a side note, I've just seen that `token-allocation` does not accept `method` as parameter

boxhock:
> They should all work right? Even if they don't use it, is an extra functionality that they should know. Same for CVI and the price, even if there's no use case now.

Not sure what you mean. It will simply throw an error if you try to use one of the unsupported data providers with CVI. (Kaiko and CoinAPI atm)

> On a side note, I've just seen that token-allocation does not accept method as parameter

Good catch, we should add that

>I guess we should also update the token-allocation adapter readme too, explaining the 2 current methods

Yeah that would be great

RodrigoAD:
> On a side note, I've just seen that token-allocation does not accept method as parameter

I'll add it on #263, I'm doing some TA changes there already

> I guess we should also update the token-allocation adapter readme too, explaining the 2 current methods

I'll add it on this PR

boxhock (approved):

PR #265, opened 2021-01-22
Reduce number of BN libraries

boxhock:


reduce/src/adapter.ts
```diff
@@ -1,7 +1,7 @@
 import objectPath from 'object-path'
 import { Requester, Validator } from '@chainlink/external-adapter'
 import { Execute } from '@chainlink/types'
-import BN from 'bignumber.js'
+import { Decimal } from 'decimal.js'
```
boxhock, line 4: Something I've noticed is that Decimal.js likes to return scientific notation. I would assume everything we use would be able to understand that, but is that something we need to consider more?

dydx-stark/src/endpoint/starkex.ts
```diff
@@ -19,12 +19,19 @@ export type PriceStarkPayload = PriceDataPoint & {
 
 const MAX_DECIMALS = 18
 
-const ZERO_BN = new BN('0')
-const TWO_BN = new BN('2')
-const TEN_BN = new BN('10')
+const ZERO_BN = BigNumber.from('0')
+const TWO_BN = BigNumber.from('2')
+const TEN_DEC = new Decimal('10')
```
krebernisak, line 24: There is no reason to use decimals here, as I do all calculations with big numbers. I convert from decimal to big number manually first, and then all further calculations are BN.

reduce/src/adapter.ts
krebernisak: Two situations that I can think of are:

- When converting from `Decimal` to `BigNumber`, as `BigNumber` does not support scientific notation
- When printing decimal numbers to the output, we don't want the scientific notation there

bootstrap/src/lib/util.ts
```diff
@@ -176,7 +176,7 @@ export const convertUnits = (coin: string, amount: string, unit?: string) => {
   switch (coin.toLowerCase()) {
     case 'btc':
       // default satoshi
-      return new BN(amount).multipliedBy(10 ** 8).toString(10)
+      return new Decimal(amount).mul(10 ** 8).toString()
```
krebernisak, line 179: Not sure how useful this function actually is, the way it's currently set up. Let me check...

krebernisak (changes requested):

krebernisak (dismissed):

reduce/src/adapter.ts
krebernisak: This is now safeguarded against in the `reduce` adapter.

krebernisak (approved):

PR #266, opened 2021-01-25
Update README.md

SeaAnenome:


krebernisak (changes requested):
Thanks, @SeaAnenome.

If you'd like this accepted please just document the input params. You can put it just before the `## Output` title as `## Input`.

There is no need to document full `curl` commands as we do not do it elsewhere, currently.

justinkaseman:
Closing this. We will be taking steps on better READMEs soon.

PR #270, opened 2021-01-26
Add Harmony adapter

gupadhyaya:
This adapter is built to fulfill oracle requests for Harmony blockchain. The repo is similar to: ethwrite-adapter and binance-smart-chain-adapter.

### DIFF
--- a/harmony/README.md
+++ b/harmony/README.md
@@ -1,0 +1,28 @@
+# Chainlink External Adapter for Harmony
+This adapter is built to fulfill Chainlink oracle requests.
+## Configurations
+The adapter uses the following environment variables
+* API_ENDPOINT: a URL to a JSON-RPC (HTTP RPC) node on Harmony, default: `https://api.s0.t.hmny.io`
+* CHAIN_ID: Harmony network chain id, default: `1 (mainnet)`
+* GAS_LIMIT: gaslimit to use, default: `6721900`
+* PRIVATE_KEY: user private key for signing the contract transaction
+
+## Input Params
+
+* address: the oracle contract to fulfill the request on
+* functionSelector: the fulfillment function selector
+* dataPrefix: the data prefix in the request
+* result or dataToSend: the value to fulfill the request with
+
+## Output
+
+```json
+{
+    "jobRunID": "278c97ffadb54a5bbb93cfec5f7b5503",
+    "data": {
+    "result": "0x5e28750a948df26931daa3c01cc3669847936d19e04b40a6819df6bef2b8827d"
+    },
+    "result": "0x5e28750a948df26931daa3c01cc3669847936d19e04b40a6819df6bef2b8827d",
+    "statusCode": 200
+}
+```
--- a/harmony/src/adapter.ts
+++ b/harmony/src/adapter.ts
@@ -1,0 +1,60 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { ExecuteWithConfig, ExecuteFactory } from '@chainlink/types'
+import { Harmony } from '@harmony-js/core'
+import { getAddress, hexToByteArray, hexlify, concat } from '@harmony-js/crypto'
+import { ChainType } from '@harmony-js/utils'
+import { Config, makeConfig, DEFAULT_API_ENDPOINT } from './config'
+
+const inputParams = {
+  address: ['address'],
+  functionSelector: ['functionSelector'],
+  dataPrefix: false,
+  dataToSend: ['dataToSend', 'result'],
+}
+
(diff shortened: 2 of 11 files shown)


harmony/README.md
```diff
@@ -0,0 +1,29 @@
+# Chainlink External Adapter for Harmony
+This adapter is built to fulfill Chainlink oracle requests.
+## Configurations
+The adapter uses the following environment variables
+* HMY_NODE_URL: a URL to a JSON-RPC (HTTP RPC) node on Harmony, default: `https://api.s0.t.hmny.io`
+* HMY_CHAIN_ID: Harmony network chain id, default: `1 (mainnet)`
```
boxhock, line 6:
```suggestion
* RPC_URL: a URL to a JSON-RPC (HTTP RPC) node on Harmony, default: `https://api.s0.t.hmny.io`
* CHAIN_ID: Harmony network chain id, default: `1 (mainnet)`
```

We should prevent prefixing the env var keys here, as then it's easier to prefix it if we're including this adapter in another package. Also replaced the node URL with a key we more commonly use for RPC URLs.

harmony/README.md
```diff
@@ -0,0 +1,29 @@
...
+* GAS_LIMIT: gaslimit to use, default: `6721900`
+* PRIVATE_KEY: user private key for signing the contract transaction
+
+## Input Params
+
+* address: the oracle contract to fulfill the request on
+* functionSelector: the fulfillment function selector
+* dataType: the data type, e.g., `uint256`, `bytes32`, `int256`
```
boxhock, line 14: Is the bytes32 encoding any different from the standard EVM bytes32 encoding? We try to leave most of these things to the adapter before the write adapter, meaning we usually expect the input here to already be encoded.

harmony/src/adapter.ts
```diff
@@ -0,0 +1,77 @@
...
+const DEFAULT_GAS_LIMIT = 6721900
+const gas = process.env.GAS_LIMIT || DEFAULT_GAS_LIMIT
+
+const hmy = new Harmony(hmyRPCEndpoint, {
+  chainType: ChainType.Harmony,
+  chainId: chainID,
+})
+const DUMMY_KEY = 'fd416cb87dcf8ed187e85545d7734a192fc8e976f5b540e9e21e896ec2bc25c3'
+const privateKey = process.env.PRIVATE_KEY || DUMMY_KEY
+hmy.wallet.addByPrivateKey(privateKey)
```
boxhock, line 24: Could you please move all of this configuration into a separate Config object, like this? https://github.com/smartcontractkit/external-adapters-js/blob/530a4552aeddc4abf4357e373d58d6d4ba6b7135/dydx-stark/src/config.ts

This will make it easier for us to give it different config params to test with, and makes it easier if we need to import this adapter in a different package.

harmony/src/adapter.ts
```diff
@@ -0,0 +1,77 @@
...
+  const dataToSend = validator.validated.data.dataToSend || ''
+
+  const value = dataType ? encode(dataType, dataToSend) : strip0x(dataToSend)
+  const transactionData = functionSelector.substring(0, 10) + strip0x(dataPrefix) + value
+
+  const transaction = hmy.transactions.newTx({
+    to: getAddress(address).checksum,
+    data: transactionData,
+    gasLimit: gas,
+    gasPrice: new hmy.utils.Unit('1').asGwei().toWei(),
```
boxhock, line 65: Will this always be a constant?

boxhock (changes requested):
Hi and thanks for the PR! Mostly looking good, just a few things that would be great to update 🙂

harmony/README.md
gupadhyaya: @boxhock it is standard evm bytes32. just followed ethwrite-adapter: https://github.com/smartcontractkit/ethwrite-adapter/blob/master/index.js#L9

harmony/src/adapter.ts
gupadhyaya: yes

harmony/README.md
gupadhyaya: removed.

gupadhyaya:
@boxhock updated, please review. let me know if I have missed anything. thanks.

boxhock:
@gupadhyaya thanks this looks good now! Do you mind please adding a short entry to the changelog as well? Could be a short sentence under the "new adapters" under [Unreleased]:
- `harmony` to write transactions to the Harmony blockchain

gupadhyaya:
> * `harmony` to write transactions to the Harmony blockchain

Added. thanks.

boxhock (approved):

Issue #269, opened 2021-01-26
Docker run cannot read from a file


justinkaseman:
While acceptance testing adapters the command
`docker run -p 8080:8080 --env-file=PATH -it proof-of-reserves-adapter:latest`
was unable to be used to start a docker container with the correct environment variables.

Having this work from would be a lot easier for node operators who needs to start 5+ containers to run composite adapters.

PR #271, opened 2021-01-26
DOCS: composite PoR README

justinkaseman:


justinkaseman:
#269 solution added to README. Leaving that issue open as we need this to be documented better.

krebernisak (approved):

PR #274, opened 2021-01-27
poa-gasprice, polygon, nomics, openexchangerates TS migration

ebarakos:


nomics/src/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+  const validator = new Validator(request, inputParams)
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || DEFAULT_ENDPOINT
+
+  let data
```
RodrigoAD, line 19: You could make a helper to send the `Requester.sucess`, something like:
```js
const _success = (payload: ResponsePayload, result: number) =>
    Requester.success(jobRunID, {
      status: 200,
      data: { sources: [], payload, result },
      result,
    })
 switch (endpoint) {
    case globalmarketcap.Name: {
      const data = await globalmarketcap.execute(config, request)
      return _success(data, data.result)
   ...
}
```

RodrigoAD (commented):
Same as the other PR's, try to make use of `config`. Specially for the API keys and URL's

nomics/src/adapter.ts
ebarakos: I moved the Requester.success inside the specific `endpoint` adapters.

RodrigoAD: You won't need `data` anymore right?

nomics/src/endpoint/globalmarketcap.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+
+  const params = {
+    key: config.apiKey,
+  }
+
+  const reqConfig = {
+    ...config.api,
+    baseURL: `https://api.nomics.com/v1/global-ticker`,
```
RodrigoAD, line 19: We could include the Base url in config, instead of the ticker, and then on each endpoint, include its corresponding url.
I guess `baseUrl` should be `'https://api.nomics.com/v1/`, and then 
- globalMarketCap -> `url: "global-ticker"`
- price -> `url: "/currencies/ticker"`

RodrigoAD (approved):
Nice Job! LGTM

Issue #183, opened 2020-12-08
Support batching for PoR adapters where available


krebernisak:
- [ ] Blockcypher

Unfortunately, this is only useful to [improve latency](https://www.blockcypher.com/dev/bitcoin/#batching):
> With regard to rate limits, each individual batch call counts as a request; for example, if you request 3 addresses in a batch, you're still using 3 API calls of resources on our end. The big advantage to batching is that you avoid 3 separate round-trip calls/reduce latency. Since the default, non-registered rate limit per second is 3, larger batches require a paid API token. To use larger batches please register.

PR #276, opened 2021-01-27
Proof of Reserve batch requests

justinkaseman:
### Description

Further investigation into using batched requests concludes that:

amberdata - no batching
blockchain_com - supports batching
blockcypher - supports batching
blockchair - supports batching
btc_com - no batching
cryptoapis - no batching
sochain - no batching

This PR makes the following changes:
- `BTC.com` recently added API_KEY and API_SECRET. The adapter has been changed to use their SDK.
- To allow the above change the `balance` ea-factory and methods are made into generics, taking an extended Config if needed, but defaulting to the normal Config type.
- Adds batched requests for `blockchain.com` and `blockcypher` (`blockchair` adapter already uses it)
- Because of `blockcypher`'s strict request per second restrictions adds the ability to throttle the number of requests per second. Setting this to `3` works on the free tier key, but takes ~10 seconds to finish for `wbtc`.

### Issues
#183 
#280


blockcypher/src/endpoint/balance.ts
```diff
@@ -28,25 +29,86 @@ const getChainId = (coin: CoinType, chain: ChainType): string => {
...
+/**
+ * Delays the amount of requests sent per second
+ *
+ * @param amount maximum number of requests per second
+ * @param data array of arbitrary data
+ * @param exec function handler of a chunk
+ *
+ * @returns array of response data
+ */
+const throttle = async (amount: number, data: any[], exec: any) => {
```
krebernisak, line 62: Rate limiting, not throttle is the concept we are looking for.

Research rate limit algorithms:

- https://blog.logrocket.com/rate-limiting-node-js/
- https://kendru.github.io/javascript/2018/12/28/rate-limiting-in-javascript-with-a-token-bucket/
- https://aliz.ai/rate-limiting-in-rxjs/

Pick one and implement it to be used like this:

```javascript
const count = 5
const slidingWindowTime = 1000
const _echo = (msg) => console.log(msg)
const _limitedEcho = utils.rateLimit(count, slidingWindowTime, _testFn)

while (true) _limitedEcho("Hello!") // This should write 5x "Hello!", then sleep until the end of the sliding window, and repeat indefinitely
```

Please also document the algorithm you picked and implemented.

krebernisak (changes requested):

ea-factories/src/factories/balance.ts
```diff
@@ -17,14 +11,20 @@ const ERROR_MISSING_ADDRESS = 'No Operation: address param is missing'
...
+) => Promise<BalancesResponse>
+
+export type BalanceConfig<C extends Config = Config> = C & {
   confirmations?: number
   isSupported: IsSupported
-  getBalance?: GetBalance
-  getBalances?: GetBalances
+  getBalance?: GetBalance<C>
+  getBalances?: GetBalances<C>
 }
```
krebernisak, line 28: I assume you noticed there is something wrong here 😅

This exploded in complexity, unnecessarily IMHO.

A hint you can use is that the API should probably be amended to look more like this:

```javascript
export type GetBalances = (accounts: Account[], confirmations?: number) => Promise<BalancesResponse>
```

We don't configure the function calls, we configure the adapter. Also not sure we even need the single returning `GetBalance`, as we can support it all by using `GetBalances`.

btc.com/src/endpoint/balance.ts
```diff
@@ -1,26 +1,31 @@
 import { balance } from '@chainlink/ea-factories'
-import { Requester } from '@chainlink/external-adapter'
-import { Config } from '@chainlink/types'
 import { isChainType, isCoinType } from '.'
+import * as blocktrail from 'blocktrail-sdk'
```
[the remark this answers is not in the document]
justinkaseman: > Blocktrail is now btc.com?
> 
> If this is true could we document it somehow, as it's confusing reading it as is.

I'm not sure how this is all settling on their side.
Blocktrail's website now links to BTC.com and BTC.com's [new docs](https://dev.btc.com/docs/js) say to use the SDK.
Their support email shows their name as bitmain.



> What would be the benefit of accessing the API through this SDK?
Originally I did it because of https://dev.btc.com/docs/js#api-authentication (they hadn't documented how to authenticate so I just went through the SDK), but I _could_ go back now and try sending the API_KEY and API_SECRET as a param.

ea-factories/src/factories/balance.ts
justinkaseman: This happened because the Config needed to be extended to include new per provider env variables such as `API_SECRET`. This same config gets used in the factory, so I made everything a generic to take in possibly extended Conifgs, but also fall back to the default Config if not specified.

Not sure that I understand how we could make the `GetBalances` balance request without the `Config`, but I'll put some thought into how I could refactor & clean this up.

blockcypher/src/endpoint/balance.ts
justinkaseman: Thanks for pointing me in the right direction with the name :sweat_smile:
I renamed this to `rate_limit`.

How do you feel about moving this to a separate issue? PoR is a high priority item for the business side and I worry making this change might take a little bit to implement. I know @RodrigoAD is working on a similar thing, he could take this over to make it very re-usable.

PR #234, opened 2021-01-14
batch adapter TS refactor

emmick4:
Batch PR for updating the following adapters to typescript:
```
coinpaprika
cryptocompare
currencylayer
dxfeed
eodhistoricaldata
etherchain
ethgasstation
fixer
fmpcloud
```
Of note, dxfeed seems to be broken, the request attempting to be made in the tests gets `{"status": "TIMEOUT"}` as the result, that was the case with the adapter before the typescript update, and when attempting the call with just `curl`. opted to include the refactor but not sure what's actually wrong there 🤷

PR #278, opened 2021-01-28
dxfeed-secondary adapter

justinkaseman:
### Description
A quick secondary adapter to provide support for the `TSLA` base to convert to `TSLA.US:TEI`

Also brings along the TS migration of `dxfeed` from #234 for easier importing.

### Caveats
dxfeed API currently down?

### Ticket
[176704940](https://www.pivotaltracker.com/story/show/176704940)


dxfeed/src/endpoint/price.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const customParams = {
+  base: ['base', 'from', 'coin'],
+}
+
+const commonSymbols: { [key: string]: string } = {
+  N225: 'NKY.IND:TEI',
+  FTSE: 'UKX.IND:TEI',
+  TSLA: 'TSLA:BFX',
+  TSLAX: 'TSLA.US:TEI',
```
thodges-gh, line 19: Let's remove this one.

dxfeed-secondary/src/endpoint/price.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const customParams = {
+  base: ['base', 'from', 'coin'],
+}
+
+const commonSymbols: { [key: string]: string } = {
+  N225: 'NKY.IND:TEI',
+  FTSE: 'UKX.IND:TEI',
+  TSLA: 'TSLA.US:TEI',
+  TSLAX: 'TSLA:BFX',
```
thodges-gh, line 15: We can remove this.

thodges-gh (changes requested):
Just a quick note to clarify the secondary-market's usage. We'll be using it only for symbols which are reported from two different markets, like TSLA. So that on the primary adapter supplying the base of TSLA pulls from one market, then the other on the secondary adapter.

thodges-gh (approved):

PR #281, opened 2021-02-01
Klaytn adapter

winnie-byun:
This PR creates an external adapter for Klaytn.


emmick4 (changes requested):
This adapter looks good, great work! Fortunately, we've recently updated our adapter framework to Typescript however and so we will need this to be updated to Typescript in the same form as other adapters (see our `develop` branch `example` adapter for a reference.) In doing this upgrade please be sure to add the adapter to the top-level package.json of this repo so it's included in the build process. Thank you!

winnie-byun:
@emmick4
I've changed it to ts. PTAL :)
* I still have linter an error because of `caver-js` pacakge. It is only implemented in js and need `require` intead of `import.
* I ran this code in https://github.com/winnie-byun/demo-klaytn-chainlink. Do I still need to make testcodes?

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const address = validator.validated.data.address
+  const data = validator.validated.data.data
+  const topcis = validator.validated.data.data
+  const value = validator.validated.data.value
+
+  return sendFulfillment(
```
justinkaseman, line 103: It might be a little bit cleaner here to get rid of the callback and use async/await here.

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
+import { Requester, Validator, AdapterError } from '@chainlink/external-adapter'
+import { Config, ExecuteWithConfig, ExecuteFactory, AdapterResponse } from '@chainlink/types'
+import { makeConfig } from './config'
+import { LegacyTransaction } from './types'
+import fs from 'fs'
+
+const Caver = require('caver-js')
+const caver = new Caver(process.env.URL)
```
justinkaseman, line 8: I'd recommend moving line 8 to inside of `klaytn/src/config.ts`'s `makeConfig`. We have an internal util that may help: 
```

// makeConfig
const util.getRequiredEnv('URL, prefix)
```
 which will throw if the URL is not provided is not provided. It can then be utilized from the Config object parameter on line 90. This means they will be logged when the External Adapter starts up (the secret will be hidden) which can be useful information when debugging.

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const privateKey = process.env.PRIVATE_KEY
```
justinkaseman, line 10: `klaytn/src/config.ts`'s `getDefaultCOnfig` will look for `API_KEY` if you want to use that environment variable name instead.

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+  value: string,
+  callback: (txhash: string, e?: Error) => Promise<AdapterResponse>,
+) => {
+  // decode data in log
+  const decoded = caver.abi.decodeLog(
+    JSON.parse(fs.readFileSync('./OracleRequestABI.json').toString()),
+    logData,
+    topics,
+  )
+  console.log('[Decoded Data] \n', decoded)
```
justinkaseman, line 27: If you want to keep this log I would make it a `DEBUG` level log using:
```
import { logger } from '@chainlink/external-adapter'

  logger.debug('[Decoded Data]: ', { decoded })
```

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+  console.log('[Received Data] \n', request)
+  const validator = new Validator(request, customParams)
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const address = validator.validated.data.address
+  const data = validator.validated.data.data
+  const topcis = validator.validated.data.data
```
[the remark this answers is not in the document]
justinkaseman: Also I think this is meant to be `topics`

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,131 @@
...
+      })
+      .on('receipt')
+      .on('error', (error: Error) => {
+        err = error
+      })
+  })
+  return callback(transactionHash, err)
+}
+
+const customParams = {
```
justinkaseman, line 83: These parameter names should match the README which says:
```
- `address` : The oracle contract to fulfill the request on
- `dataPrefix` : The data prefix in the request
- `functionSelector` : The fulfillment function selector
- `result` or `value` : The value to fulfill the request with
```

One or the other should change.
The validator will look at the request's JSON body for these property names.

justinkaseman (changes requested):
Hey @winnie-byun, thanks for the PR :smile: 

Added some comments. It would also be great to have some unit tests on here! They will be very similar to other tests found across the repository.

justinkaseman:
Ah also please add an `.eslintrc.js` file in the root with:
```
module.exports = {
  ...require('../.eslintrc.ts.js'),
}
```

klaytn/src/adapter.ts
winnie-byun: I modified the code to use `util` for env vars. However, I still don't see the flags in the log.
```
yarn start
yarn run v1.22.5
$ tsc -b
✨  Done in 1.79s.
yarn run v1.22.5
$ yarn server:dist
$ node -e 'require("./dist/index.js").server()'
{"message":"Listening on port 5002!","level":"info","instanceId":"102171fe-9e45-4f74-89cd-860ddcf3df27","timestamp":"2021-03-02T02:19:43.148Z"}
{"message":"API will be rate limited without an API key.","level":"warn","instanceId":"102171fe-9e45-4f74-89cd-860ddcf3df27","timestamp":"2021-03-02T02:19:47.292Z"}
{"message":"API will be rate limited without an API key.","level":"warn","instanceId":"102171fe-9e45-4f74-89cd-860ddcf3df27","timestamp":"2021-03-02T02:19:47.328Z"}
...
```

winnie-byun:
@justinkaseman I applied your comments. PTAL :)

justinkaseman:
@winnie-byun Thanks for the changes!

I pushed a commit for some TypeScript tweaks and linting. Take a look and see if it still works how you expect.

klaytn/test/adapter.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        testData: {
+          data: {
+            address: '0xf1e73a60687affcb1d6308fcb2ef1da3865ff797',
+            data:
+              '0x000000000000000000000000344d20e71a76b5a44fff62438cd17446a331bf3e70ae84788a565ad436a112ee40d236e6f95106e22bd45cd8b78c8fe8c93c4df0000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000344d20e71a76b5a44fff62438cd17446a331bf3ed204b3fb0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000603d9865000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000',
+            topics: [
+              '0xd8d7ecc4800d25fa53ce0372f13a416d98907a7ef3d8d3bdd79cf4fe75529c65',
+              '0x3836643062336366366262313466306338633237363234316636383361363765',
+            ],
+            value: '{"jobRunID":"1","data":{"result":"1003904"},"result":"1003904"}',
```
justinkaseman, line 24: Is there a reason why the value is structured like our AdapterResponse and not just the value itself?

(e.g. value: "1003904")

klaytn/README.md
```diff
@@ -0,0 +1,38 @@
...
+| Required? |        Name         |                                 Description                                  | Options | Defaults to |
+| :-------: | :-----------------: | :--------------------------------------------------------------------------: | :-----: | :---------: |
+|    ✅     |      `address`      |                The oracle contract to fulfill the request on                 |         |             |
+|    ✅     |       `data`        | The data that contains functionSelector and dataPrefix prefix in the request |         |             |
+|    ✅     |      `topics`       |                      The fulfillment function selector                       |         |             |
+|    ✅     | `result` or `value` |                    The value to fulfill the request with                     |         |             |
+
+---
+
+### Sample Input
```
justinkaseman, line 25: Added a space for something like
```suggestion
```json
{
  "jobRunID": "1",
  "data": {
             "address": '0xf1e73a60687affcb1d6308fcb2ef1da3865ff797',
            "data":
        "0x000000000000000000000000344d20e71a76b5a44fff62438cd17446a331bf3e70ae84788a565ad436a112ee40d236e6f95106e22bd45cd8b78c8fe8c93c4df0000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000344d20e71a76b5a44fff62438cd17446a331bf3ed204b3fb0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000603d9865000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000",
            "topics": [
              "0xd8d7ecc4800d25fa53ce0372f13a416d98907a7ef3d8d3bdd79cf4fe75529c65",
              "0x3836643062336366366262313466306338633237363234316636383361363765",
            ],
            "value": "1003904"
  }
}
```
```

Pending that comment about that value param.

File: klaytn/README.md [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+
+### Sample Output
+
+```json
+{
+  "jobRunID": "1",
+  "data": {
+    "result": "0xf5e2f6cf458ba8aa2676458888bff34536954dfb1e6aa02523d0c1143112761c"
```
justinkaseman, line 33: Would be nice to include the full result from `caver.rpc.klay.sendRawTransaction` here!

justinkaseman: Basically what gets returned when `export VERBOSE=true` is on

klaytn/test/adapter.test.ts
winnie-byun: I used chainlink's external-adapter package to build my test external adapter. Isn't this the common scenario you expect for chainlink?
https://github.com/winnie-byun/demo-klaytn-chainlink/blob/main/api_server_timestamp/index.js

klaytn/src/adapter.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ExecuteWithConfig, ExecuteFactory } from '@chainlink/types'
+import { logger } from '@chainlink/external-adapter'
+import { CustomConfig, makeConfig } from './config'
+import fs from 'fs'
+import Caver from 'caver-js'
+
+const makeTx = async (caver: any, keyring: any, input: any): Promise<txResult> => {
+  let { value } = input
+  const { address, data, topics } = input
+  value = JSON.parse(value).result
```
winnie-byun, line 11: @justinkaseman This line doesn't parse the input.
```suggestion
  const address = input.data.address
  const data = input.data.data
  const topics = input.data.topics
  let value = JSON.parse(input.data.value).result
```

winnie-byun:
@justinkaseman It looks like there is an error in `caver.rpc.klay.sendRawTransaction`. I will inform you after I find out what's wrong with the code.

justinkaseman:
Pushed a commit for using just a single number for the value param. Could you try one more time?

If that doesn't work I can set up a Klaytn node and help debug.

klaytn/src/adapter.ts
justinkaseman: Oops, good catch :sweat_smile:

klaytn/test/adapter.test.ts
justinkaseman: The test external adapter looks correct.

For this Klaytn adapter the response structure is made here:
```
    return Requester.success(jobRunID, {
      data: { result },
      result,
      status: 200,
    })
```

Regarding the value param, it should be any information needed to make the request (submit the Klaytn transaction). Since it seems like all we use is the "1003904" in:
```

  const params = caver.abi
    .encodeParameters(
      ['bytes32', 'uint256', 'address', 'bytes4', 'uint256', 'bytes32'],
      [
        decoded.requestId,
        decoded.payment,
        decoded.callbackAddr,
        decoded.callbackFunctionId,
        decoded.cancelExpiration,
        caver.utils.leftPad(value, 64), //////////// HERE
      ],
    )
    .substring(2)
```

winnie-byun: No there is no such structure and "1003904" is all we use.

I meant when the values are from [another external-adapter](https://github.com/winnie-byun/demo-klaytn-chainlink/blob/main/api_server_timestamp/index.js), what I see in klaytn external-adapter is : 
```
   value: '{"jobRunID":"1","data":{"result":"10030019"},"result":"10030019"}'
```
Is there some kind of API to get only the value?

klaytn/README.md
winnie-byun: I added the result. But `API_VERBOSE=true` still doesn't work for me. I just logged to get the result here.

winnie-byun (commented):
@justinkaseman I think it works except for parsing JSON part.
The error in `caver` was just my mistake. I put a wrong address for Klaytn client.
This code works fine :)

klaytn/test/adapter.test.ts
justinkaseman: Oh, you are planning to chain together two external adapters right?

We call those `composite adapters`. They can be seen in the [composite folder](https://github.com/smartcontractkit/external-adapters-js/tree/develop/composite). Currently they are undergoing a large refactor so that the composite adapter sends requests to the underlying external adapters.

It would look like:
- a new composite adapter that will...
- gets the time from `api_server_timestamp` 
- parse the value from the previous response and use it as `value` to request the `klaytn` adapter

[This refactored composite adapter](https://github.com/smartcontractkit/external-adapters-js/pull/349) might serve as an example until we finalize the refactor and show an example composite adapter.

justinkaseman:
@winnie-byun I think this is ready to go. Could you test one more time from your end?

winnie-byun:
@justinkaseman I've tested and it worked. (except for the value format.)
I think we can merge :)

justinkaseman:
> @justinkaseman I've tested and it worked. (except for the value format.)
> I think we can merge :)

Remind me, what is the value format part?

winnie-byun:
I am talking about the composite adapters.
My demo example uses the composite adapters and I've tested with the composite adapters.
But I have to change one line to run in this project.
Before : 
```
let { value } = input
```
After : 
```
let value = JSON.parse(input.value).result
```

boxhock:
Hey @winnie-byun, are you interested in updating this PR or are you OK with closing it for now?

boxhock:
Hey @winnie-byun , I'm going to close your PR for now as it is outdated. Feel free to re-open the PR at a later time when it's been updated, and we'll do our best to get it reviewed and processed quickly!

PR #283, opened 2021-02-02
Kaiko support for eth

RodrigoAD:
When asking to convert into ETH, Kaiko uses a different endpoint

### DIFF
--- a/composite/token-allocation/src/data-providers/kaiko.ts
+++ b/composite/token-allocation/src/data-providers/kaiko.ts
@@ -2,7 +2,9 @@ import { Requester } from '@chainlink/external-adapter'
 import { util } from '@chainlink/ea-bootstrap'
 
 const getPriceData = async (symbol: string, currency: string) => {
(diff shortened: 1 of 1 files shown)


composite/token-allocation/src/data-providers/kaiko.ts
```diff
@@ -2,7 +2,9 @@ import { Requester } from '@chainlink/external-adapter'
 import { util } from '@chainlink/ea-bootstrap'
 
 const getPriceData = async (symbol: string, currency: string) => {
-  const url = `https://us.market-api.kaiko.io/v2/data/trades.v1/spot_exchange_rate/${symbol.toLowerCase()}/${currency.toLowerCase()}`
+  const endpoint =
+    currency.toLowerCase() === 'eth' ? 'spot_direct_exchange_rate' : 'spot_exchange_rate'
```
krebernisak, line 6: Is this only for ETH? Are there any other currencies we need to consider?

RodrigoAD: AFAIK yes, is the only one considered on the main Kaiko adapter as well

krebernisak (dismissed):

krebernisak (approved):

Issue #217, opened 2021-01-09
Config improvements


justinkaseman:
### Description

Re-usable Configs were recently added to reduce shared code between EAs. This was a simple first-step and should be improved upon.

There was concern about them being too static. A generic type was added to ExecuteFactory to allow many Config types. This concern should be given more thought for the future.

These configurations should be extendable and be able to come from a variety of sources (URL params, ENV variables, argument to factory, ect.)

The compostability of these configurations in their current form should be verified.

justinkaseman:
The introduction of cycling through API keys requires an API key to be set. Some data providers do not need API keys. `defaultConfig` needs to be told if an API key is needed. One solution might be to give it a Config object of defaults. Anything not given is filled from framework defaults.

boxhock:
Is this resolved or do we have a Shortcut story for this?

justinkaseman:
> Is this resolved or do we have a Shortcut story for this?

Yeah I believe every part of this is resolved.

justinkaseman:
Closing as complete

PR #290, opened 2021-02-03
Migrating ethwrite adapter to monorepo

ebarakos:
Migration of this adapter: https://github.com/smartcontractkit/ethwrite-adapter


ethwrite/deploy_contract.ts
```diff
@@ -0,0 +1,42 @@
...
+  '0190f35b3480156100af57600080fd5b5061009161011d565b3480156100c457' +
+  '600080fd5b506100d0600435610123565b005b3480156100de57600080fd5b50' +
+  '6100d0600435610128565b3480156100f657600080fd5b506100d06004356101' +
+  '2d565b34801561010e57600080fd5b50610091610132565b60005481565b6001' +
+  '5481565b600255565b600055565b600155565b600254815600a165627a7a7230' +
+  '582062f6c7201e1a3698c586add5b8e7d1f047a1fcfd8ca2f518c06790fba3de' +
+  '22d80029'
+
+// https://docs.ethers.io/ethers.js/html/api-contract.html#deploying-a-contract
+export async function deploy(): Promise<string> {
```
RodrigoAD, line 33: This should receive provider and pk as parameters. Then from the test, you could inject the hardhat config

ethwrite/test/adapter.test.ts
```diff
@@ -0,0 +1,141 @@
+import { assert } from 'chai'
+import { Requester } from '@chainlink/external-adapter'
+import { assertSuccess, assertError, startChain } from '@chainlink/adapter-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../src/adapter'
+import { deploy } from '../deploy_contract'
+
+// using DELAYED ROOT SUITE in order to start the chain and deploy the contract
+setTimeout(async function () {
```
RodrigoAD, line 9: Not sure about this approach. Could a `before` work?

ebarakos: Spent some time to make it working with `before`, but async initialization does not work on `describe`/`context`(https://github.com/mochajs/mocha/issues/2975).

Async calls could be resolved inside `it` though.

ethwrite/src/config.ts
```diff
@@ -0,0 +1,13 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'txsend'
+
+export const DEFAULT_RPC_URL = 'http://localhost:4444'
```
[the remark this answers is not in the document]
ebarakos: Set them as defaults for running the tests, but I better set these values inside the test file.

ethwrite/deploy_contract.ts
```diff
@@ -0,0 +1,42 @@
+import { ethers } from 'ethers'
```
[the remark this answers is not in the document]
ebarakos: This helper file/function is also used, apart from testing, for deploying the contract, therefore initializing the adapter(instructions in README).

ethwrite/deploy_contract.ts
```diff
@@ -0,0 +1,42 @@
...
+export async function deploy(): Promise<string> {
+  const factory = new ethers.ContractFactory(abi, byteCode, wallet)
+  const contract = await factory.deploy()
+  console.log('Contract deployed at: ', contract.address)
+  console.log('Transaction: ', contract.deployTransaction.hash)
+  await contract.deployed()
+  return contract.address
+}
+
+deploy()
```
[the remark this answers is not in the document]
ebarakos: This is executed for the reason mentioned above.

ethwrite/README.md
```diff
@@ -0,0 +1,70 @@
...
+- Parse the desired field from that data source's response
+- Utilize this adapter to write the value to ChainB
+- Parse the transaction object from ChainB for the transaction hash
+- Write the transaction hash from ChainB to ChainA
+
+## Input Params
+
+- `exAddr`: The address for sending the transaction to.
+- `funcId`: (Optional) One of the following setter functions: `0xc2b12a73` for `bytes32`, `0xa53b1c1e` for `int256` or `0xd2282dc5` for `uint256` (defaults to `0xd2282dc5`)
+- `dataType`: (Optional) One of the following types: `bytes32`, `int256` or `uint256` (defaults to `uint256`)
```
boxhock, line 17: We normally get the data pre-encoded from the previous tasks in the job run, and IMO we should not have dataType as a param (it should be assumed it's already encoded)

ebarakos: Made it optional(default usage will not encode the result), to retain the other functionality as well.

ethwrite/read_contract.ts
```diff
@@ -0,0 +1,28 @@
+import { ethers } from 'ethers'
```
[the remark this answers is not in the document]
ebarakos: Good idea. I altered the tests so the actual value is read and checked against the intended input.

But left the the `read_contract.ts` as a standalone script for checking contract state easily from CLI, as existed in previous repo.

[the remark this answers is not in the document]
RodrigoAD: I'm not really sure that this scripts are required anymore, could you double check it? They were probably provisional with testing purposes in the old adapter, not sure if they are needed again in this version.
If they don't, you could use their functionality to unit test the adapter (under `/tests`)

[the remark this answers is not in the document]
ebarakos: `deploy_contract.ts` can be used by node operators to easily deploy the needed bytecode. If we remove this, we need to provide instructions for deploying the contract manually. But I don't have more info on it. @thodges-gh or @boxhock maybe?

`read_contract.ts` is mostly extra utility, for verifying the contract state.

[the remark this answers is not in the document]
boxhock: I don't think we'll need these files. It should be outside the scope of this adapter.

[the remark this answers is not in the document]
ebarakos: In this case, a node operator that wants to use this adapter, how can he know on which contract to refer to and/or which contract to deploy?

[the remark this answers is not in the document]
boxhock: The node operator likely isn't going to be deploying any adapters, they will just use this as part of the job spec. The address and all that is likely provided by the initiator, so the node operator doesn't necessarily have much visibility into this.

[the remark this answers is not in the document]
ebarakos: Ah, OK. Removing these then.

test-helpers/src/hardhat.ts
```diff
@@ -0,0 +1,28 @@
+import { JsonRpcServer } from 'hardhat/internal/hardhat-network/jsonrpc/server'
+import { TASK_NODE_GET_PROVIDER, TASK_NODE_CREATE_SERVER } from 'hardhat/builtin-tasks/task-names'
+import { run } from 'hardhat'
+
```
RodrigoAD, line 4: You can export the private key from here

hardhat.config.js
```diff
@@ -0,0 +1,34 @@
...
+
+module.exports = {
+  defaultNetwork: 'hardhat',
+  networks: {
+    hardhat: {
+      chainId: 31337,
+      accounts: [
+        {
+          privateKey:
+            process.env.PRIVATE_KEY ||
```
RodrigoAD, line 15: You could require the `test-helpers/src/hardhat.ts` here, and use the imported PK

ethwrite/test/adapter.test.ts
```diff
@@ -0,0 +1,200 @@
...
+import { makeExecute } from '../src/adapter'
+import { abi, deploy } from '../src/contract_helpers'
+import { ethers } from 'ethers'
+
+// using DELAYED ROOT SUITE in order to start the chain and deploy the contract
+setTimeout(async function () {
+  const chain = await startChain(4444)
+  // TODO: There are the default hardhat setup. Could be imported from a source of default/testing configuration.
+  const rpcUrl = 'http://localhost:4444'
+  const privateKey = '0x90125e49d93a24cc8409d1e00cc69c88919c6826d8bbabb6f2e1dc8213809f4c'
```
RodrigoAD, line 14: You could import the PK from the `test-helpers`

ethwrite/src/contract_helpers.ts
```diff
@@ -0,0 +1,37 @@
+import { ethers } from 'ethers'
```
RodrigoAD, line 1: As this is only used in testing, send it there (under `/test/helpers` should be fine)

ethwrite/src/config.ts
```diff
@@ -0,0 +1,19 @@
...
+  privateKey: string
+  api: any
+}
+
+export const DEFAULT_ENDPOINT = 'txsend'
+
+export const makeConfig = (): Config => {
+  return {
+    api: {},
+    rpcUrl: util.getEnv('RPC_URL'),
```
RodrigoAD, line 15: Make `rpcUrl` and `privateKey` required

ethwrite/src/config.ts
```diff
@@ -0,0 +1,19 @@
+import { util } from '@chainlink/ea-bootstrap'
+
+export type Config = {
+  rpcUrl: string
+  network?: string
+  privateKey: string
+  api: any
```
RodrigoAD, line 7: I think you can remove `api` right?

ethwrite/README.md
```diff
@@ -0,0 +1,70 @@
...
+    "r": "0x58389e578005462f7e8d0c980cd4fca9e16ab4141882f56f98deec0b572da0b3",
+    "s": "0x17a977f57a88c3dbbab89beac60beb78ae8ab4dffd676831f84d14151c5b03d5",
+    "from": "0x78C696E4cA526f17380DAc7b6C5fd54B17F3f637",
+    "hash": "0x311b1e57342f61379bc597813c41bfa7c3535cfc2a49e61b2456c602dd07708e"
+  },
+  "statusCode": 200
+}
+```
+
+## Deploy & Test
```
RodrigoAD, line 44: Update the Readme as well. This won't be necessary anymore

ethwrite/test/adapter.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  const address = await deploy(privateKey, rpcUrl)
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
+  const contract = new ethers.Contract(address, abi, provider)
+
+  describe('execute', async () => {
+    const execute = makeExecute({ rpcUrl, privateKey, api: {} })
```
RodrigoAD, line 20: Nice 👍

test-helpers/src/hardhat.ts
ebarakos: Thanks @RodrigoAD, very good idea. I was trying to directly import from hardhat config and was running into path issues.

ethwrite/src/config.ts
ebarakos: If I want to use `ExecuteWithConfig<C extends Config>`, I can't, since it is required in main Config type. And I don't feel like changing the `adapter.ts` boilerplate, since IMHO should be removed at some point.

I was thinking of changing this type to ExecuteWithConfig<C> but I am considering of checking this again separately, maybe as part of https://github.com/smartcontractkit/external-adapters-js/pull/293.

I can also just use `execute = async (input: AdapterRequest, config: Config): Promise<AdapterResponse>` for the particular adapter as in `synth-index` though. Thoughts?

RodrigoAD: You're right, the tweak is made is some places is defining the makeExecute as `const makeExecute = (config?: Config): Execute => {...`

ebarakos: Let's check changing it to either ExecuteWithConfig<C> or removing the requirement of `api` param, in another PR, what do you think?

RodrigoAD: Let's leave it as it is. There's an issue open on `Config` inflexibility, we can leave it for another PR

hardhat.config.js
```diff
@@ -0,0 +1,33 @@
+// Hardhat configuration
+// See (https://hardhat.org/config/) for more details.
+
+// const { MNEMONIC } = require('./test.env')
+// const MNEMONIC = 'novel mobile inform nurse circle spoon cricket soup crowd clip hawk glad'
```
RodrigoAD, line 5: Can we clean this comments?

RodrigoAD (commented):
I think only that last changes are needed

ethwrite/src/config.ts
justinkaseman: Tagging #217

ethwrite/test/adapter.test.ts
justinkaseman: Opened a low priority issue #321 just so we remember this

justinkaseman (approved):
Great work  :1st_place_medal:. LGTM

RodrigoAD (approved):

PR #292, opened 2021-02-03
Rate Limit on Cache

RodrigoAD:
## Context
Composite adapters now rely on price adapters. Price adapters will act as a central point of every composite adapter

## Description
- Rate Limit takes the Data Provider __capacity__ (max. requests per minute allowed) as configuration. 
- The price adapter will store in memory every request that is receiving. Every individual request will be a __participant__
- Every __participant__ will have a __heartbeat__ (number of requests received on the price adapter so far) and __lastSeen__ (last request date)
- The time the response will be stored on cache depends on the number of participants, and the capacity. The simple approach is distribute the capacity evenly between every adapter
   - If `capacity = 20` and there are 10 adapters registered, each of them will be able to make `requestPerMin = 0.9*capacity / totalParticipants => reqMin = 18 / 10 => reqMin = 1.8`
   - This translates into a max age (ms) of: `1000 / (reqMin/60) => 33.333 ms`
   - For every new participant, the max age will increase
- For every request, there will be a lastSeen check. Any participant that hasn't appeared in the last configurable `groupMaxAge` time (2 hrs by default) will be removed as a participant, leaving room for the other participants

### More accurate configuration (optional)
- Adapters request frequency is variable, some request every 10 min, others every 3 hours, others every 24 hours.
- In the previous approach, the capacity is distributed evenly between every participant. But those who request every long periods of time, will waste their assigned capacity
- To solve this, adapters will receive it's piece of capacity based on their request frequency. Those who call more often, will get more capacity
    - For this, the price adapter calculates the participant weight as: `weight = participantHeartbeat / totalHeartbeat`. This translates into an allowed request per minute as: `reqMin = 0.9*capacity * weight`

## Small Problems to consider
- It take some time for the data provider adapter to distribute properly the capacity. If every feed joins in the same minute,  it'll probably hit the limit in that minute, until it has the information from every adapter
- ~~An invalid request will add a participant, dividing the capacity~~
- With the accurate configuration, newcomers that join the adapter with high heartbeats, will struggle to find some room in the capacity


bootstrap/src/lib/cache/index.ts
```diff
@@ -9,19 +9,28 @@ import { RedisOptions } from './redis'
...
+const DEFAULT_RATE_COST = 1
+
 const env = process.env
 export const defaultOptions = () => ({
   enabled: parseBool(env.CACHE_ENABLED),
   cacheOptions: defaultCacheOptions(),
   cacheBuilder: defaultCacheBuilder(),
   key: {
     group: env.CACHE_KEY_GROUP || DEFAULT_CACHE_KEY_GROUP,
+    rateLimitGroup: env.API_KEY || undefined,
```
RodrigoAD, line 31: The API_KEY will be the group identifier. If is not specified, `maxAge` will be calculated without rlg

bootstrap/src/lib/cache/index.ts
```diff
@@ -9,19 +9,28 @@ import { RedisOptions } from './redis'
...
+    rateLimitParticipant: DEFAULT_CACHE_KEY_RATE_LIMIT_PARTICIPANT,
+    totalCapacity: parseInt(env.CACHE_RATE_CAPACITY || DEFAULT_CACHE_RATE_CAPACITY),
```
RodrigoAD, line 33: What is capacity exactly? Rate limit per second, hour or day? This should affect to the maxAge calculation

composite/token-allocation/src/adapter.ts
```diff
@@ -99,7 +99,7 @@ export const execute = async (input: AdapterRequest, config: Config): Promise<Ad
   const _success = (payload: ResponsePayload, result: number) =>
     Requester.success(jobRunID, {
       status: 200,
-      data: { sources: [], payload, result },
+      data: { sources: [], payload, result, cost: 10 },
```
RodrigoAD, line 102: Just as an example

bootstrap/src/lib/cache/index.ts
```diff
@@ -99,19 +216,32 @@ export const withCache = async (
...
     const coalescingKey = _getCoalescingKey(key)
-    const maxAge = _getMaxAge(data)
     // Add successful result to cache
     const _cacheOnSuccess = async ({ statusCode, data }: WrappedAdapterResponse) => {
       if (statusCode === 200) {
+        if (rateLimit.isEnabled()) {
+          await rateLimit.updateRateLimitGroup(data.data.cost, data.data.weight)
+        }
+        let maxAge = await _getMaxAge(request)
+        if (maxAge < 0) {
```
RodrigoAD, line 242: If max age from the request was `< 0` will get the default age

bootstrap/src/lib/cache/index.ts
```diff
@@ -70,6 +81,112 @@ export const redactOptions = (options: CacheOptions) => ({
...
+        [options.participantId]: {
+          cost,
+          weight,
+        },
+      },
+    }
+    if (cache instanceof local.LocalLRUCache) {
+      await cache.set(options.groupId, newGroup, options.groupMaxAge)
+    } else {
+      await cache.setKeepingMaxAge(options.groupId, newGroup, options.groupMaxAge)
```
RodrigoAD, line 144: Keeping the same age as it was before updating the group avoids renewing an old participant when a new one joins the group

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,107 @@
...
+    const participantsMaxAge = Object.fromEntries(
+      Object.entries(rlGroup.participants).map(([id, _]) => [
+        id,
+        _maxAgeFor(rlGroup.participants[id]),
+      ]),
+    )
+    return participantsMaxAge[participantId]
+  }
+
+  const _ifEnabled = (fn: (...args: any[]) => Promise<any>) => (_isEnabled() ? fn : () => false)
```
RodrigoAD, line 101: No function is exposed if the feature is not enabled.

bootstrap/src/lib/cache/index.ts
```diff
@@ -27,6 +31,12 @@ export const defaultOptions = () => ({
       ...(env.CACHE_KEY_IGNORED_PROPS || '').split(',').filter((k) => k), // no empty keys
     ],
   },
+  rateLimit: {
+    groupMaxAge: parseInt(env.GROUP_MAX_AGE || '') || DEFAULT_GROUP_MAX_AGE,
+    groupId: (env.API_KEY && hash(env.API_KEY)) || '',
```
justinkaseman, line 36: What about for composite adapters that use a prefix on their API_KEY? 
Currently Proof of Reserves and Outlier Detection do this.

I see a couple solutions:
1. Have them set API_KEY too (no more work but bad UX)
2. Pass through a prefix to `expose`
3. Set an optional prefix ENV VAR (is this even better UX?, maybe more intuitive naming)

This is kind of a weird case because I don't think we even currently have code that uses multiple API keys.

justinkaseman: Also makes me think about this case in a general sense - 

We have a composite adapter called price-compare that uses 2 adapters: `CMC` and `coingecko`
Which one do we key off of? `CMC`? It has no way to handle both.
Now what if the same node operator is also using `CMC` as a stand-alone adapter?

This may be an edge case that we can note and avoid until V2.
Or means that individual request caching should be thought of now.

justinkaseman:
There are a couple other places that we already know make multiple requests, like Proof of Reserves.

We should figure out if there are any more and add the `cost` to the response there too.

bootstrap/src/lib/cache/index.ts
RodrigoAD: Really good point, thanks for bringing it out. Working on it

justinkaseman (dismissed):
:+1:

bootstrap/src/lib/cache/index.ts
```diff
@@ -28,6 +32,12 @@ export const defaultOptions = () => ({
...
+    participantId: DEFAULT_CACHE_KEY_RATE_LIMIT_PARTICIPANT,
```
RodrigoAD, line 38: ParticipantId will be different for Lambda Functions

RodrigoAD (commented):
Block until Participant Id and Heartbeat are resolved

composite/token-allocation/src/adapter.ts
```diff
@@ -132,7 +110,7 @@ export const execute = async (input: AdapterRequest, config: Config): Promise<Ad
       // eslint-disable-next-line no-case-declarations
       const price = await computePrice(config, allocations, quote)
       return _success(price.payload, price.result)
-    case 'marketcap':
+    case 'marketCap':
```
krebernisak, line 113: @boxhock didn't you lowercased this recently in a hotfix IIRC?

boxhock: Yeah. We should convert the input to lowercase and then compare that

bootstrap/src/lib/cache/index.ts
```diff
@@ -97,19 +109,29 @@ export const withCache: Middleware<CacheOptions> = async (execute, options = def
...
-    const maxAge = _getMaxAge(data)
     // Add successful result to cache
-    const _cacheOnSuccess = async ({ statusCode, data, result }: AdapterResponse) => {
+    const _cacheOnSuccess = async (
+      { statusCode, data, result }: AdapterResponse,
+      maxAge?: number,
+    ) => {
       if (statusCode === 200) {
+        // Heartbeat is incremented only if valid request
+        rateLimit.incrementParticipantHeartbeat(_getKey(request), data.cost)
```
RodrigoAD, line 131: The participant heartbeat is added or increased, only if the the request has succeeded. It includes the underlying cost coming from the reponse.

bootstrap/src/lib/cache/index.ts
```diff
@@ -97,19 +109,29 @@ export const withCache: Middleware<CacheOptions> = async (execute, options = def
...
+        if (!maxAge) {
```
RodrigoAD, line 132: This basically gives the option to override the default maxAge calculation, useful when it comes in the request

bootstrap/src/lib/cache/index.ts
```diff
@@ -150,19 +172,24 @@ export const withCache: Middleware<CacheOptions> = async (execute, options = def
...
     if (entry) {
-      if (maxAge >= 0) {
+      if (!requestMaxAge || requestMaxAge >= 0) {
         logger.debug(`Cache: GET ${key}`, entry)
-        if (maxAge !== entry.maxAge) await _cacheOnSuccess(entry)
-        return { jobRunID: data.id, ...entry }
+        // If maxAge is present on the request, cache entry is updated
+        if (requestMaxAge && requestMaxAge !== entry.maxAge)
+          await _cacheOnSuccess(entry, requestMaxAge)
+        rateLimit.incrementParticipantHeartbeat(_getKey(request))
```
RodrigoAD, line 182: If the request hit the cache, we still increment the participant heartbeat. We assume is a valid request, as it exists in the cache

bootstrap/src/lib/cache/index.ts
```diff
@@ -97,19 +109,29 @@ export const withCache: Middleware<CacheOptions> = async (execute, options = def
...
     // Add successful result to cache
-    const _cacheOnSuccess = async ({ statusCode, data, result }: AdapterResponse) => {
+    const _cacheOnSuccess = async (
+      { statusCode, data, result }: AdapterResponse,
+      maxAge?: number,
+    ) => {
       if (statusCode === 200) {
+        if (!maxAge) {
+          // Heartbeat is incremented only if valid request
+          rateLimit.incrementParticipantHeartbeat(_getKey(request), data.cost)
```
RodrigoAD, line 132: We create or increment the participant heartbeat, only if the request is valid

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+  const _getParticipantEvenCapacity = (participantId: string) => {
+    const totalParticipants = participants.size || 1
+    return safeCapacity / totalParticipants / _getParticipant(participantId).cost
+  }
+
+  const _getParticipantMaxAge = (participantId: string): number => {
+    const SEC_IN_MIN = 60
+    const MS_IN_SEC = 1000
+
+    // Precise Capacity
```
RodrigoAD, line 63: Capacity calculation based on participants weight. Commented until further decisions

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+    return {
+      cost: cost || participant.cost,
+      heartbeat: participant.heartbeat + 1,
+      lastSeen: new Date().getTime(),
+    }
+  }
+
+  // hash(API_KEY):uuid()
+  const _getHeartbeatKey = () => {
+    return `${options.groupId}:${options.id}`
```
HenryNguyen5, line 92: Does this need to be a hashed api key?

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  const _getCurrentHeartbeat = (): number => {
+    return Number(totalHeartbeat.get(_getHeartbeatKey())) || 0
+  }
+
+  const _incrementTotalHeartbeat = (): number => {
+    const heartbeat = _getCurrentHeartbeat()
+    totalHeartbeat.set(_getHeartbeatKey(), heartbeat + 1)
+    return heartbeat + 1
```
HenryNguyen5, line 102: Do we need to return anything here? Seems like its used purely for its side-effect

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+      maxAge = MAX_AGE_ALLOWED
+    }
+    return maxAge
+  }
+
+  const _getUpdatedParticipant = (participantId: string, cost?: number): Participant => {
+    const participant = _getParticipant(participantId)
+    return {
+      cost: cost || participant.cost,
+      heartbeat: participant.heartbeat + 1,
```
HenryNguyen5, line 85: Should we only be incrementing heartbeat via `_incrementTotalHeartbeat`?

HenryNguyen5 (changes requested):

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+    // Precise Capacity
+    // const participantHeartbeat = _getParticipant(participantId).heartbeat
+    // const totalHeartbeat = _getCurrentHeartbeat()
+    // const participantWeight = _getWeight(participantHeartbeat, totalHeartbeat)
+    // const allowedReqPerMin = safeCapacity * participantWeight
+
+    const allowedReqPerMin = _getParticipantEvenCapacity(participantId)
+    console.log('PARTICIPANT CAPACITY: ', allowedReqPerMin)
+
+    let maxAge = Math.round(MS_IN_SEC / (allowedReqPerMin / SEC_IN_MIN))
```
HenryNguyen5, line 72: I would leave a quick comment here just to make it easier for future readers to comprehend:
`// the max age we set becomes the average age of a request coming in`
`// e.g. if we have 5 req per min, then the max age will be 250ms, since that means on average we have a request coming in every 250ms`
`// the calculated max age is bounded to the range of [DEFAULT_CACHE_MAX_AGE, MAX_AGE_ALLOWED] `

bootstrap/README.md
```diff
@@ -51,6 +51,18 @@ Cache key is derived by hashing the input object, using the SHA1 hash function,
 
 The `maxAge` input argument can be used to set per item `maxAge` parameter. If not set, or set to `0`, the cache level `maxAge` option will be used. Every time the `maxAge` input argument changes, the item will be cached with the new `maxAge` parameter. To avoid hitting the cache for a specific item set `maxAge: -1` (any value `< 0`).
 
+#### Rate Limit mitigation 
+
+When several adapters use the same data provider with the same API Key, it's possible that Rate Limit issues appear. When this happens, next configuration can be provided:
```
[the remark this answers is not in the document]
RodrigoAD: Right, this comes from the old version, needed to be updated

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (participantHeartbeat === 0) {
+      participantHeartbeat = 1
+    }
+    if (totalHeartbeat === 0) {
+      totalHeartbeat = 1
+    }
+    return (participantHeartbeat - minHeartbeat) / (totalHeartbeat - minHeartbeat)
+  }
+
+  const _getParticipant = (participantId: string): Participant => {
```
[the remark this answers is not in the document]
RodrigoAD: That sounds great, I'll do that

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+    for (const [id, heartbeat] of heartbeats) {
+      if (now - heartbeat.lastSeen > options.groupMaxAge) {
+        heartbeatsCopy.delete(id)
+        logger.debug(`Cache: Participant ${id} has expired. Removing`)
+      }
+    }
+    return heartbeatsCopy
+  }
+
+  const _withExpiration = (fn: any) => (...args: any) => {
```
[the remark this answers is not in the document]
RodrigoAD: You're right, at first I was going to use it in many places. I'll change it

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+    const updatedParticipantHearbeats = _getNonExpiredParticipants(participants)
+    participants = updatedParticipantHearbeats
+    return fn(...args)
+  }
+
+  const _ifEnabled = (fn: (...args: any[]) => any) => (_isEnabled() ? fn : () => false)
```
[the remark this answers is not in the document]
RodrigoAD: That's much better

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+export const makeRateLimit = (options: RateLimitOptions): RateLimit => {
+  const minHeartbeat = 0
+  const safeCapacity = options.totalCapacity * 0.9
+
+  const _isEnabled = (): boolean => {
+    return !!options.groupId && !!options.totalCapacity
+  }
+
+  // Weight normalizes the relevance the participant is having in the adapter
+  const _getWeight = (participantHeartbeat: number, totalHeartbeat: number): number => {
```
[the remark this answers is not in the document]
RodrigoAD: Don't really know what to do with this. It's true that the weight approach is not used, but could be used instead of the even capacity. The difference in the maxAge calculation:
- **Weight**: Each participant will get as much capacity as it needs based on its heartbeat agains the total adapter heartbeat. More heartbeat, more capacity. This has the problem explained in the PR description
- **Even**: Each participant will get the same capacity

bootstrap/src/lib/cache/index.ts
krebernisak: - What if the warmer sends a request? (we should not increment heartbeat)
- What if the entry is in cache? (we should increment heartbeat)

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+  incrementParticipantHeartbeat: (participantId: string, cost?: number) => number | false
+}
+
+type Participant = {
+  cost: number
+  heartbeat: number
+  lastSeen: number
+}
+
+const totalHeartbeat = new Map()
```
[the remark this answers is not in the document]
RodrigoAD: I don't think it will be used, but actually a simple counter should be enough

bootstrap/src/lib/cache/index.ts
```diff
@@ -28,6 +34,12 @@ export const defaultOptions = () => ({
       ...(env.CACHE_KEY_IGNORED_PROPS || '').split(',').filter((k) => k), // no empty keys
     ],
   },
+  rateLimit: {
+    groupMaxAge: parseInt(env.GROUP_MAX_AGE || '') || DEFAULT_GROUP_MAX_AGE,
+    groupId: (env.API_KEY && hash(env.API_KEY)) || '',
+    id: DEFAULT_CACHE_KEY_RATE_LIMIT_PARTICIPANT,
```
krebernisak, line 40: We should not need this anymore, right? There are no RL groups or participants anymore. We have one RL instance per adapter that keeps track of multiple request types.

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+      return currentParticipant
+    }
+
+    const participant = { cost: cost || DEFAULT_COST, heartbeat: 0, lastSeen: Date.now() }
+    participants.set(participantId, participant)
+    return participant
+  }
+
+  const increaseParticipantHeartbeat = (participantId: string) => {
```
krebernisak, line 55: When is the heartbeat decremented?

We should be keeping a sliding window of requests so we can answer questions like:

1. How many requests of type did we observe in the last 24h?
2. How many in the last hour?
3. How many in the last minute?

bootstrap/src/lib/cache/index.ts
```diff
@@ -76,7 +91,7 @@ export const withCache: Middleware<CacheOptions> = async (execute, options = def
   if (!options.enabled) return (data: AdapterRequest) => execute(data)
 
   const cache = await options.cacheBuilder(options.cacheOptions)
-
+  const rateLimit = makeRateLimit(options.rateLimit)
```
krebernisak, line 93: This is now strongly coupled to our cache, it's inside our cache, but I don't really see a reason why?

We shouldn't have made any changes to this file and instead implement the RL as a completely separate component. It should just be a middleware, something more like this:

```javascript
// read part
const reqPerMin  = rateLimit.getTroughput(input, RateLimit.MINUTE)
const maxAge = _computeMaxAge(reqPerMin)
// this is withCache(execute)
const result = await execute({ ...input, maxAge })
// write part
if (!_isWarmer(input)) rateLimit.observe(input)
```

RodrigoAD: Well before made sense, as the cache was storing everything related to Rate Limit Groups. I don't see the point of keeping it there now either.

bootstrap/src/lib/cache/index.ts
RodrigoAD: Yep, removed already

bootstrap/src/lib/cache/rateLimit.ts
```diff
@@ -0,0 +1,140 @@
...
+  totalCapacity: number
+}
+
+export interface RateLimit {
+  isEnabled: () => boolean
+  getParticipantMaxAge: (participantId: string) => number | false
+  incrementParticipantHeartbeat: (participantId: string, cost?: number) => number | false
+}
+
+type Participant = {
```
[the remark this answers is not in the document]
RodrigoAD: Updated. Please take a look to them to see if they make sense

bootstrap/README.md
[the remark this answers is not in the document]
RodrigoAD: Updated

RodrigoAD:
Closing in favor of #373

PR #293, opened 2021-02-03
Remove adapter.ts duplication

ebarakos:
Removed `adapter.ts` from `example`, `cryptoapis` and `amberdata` as PoC.

Utilized endpoint `Name`, by making it a `Names` array. This might be useful on declaring new endpoint names while also keeping backwards naming compatibility.


amberdata/src/endpoint/balance.ts
```diff
@@ -3,7 +3,7 @@ import { Config } from '@chainlink/types'
 import { Requester } from '@chainlink/external-adapter'
 import { isChainType, isCoinType, BLOCKCHAINS } from '.'
 
-export const Name = 'balance'
+export const Names = ['balance']
```
justinkaseman, line 6: I really like this concept for backwards compatibility.

What it also allows for is if we embrace that "external-adapters" are a simple wrapper that map to a data providers API we could have the naming fit the API.

For example `1forge` has an endpoint baseURL/`convert` that we use to get price data. So there we could set the name = ['price', 'convert']

Only tweak I would make is capitalizing it to denote a static property.
```suggestion
export const NAME = ['balance']
```

amberdata/test/balance.test.ts
```diff
@@ -1,7 +1,7 @@
-import { makeExecute } from '../src/adapter'
+import adapter from '../src/index'
```
justinkaseman, line 1: Could destructure this
```suggestion
import { makeExecute }  from '../src/index'
```
Just a style preference thing.

bootstrap/src/adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const execute: ExecuteWithConfig<Config> = async (request, config, endpoints) => {
+  const validator = new Validator(request, inputParams)
+  if (validator.error) throw validator.error
+
+  Requester.logConfig(config)
+
+  const jobRunID = validator.validated.id
+  const endpoint = validator.validated.data.endpoint || config.DEFAULT_ENDPOINT
+  if (endpoints) {
```
justinkaseman, line 16: Guard clauses might make this more readable, lots of if/else blocks.

bootstrap/src/adapter.ts
```diff
@@ -0,0 +1,50 @@
+import { Requester, Validator, AdapterError } from '@chainlink/external-adapter'
+import { ExecuteWithConfig, Config } from '@chainlink/types'
+
+const inputParams = {
+  endpoint: false,
+}
+
+export const execute: ExecuteWithConfig<Config> = async (request, config, endpoints) => {
```
justinkaseman, line 8: Not sure what this should be named, it's more like a function selector. Somehow execute feels weird to me.

cryptoapis/src/endpoint/bc_info.ts
```diff
@@ -1,8 +1,8 @@
 import { Requester, Validator } from '@chainlink/external-adapter'
-import { AdapterRequest, Config } from '@chainlink/types'
+import { ExecuteWithConfig, Config } from '@chainlink/types'
 import { DEFAULT_ENDPOINT } from '../config'
 
-export const Name = 'bc_info'
+export const Names = ['difficulty', 'height']
```
justinkaseman, line 5: Definitely important for backwards compatibility, but I wonder if in time these names should be housed in `ea-legos` since they are the specific data point to fetch.

In `ea-legos` it would map 'bc_info' => ['difficulty', 'height'].

So our Chainlink request would be for the data point `block_height` from `cryptoapis` which would then be transformed to call this API wrapper that knows how to make that HTTP request aka
```
{
  "endpoint": "bc_info",
  "path": "payload.headers"
}
```

I think this builds toward a future where data providers could just upload their whole API via an OpenAPI spec.

Don't think this is a right now thing, just thinking out loud.

typings/@chainlink/index.d.ts
```diff
@@ -79,6 +80,7 @@ declare module '@chainlink/types' {
   export type ExecuteWithConfig<C extends Config> = (
     input: AdapterRequest,
     config: C,
+    endpoints?: Endpoint[]
```
justinkaseman, line 83: I don't think endpoints fits in with what this type is trying to be.

Or that DEFAULT_ENDPOINT should be part of Config.

They are kind of their own thing which is the function selector picking between an API's url endpoints.

justinkaseman:
Really like the direction, thanks for making a push toward this.

Left some thoughts, also curious to hear what @krebernisak thinks.

amberdata/test/balance.test.ts
ebarakos: There are some issues on the `export =` ES6 default exports of adapters' index.ts files, that's why it can't be used as the other imports. But it should be converted to this syntax once we resolve this.

amberdata/src/endpoint/balance.ts
ebarakos: I am happy that you like it. I left the naming as I saw in some places, since we use NAME for the whole adapter package(btw, we can re-use a similar logic for routing between different adapter packages, e.g by using "source", as mentioned at the Composable Adapters document). Reverting it to NAME now until we think more regarding these.

bootstrap/src/adapter.ts
ebarakos: I removed some of the else statements.

bootstrap/src/adapter.ts
ebarakos: Yes, I thought about it as well. Mot sure yet what to do with it, I left it as it is in order to keep same reference across the common adapter interface and endpoint interfaces, until we investigate more.

HenryNguyen5:
What's the status on this PR @ebarakos @justinkaseman?

justinkaseman:
Closing as these changes have become obsolete.

PR #300, opened 2021-02-06
Docs & tweaks from TS migrated EA QA

justinkaseman:
### Description

- README enhancements
- tweaks to names for a consistent `price` endpoint
- additional use of requiring an API key in `getDefaultConfig`

from running the following TS migrated EAs using docker:

- [x] 1forge
- [x] alphavantage
- [x] amberdata-gasprice
- [x] anyblock-gasprice
- [x] binance-dex
- [x] coinapi
- [x] coinbase
- [x] coingecko
- [x] coinmarketcap
- [x] coinpaprika
- [x] cryptocompare
- [x] currencylayer
- [ ] dxfeed
- [x] eodhistoricaldata
- [x] etherchain
- [x] ethgasstation
- [x] fixer
- [x] fmpcloud
- [x] poa-gasprice
- [x] polygon
- [x] nomics
- [x] openexchangerates
- [x] kaiko
- [ ] marketstack
- [x] metalsapi
- [x] nikkei

### Caveats
`dxfeed` hit timeouts when testing
Need to add a `marketstack` API key to secrets

### Story
[176803708](https://www.pivotaltracker.com/story/show/176803708)


ebarakos (changes requested):
Good job!

Just some tiny stuff noticed:

Maybe we could change the "made" word on _"An API key that can be made on"_ references to something like "generated" or "obtained from"?

There are some case letter inconsistencies between `Price Endpoint` and `Price endpoint`. Better choose one style? Seems like the 1st is more appropriate with the general outline.

Maybe we could include Input examples instead of only Output? (Especially if you have the inputs handy, e.g from history or testing docs). What do you think on this?

justinkaseman:
Good suggestions. Consistency is always important. Made those changes.

> Maybe we could include Input examples instead of only Output? (Especially if you have the inputs handy, e.g from history or testing docs). What do you think on this?

Don't have any handy, but I like this idea and added it to the `Example` external adapter.
I'm sure I will be going back through a lot of READMEs for a second pass and will add more sample inputs then.

ebarakos (approved):

PR #306, opened 2021-02-12
Expose config injection - example on dns

RodrigoAD:


dns-query/src/index.ts
```diff
@@ -4,4 +4,10 @@ import { makeConfig } from './config'
 
 const NAME = 'DNS-Query'
 
-export = { NAME, makeConfig, makeExecute, ...expose(util.wrapExecute(makeExecute())) }
+const execute = makeExecute()
+export = {
+  NAME,
+  makeConfig,
+  makeExecute,
+  ...expose(util.wrapExecute(execute.call), '', execute.config),
```
krebernisak, line 12: I was thinking of just send the function here as before, and the middleware reads config from it directly.

```suggestion
  ...expose(util.wrapExecute(execute)),
```

bootstrap/src/index.ts
```diff
@@ -72,14 +79,14 @@ const withAsync = (execute: ExecuteWrappedResponse | ExecuteSync): ExecuteWrappe
...
-const executeSync = (execute: ExecuteWrappedResponse): ExecuteSync => {
+const executeSync = (execute: ExecuteWrappedResponse, config?: Config): ExecuteSync => {
   // TODO: Try to init middleware only once
   // const initMiddleware = withMiddleware(execute)
 
   // Return sync function
   return (data: AdapterRequest, callback: any) => {
     // We init on every call because of cache connection broken state issue
-    return withMiddleware(execute)
+    return withMiddleware(execute, config)
```
krebernisak, line 89: Just send the function to the middlewares. If some needs `execute.config` they can read it, if not they just `execute.call`. 

```suggestion
    return withMiddleware(execute)
```

bootstrap/src/index.ts
```diff
@@ -10,14 +10,21 @@ import {
...
+type Middleware = (
+  execute: ExecuteWrappedResponse,
+  config?: Config,
+) => Promise<ExecuteWrappedResponse>
 
 // Try to initialize, pass through on error
-const skipOnError = (middleware: Middleware) => async (execute: ExecuteWrappedResponse) => {
+const skipOnError = (middleware: Middleware) => async (
+  execute: ExecuteWrappedResponse,
+  config?: Config,
```
krebernisak, line 24: No need to add to every middleware, config is now part of execute fn.

bootstrap/src/index.ts
```diff
@@ -56,40 +67,51 @@ const withLogger = (execute: ExecuteWrappedResponse) => async (data: AdapterRequ
...
   // TODO: Try to init middleware only once
   // const initMiddleware = withMiddleware(execute)
 
   // Return sync function
   return (data: AdapterRequest, callback: any) => {
     // We init on every call because of cache connection broken state issue
     return withMiddleware(execute)
-      .then((executeWithMiddleware) => executeWithMiddleware(data))
+      .then((executeWithMiddleware) => executeWithMiddleware.call(data))
+      .then(wrapResponse)
```
RodrigoAD, line 105: Wrapping the response at the end, instead of doing it in every adapter

PR #309, opened 2021-02-16
Missing params on response

RodrigoAD:
After removing `utils.wrapAdapter`, we just need to pass the response on the callback


krebernisak (dismissed):

justinkaseman (approved):
Yep yep

PR #310, opened 2021-02-17
config part of execute

RodrigoAD:


typings/@chainlink/index.d.ts
```diff
@@ -68,7 +68,11 @@ declare module '@chainlink/types' {
   /* BOOTSTRAP */
   export type ExecuteSync = (input: AdapterRequest, callback: Callback) => void
 
-  export type Execute = (input: AdapterRequest) => Promise<AdapterResponse>
+  export type ExecuteCall = (input: AdapterRequest) => Promise<AdapterResponse>
```
RodrigoAD, line 71: I need this type for the middleware

typings/@chainlink/index.d.ts
```diff
@@ -26,7 +26,7 @@ declare module '@chainlink/types' {
     network?: string
     returnRejectedPromiseOnError?: Boolean
     verbose?: boolean
-    api: Partial<AxiosRequestConfig>
+    api?: Partial<AxiosRequestConfig>
```
RodrigoAD, line 29: Config is super strict, we either leave it flexible with optional api, or extend defaultConfig on every adapter

dns-query/src/adapter.ts
```diff
@@ -41,5 +41,7 @@ const execute: ExecuteWithConfig<Config> = async (input, config) => {
   })
 }
 
-export const makeExecute: ExecuteFactory<Config> = (config?: Config) => (input) =>
-  execute(input, config || makeConfig())
+export const makeExecute: ExecuteFactory<Config> = (config?: Config) => ({
```
RodrigoAD, line 44: This transformation would be needed on every adapter

PR #313, opened 2021-02-17
Tiingo adapter

christianagnew:


tiingo/src/config.ts
```diff
@@ -0,0 +1,17 @@
+import { Requester } from '@chainlink/external-adapter'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'eod'
+export const DEFAULT_BASE_URL = 'https://api.tiingo.com/tiingo/'
+
+export const makeConfig = (prefix?: string): Config => {
+  const config = Requester.getDefaultConfig(prefix)
```
justinkaseman, line 8: Is the API key required for usage?

If so pass
```suggestion
  const config = Requester.getDefaultConfig(prefix, true)
```

tiingo/src/config.ts
```diff
@@ -0,0 +1,17 @@
...
+  config.api = {
+    ...config.api,
+    baseURL: config.api.baseUrl || DEFAULT_BASE_URL,
+    params: {
+      token: config.apiKey,
```
justinkaseman, line 13: Could you move this bit into `eod.ts`?

The API_KEY will be caught in logs otherwise.

tiingo/src/endpoint/eod.ts
```diff
@@ -0,0 +1,30 @@
...
+  const field = validator.validated.data.field || 'close'
+  const url = `daily/${ticker.toLowerCase()}/prices`
+
+  const reqConfig = { ...config.api, url }
+
+  const response = await Requester.request(reqConfig)
+  const result = Requester.validateResultNumber(response.data, [0, field])
+
+  return Requester.success(jobRunID, {
+    data: { result },
```
justinkaseman, line 26:
```suggestion
    data: config.verbose ? { ...response.data, result } : { result },
```

I'll have to add this to the example adapter, but a nice to have in case people want the full response body.

tiingo/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose, util } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'TIINGO'
+
+export = { NAME, makeExecute, makeConfig, ...expose(util.wrapExecute(makeExecute())) }
```
justinkaseman, line 7: A very recent change got rid of the need for `wrapExecute`.

```suggestion
export = { NAME, makeExecute, makeConfig, ...expose(makeExecute()) }
```

tiingo/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Configuration
+
+This adapter supports the following environment variables:
+
+- `API_KEY`: Your Tiingo API key
+
+## Input Params
+
+- `ticker`: The stock ticker to query
```
justinkaseman, line 11: Since this is a required param could we mark it as such?

Trying out a new README template that might be a little more clear:
https://github.com/smartcontractkit/external-adapters-js/blob/develop/example/README.md

justinkaseman (changes requested):
Thanks for the PR! :smile:

A couple small tweaks based on recent practices.

tiingo/test/adapter.test.ts
```diff
@@ -0,0 +1,83 @@
+import { assert } from 'chai'
+import { Requester } from '@chainlink/external-adapter'
+import { assertSuccess, assertError } from '@chainlink/adapter-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../src/adapter'
+
+describe('execute', () => {
+  process.env.API_KEY = process.env.API_KEY ?? 'aa8ea0dd8555e13832a637932449d009f2ca441f'
```
justinkaseman, line 8:
```suggestion
  process.env.API_KEY = process.env.API_KEY ?? 'test_API_key'
```
You'll want to keep API keys out of the code base.

We still need to connect this bit to a secrets store for local integration testing.

tiingo/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@chainlink/tiingo-adapter",
+  "version": "0.0.3",
```
justinkaseman, line 3:
```suggestion
  "version": "0.0.1",
```

justinkaseman (approved):
Thanks for the changes, 2 more quick ones and then LGTM.

tiingo/test/adapter.test.ts
christianagnew: Ahhh that makes more sense, was a bit confused when I thought you meant to include a key

justinkaseman (commented):
Could you rebase with develop? Then good to go!

justinkaseman (approved):
Great! Thanks @christianagnew

PR #316, opened 2021-02-18
Publish Docker images to SDLC Public ECR instead of Prod Public ECR

emedvedev:
This PR switches the Docker push workflow to use the Public ECR registry belonging to the SDLC account.


emedvedev:
@HenryNguyen5 ready for review. Tested it on a fake release, works as intended, uploads the images to Public ECR on SDLC.

.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ name: Release
 
 env:
   cache-name: release-ci
-  publicecr-name: chainlink
+  publicecr-name: z0b1w9r9
```
HenryNguyen5, line 11: Did you want to swap alias over then merge this in?

HenryNguyen5 (changes requested):

.github/workflows/release.yml
HenryNguyen5: Actually it should be fine, the alias is more for external consumers than us anyway

HenryNguyen5 (approved):

.github/workflows/release.yml
emedvedev: I mainly swapped it because we don't have the Secure SDLC alias thing figured out with AWS support yet :) The ID should work regardless of the alias, so it's probably even more future-proof (if slightly more confusing).

PR #320, opened 2021-02-19
added preset ticker for GRT on coinpaprika and coingecko

ebarakos:


coingecko/src/endpoint/price.ts
```diff
@@ -14,10 +14,11 @@ const customParams = {
   coinid: false,
 }
 
-const presetTickers: { [ticket: string]: string } = {
+const presetTickers: { [ticker: string]: string } = {
```
boxhock, line 17: Good catch haha

boxhock (approved):

PR #323, opened 2021-02-22
Data providers batch support

RodrigoAD:
This is the first of the Rate Limit PR's with data reliance. It gives full support to composite adapters to different endpoints, such as batching and market cap. 

Once this gets merged, Token Allocations would go next #325 

New price adapters Readme's:
- [Amberdata](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/amberdata/README.md)
- [Coingecko](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/coingecko/README.md)
- [Coinmarketcap](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/coinmarketcap/README.md)
- [Coinpaprika](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/coinpaprika/README.md)
- [Cryptocompare](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/cryptocompare/README.md)
- [Nomics](https://github.com/smartcontractkit/external-adapters-js/blob/2d5b750c9689702d41a26c75535ea72def652f83/nomics/README.md)


boxhock (changes requested):
A few of my changes requested on #325 also seem to apply here

coingecko/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const response: Record<string, any> = await getPriceData(config, ids, quote, true)
+  const marketCaps = Object.fromEntries(
+    Object.entries(response).map(([coinId, data]) => [
+      idToSymbol[coinId],
+      Requester.validateResultNumber(data, [`${quote.toLowerCase()}_market_cap`]),
+    ]),
+  )
+
+  const result = symbols.length === 1 && marketCaps[symbols[0]]
```
boxhock, line 52: This would return a boolean, no? Or is this some syntax I'm not familiar with :sweat_smile: 

Do we want something like this instead?
```suggestion
  const result = symbols.length === 1 ? marketCaps[symbols[0]] : ''
```

Same applies to the other result keys in different adapters.

RodrigoAD: If the length is 1, then will return the value.`false` in any other case, which I don't like either. 
Basically `result: false` will happen when multiple assets are requested, as there is no a single result. I will change it to the empty string, seems more natural

boxhock (approved):

PR #325, opened 2021-02-23
Token allocation data reliance

RodrigoAD:
Token Allocations will now depend on data from external adapters, instead of data providers.

This is the second of the Rate Limit related PR's with data reliance.


amberdata/src/adapter.ts
```diff
@@ -21,6 +21,9 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     case price.Name: {
```
boxhock, line 21: I can't comment on the line above, but we should convert the endpoint to lowercase before comparing, just so we avoid issues if the endpoint is defined as "marketCap".

coinpaprika/src/endpoint/multi.ts
```diff
@@ -0,0 +1,81 @@
...
+    ...config.api,
+    url,
+    params,
+  }
+
+  const response = await Requester.request(options, customError)
+
+  const payload = getPayload(response.data, symbols, quote, withMarketCap)
+
+  const result = ''
```
boxhock, line 75: This seems not right?

composite/token-allocation/src/config.ts
```diff
@@ -21,24 +15,56 @@ enum DataProvider {
...
-  [DataProvider.Coinpaprika]: coinpaprika,
-  [DataProvider.Nomics]: nomics,
-  [DataProvider.Coinmarketcap]: coinmarketcap,
-  [DataProvider.Coingecko]: coingecko,
-  [DataProvider.Coinapi]: coinapi,
-  [DataProvider.Kaiko]: kaiko,
+const providers: Record<string, DataProviderConfig> = {
+  [DataProvider.Amberdata]: {
+    batchingSupport: false,
+    defaultDataUrl: 'http://localhost:8080',
```
boxhock, line 21: What is this supposed to be the URL to? The adapter?

cryptocompare/src/adapter.ts
```diff
@@ -20,6 +20,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     case price.NAME: {
```
boxhock, line 20: Same comment with updating the endpoint comparison to lowercase.

cryptocompare/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,66 @@
...
+  const options = {
+    ...config.api,
+    url,
+    params,
+  }
+
+  const response = await Requester.request(options, customError)
+  const payload = getPayload(symbols, response.data, tsyms)
+
+  const result = ''
```
boxhock, line 60: This also seems wrong.

cryptocompare/src/endpoint/multi.ts
```diff
@@ -0,0 +1,66 @@
...
+  const options = {
+    ...config.api,
+    url,
+    params,
+  }
+
+  const response = await Requester.request(options, customError)
+  const payload = getPayload(symbols, response.data, tsyms)
+
+  const result = ''
```
boxhock, line 60: Again

boxhock (changes requested):

coinpaprika/src/endpoint/multi.ts
RodrigoAD: I'll just remove it, is actually useless in this case.

composite/token-allocation/src/config.ts
RodrigoAD: Yep. This is yet to be defined, but should get the `url` from the discovery service

composite/token-allocation/src/config.ts
```diff
@@ -21,24 +15,47 @@ enum DataProvider {
...
+    batchingSupport: false,
+  },
 }
 
 export const DEFAULT_TOKEN_DECIMALS = 18
 export const DEFAULT_TOKEN_BALANCE = 1
 
 export const makeConfig = (prefix = '', provider = ''): Config => {
   const dataProvider = provider || util.getRequiredEnv('DATA_PROVIDER', prefix)
+  const dataProviderUrl = util.getRequiredEnv('DATA_PROVIDER_URL', prefix)
```
RodrigoAD, line 52: Until we have the service discovery, adapter url will have to be set through env variable

RodrigoAD:
Ready to review

composite/token-allocation/src/config.ts
```diff
@@ -21,24 +15,47 @@ enum DataProvider {
...
+  [DataProvider.Cryptocompare]: {
+    batchingSupport: true,
+    batchEndpoint: 'multi',
+  },
+  [DataProvider.Coinpaprika]: {
+    batchingSupport: true,
+    batchEndpoint: 'multi',
+  },
+  [DataProvider.Nomics]: {
+    batchingSupport: true,
```
krebernisak, line 31: Why did we focus on batching these requests? Is there a specific reason I am missing?

IMHO this introduces many new problems without really bringing any improvements. Batching is useful, but the way our system is currently designed is posing a problem.

Consider RL & warmed adapter, and an index token we need to price. If we don't batch, the adapter has a lot simpler state to work with. It tracks single data types, so when your index changes it can just start tracking the new data point, and with time close the subscription to the old one. All costs are 1, so by changing the index, we have a temporary dangling cost of M*1 (M tokens charged).

In the case of batching, we track the whole batch request. If your needs change by just 1 token, we need to start tracking the whole new batch and introduce a temporary dangling cost of N1 (size of the first batch).

So this is costing us a lot more in the end. Consider we timeout warmer subscriptions if we don't see a request for data in 2h. It can cost us up to (2/maxAge) * N to keep this dangling subscription.

We also don't reuse cached data between indexN1 & changed indexN2. We also don't reuse cached data between tokens listed in the index and feeds requesting the price of the token, standalone.

Not to mention code coupling and codebase problems this brings. What happens when Kaiko adds batching?

So in the end, batching will introduce a lot of inefficiency and cost, because of our current design, to save some free and fast HTTP round trips. I don't see it's worth it.

RodrigoAD: Yep that makes sense, didn't consider warmer service here...

krebernisak: I missed a case that I thought was nonexistent or very rare, but @boxhock just mentioned that API providers like CryptoCompare & CoinAPI count up to 100 in a batch as 1 RL token.

If that is the case then batching is valuable, but we still have the problems with it mentioned above that we need to figure out how to avoid.

RodrigoAD: Well that's the reason behind all this batching. CMC also count them as 1.

RodrigoAD:
Ready for review

krebernisak (approved):

PR #326, opened 2021-02-23
Composite data reliance

RodrigoAD:
This PR's collects every change related to composite adapters that uses EA data:

- [x] Data Providers adapters support market cap and batching
- [x] Token Allocations request to Data Provider adapters (#325)
- [ ] Cache calculates max age based on other adapters heartbeat (#292 deprecated) (#347 )
- [ ] Defi Pulse works correctly (TA dependency) (#348)
- [ ] APY Finance works correctly (TA dependency) (#348)
- [ ] Synthx works correctly (TA dependency) (#348)
- [ ] CVI works correctly (TA dependency)
- [x] Market Closure request to Data Provider adapters (#344)
- [x] Proof of Reserves request to Data Provider adapters


coingecko/src/endpoint/price.ts
```diff
@@ -21,62 +21,104 @@ const presetTickers: { [ticker: string]: string } = {
...
+      idToSymbol[coinId],
+      Requester.validateResultNumber(data, [quote.toLowerCase()]),
+    ]),
+  )
+
+  const result = symbols.length === 1 ? prices[symbols[0]] : ''
+  const payload = getPayload(symbols, prices, quote)
   return Requester.success(jobRunID, {
-    data: config.verbose ? { ...response.data, result } : { result },
+    data: config.verbose ? { ...response.data, result, payload } : { result, payload },
```
justinkaseman, line 121: I like the concept of having a different key for batched responses.

coingecko/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+    }
+    return [key, val]
+  })
+
+  const payload: ResponsePayload = Object.fromEntries(payloadEntries)
+  return payload
+}
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
```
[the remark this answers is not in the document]
justinkaseman: If we ever need to add say `24h_vol` then it is just a few extra lines.

coingecko/src/endpoint/price.ts
```diff
@@ -21,62 +21,104 @@ const presetTickers: { [ticker: string]: string } = {
...
+        [quote.toUpperCase()]: {
+          price: prices[symbol],
+        },
+      },
+    }
+    return [key, val]
+  })
 
+  const payload: ResponsePayload = Object.fromEntries(payloadEntries)
+  return payload
```
justinkaseman, line 91: I think we should return batch payloads in as similar of a structure as possible to their input params.

So in this case of price if you put in:
```
{
base: ['btc', 'eth', 'uni'],
quote: 'btc'
}
```

It would return values with the same order
```
payload: [413414, 241444, 223]
```
Turning it into a map would happen in the composite adapter.

coinmarketcap/src/endpoint/price.ts
```diff
@@ -28,60 +28,104 @@ const priceParams = {
...
+      ? _validatePrice(Object.values(response.assets)[0])
+      : ''
+
+  const payloadEntries = Object.entries(response.assets).map(([symbol, price]) => {
+    const val = {
+      quote: {
+        [convert]: { price: _validatePrice(price) },
+      },
+    }
+    return [symbol.toUpperCase(), val]
```
justinkaseman, line 123: Same thoughts as coingecko on batch response shape.

coinmarketcap/src/endpoint/marketcap.ts
```diff
@@ -1,37 +1,43 @@
...
+export const NAME = 'marketcap'
 
-const marketcapParams = {
-  market: ['market', 'to', 'quote'],
+const priceParams = {
+  symbol: ['base', 'from', 'coin', 'sym', 'symbol'],
+  quote: ['quote', 'to', 'market', 'convert'],
 }
 
 export const execute: ExecuteWithConfig<Config> = async (request, config) => {
```
justinkaseman, line 12: Same thoughts as other adapters that a lot of code could be reduced by using a `path` in price.

coinpaprika/README.md
```diff
@@ -131,3 +131,107 @@ Returns the global market capitilization from the [global endpoint](https://api.
   "statusCode": 200
 }
 ```
+
+## Multi Endpoint
```
justinkaseman, line 135: I don't love the concept of multi endpoints. I think this is adding complexity to the underlying EAs which we should avoid.

If we need two point of data from the coinpaprika `price` endpoint we should be able to just fetch a raw data payload.

**Or a solution very in-line with changes we are making** is to use a `path` parameter on the `price` endpoint for getting `price` and then `marketcap`. The cache should serve both immediately so there is low latency. One issue here is we build the cache key from the params so it would unnecessarily make two participants. Instead we could just filter out `path` from the params as a quick fix. It would see both requests hitting the `price` endpoint.

cryptocompare/README.md
```diff
@@ -37,3 +37,103 @@
   "statusCode": 200
 }
 ```
+
+## Multi Endpoint
```
justinkaseman, line 41: Same dislike of multi endpoints as coinpaprika

nomics/README.md
```diff
@@ -337,3 +337,44 @@
   "statusCode": 200
 }
 ```
+
+## Marketcap Endpoint
```
justinkaseman, line 341: Same thoughts as other adapters about just sharing the endpoint using a `path` param

amberdata/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,37 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { ExecuteWithConfig, Config } from '@chainlink/types'
+
+export const Name = 'marketcap'
```
justinkaseman, line 4:
```suggestion
export const NAME = 
```
Also we should keep this consistent

justinkaseman (commented):
Left some thoughts.

I think it looks good from a "it works" perspective - but my concerns are the growing complexity and difficulty in maintainability.

amberdata/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const customError = (data: any) => {
+  return Object.keys(data.payload).length === 0
+}
+
+const customParams = {
+  base: ['base', 'from', 'coin'],
```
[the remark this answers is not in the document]
RodrigoAD: It's the endpoint used in token allocations. It accepts symbols as well

coingecko/src/endpoint/marketCap.ts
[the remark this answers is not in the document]
RodrigoAD: Yep, that's good

coingecko/src/endpoint/price.ts
RodrigoAD: I don't agree. Current Payload format is very rich and flexible. Is the one CMC uses and is currently used inside Token Allocations.

coingecko/src/endpoint/price.ts
```diff
@@ -21,62 +21,104 @@ const presetTickers: { [ticker: string]: string } = {
...
+
+  const response: Record<string, any> = await getPriceData(config, ids, quote)
+  const prices = Object.fromEntries(
+    Object.entries(response).map(([coinId, data]) => [
+      idToSymbol[coinId],
+      Requester.validateResultNumber(data, [quote.toLowerCase()]),
+    ]),
+  )
+
+  const result = symbols.length === 1 ? prices[symbols[0]] : ''
```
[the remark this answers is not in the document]
RodrigoAD: Yep that's something we need to agree on

coingecko/src/endpoint/price.ts
RodrigoAD: For batched and for anything that can't be included into a single result

coinpaprika/README.md
RodrigoAD: In the case of cryptocompare, added the `multi` basically to avoid any endpoint changes on the current adapters, as the current `price` endpoint doesn't support batching. 
For the paprika case, I'll review it again, it seems that isn't necessary

amberdata/src/endpoint/marketCap.ts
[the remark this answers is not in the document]
justinkaseman: They need better docs :grin:

nomics/README.md
```diff
@@ -337,3 +337,26 @@
...
```
justinkaseman, line 341: If we go toward calling this `currencies` or `ticker` we could name the Endpoint as such and then just add a line for
```
Using the following values as `endpoint` will fill in the corresponding field parameter:
|  endpoint   |    field     |
| :---------: | :----------: |
|   `price`   |   `price`    |
| `marketcap` | `market_cap` |

```

cryptocompare/README.md
```diff
@@ -37,3 +37,35 @@
...
+
+## Marketcap Endpoint
+
+### Input Params
+
+| Required? |          Name           |               Description                | Options | Defaults to |
+| :-------: | :---------------------: | :--------------------------------------: | :-----: | :---------: |
+|    ✅     | `base`, `from`, `coin`  |   The symbol of the currency to query    |         |             |
+|    ✅     | `quote`, `to`, `market` | The symbol of the currency to convert to |         |             |
+
```
justinkaseman, line 49:
```suggestion

### Sample Input
```

cryptocompare/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  const response = await Requester.request(options, customError)
+  const marketCaps = symbols.map((symbol) => {
+    const key = symbol
+    const result = Requester.validateResultNumber(response.data, [
+      'RAW',
+      symbol.toUpperCase(),
+      tsyms,
+      'MKTCAP',
```
justinkaseman, line 42: This could be `field`. Lots of possible data returned by this endpoint.

cryptocompare/src/endpoint/marketCap.ts
```diff
@@ -0,0 +1,54 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { ExecuteWithConfig, Config } from '@chainlink/types'
+
+export const NAME = 'marketcap'
```
justinkaseman, line 4: Could call this closer to the API endpoint: `pricemultifull`

composite/token-allocation/src/config.ts
```diff
@@ -1,44 +1,19 @@
...
+import { getDataProvider } from './dataProvider'
+import { Config } from './types'
 
 export const DEFAULT_TOKEN_DECIMALS = 18
 export const DEFAULT_TOKEN_BALANCE = 1
 
-export const makeConfig = (prefix = '', provider = ''): Config => {
-  const dataProvider = provider || util.getRequiredEnv('DATA_PROVIDER', prefix)
+export const makeConfig = (prefix = ''): Config => {
+  const dataProviderUrl = util.getRequiredEnv('DATA_PROVIDER_URL', prefix)
```
justinkaseman, line 10: Now that composite adapters use external adapters as services we can make this one `token-allocation` composite adapter handle all 8 DataProviders at once.

A node operator should be able to provide the URL for any of the underlying adapters using a prefix:
```
 AMBERDATA_DATA_PROVIDER_URL
 CRYPTOCOMPARE_DATA_PROVIDER_URL
COINPAPRIKA_DATA_PROVIDER_URL
NOMICS_DATA_PROVIDER_URL
AMBERDATA_DATA_PROVIDER_URL
COINMARKETCAP_DATA_PROVIDER_URL
COINGECKO_DATA_PROVIDER_URL
COINAPI_DATA_PROVIDER_URL
KAIKO_DATA_PROVIDER_URL
```

`token-allocation` `adapter.ts` takes a new `provider` parameter to switch between these URLs based on the desired data provider.

Now Infra would only need to deploy one `token-allocation` adapter and Feeds will provide the `provider` in the Bridge's URL query parameters.

See this comment & PR as an example:
https://github.com/smartcontractkit/external-adapters-js/pull/349#discussion_r588778806

coinpaprika/src/endpoint/price.ts
```diff
@@ -52,12 +58,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     params,
   }
 
+  const resultPaths: { [key: string]: string[] } = {
```
justinkaseman, line 61: See Nomics comments

coinpaprika/src/endpoint/price.ts
```diff
@@ -2,11 +2,16 @@ import { Requester, Validator } from '@chainlink/external-adapter'
 import { ExecuteWithConfig, Config } from '@chainlink/types'
 
 export const NAME = 'price'
```
justinkaseman, line 4: Could we call this endpoint `tickers`?

coinpaprika/src/endpoint/price.ts
```diff
@@ -2,11 +2,16 @@ import { Requester, Validator } from '@chainlink/external-adapter'
...
+export enum Paths {
```
justinkaseman, line 5: See Nomics comments

coinmarketcap/src/endpoint/price.ts
```diff
@@ -76,9 +82,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   }
 
   const key = params.id || _keyForSlug(response.data, params.slug || '') || params.symbol
-  const path = ['data', key, 'quote', convert, 'price']
+  const resultPaths: { [key: string]: string[] } = {
```
justinkaseman, line 85: See Nomics comments

coingecko/src/endpoint/price.ts
```diff
@@ -69,10 +77,15 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     params,
   }
 
+  const param: { [key: string]: string } = {
```
justinkaseman, line 80: See Nomics comments

amberdata/src/endpoint/token.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const execute: ExecuteWithConfig<Config> = async (input, config) => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+
+  const coin = validator.validated.data.base
+  const endpoint: TokenEndpoints = validator.validated.data.endpoint
+  const path = paths[endpoint] || 'marketCapUSD'
```
justinkaseman, line 32: I don't think we should rely on `endpoint` like this at this level. See Nomics comments.

coingecko/src/endpoint/price.ts
```diff
@@ -8,10 +8,16 @@ const customError = (data: any) => {
...
+export enum Paths {
+  Price = 'price',
+  MarketCap = 'marketcap',
+}
+
 const customParams = {
   base: ['base', 'from', 'coin'],
   quote: ['quote', 'to', 'market'],
   coinid: false,
+  path: false,
```
justinkaseman, line 20: See Nomics comments

coingecko/src/endpoint/price.ts
```diff
@@ -56,11 +62,13 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   const coinid = validator.validated.data.coinid
   const url = '/simple/price'
   const market = validator.validated.data.quote
+  const path: string = validator.validated.data.path || Paths.Price
```
justinkaseman, line 65: See Nomics comments

justinkaseman (changes requested):

justinkaseman:
Thanks for making those requested changes. I have some more for you :sweat_smile:

justinkaseman:
Merging :tada: . Additional requested changes will be handled in a separate PR.

PR #343, opened 2021-03-04
CMC should prefer IDs over slugs

boxhock:
As even slugs can change over time, CMC recommends us to use IDs to map the symbols instead of slugs.


RodrigoAD (approved):

justinkaseman (commented):
LGTM.

Eventually we should favor fetching the actual IDs from https://coinmarketcap.com/api/documentation/v1/#operation/getV1CryptocurrencyMap though right?

boxhock:
@justinkaseman Yes, just being careful about rate-limits right now

PR #344, opened 2021-03-04
MarketClosure data reliance

boxhock:


composite/market-closure/src/checks/schedule.ts
```diff
@@ -6,11 +6,11 @@ const customParams = {
   schedule: false,
 }
 
-export const isMarketClosedFactory = (input: AdapterRequest): (() => Promise<boolean>) => {
+export const isMarketClosedFactory = (input: AdapterRequest): Promise<boolean> => {
   const validator = new Validator(input, customParams)
   if (validator.error) throw validator.error
   const schedule = validator.validated.data.schedule || {}
-  return async () => isMarketClosed(schedule as Schedule)
+  return new Promise<boolean>((resolve) => resolve(isMarketClosed(schedule as Schedule)))
```
RodrigoAD, line 13: - You want to return a promise to be consistent with some Types I guess right?
- An easier way to make it return a Promise, could be:
    -  Make the function `async`. That automatically returns a Promise
    - Or, `return Promise.resolve(isMarketClosed(...))`

Also, it seems that you could merge `isMarketClosedFactory` and `isMarketClosed` into one single function `isMarketClosed`.

composite/market-closure/src/checks/tradinghours.ts
```diff
@@ -10,10 +10,10 @@ const commonKeys: Record<string, string> = {
   N225: 'xjpx',
 }
 
-export const isMarketClosedFactory = (input: AdapterRequest): (() => Promise<boolean>) => {
+export const isMarketClosedFactory = (input: AdapterRequest): Promise<boolean> => {
   const validator = new Validator(input, customParams)
   if (validator.error) throw validator.error
-  return async () => await isMarketClosed(validator.validated.data.symbol)
+  return isMarketClosed(validator.validated.data.symbol)
```
RodrigoAD, line 16: I think you can merge this two as well

composite/market-closure/package.json
```diff
@@ -31,8 +31,7 @@
     "typescript": "^3.9.7"
   },
   "dependencies": {
-    "@chainlink/fcsapi-adapter": "^0.0.3",
-    "@chainlink/finnhub-adapter": "^0.0.3",
+    "axios": "^0.21.1",
```
justinkaseman, line 34: What if we used `@external-adapter/Requester` for HTTP calls? Keeps us in the pattern of having a HTTP "module".

Only drawback I could imagine is if we didn't want retries, but since everything _should_ be instant with cache warming it seems like a good thing to have.

composite/market-closure/src/checks/schedule.ts
boxhock: Ah, of course - I think that's what I intended, don't know how I ended up with what I have here hahah

composite/market-closure/package.json
boxhock: So the only reason for pulling in `axios` here was to get the `AxiosRequestConfig` type in dataProvider.ts, as opposed to using the `any` type. We're actually using `Requester` for the HTTP calls.

Do you think we should go for `any` instead of pulling it in as a dependency here?

justinkaseman: Ohh yeah, that's what I get for commenting before doing a full review :sweat_smile: 

No, I think having the type is always preferred. Maybe instead we change line 23 of @chainlink types to something like
```
  export { AxiosRequestConfig } from 'axios'
```
because every composite adapter will likely need to see the type
or eventually create some composite adapter patterns and types.

Relying on @chainlink/external-adapter's axios dependency is probably better.

composite/market-closure/README.md
```diff
@@ -26,8 +25,8 @@ The adapter takes the following environment variables:
 | :-------: | :-------------------------------: | :-------------------------------------------------------------------------------------------: | :-----: | :---------: |
 |    ✅     |  `referenceContract`, `contract`  |                 The Aggregator contract to call for its latest round's price                  |         |             |
 |    ✅     |            `multiply`             | To handle big numbers, the amount to divide the output from reading the reference contract by |         |             |
-|    ✅     | `base`, `asset`, `from`, `symbol` |                                 The target currency to query                                  |         |             |
 |    🟡     |            `schedule`             |                   A schedule of market times to check whether they are open                   |         |             |
+|    ✅     |                                   |                         Any params required by the underlying adapter                         |         |             |
```
justinkaseman, line 29: I think they will also get confused here. Maybe separate it a little bit. [Here's how I structured mine](https://github.com/smartcontractkit/external-adapters-js/blob/473d8e328be93c2f271ffe6fa74608ad748c292d/composite/proof-of-reserves/README.md)

composite/market-closure/README.md
```diff
@@ -16,8 +16,7 @@ The adapter takes the following environment variables:
 | :-------: | :-----------------------------------------: | :-------------------------------------------: | :------------------------: | :---------: |
 |    ✅     |                `CHECK_TYPE`                 |   The provider to check if a market is open   | `schedule`, `tradinghours` |             |
 |    🟡     | `CHECK_API_KEY` (when using `tradinghours`) |  An API key when needed by a check provider   |                            |             |
-|    ✅     |               `PRICE_ADAPTER`               | The provider to use for retrieving price data |    `finnhub`, `fcsapi`     |             |
-|    🟡     |                  `API_KEY`                  |  An API key when needed by a price provider   |                            |             |
+|    ✅     |             `DATA_PROVIDER_URL`             | The adapter URL for the price data provider   | `finnhub`, `fcsapi`        |             |
```
justinkaseman, line 19: Nitpicking, but I find this confusing. I foresee feeds team reading this and setting `DATA_PROVIDER_URL = finnhub` instead of `FINNHUB_DATA_PROVIDER_URL=https://whatever`

composite/market-closure/src/config.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getDataProvider, PriceAdapter } from './dataProvider'
+import { Check, getCheckImpl, getCheckProvider } from './checks'
+
+export type Config = {
+  priceAdapter: PriceAdapter
+  checkAdapter: Check
+}
+
+export const makeConfig = (prefix = ''): Config => {
+  const dataProviderUrl = util.getRequiredEnv('DATA_PROVIDER_URL', prefix)
```
justinkaseman, line 12: So doing it this way feeds would still have to run one market-closure adapter for finnhub and one market-closure adapter for fcsapi right?

What if instead we made this more flexible. It could take in a new param `provider` of either `finnhub` or `fcsapi` that we use as the prefix for the `DATA_PROVIDER_URL`. The `baseURL` is switched depending on what provider you want.

We could do the same with `check`.

I used this pattern in mine, https://github.com/smartcontractkit/external-adapters-js/pull/349/files#diff-3312b3574a4a6a6050bc295ea85bc593cdde939e8abf11a51df6a46151bfb65dR11. See what you think.

This would let feeds just run one single `market-closure` adapter.

composite/market-closure/src/index.ts
```diff
@@ -1,6 +1,8 @@
 import { expose } from '@chainlink/ea-bootstrap'
-import { execute } from './adapter'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
 
 const NAME = 'MarketClosure'
```
justinkaseman, line 5: Could we make this something like `MARKET_CLOSURE`? Or `MARKET-CLOSURE`.. or `MARKETCLOSURE`. Not sure what pattern we want but we should be consistent.

composite/market-closure/package.json
boxhock: Yeah I agree, that would be better

composite/market-closure/src/config.ts
boxhock: We unfortunately can't split this on input params, as the CL node will send the exact same input params to each adapter. Going to leave a comment on your other PR as well.

composite/market-closure/README.md
boxhock: I've made an attempt to make the documentation more clear. Please see if it makes more sense now

RodrigoAD (approved):
It LTGM

PR #345, opened 2021-03-04
Add Amberdata case for DIGG/BTC

boxhock:


amberdata/src/endpoint/price.ts
```diff
@@ -19,12 +25,24 @@ export const execute: ExecuteWithConfig<Config> = async (input, config) => {
 
   const coin = validator.validated.data.base
   const market = validator.validated.data.quote
-  const url = `/api/v2/market/spot/prices/pairs/${coin.toLowerCase()}_${market.toLowerCase()}/latest`
+  const includes = validator.validated.data.includes || []
```
RodrigoAD, line 28: Why an array? Is expected to handle more than one? They will have to include a base anyway right?

boxhock: It's so the format is ready for any possible future uses. We can't use "base", because other adapters will expect a different value here.

This was the format that was agreed on just so we can get this out for now.

RodrigoAD (approved):

PR #347, opened 2021-03-05
Rate limit event driven

RodrigoAD:


bootstrap/src/index.ts
```diff
@@ -35,6 +37,21 @@ const withStatusCode: Middleware = async (execute) => async (input) => {
   return { ...rest, statusCode, data }
 }
 
+export const withRateLimit: Middleware = async (execute) => async (input) => {
+  const totalReqPerMin = rateLimit.getTroughput(rateLimit.Interval.MINUTE)
+  const participantReqPerMin = rateLimit.getTroughput(rateLimit.Interval.MINUTE, input)
+  const cost = rateLimit.getParticipantCost(participantReqPerMin)
+  const maxReqPerMin = rateLimit.getMaxReqAllowed(
+    totalReqPerMin.length + 1,
+    participantReqPerMin.length + 1,
```
RodrigoAD, line 46: If the request succeed, we could assume that the request will be 1 more than the current one.

bootstrap/src/lib/cache-warmer/reducer.ts
```diff
@@ -92,4 +92,3 @@ export const rootReducer = combineReducers({
   subscriptions: subscriptionsReducer,
   warmups: warmupReducer,
 })
-export type RootState = ReturnType<typeof rootReducer>
```
krebernisak, line 96: We need this, the cache warmer reducer needs to export its state type.

bootstrap/src/lib/cache-warmer/side-effects.ts
```diff
@@ -21,7 +21,7 @@ import {
   warmupUnsubscribed,
 } from './actions'
 import { WARMUP_REQUEST_ID, Config, get } from './config'
-import { RootState } from './reducer'
+import { RootState } from '../../index'
```
krebernisak, line 24: Please revert changes to this file.

We need to be aware of what depends on what. This warmer component needs to be contained, it should not depend on the application, but the application depends on the warmer component.

RodrigoAD: Yep, I know, but this was breaking the cache warmer. We need to tell somehow to `createEpicMiddleware` that Root State now has rate-limit as well, as both stores and reducers are combined.

bootstrap/src/lib/cache/index.ts
```diff
@@ -150,8 +150,6 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
 
     if (entry) {
       if (maxAge >= 0) {
-        logger.debug(`Cache: GET ${key}`, entry)
```
krebernisak, line 155: Why did you remove this? I think we'd need this feature to update the already cached item with the new TTL.

If you don't want for warmer to recache the item, then don't set a `maxAge` to the input and that will achieve the same thing.

RodrigoAD: This is keeping the cache values forever. As maxAge from the input is constantly changing, it updates the cache entry with the same values, but with extended maxAge

RodrigoAD: I've included the max age coming from the RateLimit middleware into `data: { rateLimitMaxAge: ... }`. If the request contains `maxAge` would override the default value

bootstrap/src/lib/metrics/index.ts
```diff
@@ -0,0 +1,39 @@
...
+const collectDefaultMetrics = client.collectDefaultMetrics
+const Registry = client.Registry
+export const register = new Registry()
+
+collectDefaultMetrics({ register })
+
+export const httpRequestsTotal = new client.Counter({
+  name: 'http_requests_total',
+  help: 'The number of http requests this external adapter has serviced for its entire uptime',
+  labelNames: ['method', 'statusCode', 'apiKey', 'retry'] as const,
```
HenryNguyen5, line 12: I would merge `cacheWarmerRequests`, `httpRequestsCacheHits`, `httpRequestsDataProviderHits` into this metric as labels, like the following:
```ts
labelNames: ['method', 'statusCode', 'apiKey', 'retry', 'isCacheHit', 'isWarmerRequest', 'isDataProviderHit'] as const,
```

HenryNguyen5: This way, you can set those additional labels as boolean values

HenryNguyen5 (changes requested):

bootstrap/src/lib/metrics/index.ts
HenryNguyen5: Now that i think about it, we dont even need `isDataProviderHit` since we can infer that from `isCacheHit`

RodrigoAD: Perfect, didn't know that

RodrigoAD:
Base branch got so many changes that this PR became unmaintainable. 
- Most of the functionality from this PR got merged here: #366 
- Closing in favor of #373

PR #349, opened 2021-03-05
Composite PoR uses underlying data adapters

justinkaseman:
### Description
Refactors Proof of Reserves composite adapter to use underlying data providers
Starts off an example composite adapter


composite/example/README.md
```diff
@@ -0,0 +1 @@
+# Chainlink example composite adapter
```
justinkaseman, line 1: Just a placeholder example. Still needs to be filled in.

external-adapter/src/validator.ts
```diff
@@ -6,13 +6,15 @@ import { AdapterErrorResponse } from '@chainlink/types'
 export class Validator {
   input: any
   customParams: any
+  options: Record<string, any[]>
   validated: any
   error: AdapterError | undefined
   errored: AdapterErrorResponse | undefined
 
-  constructor(input = {}, customParams = {}) {
+  constructor(input = {}, customParams = {}, options = {}) {
```
justinkaseman, line 14: Adding an additional parameter to our Validator that validates options.

So if I had the input
```
{
id: 1
data: {
  coin: "BTC"
}
}
```
and want to make sure that `coin` = `BTC` or `ETH` I could pass
```
{
 coin: ["BTC", "ETH"]
}
```

composite/proof-of-reserves/src/balance.ts
```diff
@@ -9,9 +9,7 @@ import btcCom from '@chainlink/btc.com-adapter'
 import cryptoapis from '@chainlink/cryptoapis-adapter'
 import sochain from '@chainlink/sochain-adapter'
 
-const ENV_BTC_INDEXER_ADAPTER = 'BTC_INDEXER_ADAPTER'
-
-const adapters: AdapterImplementation[] = [
+export const adapters: AdapterImplementation[] = [
```
justinkaseman, line 12: I kept the dependencies on the underlying adapters purely to use them as a source of truth for `.NAME`, though I could see removing this. Thoughts?

composite/proof-of-reserves/src/adapter.ts
```diff
@@ -1,56 +1,42 @@
...
   return output
 }
 
-// Get address set for protocol
-const runProtocolAdapter = async (input: AdapterRequest) => {
-  const execute = getProtocolImpl({ type: getProtocol() })
-  return runAdapter(execute, input, '_onProtocol')
+const inputParams = {
+  protocol: true,
+  indexer: true,
```
justinkaseman, line 22: Keeping this composite adapter very flexible. There are 2 protocol adapters and 7 balance adapters. Instead of running 14 different PoR adapters we can just run 1.

The two main concepts are: 
- input params to switch between underlying adapters
- `DATA_PROVIDER_URL` uses the param prefix to find the URL to the location

boxhock: It will not be possible to split the adapters based on input params, as the Chainlink node will have to send the exact same params to all adapters. Last time this issue came up I proposed we split this on query string params on the bridge instead, which would make it possible to deploy a single adapter, and split the protocol/balance adapters based on the QS params you set when creating the bridge. Maybe this is something we should start prioritizing instead?

boxhock (changes requested):

composite/proof-of-reserves/src/adapter.ts
boxhock: After our talk this looks good. Going to copy this to the other composite adapters as well 👍

composite/proof-of-reserves/src/adapter.ts
```diff
@@ -1,56 +1,42 @@
...
+import { runBalanceAdapter } from './balance'
+import { runReduceAdapter } from './reduce'
+
+export const makeRequestFactory = (config: Config, prefix: string) => async (
+  input: AdapterRequest,
+) => Requester.request({ ...config.api, method: 'post', url: getURL(prefix, true), data: input })
 
 // Run, log, throw on error
-const runAdapter = async (execute: Execute, input: AdapterRequest, tag: string) => {
+export const callAdapter = async (execute: any, input: AdapterRequest, tag: string) => {
```
boxhock, line 14: I might be missing it, but why are we changing the type of `execute` from Execute to any?

justinkaseman: Shouldn't be! Changed.

composite/proof-of-reserves/src/balance.ts
```diff
@@ -21,27 +19,23 @@ const adapters: AdapterImplementation[] = [
...
+  input: AdapterResponse,
+) => {
+  const execute = makeRequestFactory(config, indexer)
+  const next = {
+    id: input.jobRunID,
+    data: {
+      result: input.data.result,
+      dataPath: 'result',
+      endpoint: 'balance',
+      confirmations: 6,
```
boxhock, line 37: Can we make this configurable, just if we need it at some point?

boxhock (approved):

PR #355, opened 2021-03-09
Dnsproof adapter

christianagnew:


justinkaseman:
Hey @christianagnew!

We already have some adapters that handle DNS record checking.

Take a look at [composite/dns-record-check](https://github.com/smartcontractkit/external-adapters-js/tree/develop/composite/dns-record-check) and its underlying adapter [dns-query](https://github.com/smartcontractkit/external-adapters-js/blob/develop/dns-query/README.md) to see if they will fit your needs.

PR #360, opened 2021-03-12
Clean Redis Shutdown

saschagoebel:
This has not been tested, but it's mostly copypasta from one of my projects where I use Redis for caching as well. Fixes these issues:

- client.end should not be used in production and will sometimes lose data

- Close the Redis connection even when the request failed. Otherwise the open connections will stack until the adapter is restarted.


krebernisak:
Thanks, @saschagoebel!

As there were some CI errors, and to speed up this process I'm closing this PR in favor of [Improve Redis connection cleanup #361](https://github.com/smartcontractkit/external-adapters-js/pull/361/files).

PR #363, opened 2021-03-13
Feature/outlier detection refactor

justinkaseman:
### Description
- Reworks composite/`outlier-detection-refactor` to use underlying data provider external adapters
- Lots of TypeScript tweaks


packages/composites/outlier-detection/README.md
```diff
@@ -9,74 +9,134 @@ If all checks passes, the adapter returns the median from the source data provid
...
-- `RPC_URL`: ETH RPC URL to read the reference data value. Required by runlog requests.
+1. A data source adapter
 
-This adapter also takes any configuration variables of the underlying adapters. Simply prefix the environment variables.
-E.g. dxFeed endpoint URL:
+   | Required? |                  Name                  |                      Description                      | Options | Defaults to |
+   | :-------: | :------------------------------------: | :---------------------------------------------------: | :-----: | :---------: |
+   |           |        `XBTO_DATA_PROVIDER_URL`        |        The location of a XBTO external adapter        |         |             |
+   |           | `GENESIS_VOLATILITY_DATA_PROVIDER_URL` | The location of a Genesis Volatility external adapter |         |             |
+   |           |       `DXFEED_DATA_PROVIDER_URL`       |       The location of a DXFeed external adapter       |         |             |
```
boxhock, line 18: I think this comment applies to the whole PR and not just this, but I think we should remove any specific references to underlying adapters (except examples in the Readme).

This adapter is essentially just checking one group of adapters against a different group of adapters and/or the on-chain value, and we should be able to extend that to use any underlying adapters.

packages/composites/outlier-detection/src/source.ts
```diff
@@ -1,39 +1,9 @@
...
+import GenesisVolatility from '@chainlink/genesis-volatility-adapter'
+import XBTO from '@chainlink/xbto-adapter'
+import dxFeed from '@chainlink/dxfeed-adapter'
 
-export enum SourceDataProvider {
-  XBTO = 'xbto',
-  GenesisVolatility = 'genesisvolatility',
-  dxFeed = 'dxfeed',
-}
+export const adapters: AdapterImplementation[] = [GenesisVolatility, XBTO, dxFeed]
```
boxhock, line 7: Part of my comment above. We should just remove any references of a specific set of adapters and instead allow any adapters to be used.

boxhock (changes requested):
Good job! There are a lot of changes in this PR 😅  But I think I found some initial stuff I think we should change

packages/composites/outlier-detection/README.md
justinkaseman: Interesting idea, yeah that makes a lot of sense.

So checking the on-chain value would become more optional then right?

boxhock: Yes – checking the on-chain value should be optional. The reason for fetching it (either from meta data or from RPC) is that the adapter should return the on-chain value if outlier detection is triggered.

PR #365, opened 2021-03-16
Add basic prom metrics

HenryNguyen5:


bootstrap/src/lib/metrics/index.ts
```diff
@@ -0,0 +1,52 @@
+import * as client from 'prom-client'
+
+client.collectDefaultMetrics()
+
+export const METRICS_ENABLED = !!process.env.METRICS
```
RodrigoAD, line 5: From the Readme, should this be `process.env.METRICS_ENABLED`?

bootstrap/src/lib/server.ts
```diff
@@ -49,8 +54,18 @@ export const initHandler = (
...
     process.exit()
   })
 }
+
+function setupMetricsServer() {
+  const metricsApp = express()
+  const metricsPort = process.env.METRICS_PORT || 9080
+
+  metricsApp.get('/metrics', (_, res) => {
+    res.send(client.register.metrics())
```
RodrigoAD, line 67: I think this returns a Promise. `res.send(await client.register.metrics())` should do it

bootstrap/src/index.ts
```diff
@@ -48,7 +49,40 @@ const withLogger: Middleware = async (execute) => async (input: AdapterRequest)
...
+      metrics.httpRequestsTotal.labels(labels).inc()
+    }
+  }
+
+  const record = recordMetrics()
+  try {
+    const result = await execute(input)
+    record(
+      result.statusCode,
+      result.data.maxAge
```
RodrigoAD, line 72: Thought `maxAge` was included in data, but is included at the top level, at `result.maxAge`

bootstrap/src/lib/server.ts
HenryNguyen5: Nice catch!

bootstrap/src/lib/metrics/index.ts
HenryNguyen5: Yep

bootstrap/src/index.ts
HenryNguyen5: Ill check both, just incase

RodrigoAD (dismissed):
Nice 😀

bootstrap/src/lib/metrics/index.ts
```diff
@@ -0,0 +1,57 @@
+import * as client from 'prom-client'
+
+client.collectDefaultMetrics()
+client.register.setDefaultLabels(
+  // we'll inject both name and versions in
+  // when EAEE gets merged, because it'll be a lot easier
+  // to refactor with full type coverage support
+  { app_name: process.env.METRICS_NAME || 'N/A', app_version: 'N/A' },
+)
+export const METRICS_ENABLED = !!process.env.EXPERIMENTAL_METRICS_ENABLED
```
krebernisak, line 10: There is a `utils/parseBool` that you can consider using here. 
Just so someone would not set an explicit `EXPERIMENTAL_METRICS_ENABLED=false`.

krebernisak (dismissed):
@HenryNguyen5 our prophet making us see the bigger picture 🙏 😄

bootstrap/src/index.ts
```diff
@@ -48,7 +49,40 @@ const withLogger: Middleware = async (execute) => async (input: AdapterRequest)
...
+    )
+    return result
+  } catch (error) {
+    record()
+    throw error
+  }
+}
+
+const middleware = [withLogger, skipOnError(withCache), withStatusCode].concat(
+  metrics.METRICS_ENABLED ? [withMetrics] : [],
```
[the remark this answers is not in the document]
HenryNguyen5: I moved it in the middleware, but it ended up looking more complex since we have to `await execute` and return the result still

bootstrap/src/lib/metrics/index.ts
HenryNguyen5: Sounds good!

se3000 (approved):

PR #374, opened 2021-03-19
Cache and Rate limit metrics

RodrigoAD:
## Cache and RL Metrics

- 3 new metrics added:
  - `rate_limit_credits_spent_total`: Counter that gets track of the number of requests the adapter is making to the provider
  - `cache_execution_duration_seconds`: Histogram that observes the duration of the cache module execution.
  - `cache_data_staleness`: Histogram that observes how old is the data returned
- Added `docker-compose.yml` to setup prometheus and grafana for a dev environment
- Token allocations sends the jobRunId into the adapter. Needed for metrics

### DIFF
--- a/packages/composites/token-allocation/src/dataProvider.ts
+++ b/packages/composites/token-allocation/src/dataProvider.ts
@@ -24,8 +24,7 @@ export const adapters: AdapterImplementation[] = [
 
 export type Source = typeof adapters[number]['NAME']
 
-const getPrices = (apiConfig: any) => async (
-  jobRunID: string,
+const getPrices = (apiConfig: any) => (jobRunID: string) => async (
   symbols: string[],
   quote: string,
   withMarketCap = false,
--- a/packages/composites/token-allocation/src/dataProvider.ts
+++ b/packages/composites/token-allocation/src/dataProvider.ts
@@ -1,28 +1,6 @@
 import { Requester } from '@chainlink/ea-bootstrap'
-import { PriceAdapter, ResponsePayload } from './types'
-import { AdapterImplementation, RequestConfig } from '@chainlink/types'
-// data provider external adapters
-import Amberdata from '@chainlink/amberdata-adapter'
-import CoinApi from '@chainlink/coinapi-adapter'
-import CoinGecko from '@chainlink/coingecko-adapter'
-import CoinMarketCap from '@chainlink/coinmarketcap-adapter'
-import CoinPaprika from '@chainlink/coinpaprika-adapter'
-import CryptoCompare from '@chainlink/cryptocompare-adapter'
-import Kaiko from '@chainlink/kaiko-adapter'
-import Nomics from '@chainlink/nomics-adapter'
-
-export const adapters: AdapterImplementation[] = [
-  Amberdata,
-  CoinApi,
-  CoinGecko,
-  CoinMarketCap,
-  CoinPaprika,
-  CryptoCompare,
-  Kaiko,
-  Nomics,
-]
-
-export type Source = typeof adapters[number]['NAME']
+import { RequestConfig } from '@chainlink/types'
+import { ResponsePayload } from './types'
 
 export const getPriceProvider = (apiConfig: RequestConfig, jobRunID: string) => async (
   symbols: string[],
--- a/packages/composites/token-allocation/src/dataProvider.ts
+++ b/packages/composites/token-allocation/src/dataProvider.ts
@@ -32,7 +32,10 @@ const getPrices = (apiConfig: any) => async (
 ): Promise<ResponsePayload> => {
   const results = await Promise.all(
     symbols.map(async (base) => {
-      const data = { id: jobRunID, data: { base, quote, endpoint: withMarketCap ? 'marketcap' : 'price' } }
+      const data = {
+        id: jobRunID,
+        data: { base, quote, endpoint: withMarketCap ? 'marketcap' : 'price' },
+      }
       const response = await Requester.request({ ...apiConfig, data: data })
       return response.data.result
     }),
--- a/packages/composites/token-allocation/src/dataProvider.ts
+++ b/packages/composites/token-allocation/src/dataProvider.ts
@@ -25,13 +25,14 @@ export const adapters: AdapterImplementation[] = [
 export type Source = typeof adapters[number]['NAME']
 
 const getPrices = (apiConfig: any) => async (
+  jobRunID: string,
   symbols: string[],
   quote: string,
   withMarketCap = false,
 ): Promise<ResponsePayload> => {
   const results = await Promise.all(
     symbols.map(async (base) => {
-      const data = { data: { base, quote, endpoint: withMarketCap ? 'marketcap' : 'price' } }
+      const data = { id: jobRunID, data: { base, quote, endpoint: withMarketCap ? 'marketcap' : 'price' } }
       const response = await Requester.request({ ...apiConfig, data: data })
       return response.data.result
     }),
--- a/packages/composites/token-allocation/src/dataProvider.ts
+++ b/packages/composites/token-allocation/src/dataProvider.ts
@@ -1,6 +1,6 @@
 import { Requester } from '@chainlink/ea-bootstrap'
 import { PriceAdapter, ResponsePayload } from './types'
-import { AdapterImplementation } from '@chainlink/types'
+import { AdapterImplementation, RequestConfig } from '@chainlink/types'
 // data provider external adapters
 import Amberdata from '@chainlink/amberdata-adapter'
 import CoinApi from '@chainlink/coinapi-adapter'
@@ -24,7 +24,7 @@ export const adapters: AdapterImplementation[] = [
 
 export type Source = typeof adapters[number]['NAME']
 
-const getPrices = (apiConfig: any) => (jobRunID: string) => async (
+export const getPriceProvider = (apiConfig: RequestConfig, jobRunID: string) => async (
   symbols: string[],
   quote: string,
   withMarketCap = false,
@@ -50,10 +50,4 @@ const getPrices = (apiConfig: any) => (jobRunID: string) => async (
   })
 
   return Object.fromEntries(payloadEntries)
-}
-
-export const getDataProvider = (apiConfig: any): PriceAdapter => {
-  return {
-    getPrices: getPrices(apiConfig),
-  }
 }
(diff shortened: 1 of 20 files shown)


packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -185,14 +199,19 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
...
+    }
+    return { ...result, debug }
   }
 
   // Middleware wrapped execute fn which cleans up after
   return async (input) => {
     const result = await _executeWithCache(input)
     // Clean the connection
-    await cache.close()
+    // await cache.close()
```
RodrigoAD, line 214: Should we get rid of this @krebernisak?

krebernisak: Let's sync offline on that! I've noticed this comment but thought you were experimenting with something.

We need to close the connection after ourselves, or we are just going to open new connections indefinitely and slow down the Redis node or make it crash. So now I'm interested in how this line affects the warmer.

packages/composites/token-allocation/src/dataProvider.ts
```diff
@@ -2,13 +2,17 @@ import { Requester } from '@chainlink/external-adapter'
...
+  jobRunID: string,
   symbols: string[],
   quote: string,
   withMarketCap = false,
 ): Promise<ResponsePayload> => {
   const results = await Promise.all(
     symbols.map(async (base) => {
-      const data = { data: { base, quote, endpoint: withMarketCap ? 'marketcap' : 'price' } }
+      const data = {
+        id: jobRunID,
```
krebernisak, line 13: Why do we need this `jobRunID` here?

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -175,7 +177,19 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
         logger.debug(`Cache: GET ${key}`, entry)
         const reqMaxAge = _getRequestMaxAge(data)
         if (reqMaxAge && reqMaxAge !== entry.maxAge) await _cacheOnSuccess(entry)
-        return { jobRunID: data.id, ...entry }
+        const ttl = await cache.pttl(key)
```
krebernisak, line 180: This looks like a typo here. It should be `cache.ttl`.

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -175,7 +177,19 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
...
+        const debug = {
+          cacheHit: true,
+          staleness,
+          performance: endMetrics(true, staleness),
+          providerCost: 0,
+        }
+        return {
+          jobRunID: data.id,
+          ...entry,
+          debug,
```
krebernisak, line 191: We should make sure we don't override the debug info other parts of the system added.

packages/core/bootstrap/src/lib/cache/metrics.ts
```diff
@@ -0,0 +1,39 @@
+import * as client from 'prom-client'
+import { MAXIMUM_MAX_AGE } from './index'
+
+enum CacheTypes {
+  Redis = 'redis',
+  Local = 'local',
+}
+
+export const cacheExecutionDurationSeconds = new client.Histogram({
```
krebernisak, line 9: Consider changing this form seconds to milliseconds as it could be a better unit to measure duration here.

Also, let's make a style change and explicitly lowercase the metrics instances, so it's recognisable they are metrics(different than other vars) and we call them the same way we name them (e.g.,  `cache_execution_duration_seconds`).

packages/core/bootstrap/src/lib/rate-limit/index.ts
```diff
@@ -50,11 +50,11 @@ const logRemainingCapacity = (state: Heartbeats, interval: IntervalNames): void
   const capacity = config.get().totalCapacity / cost
   const remainingCapacity = capacity - dataProviderRequests.length
   if (remainingCapacity <= 0) {
-    logger.error('Rate Limit: Data Provider tokens not available')
+    logger.debug('Rate Limit: Data Provider tokens not available')
```
krebernisak, line 53: Maybe `warn` is better than `debug` here, if you don't consider it an error.

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -175,6 +175,14 @@ export const getHashOpts = (): Required<Parameters<typeof objectHash>>['1'] => (
       .includes(props),
 })
 
+export const isDebug = (): boolean => {
```
krebernisak, line 178: I think we should split those two concepts, `DEBUG=true` vs `LOG_LEVEL=debug`.

We could set a production service to `LOG_LEVEL=debug` to get some more info, but we probably don't want to enable all debug options in that case. It could be beneficial to have those two as two separate concepts.

packages/core/bootstrap/src/lib/cache/index.ts
RodrigoAD: To get the ttl is milliseconds, `redis` uses the `pttl` [command](https://redis.io/commands/PTTL).

krebernisak: Ok, not sure why they have two commands to get the same info in different resolutions heh

Maybe we call it `ttl`, but internally in Redis cache call `pttl` and for in memory cache we make sure we return the `ttl` in ms.

packages/composites/token-allocation/src/dataProvider.ts
RodrigoAD: `jobRunID` is not sent through automatically now. It will also be useful to get track of it on relying adapters to filter by job

packages/composites/token-allocation/src/dataProvider.ts
```diff
@@ -1,14 +1,17 @@
 import { Requester } from '@chainlink/external-adapter'
 import { PriceAdapter, ResponsePayload } from './types'
 
-const getPrices = (apiConfig: any) => async (
+const getPrices = (apiConfig: any) => (jobRunID: string) => async (
```
krebernisak, line 4: Heh, there are three layers to this function now...
It's getting a little complex, so try to figure out a way to simplify here.

packages/core/bootstrap/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: "3.7"
```
krebernisak, line 1: Please add some info about this to README, and possibly a package.json script shortcut.

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -175,7 +177,19 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
...
```
krebernisak, line 191: If any other middleware adds debug info to `entry`, they will now be overwritten.

Consider doing something like:

 ```javascript
return {
  jobRunID: data.id,
  ...entry,
  debug: {
    ...(entry.debug || {}),
    ...debug
  }
}

krebernisak: Like you did below for result, the same should be done for this flow.

packages/core/bootstrap/src/lib/cache/metrics.ts
```diff
@@ -0,0 +1,43 @@
...
+export const observeMetrics = (id: string, participantId: string) => {
+  const cacheType = process.env.CACHE_TYPE === 'redis' ? CacheTypes.Redis : CacheTypes.Local
+  const defaultLabels = {
+    job_run_id: id,
+    participant_id: participantId,
+    experimental: 'true',
+    cache_type: cacheType,
+  }
+
+  const start = process.hrtime()
```
krebernisak, line 33: Check out `prom-client` helper functions like `histogram.startTimer()`, as an option that we can use instead of taking manual readings.

packages/core/bootstrap/src/lib/rate-limit/index.ts
```diff
@@ -95,12 +97,7 @@ export const withRateLimit = (store: Store<RootState>): Middleware => async (exe
   state = store.getState()
   logRemainingCapacity(state.heartbeats, IntervalNames.MINUTE)
 
-  const isCacheHit = !!result.maxAge
-  if (!isCacheHit) {
-    metrics.rateLimitCreditsSpentTotal
-      .labels({ id: input.id, participantId: requestTypeId, experimental: 'true' })
-      .inc(result.data.cost || 1)
-  }
+  metrics.observeMetrics(input.id, requestTypeId, result)
```
krebernisak, line 100: I much rather prefer the style from the left, than using a helper function like this, as it gives a lot more information about what is actually happening here. Makes the code more readable and understandable. Just my 2c.

krebernisak (approved):
Looking good! I've added a few comments for your consideration :)

packages/core/bootstrap/src/lib/cache/index.ts
RodrigoAD: `entry` is the cache data, it should never have `debug` in it, no?

packages/core/bootstrap/src/lib/cache/metrics.ts
RodrigoAD: There's no helper to get the measure in `ms`... `prom-client` uses `process.hrtime` the same way. Maybe we could send this into a metrics helper?

krebernisak: This article makes sense: https://www.robustperception.io/who-wants-seconds

> So when choosing units for instrumentation, use base units like seconds and bytes.

So let's just use seconds as a standard for everything time-related.

packages/core/bootstrap/src/lib/rate-limit/index.ts
```diff
@@ -95,12 +97,16 @@ export const withRateLimit = (store: Store<RootState>): Middleware => async (exe
...
-      .inc(result.data.cost || 1)
+  const defaultLabels = {
+    job_run_id: input.id,
+    participant_id: requestTypeId,
+    experimental: 'true',
   }
+  let cost = Number(result.debug?.providerCost)
+  if (isNaN(cost)) {
+    cost = 1
+  }
```
krebernisak, line 108:
```suggestion
  const cost = Number(result.debug?.providerCost) || 1
```

krebernisak (approved):

PR #348, opened 2021-03-05
updated composite readmes

RodrigoAD:
Updated Composite Readme specifying the `DATA_PROVIDER_URL` required env

PR #375, opened 2021-03-19
Token Allocation source parameter

justinkaseman:
### Description
1. Refactors composite Token Allocation to use prefixed `[name]_DATA_PROVIDER_URL` environment variables and a source parameter to switch between underlying adapters

2. Cherry picks README updates from #348 

3. Fixes issue with a circular type import that would throw the following error when compiling
```
Cannot write file ' ... ' because it would overwrite input file.
```
4. Deletes some files that were left over fragments from the rebase with develop


packages/composites/token-allocation/src/config.ts
```diff
@@ -1,20 +1,42 @@
...
+    : util.getEnv(ENV_DATA_PROVIDER_URL, prefix)
+
 export const makeConfig = (prefix = ''): Config => {
-  const dataProviderUrl = util.getRequiredEnv('DATA_PROVIDER_URL', prefix)
-  const defaultConfig = getDefaultConfig(prefix)
-  defaultConfig.api.baseURL = defaultConfig.api.baseURL || dataProviderUrl
-  defaultConfig.api.method = 'post'
+  const sources: SourceRequestOptions = {}
+
+  for (const a of adapters) {
```
justinkaseman, line 20: Building possible requests based on what adapters are "discovered"

packages/composites/token-allocation/src/dataProvider.ts
```diff
@@ -1,5 +1,28 @@
 import { Requester } from '@chainlink/external-adapter'
 import { PriceAdapter, ResponsePayload } from './types'
+import { AdapterImplementation } from '@chainlink/types'
+// data provider external adapters
+import Amberdata from '@chainlink/amberdata-adapter'
```
justinkaseman, line 5: Importing the underlying adapter to use the `NAME` as a source of truth

packages/core/types/@chainlink/index.d.ts
```diff
@@ -58,8 +58,9 @@ declare module '@chainlink/types' {
   export type Callback = (statusCode: number, data?: any) => void
   export type AdapterHealthCheck = (callback: Callback) => any
 
-  import type { RequestConfig } from '@chainlink/external-adapter'
-  export type RequestConfig = RequestConfig
+  // TODO: fix circular dependency
```
justinkaseman, line 61: Here was our problem with that error on `yarn setup`. I think our solution is probably to combine `bootstrap` and `external-adapter`

packages/sources/coinmarketcap/package.json
```diff
@@ -33,5 +34,8 @@
     "start": "yarn server:dist"
   },
   "devDependencies": {},
-  "dependencies": {}
+  "dependencies": {
```
justinkaseman, line 37: I don't think we _need_ these, but might be nice for documenting dependencies?

@HenryNguyen5 What do you think here? How should we best handle intra-mono-repo dependencies?

packages/composites/token-allocation/src/adapter.ts
```diff
@@ -68,35 +68,39 @@ const toValidAllocations = (allocations: any[]): TokenAllocations => {
...
+  source: true,
   allocations: true,
   quote: false,
   method: false,
 }
 
 export const execute = async (input: AdapterRequest, config: Config): Promise<AdapterResponse> => {
-  const validator = new Validator(input, inputParams)
+  const paramOptions = makeOptions(config)
+  const validator = new Validator(input, inputParams, paramOptions)
```
justinkaseman, line 96: Using the new Validator options

packages/sources/coinmarketcap/package.json
HenryNguyen5: This is the right way to do it, packages should be treated as individual units, the notion of monorepo is purely just for us. If we ever publish these packages (which we should), then people trying to download these packages and use them would be unable to since theyre missing deps. These packages should ideally be versioned properly too, but thats something to address later on

packages/composites/token-allocation/src/types.ts
```diff
@@ -27,8 +28,10 @@ export type PriceAdapter = {
   ) => Promise<ResponsePayload>
 }
 
+export type SourceRequestOptions = { [source: string]: RequestConfig }
```
RodrigoAD, line 31: I think each source is a `PriceAdapter` type

packages/composites/token-allocation/src/adapter.ts
```diff
@@ -68,35 +68,39 @@ const toValidAllocations = (allocations: any[]): TokenAllocations => {
...
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
   const { quote = config.defaultQuote, method = config.defaultMethod } = validator.validated.data
   const allocations = toValidAllocations(validator.validated.data.allocations)
+  const source = validator.validated.data.source
+  const request = config.sources[source]
```
RodrigoAD, line 103: Could we name this something like `sourceConfig`? After the type change, this should be of the `PriceAdapter` type, so we can avoid that `any` types

packages/composites/synth-index/README.md
```diff
@@ -4,24 +4,10 @@ The adapter calculates a Synthetix Index value in the currency selected
 
 ## Configuration
 
+- `DATA_PROVIDER_URL`: The adapter URL for the price data provider
```
RodrigoAD, line 7: We'll probably want the `*_DATA_PROVIDER_URL` table in every composite that relies on Token Allocations

packages/composites/token-allocation/src/config.ts
```diff
@@ -1,20 +1,42 @@
...
+    const url = getURL(a.NAME.toUpperCase())
+    if (url) {
+      const defaultConfig = getDefaultConfig(prefix)
+      defaultConfig.api.baseURL = url
+      defaultConfig.api.method = 'post'
+      sources[a.NAME] = getDataProvider(defaultConfig.api)
```
RodrigoAD, line 26: To avoid case errors, make the sources keys either uppercase or lowercase, and make the conversion as well when you want to access any of them

RodrigoAD (commented):
Thanks man for the PR, left you a couple of comments

packages/composites/synth-index/README.md
justinkaseman: I'll do the other composites that rely on Token Allocations in another PR. If this looks good let's get this one in so that you guys don't hit the errors when running `yarn setup`.

packages/core/types/@chainlink/index.d.ts
krebernisak: I agree, let's remove `external-adapter` package and merge it into `ea-bootstrap` 👍

packages/composites/token-allocation/src/config.ts
```diff
@@ -1,20 +1,43 @@
...
 
   return {
-    priceAdapter: getDataProvider(defaultConfig.api),
+    sources,
     defaultMethod: util.getEnv('DEFAULT_METHOD', prefix) || 'price',
     defaultQuote: util.getEnv('DEFAULT_QUOTE') || 'USD',
   }
 }
+
+export const makeOptions = ({ sources }: Config) => {
```
krebernisak, line 38: Do we really need this method, and can we do this in `makeConfig`?
From the user perspective/api, it seems confusing to both have `makeOptions` & `makeConfig`.

justinkaseman: Yeah I see what you mean. I'll return it as part of the config.

PR #377, opened 2021-03-22
WebSockets

krebernisak:
# WebSockets - Let's connect!

This is a draft PR that explores a path to add WebSocket support to the EA framework.
All comments, suggestions, and proposals are greatly appreciated!

This implementation adds a WebSocket driver component that supports multiple distinct WS connections and multiplexing subscriptions for each connection while handling errors/panics, reconnections, and dropping unused connections/subscriptions. It also exposes real-time metrics to Prometheus.

The general idea is to add a new optional interface to the framework for adapters to implement. The ones with WS support can then expose WS subscription builders for specific data points (request types). Internally, similarly to the functionality of the cache-warmer, a background WS subscription is made, and the resulting stream of data is used to update and keep the data fresh for consumers.

Adapter supporting WS should expose, for each supported data subscription:

- fn to generate subscribe message payload
- fn to generate unsubscribe message payload
- fn to filter/select message from this subscription
- fn to map from a subscription message to a CL data point

With this first iteration, consumers are not aware of these WS connections happening in the background, and they read the data from the service as before. Hopefully, the data should be less stale, and RL tokens consumption of underlying provider API should drop for providers with WS support. In the future, the EA service could also expose a WS endpoint so it can make the persistent bidirectional connection to the CL node.

## TODO framework:

- [x] Basic WebSocket driver implementation with multiplex subscription support
  - [ ] Panic & error handling
  - [x] Automatically reconnect dropped connection/subscriptions
  - [x] Automatically close dead (unresponsive) connection/subscriptions
- [x] WebSocket driver Prometheus metrics
  - [x] Basic metrics like `ws_connection_active`, `ws_subscription_total`, `ws_subscription_active` & `ws_message_total`
- [x] Pipe WS output data stream to storage (keep the cached data fresh)
- [x] Add robust interface so adapters could expose WS subscription specification
  - [x] Mock WebSocket Middleware subscribing to Coinbase ticker WS
- [x] Add WS rate limit tiers to rate-limit info package (e.g., tier 0: 1/10 connection/subscriptions)
- [ ] Add subscription priority list config, so in case of reaching WS limits we prioritise some subscriptions over others
- [ ] Cache-warmer & WS components should play nicely together (cache-warmer should not fetch data for which we have active subscriptions)

## TODO source adapters:

Additionally, we currently have a few one-off WS connection implementations like in the case of `tradingeconomics-stream`. To support WS, that specific adapter was never run serverless but as a persistent containerized service. This PR is going to try to replace those types of implementations with the new WebSocket driver component, which will hopefully be more robust, have more features, will abstract state & lifecycle management from adapters, and enable adapter developers to easily add WS connection without having to duplicate the implementation.

- [x] Add WebSocket support to `tradingeconomics`
  - [x] Remove `tradingeconomics-stream`
- [x] Add WebSocket support to `coinbase`
- [x] Add WebSocket support to `amberdata`
- [x] Add WebSocket support to `coinapi`
- [x] Add WebSocket support to `cryptocompare`


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,183 @@
...
+      // Stream of WS connected & disconnected events
+      const open$ = openObserver.pipe(map((event) => connected({ event, config } as any)))
+      const close$ = closeObserver.pipe(map((event) => disconnected({ event, config } as any)))
+
+      // Close the WS connection on disconnect
+      const disconnect$ = action$.pipe(
+        filter(disconnect.match),
+        filter(({ payload }) => payload.connectionInfo.key === connectionKey),
+        tap(() => wsSubject.closed || wsSubject.complete()),
+        filter(() => false), // do not duplicate events
```
HenryNguyen5, line 70: I think this should be replaced with a no-op action. Such as `WS/UNSUBSCRIBED_SUCCESS`. Usually, with a single async action there is a minimum set of 2 actions, with 3 actions being the norm for error handling. E.g.

For unsubscribing, we would have the following set of actions:
`WS/UNSUBSCRIBE_REQUESTED`
`WS/UNSUBSCRIBE_SUCCESSFUL`
`WS/UNSUBSCRIBE_FAILED`

To reduce the amount of boilerplate this generates, we can use the `createSlice` helper function provided by redux toolkit which automatically creates actions based on the configured reducer keys

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,183 @@
...
+        map(({ payload }) => ({ payload, subscriptionKey: payload.subscriptionInfo.key })),
+        withLatestFrom(state$),
+        filter(([{ subscriptionKey }, state]) => {
+          // if subscription does not exist, then continues
+          return !state.ws.subscriptions.active[subscriptionKey]
+        }),
+        // on a subscribe action being dispatched, open a new WS subscription if one doesn't exist yet
+        mergeMap(([{ subscriptionKey, payload }]) => {
+          // Helper function to build actions
+          const _wsSubscriptionPayload = (message: any) => ({
```
HenryNguyen5, line 86: Underscores mean unused in JS-land, we should use standing camelCase here :), if this function ever actually becomes unused and we forget to remove it, both tsc and eslint wont warn us because of the underscore convention

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,183 @@
...
+          ),
+        ),
+      )
+
+      // Return the new connection stream
+      return ws$
+    }),
+  )
+
+export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$) =>
```
HenryNguyen5, line 140: This seems more well suited for a reducer, or our own redux middleware

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  // Helper function to build actions
+  const _wsSubscriptionPayload = (message: any) => ({
+    connectionInfo,
+    subscriptionInfo,
+    message,
+  })
+
+  store.dispatch(subscribe(_wsSubscriptionPayload(subscribeMsg)))
+
+  const unsubscribeMsg = { type: 'unsubscribe', channels: [{ name: 'ticker', product_ids }] }
```
HenryNguyen5, line 38: Should this be here?

packages/core/bootstrap/src/lib/ws/metrics.ts
```diff
@@ -0,0 +1,43 @@
+import * as client from 'prom-client'
+
+export const ws_connection_active = new client.Gauge({
```
HenryNguyen5, line 3: We can probably remove all the experimental labels since all metrics are experimental currently

HenryNguyen5 (changes requested):

packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -0,0 +1,93 @@
...
+    builder.addCase(actions.unsubscribed, (state, action) => {
+      logger.info(`WS: Unsubscription ${JSON.stringify(action.payload.subscriptionMsg)}`)
+      // Remove subscription
+      const key = getSubsId(action.payload.subscriptionMsg)
+      delete state.active[key]
+      // Decrement num of active subscriptions
+      state.total--
+    })
+
+    builder.addCase(actions.unsubscribedAll, (state) => {
```
RodrigoAD, line 81: This could be the disconnection action

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,228 @@
...
+          return !state.ws.subscriptions.active[subscriptionKey]
+        }),
+        // on a subscribe action being dispatched, open a new WS subscription if one doesn't exist yet
+        mergeMap(([{ subscriptionKey, payload }]) =>
+          wsSubject
+            .multiplex(
+              () => payload.subscriptionMsg,
+              () => wsHandler.unsubscribe(payload.subscriptionMsg), // TODO: Test this
+              (message) => {
+                if (wsHandler.isError(message)) {
```
RodrigoAD, line 108: Handle the error and retry

packages/sources/coinapi/src/config.ts
```diff
@@ -25,3 +27,35 @@ export const makeConfig = (prefix?: string): Config => {
   }
   return config
 }
+
+// TODO: Needs to be tested
```
RodrigoAD, line 31: Can be a separate story

packages/sources/tradingeconomics/src/config.ts
```diff
@@ -18,20 +20,52 @@ import { util } from '@chainlink/ea-bootstrap'
...
+    ...defaultConfig,
+    url: util.getEnv('API_URL', prefix) || DEFAULT_WS_API_ENDPOINT,
     key: util.getRequiredEnv('API_CLIENT_KEY', prefix),
     secret: util.getRequiredEnv('API_CLIENT_SECRET', prefix),
     symbols: util.getRequiredEnv('SYMBOLS', prefix),
     reconnectTimeout: Number(util.getEnv('RECONNECT_TIMEOUT', prefix) || 3000),
   }
 }
+
+// TODO: Needs testing
```
RodrigoAD, line 45: Can be a separate story

packages/sources/amberdata/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute } from './adapter'
-import { makeConfig, NAME } from './config'
+import { makeConfig, makeWSHandler, NAME } from './config'
 
-export = { NAME, makeExecute, makeConfig, ...expose(makeExecute()) }
+export = { NAME, makeExecute, makeConfig, ...expose(makeExecute(), makeWSHandler(makeConfig())) }
```
justinkaseman, line 5: Token Allocations is importing this package, which on import runs `makeWSHandler(makeConfig())`, which throws an error in `makeConfig`because `API_KEY` isn't provided.

RodrigoAD: Aha I see, thanks!

packages/core/bootstrap/src/lib/store/index.ts
```diff
@@ -14,7 +14,11 @@ import { nanoid } from '@reduxjs/toolkit'
 import { composeWithDevTools } from 'remote-redux-devtools'
 import { isDebug, isDebugLogLevel } from '../util'
 
-export const toActionPayload = <A extends ActionBase>(data: any): A => ({
+export const asAction = <T>() => (p: T) => ({
```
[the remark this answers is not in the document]
krebernisak: Yeah makes sense, but there weren't a lot of these util functions so I didn't create a new file just for this one.

packages/core/bootstrap/src/lib/store/index.ts
```diff
@@ -14,7 +14,11 @@ import { nanoid } from '@reduxjs/toolkit'
...
+  payload: toActionPayload<T>(p),
+})
+
+export const toActionPayload = <T>(data: T): ActionBase & T => ({
```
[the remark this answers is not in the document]
krebernisak: Yes, I was looking for a better way to construct these actions with some default metadata. We could use your Redux experience here :)

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,299 @@
...
+            take(1),
+            mapTo(subscriptionReset({})),
+          )
+    
+          // start the current unsubscription timer
+          const timeout$ = of(unsubscribe({ ...payload })).pipe(delay(config.subscriptionTTL), log('WS: Unsubscription due to inactive feed'))
+    
+          // if a re-subscription comes in before timeout emits, then we emit nothing
+          // else we unsubscribe from the current subscription
+          return race(reset$, timeout$)
```
krebernisak, line 204: Just filter here so heartbeats don't go through after the race.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,299 @@
...
+      const messageReceived$ = action$.pipe(
+        filter(messageReceived.match)
+      )
+
+      const unresponsiveOnTimeout$ = messageReceived$.pipe(
+        mergeMap(({ payload: { input, subscriptionKey } }) => {
+          const reset$ = messageReceived$.pipe(
+            filter(({ payload }) => subscriptionKey === payload.subscriptionKey),
+            take(1),
+            mapTo(subscriptionReset({})),
```
krebernisak, line 217: Same comment as above.

packages/core/bootstrap/src/lib/ws/actions.ts
```diff
@@ -0,0 +1,42 @@
...
+  input: AdapterRequest
+}
+
+
+
+export const subscribe = createAction('WS/SUBSCRIBE!', asAction<WSSubscriptionPayload>())
+export const subscribed = createAction('WS/SUBSCRIBED', asAction<WSSubscriptionPayload>())
+export const unsubscribe = createAction('WS/UNSUBSCRIBE!', asAction<WSSubscriptionPayload>())
+export const unsubscribed = createAction('WS/UNSUBSCRIBED', asAction<WSSubscriptionPayload>())
+export const subscriptionReset = createAction('WS/SUBSCRIPTION_RESET!', asAction())
```
krebernisak, line 32: Seems unnecessary.

```suggestion
```

packages/core/bootstrap/src/lib/ws/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  /** Number of ms if no data received this connection is considered dead */
+  connectionTTL: number
+  // TODO: exponential back-off
+  /** Maximum number of connection retries */
+  connectionRetryLimit: number
+  /** Minimum delay between connection retries */
+  connectionRetryDelay: number
+
+  /** Maximum amount of subscriptions per connection */
+  subscriptionLimit: number
```
krebernisak, line 18: I don't see this used? We should populate the limit from our rate limits package, or it's set manually by the user.

packages/core/bootstrap/src/lib/ws/types.ts
```diff
@@ -0,0 +1,30 @@
+
+export interface WSConfig {
+  /** Info about the connection */
+  connectionInfo: {
+    key: string
+  }
+  /** Maximum amount of parallel connections */
+  connectionLimit: number
```
krebernisak, line 8: I don't see this used? We should populate the limit from our rate limits package, or it's set manually by the user.

packages/core/bootstrap/src/lib/ws/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  /** Number of ms if no data received this connection is considered dead */
+  connectionTTL: number
+  // TODO: exponential back-off
+  /** Maximum number of connection retries */
+  connectionRetryLimit: number
```
krebernisak, line 13: Some of these are unused, for now.
Please open a GitHub issue and make a list of things that still need to be implemented.

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,48 @@ declare module '@chainlink/types' {
     warning?: string
   }
 
+
+  export type MakeWSHandler = () => WSSubscriptionHandler
```
krebernisak, line 106:
```suggestion
  export type WSHandlerFactory = () => WSSubscriptionHandler
```

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,48 @@ declare module '@chainlink/types' {
...
+     * Connection information
+     */
+    connection: {
+      url: string
+      protocol?: any
+    }
+    /**
+     * Gets the subscription message necessary to subscribe to the feed channel
+     */
+    subscribe: (input: AdapterRequest) => any
```
krebernisak, line 119: What if the WS subscription is not supported for this input type:

```suggestion
    subscribe: (input: AdapterRequest) => any | undefined
```

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,48 @@ declare module '@chainlink/types' {
...
+    /**
+     * Gets the unsubscription message necessary to unsubscribe to the feed channel
+     */
+    unsubscribe: (input: any) => any
```
krebernisak, line 123: What if the WS subscription is not supported for this input type:

```suggestion
    unsubscribe: (input: any) => any | undefined
```

packages/sources/amberdata/src/config.ts
```diff
@@ -21,3 +24,41 @@ export const makeConfig = (prefix = ''): Config => {
   config.api.baseURL = config.api.baseURL || DEFAULT_API_ENDPOINT
   return config
 }
+
+export const makeWSHandler = (defaultConfig?: Config): MakeWSHandler => {
+  const subscriptions: any = {}
+  const getPair = (input: AdapterRequest): string => {
+    const validator = new Validator(input, endpoint.price.customParams)
+    if (validator.error) {
+      return ''
```
krebernisak, line 33: We should not swallow an error like this. If we can not construct a subscribe/unsubscribe message, we should return `undefined` that the WS component should handle.

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,48 @@ declare module '@chainlink/types' {
...
+     */
+    isError: (message: any) => boolean
+    /**
+     * Based on the incoming message, returns its corresponding subscription message
+     */
+    subsFromMessage: (message: any) => any
+    /**
+     * Optional. If defined, the response will be formatted using this function
+     */
+    toAdapterResponse?: (message: any) => any
```
krebernisak, line 143: Why do we need both `parse` and `toAdapterResponse`?

If there is not something I'm missing, I suggest we use just the latter:

```suggestion
    toResponse: (message: any) => AdapterResponse
```

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -128,10 +129,13 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
     const key = _getKey(data)
     const coalescingKey = _getCoalescingKey(key)
     const endMetrics = metrics.observeMetrics(data.id, key, data.debug?.feedId)
-    const maxAge = _getRequestMaxAge(data) || _getDefaultMaxAge(data)
+    let maxAge = _getRequestMaxAge(data) || _getDefaultMaxAge(data)
     // Add successful result to cache
     const _cacheOnSuccess = async ({ statusCode, data, result }: AdapterResponse) => {
       if (statusCode === 200) {
+        if (maxAge < 0) {
```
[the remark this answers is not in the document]
RodrigoAD: When `maxAge` is on the request, and has a `-1` value, it should refresh the cache, but it doesn't mean that the new max age needs to be -1

packages/core/types/@chainlink/index.d.ts
RodrigoAD: `parse` get the message relevant information, `toAdapterResponse` is how the response will be saved in cache. It's possible that some adapter wants an specific response format

krebernisak: But why would the WS component need the "message relevant information". This can all be abstracted by our `AdaperResponse` type that adapters would map messages to.

packages/core/bootstrap/src/lib/ws/config.ts
```diff
@@ -0,0 +1,51 @@
...
+  const url = getRequiredEnv(ENV_WS_CONNECTION_URL, prefix)
+  const protocol = getEnv(ENV_WS_CONNECTION_PROTOCOL, prefix) // TODO: load array
+  const key = getEnv(ENV_WS_CONNECTION_KEY, prefix) || url // TODO: generate key
+  return { key, url, protocol }
+}
+
+/** Load WSConfig from environment variables */
+export const envLoad_WSConfig = (prefix = ''): WSConfig => ({
+  connectionInfo: envLoad_WSConnectionInfo(prefix),
+  connectionLimit: Number(getEnv(ENV_WS_CONNECTION_LIMIT, prefix)) || DEFAULT_WS_CONNECTION_LIMIT,
```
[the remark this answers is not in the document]
krebernisak: Yeah, I agree we can type this better and remove some of the boilerplate. But as we are short on time, and this improvement could also be made to some other places, let's defer for now. 

@RodrigoAD please make a separate GH Issue to improve how we type and build configs for different components, and link these two Henry's comments.

packages/core/bootstrap/src/lib/ws/actions.ts
```diff
@@ -0,0 +1,42 @@
+import { AdapterRequest, WSSubscriptionHandler } from '@chainlink/types'
+import { createAction } from '@reduxjs/toolkit'
+import { asAction } from '../store'
+import { WSConfig, WSConnectionInfo } from './types'
+
+/** CONNECTIONS */
+export interface WSConfigPayload {
+  config: WSConfig
+  wsHandler: WSSubscriptionHandler
```
krebernisak, line 9: Do we use this?

A handler is an object/function, meaning not serializable, and we should not put any non-serializable data in an event.

```suggestion
```

packages/core/bootstrap/src/lib/ws/epics.ts
krebernisak: @RodrigoAD please add another GitHub issue: Improve Redux command handling. 
I agree every command should be followed by a `SUCCESS` or `FAIL` event.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,299 @@
...
+                  logger.error(`WS: Subscription Error: ${JSON.stringify(message)}`)
+                  return false
+                }
+                return getSubsId(wsHandler.subsFromMessage(message)) === subscriptionKey
+              },
+            )
+            .pipe(
+              withLatestFrom(state$),
+              map(([message, state]) => {
+                // TODO: Use merge map to not discard the first message
```
krebernisak, line 131: Please add this in, as discussed.

packages/core/types/@chainlink/index.d.ts
RodrigoAD: Yep I see

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,299 @@
...
+          )
+      )
+
+      const messages$ = action$.pipe(
+        filter(messageReceived.match),
+        filter((action) => wsHandler.filter(action.payload.message)),
+        mergeMap(async (action) => {
+          const response = wsHandler.parse(action.payload.message)
+          if (!response) return action
+          const adapterResponse = wsHandler.toAdapterResponse || ((result: any) => Requester.success('1', { data: { result } }))
```
krebernisak, line 163: Simplify this a little bit by removing the `parse` fn.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,299 @@
...
+              maxAge: -1
+            },
+            debug: {
+              ws: true
+            }
+          }
+          await cache(input)
+          return action
+        }),
+        catchError((error) => of(error).pipe(log(`WS: ${error.message}`))),
```
krebernisak, line 179: If we have an error here, this will close the stream as you just log and propagate/throw the error again.

Remove this line, and add a try/catch in the `mergeMap` above, where we'll catch the cache errors, and log them for now.

packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -0,0 +1,87 @@
...
+  total: number
+}
+
+const initConnectionsState: ConnectionsState = { active: {}, total: 0 }
+
+export const connectionsReducer = createReducer<ConnectionsState>(
+  initConnectionsState,
+  (builder) => {
+    builder.addCase(actions.connected, (state, action) => {
+      logger.info('WS: New connection')
```
krebernisak, line 25: Please remove all logging for reducers, as they should be side-effect-free, and move them to epics where events are triggering.

packages/sources/coinapi/src/config.ts
```diff
@@ -25,3 +27,36 @@ export const makeConfig = (prefix?: string): Config => {
...
+      subscribe: (input) => {
+        const validator = new Validator(input, endpoint.price.customParams)
+        if (validator.error) {
+          return 
+        }
+        const base = validator.overrideSymbol(NAME).toLowerCase()
+        const quote = validator.validated.data.quote.toLowerCase()
+        return getSubscription([base, quote])
+      },
+      unsubscribe: () => '',
```
krebernisak, line 53: How do we unsubscribe?

packages/core/bootstrap/src/lib/provider-limits/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  const providerLimit = getProviderLimits(provider, tier, 'http')
+  if (!providerLimit) {
+    const msg = `Rate Limit: Provider: "${provider}" and Tier: "${tier}" doesn't match any provider spec in limits.json`
+    logger.warn(msg)
+    throw new Error(msg)
+  }
+  return calculateRateLimit(providerLimit as HTTPTier)
+}
+
+export const getWSLimits = (
```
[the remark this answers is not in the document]
RodrigoAD: Not yet, but WS module should use it when handle subs limits

packages/core/bootstrap/src/lib/provider-limits/index.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export const getRateLimit = (
+  provider: string,
+  tier: string,
+): ProviderRateLimit => {
+  const providerLimit = getProviderLimits(provider, tier, 'http')
+  if (!providerLimit) {
+    const msg = `Rate Limit: Provider: "${provider}" and Tier: "${tier}" doesn't match any provider spec in limits.json`
+    logger.warn(msg)
```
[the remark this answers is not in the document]
RodrigoAD: Right

packages/sources/cryptocompare/src/config.ts
```diff
@@ -25,3 +26,42 @@ export const makeConfig = (prefix?: string): Config => {
...
+  const getPair = (input: AdapterRequest): string => {
+    const validator = new Validator(input, endpoint.price.customParams)
+    if (validator.error) {
+      return ''
+    }
+    const base = validator.overrideSymbol(NAME).toUpperCase()
+    const quote = validator.validated.data.quote.toUpperCase()
+    return `${base}~${quote}`
+  }
+  const getSubscription = (pair: string, subscribe = true) => ({ action: subscribe ? 'SubAdd' : 'SubRemove', subs: [`${subscriptions.aggregate}~CCCAGG~${pair}`] })
```
krebernisak, line 46:
```suggestion
  const getSubscription = (pair: string, action = 'SubAdd' | 'SubRemove') => ({ action, subs: [`${subscriptions.aggregate}~CCCAGG~${pair}`] })
```

packages/core/bootstrap/src/lib/ws/types.ts
RodrigoAD: #467

packages/core/bootstrap/src/lib/ws/types.ts
RodrigoAD: #467

packages/core/bootstrap/src/lib/ws/types.ts
RodrigoAD: #467

packages/sources/coinapi/src/config.ts
RodrigoAD: Didn't find anything on their docs

packages/core/bootstrap/src/lib/ws/actions.ts
RodrigoAD: #469

packages/core/bootstrap/src/lib/ws/epics.ts
RodrigoAD: #470

packages/core/bootstrap/src/lib/ws/config.ts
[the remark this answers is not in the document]
RodrigoAD: #471

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -124,21 +131,22 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
...
             .pipe(
               withLatestFrom(state$),
-              map(([message, state]) => {
-                // TODO: Use merge map to not discard the first message
+              mergeMap(([message, state]) => {
                 if (!state.ws.subscriptions[subscriptionKey]?.active) {
-                  return subscribed(payload)
+                  return of(subscribed(payload), messageReceived({ message, subscriptionKey })).pipe(
+                    log(`WS: New subscription ${JSON.stringify(payload.subscriptionMsg)}`),
+                    take(1),
```
krebernisak, line 138: Why `take(1)` here? This will only let the first `subscribed` through, and discard the next `messageReceived`.

I suggest removing this line.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -154,16 +162,22 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
       const messages$ = action$.pipe(
         filter(messageReceived.match),
         filter((action) => wsHandler.filter(action.payload.message)),
+        withLatestFrom(state$),
+        map(([action, state]) => {
+          const input = state.ws.subscriptions[action.payload.subscriptionKey]?.input || {}
+          if (!input) logger.warn(`WS: Could not find subscription from incoming message`)
+          return { ...action, input }
+        }),
```
krebernisak, line 170: I'd prefer we don't add to the action here, but we remove this whole map operator and move the logic to the `mergeMap` below.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -210,7 +224,13 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
       )
 
       const unresponsiveOnTimeout$ = messageReceived$.pipe(
-        mergeMap(({ payload: { input, subscriptionKey } }) => {
+        withLatestFrom(state$),
+        map(([action, state]) => {
+          const input = state.ws.subscriptions[action.payload.subscriptionKey]?.input || {}
+          if (!input) logger.warn(`WS: Could not find subscription from incoming message`)
+          return { ...action, input }
+        }),
```
krebernisak, line 232: Same as above, just fetch the input cleanly in the `mergeMap` below, instead of passing it with an action.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,319 @@
...
+      const messages$ = action$.pipe(
+        filter(messageReceived.match),
+        filter((action) => wsHandler.filter(action.payload.message)),
+        withLatestFrom(state$),
+        map(([action, state]) => {
+          const input = state.ws.subscriptions[action.payload.subscriptionKey]?.input || {}
+          if (!input) logger.warn(`WS: Could not find subscription from incoming message`)
+          return { ...action, input }
+        }),
+        mergeMap(async (action) => {
```
krebernisak, line 171:
```suggestion
        mergeMap(async ([action, state]) => {
          const input = state.ws.subscriptions[action.payload.subscriptionKey]?.input || {}
          if (!input) logger.warn(`WS: Could not find subscription from incoming message`)
```

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,319 @@
...
+              debug: {
+                ws: true
+              }
+            }
+            await cache(input)
+            return action
+          } catch (e) {
+            logger.error(`WS: ${e.message}`)
+            return action
+          }
```
krebernisak, line 192:
```suggestion
          } catch (e) {
            logger.error(`WS: ${e.message}`)
          }
          return action
```

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,319 @@
...
+      // Once a request happens, a subscription timeout starts. If no more requests ask for this susbcription before the time runs out, it will be unsubscribed
+      const unsubscribeOnTimeout$ = heartbeat$.pipe(
+        // when a subscription comes in
+        mergeMap(({ payload, subscriptionKey }) => {
+          // we look for matching subscriptions of the same type
+          // which deactivates the current timer
+          const reset$ = heartbeat$.pipe(
+            filter(({ subscriptionKey: keyB }) => subscriptionKey === keyB),
+            take(1),
+            filter(() => false)
```
krebernisak, line 210: Same comment as above!

krebernisak, line 210: This filter will not let any events through!

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -0,0 +1,319 @@
...
+        map(([action, state]) => {
+          const input = state.ws.subscriptions[action.payload.subscriptionKey]?.input || {}
+          if (!input) logger.warn(`WS: Could not find subscription from incoming message`)
+          return { ...action, input }
+        }),
+        mergeMap(({ payload: { subscriptionKey }, input }) => {
+          const reset$ = messageReceived$.pipe(
+            filter(({ payload }) => subscriptionKey === payload.subscriptionKey),
+            take(1),
+            filter(() => false)
```
krebernisak, line 237: This filter will not let any events through!

packages/sources/coinapi/src/config.ts
```diff
@@ -25,3 +27,34 @@ export const makeConfig = (prefix?: string): Config => {
...
+        url: defaultConfig.api.baseWsURL || DEFAULT_WS_API_ENDPOINT
+      },
+      subscribe: (input) => {
+        const validator = new Validator(input, endpoint.price.customParams)
+        if (validator.error) return
+        const base = validator.overrideSymbol(NAME).toLowerCase()
+        const quote = validator.validated.data.quote.toLowerCase()
+        return getSubscription([base, quote])
+      },
+      unsubscribe: () => '',
```
krebernisak, line 51: Let's rather return `undefined` here as a signal that the unsubscribe msg is undefined :)

packages/core/bootstrap/src/lib/ws/epics.ts
RodrigoAD: Yep right, I wanted the log to be printed only for the first event

krebernisak: Yeah, this will not do it!

You would need to `subscribed(payload).pipe(log....)`, but mixing core logic with logs becomes overwhelming at one point. I suggest you move logs to a separate pipeline altogether, as I did with metrics. Metrics are handled in a completely separate pipeline.

RodrigoAD: Opening an issue on ws logging

HenryNguyen5 (approved):

PR #382, opened 2021-03-24
Build process for packages structure

justinkaseman:
### Description
Changes the build process to handle our new packages directory structure

### Changes
- Adds hard dependencies for all packages to move toward semantic versioning
- Brings @chainlink/external-adapter into @chainlink/bootstrap
- Consistent published package names for `/core`
- Migrates package manager from Yarn => Yarn 2
See more here: https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern
- Dockerfile uses `/package` in the path to copy files
- Clean up Makefile and utilize `yarn workspace` runner instead of `wsrun`
- Rework adapters.json for each package folder: `core`, `examples`, `sources`, `composites`


Makefile
```diff
@@ -19,31 +19,23 @@ new:
...
+	yarn workspace @chainlink/${name}-adapter setup
 	yarn --frozen-lockfile --production
 
 build:
+	#######################################
+	# Compile JavaScript to a single file
+	#######################################
 	npx @vercel/ncc@0.25.1 build packages/$(adapter) -o packages/$(adapter)/dist
-
-build-2-step:
```
justinkaseman, line 36: Removed these, to be handled in a different PR,

packages/sources/xbto/tsconfig.json
```diff
@@ -3,10 +3,11 @@
...
     "outDir": "dist",
     "rootDir": "src",
+    "types": ["mocha"],
     "typeRoots": ["../../../node_modules/@types", "../../core/types", "./types"],
     "composite": true
   },
   "include": ["src/**/*"],
   "exclude": ["dist", "**/*.spec.ts", "**/*.test.ts"],
-  "references": [{ "path": "../../core/external-adapter" }, { "path": "../../core/bootstrap" }]
+  "references": [{ "path": "../../core/bootstrap" }]
```
krebernisak, line 12: Why did we add these references in? Is that to support topologically building dependencies?

We were using `wsrun` for that with Yarn 1, but now with Yarn 2 there is a [foreach  -t](https://yarnpkg.com/cli/workspaces/foreach) option that should do just that.

justinkaseman: [References paths](https://www.typescriptlang.org/docs/handbook/project-references.html) are supposed to speed up build times and improve intellisense performance by referencing already compiled code when building dependent projects.

Is that what the `-t` flag on `foreach` is doing too?

PR #384, opened 2021-03-25
Overrides Input parameter

RodrigoAD:
## Description
`overrides` param can be included into the request body. If the symbol provided as base exists in `overrides`, it should use that as symbol.

There is param validation added into `Validator` that checks if `overrides` has the proper format

Adapter that uses overrides:
- 1forge
- amberdata
- alphavantage
- coinapi
- coingecko
- coinmarketcap
- coinpaprika
- cryptocompare
- currencylayer
- dxfeed
- dxfeed-secondary
- finage
- fixer
- iexcloud
- kaiko
- metalsapi
- nomics
- openexchangerates
- polygon
- tradermade


1forge/src/config.ts
```diff
@@ -1,6 +1,8 @@
 import { Requester } from '@chainlink/external-adapter'
 import { Config } from '@chainlink/types'
 
+export const NAME = '1FORGE'
```
RodrigoAD, line 4: Moved Adapter name from `index.ts` to `config.ts` so can be accessed inside the adapter

presetSymbols.json
```diff
@@ -0,0 +1,24 @@
+{
```
RodrigoAD, line 1: Should we keep this file in the repo?

krebernisak: We need to keep these in for now as removing them means changing the external API, and feeds will break on adapter updates.

external-adapter/src/validator.ts
```diff
@@ -32,20 +34,66 @@ export class Validator {
...
+  validateOverrides() {
+    if (!this.input.data?.overrides) return
+    try {
+      this.validated.overrides = this.formatOverride(this.input.data.overrides)
+    } catch (e) {
+      this.throwInvalidInput(e)
+    }
+  }
+
+  throwInvalidInput(error: any) {
```
krebernisak, line 50: This function does not throw, consider renaming.

coinpaprika/src/endpoint/price.ts
```diff
@@ -25,25 +20,39 @@ const convertFromTicker = async (ticker: string, coinId: string | undefined) =>
...
         x.symbol.toLowerCase() === ticker.toLowerCase() && x.rank !== 0,
     )
-  // if (!coin) return callback('Could not find coin', null)
+  if (typeof coin?.id === 'undefined') {
+    throw new Error('Coin id not found')
+  }
   return coin.id.toLowerCase()
 }
 
+const overrideSymbol = (overrides: Override | undefined, symbol: string): string | undefined => {
```
krebernisak, line 29: Isn't this method already defined in `validator.overrideSymbol`?

coingecko/src/endpoint/price.ts
```diff
@@ -40,30 +24,42 @@ const convertFromTicker = async (config: Config, ticker: string, coinId: string)
...
 
   if (typeof coin === 'undefined') {
-    return undefined
+    throw new Error('Coin id not found')
   }
 
   return coin.id.toLowerCase()
 }
 
+const overrideSymbol = (overrides: Override | undefined, symbol: string): string | undefined => {
```
krebernisak, line 36: Isn't this method already defined in `validator.overrideSymbol`?

coinpaprika/src/endpoint/price.ts
RodrigoAD: This is a special case, happens in paprika and coingecko. We don't want to get the `overrideSymbol` by default, as if doesn't exists, we want to query their ID, and for that, we need the symbol that is on `input.data`

external-adapter/src/validator.ts
RodrigoAD: Right

coinpaprika/src/endpoint/price.ts
krebernisak: Because `validator.overrideSymbol` will throw? But I'm not sure that is the right thing to do. These overrides should not be required IMHO, so if we have an override we override if not we should continue as is.

RodrigoAD: Is a special case for paprika and coingecko. If the asked symbol is not in `overrides`, we need its `id` from the data provider. For this, we need the requested symbol

thodges-gh: Can we make `overrideSymbol` overloaded in the validator to account for this? Should be cleaner overall.

justinkaseman (dismissed):
LGTM :rocket:

external-adapter/src/validator.ts
```diff
@@ -32,20 +35,75 @@ export class Validator {
...
     }
   }
 
+  validateOverrides() {
+    if (!this.input.data?.overrides) {
+      this.validated.overrides = this.formatOverride(presetSymbols)
+      return
+    }
+    try {
+      this.validated.overrides = this.formatOverride({
```
justinkaseman, line 48: I think we should deep merge here instead of shallow merging. I'm worried about some case where someone overrides say
```
{
  "tradermade": {
    "FTSE": "UK200"
  }
}
```
but now this overwrites all of the other overrides.

Should be pretty easy to do with lodash [.merge](https://lodash.com/docs/4.17.15#merge)

external-adapter/test/validator.test.ts
```diff
@@ -122,4 +122,44 @@ describe('Validator', () => {
     assert.equal(validator.validated.id, input.id)
     assert.isUndefined(validator.error)
   })
+
```
justinkaseman, line 125: Yay tests! :1st_place_medal:

1forge/README.md
```diff
@@ -27,6 +27,8 @@
 |    ✅     | `base`, `from` |      The symbol of the currency to query      | [List](https://1forge.com/currency-list) |             |
 |    ✅     | `quote`, `to`  |   The symbol of the currency to convert to    | [List](https://1forge.com/currency-list) |             |
 |    🟡     |   `quantity`   | An additional amount of the original currency |                                          |             |
+|    🟡     |   `overrides`   | If base provided is found in overrides, that will be used  | [Format](../presetSymbols.json)|             |
```
justinkaseman, line 30: Maybe we should have some root level documentation on shared `input parameters` like we do for `meta`. Don't remove these - but something to think about to make these docs "scale".

justinkaseman (approved):
:+1:

krebernisak (approved):

PR #389, opened 2021-03-26
Use ratelimit config

justinkaseman:


packages/core/bootstrap/src/lib/rate-limit/config.ts
```diff
@@ -15,8 +16,12 @@ export interface Config {
...
 
 export function get(): Config {
+  const capacity = parseInt(getEnv('RATE_LIMIT_CAPACITY') || '')
+  const provider = getEnv('API_PROVIDER') || ''
+  const tier = parseInt(getEnv('API_TIER') || '')
+  const providerConfig = getRateLimit(provider, tier)
   return {
     hashOpts: getHashOpts(),
-    totalCapacity: parseInt(process.env.RATE_LIMIT_CAPACITY || ''),
+    totalCapacity: capacity || providerConfig?.quota,
```
justinkaseman, line 25: `RATE_LIMIT_CAPACITY` takes priority as an override. Otherwise looks for `API_PROVIDER` and `API_TIER`

PR #398, opened 2021-03-30
Merge and rename tradingeconomics-stream as tradingeconomics

krebernisak:


packages/sources/tradingeconomics/src/adapter.ts
```diff
@@ -1,36 +1,90 @@
...
+}
+
+const commonSymbols: { [symbol: string]: string } = {
+  N225: 'NKY:IND',
+  FTSE: 'UKX:IND',
+}
+
+export const execute = async (input: AdapterRequest, config: Config) => {
+  // First start background service
+  if (!isServiceStarted) {
```
krebernisak, line 42: @boxhock this PR is hard to review, but I essentially only changed the subscription flow by adding these few lines here and removing them from `index.ts`.

boxhock: I like the way of doing this :+1:

boxhock (approved):

PR #399, opened 2021-03-30
Move write adapters to /targets subfolder

krebernisak:
A quick refactor to separate out write adapters to their own section. 
This will make us notice the different type of adapters we need, so we can incorporate it in our design.


justinkaseman (approved):
Great stuff, LGTM

PR #401, opened 2021-03-30
make scripts to ts scripts

justinkaseman:
### Description
To give us more fine grained control and keep the code-base isomorphic to JS this PR migrates our make scripts to TS.

There are two scripts:
```
yarn docker [type path] [folder name]
```
and 
```
yarn new [type] [name]
```


scripts/new.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const n: string = process.argv[3]
+  if (!n) return log.red('Missing second argument: name')
+
+  shell.mkdir(`packages/${type}s/${n}`)
+  shell.cp('-R', `packages/examples/${type}/*`, `packages/${type}s/${n}`)
+  // cp -R will not copy hidden & special files, so we copy manualy
+  shell.cp(`packages/examples/${type}/.eslintrc.js`, `packages/${type}s/${n}`)
+  shell
+    .cat('.github/strategy/adapters.json')
```
krebernisak, line 21: Not sure what to make of this, pun intended haha

On the one hand, we currently have a need for only a simple build tool, and declarative Makefile is a lot simpler and more readable than imperative JS code like this. On the other hand, this could open a way for us to start building an automaton CLI for EAs, which was on my mind for quite some time.

A while ago I started to prototype a similar tool codenamed `@chainlink/uplink` that used Pulumi to spawn quick CL + EA epithermal environments. It also had the ability to build and publish Docker images to AWS or Docker Hub but eventually, other things took over and development stopped.

Let's start small and make this work for the current release, then start a wiki page where we could start planning/discussing what are some automaton ideas that would fit into a tool like this.

krebernisak (approved):

PR #402, opened 2021-03-31
Epic EAEE Acceptance Fixes

RodrigoAD:
- Feed Id label on metrics
- Cost fixed on RL 
- Standardize time metrics to be in secs
- General dashboard added

### DIFF
--- a/packages/core/bootstrap/src/lib/rate-limit/index.ts
+++ b/packages/core/bootstrap/src/lib/rate-limit/index.ts
@@ -102,8 +102,8 @@ export const withRateLimit = (store: Store<RootState>): Middleware => async (exe
     participant_id: requestTypeId,
     experimental: 'true',
   }
-  const cost = Number(result.debug?.providerCost) || 1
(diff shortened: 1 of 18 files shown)


packages/core/bootstrap/src/lib/rate-limit/index.ts
```diff
@@ -102,8 +102,8 @@ export const withRateLimit = (store: Store<RootState>): Middleware => async (exe
     participant_id: requestTypeId,
     experimental: 'true',
   }
-  const cost = Number(result.debug?.providerCost) || 1
-  metrics.rateLimitCreditsSpentTotal.labels(defaultLabels).inc(cost)
+  const cost = Number(result.debug?.providerCost)
```
HenryNguyen5, line 105: Can cost be fractional? Otherwise we should use `parseInt`

RodrigoAD: It should never be fractional, but we could use `parseInt` just in case

krebernisak (approved):

PR #403, opened 2021-03-31
Refactor docs for EAEE

aalu1418:
Updating README docs for various adapters for consistent layout and information structure
+ other small modifications for information/readability
+ applying template to composite adapters
+ applying template to adapters used in composite adapters

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,10 @@ Runs the setup step for all adapters. Typically this step just compiles TypeScri
 ### Running Locally
 
 External adapters should be run as long-lived processes. They can be found under various folders:
-* [`packages/sources`](./packages/sources): adapters for reading data from API sources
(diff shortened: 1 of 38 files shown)


aalu1418:
Question about the Bitcoin Composite Adapter:
https://github.com/smartcontractkit/external-adapters-js/tree/epic/EAEE/packages/composites/bitcoin-json-rpc

Is the output the full json object or would it only be the specific parameter from `endpoint`?

krebernisak (approved):
Thanks, @aalu1418! Looks great :)

For the next step on the topic of docs and API in general, we should think about exposing descriptive config & input/response info directly from the adapters, and potentially generate docs similar to this automatically. An adapter could expose  
JSON data where everything is explained and which we can use to generate docs.

But, we should probably look to use already available tools, and for example, generate Swagger specification.

PR #405, opened 2021-03-31
Cleanup yarn berry

HenryNguyen5:
- Make `@chainlink/types` its own package and include it as a dev dependency where required
- Split up "prepublishOnly" into "prepublish" and "prepack" as the former lifecycle hook is no longer supported
- Remove "remotedev-server" dependency in favour of using a dockerized version

Most issues that were detected from `yarn dlx @yarnpkg/doctor` were fixed, only a few left https://yarnpkg.com/getting-started/migration#before-we-start


krebernisak (approved):
I checked and double-checked everything! Looks good! 😄

justinkaseman (approved):
You are the best. LGTM.

PR #408, opened 2021-04-01
chore/jest migration

HenryNguyen5:


packages/composites/synth-index/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,41 @@
...
+const makeMockConfig = () => {
+  return {
+    defaultNetwork: 'mainnet',
+    taConfig: ta.makeConfig(''),
+  }
+}
+
+describe('execute', () => {
+  const jobID = '1'
+  process.env.DATA_PROVIDER_URL = 'ignoreable'
```
justinkaseman, line 16: Only want to fill this in if it isn't provided

```suggestion
  if (!process.env.DATA_PROVIDER_URL) process.env.DATA_PROVIDER_URL = 'ignoreable'
```

packages/composites/token-allocation/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,34 @@
+import { Requester } from '@chainlink/ea-bootstrap'
+import { assertError } from '@chainlink/ea-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../../src/adapter'
+import { makeConfig } from '../../src/config'
+
+describe('execute', () => {
+  const jobID = '1'
+  process.env.DATA_PROVIDER_URL = 'ignoreable'
```
justinkaseman, line 9: Same - Only want to fill this in if it isn't provided

```suggestion
  if (!process.env.DATA_PROVIDER_URL) process.env.DATA_PROVIDER_URL = 'ignoreable'
```

packages/composites/synth-index/test/integration/adapter.test.ts
HenryNguyen5: We should make this a separate PR if we can, i tried to avoid actual code changes as much as possible

packages/composites/token-allocation/test/integration/adapter.test.ts
HenryNguyen5: We should make this a separate PR if we can, i tried to avoid actual code changes as much as possible

packages/composites/synth-index/test/integration/adapter.test.ts
justinkaseman: Ah yeah 100%.

justinkaseman (approved):
Without you we would still be waiting 30 minutes on each commit like cavemen. Thank you for bringing us :fire: .

PR #412, opened 2021-04-02
chore/cleanup deps 2

HenryNguyen5:
Changes
- Runs lint from root now, we can add a root script later on to support per-package linting again if needed
- Hoists all of our dev deps of `prettier`, `eslint`, `ts-node` and uses them at the top level. Reduces package.json manifest noise and improves linting performance significantly.
- Simplifies eslint config, updating to the latest standard presented by typescript-eslint, removes `standard`(opinionated) and `import`(covered by ts and yarn) configs.
- Removes dangling mocha dep
- Removes tslib harmony deps, we shouldnt need this as tslib is a dep of every package now
- Removes sinon from packages that dont use it


krebernisak (approved):
@HenryNguyen5, this looks great!

I bet the speedup is substantial, and there is no point in having duplicated setup like this with the current structure of the repo.

Going further, I'd like us to think about how can we support adapters external to this repo, as maybe it is sometimes slow to get a PR accepted to this repo from external contributors, and eventually adapters are going to spread out from this repo. I'm not sure long term we want to be in the situation where we need to vet and include every possible adapter here.

Truffle as a quick example, there is no Truffle contracts monorepo where they vet and include a bunch of smart contracts. Instead, they have a core framework (batteries included) + CLI +  boxes. You just `truffle unbox` and your scaffolding project is ready.

So if we try to build something similar, that scaffolding project(s) will need to have a different setup than adapters included in this monorepo.

```bash
npx @chainlink/ea-bootstrap create  ...
```

PR #416, opened 2021-04-06
fixes to the new.ts file for generating new adapters

aalu1418:
* check to make sure `jq` is installed so the remaining script can run
* removed commands for copying `.eslintrc.js` file from the examples folder (doesn't exist)
* added missing `s` in editing the `.github/strategy/adapters.json` file
* fixed `shell.sed()` parameter order


justinkaseman (approved):
Good stuff. LGTM.

PR #417, opened 2021-04-06
bitcoin endpoints

emmick4:
This PR adds height and difficulty endpoints from the blockchain.com and blockstream.info APIs.


blockstream/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'ETHGASSTATION'
```
thodges-gh, line 5:
```suggestion
const NAME = 'BLOCKSTREAM'
```

thodges-gh (changes requested):

blockstream/README.md
```diff
@@ -0,0 +1,98 @@
...
+      "580": 0.6,
+      "600": 0.6,
+      "620": 0.6
+    },
+    "result": 57000000000
+  },
+  "result": 57000000000,
+  "statusCode": 200
+}
+```
```
thodges-gh, line 98: These docs look to be carried over from EthGasStation as well and need to be updated for Blockstream.

thodges-gh (changes requested):

blockchain.com/src/endpoint/height.ts
```diff
@@ -0,0 +1,27 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
+import { Config, ExecuteWithConfig } from '@chainlink/types'
+
+export const Name = 'height'
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+  console.log('EXECUTING!')
```
RodrigoAD, line 7: You forgot to remove this

blockchain.com/src/endpoint/difficulty.ts
```diff
@@ -0,0 +1,26 @@
...
+    ...config.api,
+    baseURL: config.api.baseURL,
+    url: 'q/getdifficulty',
+  }
+
+  const response = await Requester.request(reqConfig)
+  const result = response.data
+
+  return Requester.success(jobRunID, {
+    data: config.verbose ? { ...response.data, result } : { result },
```
RodrigoAD, line 22: Double check the verbose output, `...response.data` and `result` I think are the same

blockchain.com/src/endpoint/height.ts
```diff
@@ -0,0 +1,27 @@
...
+    ...config.api,
+    baseURL: config.api.baseURL,
+    url: 'q/getblockcount',
+  }
+
+  const response = await Requester.request(reqConfig)
+  const result = response.data
+
+  return Requester.success(jobRunID, {
+    data: config.verbose ? { ...response.data, result } : { result },
```
RodrigoAD, line 23: Same as before here

blockstream/README.md
```diff
@@ -0,0 +1,98 @@
+# Chainlink External Adapter for blockstream
+
+### Environment Variables
+
+| Required? |  Name   | Description | Options | Defaults to |
+| :-------: | :-----: | :---------: | :-----: | :---------: |
+|         | API_KEY |             |         |             |
```
RodrigoAD, line 7: Is not required?

blockchain.com/src/endpoint/height.ts
justinkaseman, line 7: These are going to gum up the log files since they log with every request. Use `logger.debug` here instead.

blockstream/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@chainlink/blockstream-adapter",
+  "version": "0.0.3",
```
justinkaseman, line 3:
```suggestion
  "version": "0.0.1",
```

blockstream/test/adapter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const requests = [
+      {
+        name: 'empty data',
+        testData: { data: {} },
+      },
+      {
+        name: 'id not supplied',
+        testData: {
+          data: {
+            endpoint: 'ethgasAPI',
```
justinkaseman, line 22: These docs also look to be carried over from EthGasStation and need to be updated for Blockstream.

blockstream/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@chainlink/blockstream-adapter",
+  "version": "0.0.1",
+  "description": "Chainlink blockstream adapter.",
+  "keywords": [
+    "Chainlink",
+    "LINK",
+    "blockchain",
+    "oracle",
+    "ethgasstation"
```
aalu1418, line 10: switch `ethgasstation` to `blockstream`?

blockstream/README.md
```diff
@@ -0,0 +1,180 @@
...
+| :-------: | :-----: | :---------: | :-----: | :---------: |
+|           | API_KEY |             |         |             |
+
+---
+
+### Input Parameters
+
+| Required? |   Name   |     Description     |            Options             | Defaults to |
+| :-------: | :------: | :-----------------: | :----------------------------: | :---------: |
+|           | endpoint | The endpoint to use | difficulty, height             |             |
```
aalu1418, line 15: what's the default endpoint?

blockstream/src/endpoint/height.ts
```diff
@@ -0,0 +1,31 @@
+import { Requester, Validator } from '@chainlink/external-adapter'
```
aalu1418, line 1: this is nearly identical code to the `difficulty.ts` file...would it make sense to combine them into one to make it more maintainable?

blockstream/README.md
aalu1418: Doesn't seem like it? The EA uses this endpoint: https://blockstream.info/api/blocks which is publicly accessible

thodges-gh: Yeah it's not required.

blockstream/README.md
justinkaseman: Gotchu fam
```suggestion
|           | endpoint | The endpoint to use | difficulty, height             |        difficulty     |
```

blockstream/package.json
justinkaseman:
```suggestion
    "blockstream"
```

blockstream/README.md
thodges-gh: Ah, needs to be required. Will update.

blockstream/src/endpoint/height.ts
thodges-gh: I think the general idea is to make them separate to make them maintainable in case the API changes.

justinkaseman: Yes, good intuition. This is exactly our newest pattern. Will write up a little explanation in a sec.

justinkaseman: Okay so the new pattern we are trying to run with is:

1. The blockstream External Adapter is a source adapter. It handles READs to go get data from the blockstream API. It is essentially a glorified proxy, with some fancy middleware for caching data. Being a proxy to hit an API we want to keep it as close to the actual API as possible. So we have `endpoints` for the actual API endpoints: `blocks` in this case. By using the `field` parameter we can control which piece of data gets returned from a call to this endpoint.

2. We have these things called an "endpoint" which come in as a parameter from the Chainlink node. These were poorly named and should have been called something like "dataPoint". What it really represents is saying "I want this piece of data for what blockstream says the BTC block difficulty is". The blockstream adapter contains the API specific implementation of how to get that data from blockstream's API.

 In execution this all ends up being just a giant function selector - given some config data go get this data point.

PR #418, opened 2021-04-06
Migrate paypal EA to monorepo

aalu1418:
Migrate paypal adapter from https://github.com/smartcontractkit/paypal-adapter into the EA monorepo

- [x] Implement `sendPayout`
- [x] Implement `getPayout`
- [x] Testing/debugging
- [x] Test cases
- [x] Documentation updates


emmick4 (approved):
👍  looks good to me!

packages/sources/paypal/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Environment Variables
+
+| Required? |      Name       |                                       Description                                        |      Options      | Defaults to |
+| :-------: | :-------------: | :--------------------------------------------------------------------------------------: | :---------------: | :---------: |
+|    ✅     |   `CLIENT_ID`   | [API access credential](https://developer.paypal.com/docs/api/overview/#get-credentials) |                   |             |
+|    ✅     | `CLIENT_SECRET` | [API access credential](https://developer.paypal.com/docs/api/overview/#get-credentials) |                   |             |
+|           |     `MODE`      |                 Parameter for determining a live or sandbox environment                  | `sandbox`, `live` |  `sandbox`  |
```
justinkaseman, line 14: Maybe this should be an input param rather than an env var. If I was a user seems like the flow would always be to try sandbox and then switch to live.

packages/sources/paypal/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Config, ExecuteWithConfig, ExecuteFactory } from '@chainlink/types'
+import { makeConfig, DEFAULT_ENDPOINT } from './config'
+import { sendPayout, getPayout } from './endpoint'
+
+const inputParams = {
+  endpoint: false,
+}
+
+const paramOptions = {
+  endpoint: ['getpayout', 'sendpayout'],
```
justinkaseman, line 11: Maybe we should call these something a little bit more generic. 

See [this comment](https://github.com/smartcontractkit/external-adapters-js/pull/417#discussion_r609118508) for a bit more context.

We haven't fully fleshed out this pattern, but one path forward I could see is:
```suggestion
  endpoint: [getpayout.NAME, sendpayout,NAME, 'read', 'write'],
```
We still want to use allow `getpayout` to stay close to the API, but we also want to be able to use a unified API endpoint if we envision more adapters getting made with similar functionality (think: Cash App adapter, Venmo adapter). Use case: I'm an artist and I want a smart contract that will send an NFT once paid across one of my various money apps.

Rodrigo & Evangelos have examples where he put both of these endpoint names in the endpoints `NAME` itself (as an Array).

I think we should move toward keeping this adapter un-aware of the unified `endpoint` and instead map them in another package or middleware -- but this is out of scope for just getting this in. Just something to have in your mind.

justinkaseman:
Awesome - very high quality work. 

LGTM. Left two comments that are more future proofing "nice to haves".

packages/sources/paypal/README.md
aalu1418: Hmmmm interesting suggestions. Here's what I was thinking:
* I like it defaulting to `sandbox` because it means the user or EA operator has to make a very clear choice to send a 'live' transaction
* In terms moving the `mode` parameter to an input param, i don't think it would really work to flip between the two environments because the `CLIENT_ID` and `CLIENT_SECRET` are specific to one environment or the other (based on my understanding).

justinkaseman: Ah, shoot. Alright if that is the case then LGTM.

PR #425, opened 2021-04-08
add support for RAI to kaiko, nomics

nzleet:


external-adapter/src/overrides/presetSymbols.json
```diff
@@ -13,13 +13,15 @@
...
+    "MDX": "MDX2",
+    "RAI": "RAI3"
   },
   "coinpaprika": {
     "GRT": "grt-the-graph"
   },
   "kaiko": {
-    "uni": "uniswap"
+    "uni": "uniswap",
+    "RAI": "RAIR"
```
nzleet, line 24: This may need to be lowercase? not sure

thodges-gh: It shouldn't matter, the adapter will make it all lower case anyway.

thodges-gh (approved):

PR #433, opened 2021-04-12
Websocket adapter handler

RodrigoAD:
- Adapters can expose a websocket handler so the WS middleware can handle the connection, subscribe, filter and parse incoming messages from the stream.
- A simple hack is made to pipe the incoming ws information into the cache

The following adapters support websockets:
  - [x] Coinbase
  - [x] Amberdata
  - [x] Trandingeconomic (Needs to be tested)
  - [x] Coinapi (Needs to be tested)
  - [x] Cryptocompare
 
 Other
- [ ] Document WS Handler Interface in Code
- [x] Default WS Connection details included in `Config`
- [ ] Handle error messages
- [x] Handle unsubscriptions. Coming from the WS server or the client


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -33,6 +38,8 @@ import {
   ws_subscription_total,
 } from './metrics'
 
+const debug = (message: any, withParams = false) => tap((value: any) => withParams ? console.log(message, value) : console.log(message))
```
HenryNguyen5, line 41: Should we be using the app provided logger rather than console log?

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -106,6 +115,7 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
                 merge(
                   action$.pipe(
                     filter(unsubscribe.match),
+                    debug('We never enter here...'),
```
HenryNguyen5, line 118: 😭

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -121,8 +131,40 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
         }),
       )
 
+      const messages$ = action$.pipe(
```
HenryNguyen5, line 134: I would put this functionality into the `multiplex` pipeline above, since its tightly bound to that anyway when the `messageReceived` function is called

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -121,8 +131,40 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
...
+        filter(messageReceived.match),
+        mergeMap((action) => 
+          of(action).pipe(
+            filter((action) => !wsHandler.filter(action.payload.message)),
+            switchMap(async (action) => {
+              const response = wsHandler.parse(action.payload.message)
+              const execute: Execute = () => { return Promise.resolve(wsHandler.toAdapterResponse(response)) }
```
HenryNguyen5, line 141:
```suggestion
              const execute: Execute = () => Promise.resolve(wsHandler.toAdapterResponse(response))
```

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -121,8 +131,40 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
         }),
       )
 
+      const messages$ = action$.pipe(
+        filter(messageReceived.match),
+        mergeMap((action) => 
+          of(action).pipe(
+            filter((action) => !wsHandler.filter(action.payload.message)),
+            switchMap(async (action) => {
```
HenryNguyen5, line 139: What's the reasoning for a `switchMap` here? Do we want to drop the current caching action if a new message comes in before it finishes? If we don't, we should use `mergeMap` instead

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -11,18 +11,20 @@ export * as reducer from './reducer'
...
 
 // TODO: WebSockets
-export const withWebSockets = (store: Store<RootState>): Middleware => async (execute) => async (
+export const withWebSockets = (store: Store<RootState>) => (wsHandler?: WSSubscriptionHandler): Middleware => async (execute) => async (
   input: AdapterRequest,
 ) => {
+  if (!wsHandler) return await execute(input)
   // TODO: Warmer
-  const { connect, subscribe, unsubscribe, disconnect } = actions
+  // const { connect, subscribe, unsubscribe, disconnect } = actions
```
HenryNguyen5, line 19: We should clean up the comments in this file

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,18 @@ declare module '@chainlink/types' {
     warning?: string
   }
 
+  export type WSSubscriptionHandler = {
+    connection: {
+      url: string | any
```
HenryNguyen5, line 107: What other type can a url be?

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,18 @@ declare module '@chainlink/types' {
...
+      protocol?: any
```
HenryNguyen5, line 108: We should clean up the typing in this file, LMK if you need help doing so!

packages/core/types/@chainlink/index.d.ts
```diff
@@ -102,6 +102,18 @@ declare module '@chainlink/types' {
     warning?: string
   }
 
+  export type WSSubscriptionHandler = {
```
HenryNguyen5, line 105: Its TS convention to use `interface` over `type` when possible

packages/sources/amberdata/src/config.ts
```diff
@@ -21,3 +24,36 @@ export const makeConfig = (prefix = ''): Config => {
...
+
+export const makeWSHandler = (): WSSubscriptionHandler => {
+  const config = makeConfig()
+  return {
+    connection: {
+      url: config.api.baseWsURL || DEFAULT_WS_API_ENDPOINT,
+      protocol: { headers: { ...config.api.headers }}
+    },
+    subscribe: (input) => {
+      // TODO: What to do with different enpoints? Should every endpoint have its own WS Handler?
```
HenryNguyen5, line 36: I believe that's what the multiplexing behaviour does for us for free, every endpoint gets its own subscription, which is multiplexed onto the single connection we have to the api provider

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -11,18 +11,20 @@ export * as reducer from './reducer'
 export * as epics from './epics'
 
 // TODO: WebSockets
-export const withWebSockets = (store: Store<RootState>): Middleware => async (execute) => async (
+export const withWebSockets = (store: Store<RootState>) => (wsHandler?: WSSubscriptionHandler): Middleware => async (execute) => async (
```
HenryNguyen5, line 14: Any reason for currying here? Seems like we could just use a 2 parameter function instead

HenryNguyen5 (changes requested):
Looking great! Just a few points of feedback, I ran through a review for the base `feature/websockets` too, maybe take a quick look there since there will be relevant feedback in that PR too.

packages/core/bootstrap/src/lib/ws/epics.ts
RodrigoAD: This was mostly for this PR, I don't think it will stay after is Ready to merge

packages/core/types/@chainlink/index.d.ts
RodrigoAD: Totally, only string

packages/core/types/@chainlink/index.d.ts
RodrigoAD: We should yep, I'll open an issue

packages/sources/amberdata/src/config.ts
RodrigoAD: This was more about the subscription message necessary for each endpoint. Probably `price` and `marketCap` would have different subscription messages

packages/core/bootstrap/src/lib/ws/index.ts
RodrigoAD: No specific reason yet, but I feel is good to have this flexibility. Makes the middleware composable and more testable.

HenryNguyen5: Sure, probably wont matter too much as soon as we get DI setup anyway

PR #436, opened 2021-04-12
Fixing metrics throwing error for non-pricefeed adapters

### LINKED ISSUE #421: Non-price feed error throws an error when queried due to metrics checking for price feed parameters
Non-blocking error is thrown each time a non-price feed adapter is queried.
```
{"error":{"jobRunID":"1","status":"errored","statusCode":400,"name":"AdapterError"},"level":"error","message":"Error validating input.","instanceId":"bd50ff51-dcc2-459b-924c-41a7e8f8783a","timestamp":"2021-04-07T14:20:40.029Z"}
{"message":"Unable to validate feed name","level":"debug","instanceId":"bd50ff51-dcc2-459b-924c-41a7e8f8783a","timestamp":"2021-04-07T14:20:40.029Z"}
```

The parameters are validated as part of the middleware/metrics system in the /bootstrap/src/lib/metrics/util.ts file and runs the Validator which is throwing the first line of the error above.
```
/**
 * Get feed id name based on input params
 * @param input The adapter input request
 * @returns {string}
 */
export const getFeedId = (input: AdapterRequest): string => {
  const commonFeedParams = {
    base: ['base', 'from', 'coin', 'symbol', 'asset'],
    quote: ['quote', 'to', 'convert'],
  }
  try {
    const validator = new Validator(input, commonFeedParams)
    if (validator.error) {
      logger.debug('Unable to validate feed name')
      return JSON.stringify(input)
    }
    return `${validator.validated.data.base.toUpperCase()}/${validator.validated.data.quote.toUpperCase()}`
  } catch (e) {
    return JSON.stringify(input)
  }
}
```

Can be confusing to operators to see an error when running non-price feed EAs.

aalu1418:
also included small fix to paypal adapter throwing errors

closes #421

### DIFF
--- a/packages/core/bootstrap/src/lib/metrics/util.ts
+++ b/packages/core/bootstrap/src/lib/metrics/util.ts
@@ -39,14 +39,20 @@ export const getFeedId = (input: AdapterRequest): string => {
     base: ['base', 'from', 'coin', 'symbol', 'asset'],
     quote: ['quote', 'to', 'convert'],
   }
-  try {
+
+  // check if string is within array
+  function includesCheck(param: string) {
+    return Object.keys(input.data).includes(param)
+  }
+
+  // run through validdator if input.data object has keys that match potential base and quote parameters
+  if (commonFeedParams.base.some(includesCheck) && commonFeedParams.quote.some(includesCheck)) {
     const validator = new Validator(input, commonFeedParams)
     if (validator.error) {
       logger.debug('Unable to validate feed name')
       return JSON.stringify(input)
     }
     return `${validator.validated.data.base.toUpperCase()}/${validator.validated.data.quote.toUpperCase()}`
-  } catch (e) {
-    return JSON.stringify(input)
   }
+  return JSON.stringify(input)
 }
--- a/packages/sources/paypal/src/endpoint/getPayout.ts
+++ b/packages/sources/paypal/src/endpoint/getPayout.ts
@@ -1,4 +1,4 @@
-import { Validator, AdapterError } from '@chainlink/ea-bootstrap'
+import { Requester, Validator, AdapterError } from '@chainlink/ea-bootstrap'
 import { Config, ExecuteWithConfig } from '@chainlink/types'
 import * as paypal from '@paypal/payouts-sdk'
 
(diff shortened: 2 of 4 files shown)


packages/core/bootstrap/src/lib/metrics/util.ts
```diff
@@ -39,14 +39,20 @@ export const getFeedId = (input: AdapterRequest): string => {
...
+    return Object.keys(input.data).includes(param)
+  }
+
+  // run through validdator if input.data object has keys that match potential base and quote parameters
+  if (commonFeedParams.base.some(includesCheck) && commonFeedParams.quote.some(includesCheck)) {
     const validator = new Validator(input, commonFeedParams)
     if (validator.error) {
       logger.debug('Unable to validate feed name')
       return JSON.stringify(input)
     }
```
aalu1418, line 54: Is this line still necessary if the parameters are checked to be present in line 49? (will it ever error?)

RodrigoAD: I think it won't. Actually, `validator.error` is taking care of the error, so we can remove it 👍

packages/core/bootstrap/src/lib/metrics/util.ts
```diff
@@ -39,14 +39,20 @@ export const getFeedId = (input: AdapterRequest): string => {
...
   }
-  try {
+
+  // check if string is within array
+  function includesCheck(param: string) {
+    return Object.keys(input.data).includes(param)
+  }
+
+  // run through validdator if input.data object has keys that match potential base and quote parameters
+  if (commonFeedParams.base.some(includesCheck) && commonFeedParams.quote.some(includesCheck)) {
```
RodrigoAD, line 49: Perfect

packages/sources/paypal/src/endpoint/getPayout.ts
```diff
@@ -41,7 +41,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   try {
     response = await config.api.client.execute(paypal_req)
   } catch (e) {
-    throw Requester.errored(jobRunID, e)
+    e.message = JSON.stringify({ statusCode: e.statusCode, ...JSON.parse(e.message) })
```
RodrigoAD, line 44: Why is this? `e.statusCode` could go into `AdapterError`, no?

packages/sources/paypal/src/endpoint/getPayout.ts
```diff
@@ -41,7 +41,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
...
+    throw new AdapterError({
+      jobRunID,
+      message: e.message,
+      statusCode: 500,
+    })
```
RodrigoAD, line 49: - It would be nice to use `Requester.success` when the call succeeds, so we make sure everything stays uniform.
- About the error, why not forward the same error status code from our adapter? At the end, the paypal client is just making an HTTP request as well
```suggestion
try {
    const response = await config.api.client.execute(paypal_req)
    return Requester.success(jobRunID, { ...response, status: response.statusCode })
  } catch (e) {
    throw Requester.errored(jobRunID, e, e.statusCode)
  }
```

aalu1418: Good points! I will make some changes to use the standard `Requester.success` and `Requester.errored`

My original intention was to standardize the error code that the adapter returned since i saw different errors ranging from 400, 404, and 429 which made it hard to test. But I think I might need to rethink how the test cases are written.

Also for the `success` return, the response structure doesn't match the expected input structure so that's why it's custom. But I think you're right in that it would be useful to standardize the output.

RodrigoAD (approved):
Nice! Thanks!

PR #442, opened 2021-04-14
Adapters for Intrinio, Unibit, Twelvedata

aalu1418:
Adapters for:
- [x] Intrinio - ws (`iex` only) - live feed data
- [x] Unibit - rest - previous EOD data
- [x] Twelvedata - ~ws (to be completed in a different PR)~ rest - previous EOD data


intrinio/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@chainlink/intrinio-adapter",
+  "version": "0.0.3",
```
justinkaseman, line 3:
```suggestion
  "version": "0.0.1",
```

twelvedata/src/endpoint/price.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+  const validator = new Validator(request, customParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const symbol = validator.overrideSymbol(AdapterName).toUpperCase()
+
+  const url = `price?symbol=${symbol}&apikey=${config.apiKey}`
```
justinkaseman, line 20: You could use `params` on options to pass these in. Might look a little cleaner.

So in this case it would be:
```
const url = 'price'

const params = {
symbol,
apikey: config.apiKey
}

const options = {
...config.api,
params,
url,
}
```

justinkaseman (approved):
Great job! LGTM :+1:

PR #448, opened 2021-04-21
Chore/intrinio tiingo mods

aalu1418:
Modifications to intrinio:
* changing the quote type from `bid` & `ask` to `last` price only (`bid` and `ask` are only present during trading hours and do not reflect actual trade prices)
* added additional accepted parameters in the request body

Modifications to tiingo
* default endpoint is `eod` (other submitted endpoints will follow `eod`)
* additional accepted parameters in the request body

### DIFF
--- a/intrinio/src/adapter.ts
+++ b/intrinio/src/adapter.ts
@@ -47,7 +47,7 @@ export const startService = (config: Config): void => {
 }
 
 const customParams = {
-  base: ['base', 'from', 'coin', 'market', 'asset'],
+  base: ['base', 'from', 'coin', 'asset'],
 }
 
 export const execute = async (input: AdapterRequest, config: Config) => {
--- a/intrinio/src/adapter.ts
+++ b/intrinio/src/adapter.ts
@@ -31,10 +31,10 @@ const subscribe = (assets: string[], config: Config) => {
         break
       case PROVIDER_OPTIONS[0]: //iex
       default:
-        if (quote.type == 'last') return
+        if (quote.type != 'last') return //only using the `last` price (saved in both bid & ask slot so average works out to be equal)
         prices[quote.ticker] = {
-          ...prices[quote.ticker],
-          [quote.type]: quote.price,
+          bid: quote.price,
+          ask: quote.price,
         }
         break
     }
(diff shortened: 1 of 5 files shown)


intrinio/src/adapter.ts
```diff
@@ -47,7 +47,7 @@ export const startService = (config: Config): void => {
 }
 
 const customParams = {
-  base: ['base', 'from', 'asset'],
+  base: ['base', 'from', 'coin', 'market', 'asset'],
```
thodges-gh, line 50: 'market' is not relevant here. When supplying params for a pair, the keys base, from, and coin are used together. Then quote, to, and market are used for the denomination.

tiingo/src/endpoint/eod.ts
```diff
@@ -4,7 +4,7 @@ import { ExecuteWithConfig, Config } from '@chainlink/types'
 export const NAME = 'eod'
 
 const customParams = {
-  ticker: true,
+  ticker: ['ticker', 'base', 'from', 'coin', 'market'],
```
thodges-gh, line 7: Same thing here.

thodges-gh (changes requested):

tiingo/src/adapter.ts
```diff
@@ -13,19 +13,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
...
-      return await eod.execute(request, config)
-    }
+    case eod.NAME:
     default: {
-      throw new AdapterError({
-        jobRunID,
-        message: `Endpoint ${endpoint} not supported.`,
-        statusCode: 400,
-      })
+      return await eod.execute(request, config)
```
justinkaseman, line 21: Just out of curiousity, was a different endpoint name being requested that caused this to get a default added?

aalu1418: I believe they were using a single job spec across various adapters. Some adapters used `endpoint: stock` which caused the tiingo EA to throw an error. So now it defaults to the only endpoint no matter what the `endpoint` parameter is set to

justinkaseman (approved):
:+1:

thodges-gh (approved):

PR #451, opened 2021-04-22
Composite aggregator adapter

aalu1418:
Can use any underlying adapters to create a data array which is then aggregated via the reduce EA

Main changes:
* New composite adapter: aggregator
* Bug fixes to the reduce adapter (prevent removing 0 from list, `new Decimal` errors if param is undefined)

### DIFF
--- a/reduce/src/adapter.ts
+++ b/reduce/src/adapter.ts
@@ -41,7 +41,7 @@ export const execute: Execute = async (request) => {
   const _get = (val: unknown): number => objectPath.get(val, path)
 
   // Filter undesired values
-  inputData = inputData.filter((val) => !!_get(val))
+  inputData = inputData.filter((val) => !isNaN(_get(val)))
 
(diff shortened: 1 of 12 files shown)


reduce/src/adapter.ts
```diff
@@ -41,7 +41,7 @@ export const execute: Execute = async (request) => {
   const _get = (val: unknown): number => objectPath.get(val, path)
 
   // Filter undesired values
-  inputData = inputData.filter((val) => !!_get(val))
+  inputData = inputData.filter((val) => !isNaN(_get(val)))
```
aalu1418, line 44: This was an interesting change to make. I understand that using `!!` can eliminate `0` value outliers, and for price feeds I think that makes sense. Changed it to `!isNaN` since it affected taking an average if the array included `0`, and that other non price-feed use cases may require including 0 in the average.

PR #453, opened 2021-04-23
Higher decimal.js default precision

RodrigoAD:
By default `decimal.js` uses a precision of 20. As we could lose some precision with really large numbers, updating it to a higher precision

### DIFF
--- a/packages/composites/token-allocation/src/adapter.ts
+++ b/packages/composites/token-allocation/src/adapter.ts
@@ -6,6 +6,8 @@ import { AdapterError, Requester, Validator } from '@chainlink/ea-bootstrap'
 import { BigNumber } from 'ethers'
 import { getPriceProvider } from './dataProvider'
 
+Decimal.set({ precision: 100 })
+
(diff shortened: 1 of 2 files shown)


packages/composites/token-allocation/src/adapter.ts
```diff
@@ -6,6 +6,8 @@ import { AdapterError, Requester, Validator } from '@chainlink/ea-bootstrap'
 import { BigNumber } from 'ethers'
```
justinkaseman, line 6: Could we get rid of this now and just use Decimal in its place?

RodrigoAD: We agreed some time ago, BigNumber should be for integer operations, and Decimal for floating

justinkaseman: That works, pushing a README update so that we have this pattern documented, otherwise it will get lost.

justinkaseman (approved):
:+1:

PR #460, opened 2021-04-27
Batch requests

justinkaseman:
### Description
Brings back batched price requests & marketcap requests for the following providers that are used in `Token Allocation` composite adapter:

- [x] CoinMarketCap
- [x] Coingecko
- [x] CryptoCompare
- [x] Nomics

Adds batching capabilities to:
- [x] CoinAPI (only supports price, not marketcap)

This functionality is only added for `price` endpoints using the `base` input parameter.

To allow this, the `override` functionality now handles array inputs.

Also brings override functionality into `nomics`.


packages/sources/coinapi/src/endpoint/assets.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const options = {
+    ...config.api,
+    url,
+    params: { ...config.api.params, ...params },
+  }
+
+  const response = await Requester.request<AssetsResponse>(options, customError)
+
+  if (Array.isArray(symbol)) {
```
HenryNguyen5, line 54: This should probably be a function called `handleBatchedResponse` or something like that, otherwise there's no context to others reading the source code of whats happening

packages/sources/coinapi/test/integration/assets.test.ts
```diff
@@ -0,0 +1,93 @@
...
+import { assertError, assertSuccess } from '@chainlink/ea-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../../src/adapter'
+
+describe('execute', () => {
+  const jobID = '1'
+  const execute = makeExecute()
+  process.env.API_KEY = process.env.API_KEY ?? 'test_api_key'
+
+  describe('successful calls @integration', () => {
```
HenryNguyen5, line 11: No need for `@integration`, integration tests are filtered by path now

packages/sources/coinapi/test/integration/assets.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        name: 'from/to',
+        testData: { id: jobID, data: { from: 'ETH', endpoint: 'assets' } },
+      },
+      {
+        name: 'coin/market',
+        testData: { id: jobID, data: { coin: 'ETH', endpoint: 'assets' } },
+      },
+    ]
+
+    requests.forEach((req) => {
```
HenryNguyen5, line 27: I would use https://jestjs.io/docs/api#describeeachtablename-fn-timeout since jest can handle these natively

packages/sources/coingecko/src/endpoint/price.ts
```diff
@@ -79,8 +65,23 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   }
 
   const response = await Requester.request(options, customError)
+
+  if (Array.isArray(symbol)) {
```
HenryNguyen5, line 69: I would move this out to a named function

packages/sources/coingecko/src/endpoint/price.ts
```diff
@@ -21,38 +22,23 @@ const customParams = {
...
-  return coin.id.toLowerCase()
-}
-
 export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   const validator = new Validator(request, customParams)
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
-  const symbol = validator.overrideSymbol(AdapterName)
+  const symbol = validator.overrideSymbol(AdapterName) as string
```
HenryNguyen5, line 30: Shouldn't `overrideSymbol` always return a string?

HenryNguyen5 (changes requested):
Looking pretty good!

The feedback i requested on the first few adapters also applies to the rest of them

packages/composites/token-allocation/src/dataProvider.ts
```diff
@@ -2,11 +2,40 @@ import { Requester } from '@chainlink/ea-bootstrap'
...
-  withMarketCap = false,
-): Promise<ResponsePayload> => {
+const batchingSupport: { [name: string]: boolean } = {
+  COINGECKO: true,
+  COINMARKETCAP: true,
+  CRYPTOCOMPARE: true,
+  NOMICS: true,
+}
+
+export const getPriceProvider = (
```
HenryNguyen5, line 12: We should probably have a unit test for this no?

HenryNguyen5, line 12: How do you feel about documenting this function, since now it has grown in complexity and handles multiple use cases

HenryNguyen5 (changes requested):

packages/composites/token-allocation/src/dataProvider.ts
```diff
@@ -2,11 +2,40 @@ import { Requester } from '@chainlink/ea-bootstrap'
...
+  jobRunID: string,
+  source: string,
+  apiConfig: RequestConfig,
+) => async (symbols: string[], quote: string, withMarketCap = false): Promise<ResponsePayload> => {
+  if (
+    batchingSupport[source.toUpperCase()] ||
+    (source.toUpperCase() === 'COINAPI' && quote.toUpperCase() === 'USD') // Special case for CoinAPI which only can batch USD quotes
```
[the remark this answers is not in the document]
justinkaseman: Since `COINAPI` isn't included in the variable where I am holding which EAs support batching (`batchingSupport`) this lets it pass through if the EA when  CoinAPI and the quote is USD, otherwise it doesn't batch.

Maybe it would be more clear to include `COINAPI` in batchingSupport and bail if quote isn't USD.

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -93,6 +102,27 @@ export class Requester {
...
+   * @param result a single result value
+   * @param results multiple results for internal batch requests
+   */
+
+  static withResult<T>(
+    response: AxiosResponse<T>,
+    result?: number,
+    results?: { [fsym: string]: number },
+  ): AxiosResponseWithResult<T> | AxiosResponseWithResultResponse<T> {
+    const isObj = Object.prototype.toString.call(response.data) === '[object Object]'
```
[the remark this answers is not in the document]
justinkaseman: Here I'm checking if the response data is a true object.

Many things in JS are actually objects. Like:
```
typeof ['hello wurld'] === 'object // true
```

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -93,6 +102,27 @@ export class Requester {
...
+    const responseWithResult = isObj
+      ? (response as AxiosResponseWithResult<T>)
+      : ({ ...response, data: { payload: response.data } } as AxiosResponseWithResultResponse<T>)
+    if (result) responseWithResult.data.result = result
```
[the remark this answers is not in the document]
justinkaseman: Oooh fancy, how have I never seen this before

packages/sources/coingecko/src/endpoint/price.ts
justinkaseman: Good suggestion. Changing.

packages/sources/coingecko/src/endpoint/price.ts
justinkaseman: It can also return an array now, because for batched `symbol` inputs they come in as an array. So we may override each one.

packages/core/bootstrap/src/lib/external-adapter/requester.ts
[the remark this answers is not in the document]
HenryNguyen5: Ah i see what you want now, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#real-world_usage

packages/sources/coinapi/src/endpoint/assets.ts
```diff
@@ -0,0 +1,64 @@
...
+  data_trade_end: string
+  data_symbols_count: number
+  volume_1hrs_usd: number
+  volume_1day_usd: number
+  volume_1mth_usd: number
+  price_usd: number
+  id_icon: string
+}[]
+
+const customError = (data: AssetsResponse) => data.length === 0
```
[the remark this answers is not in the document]
justinkaseman: Don't think this actually happens. Removing.

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -228,3 +228,41 @@ export const permutator = (options: string[], delimiter?: string): string[] | st
...
+ *
+ * @url
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#real-world_usage
+ *
+ * @param obj The value to type check
+ * @param fullClass (Optional) Whether to use polyfill for checking null
+ *
+ * @returns String describing type of obj
+ */
+export function type(obj: any, fullClass?: boolean): string {
```
HenryNguyen5, line 244: type is a reserved keyword in TS-land, maybe `narrowObjectType` or `specificObjectType` would be better

packages/sources/coinapi/test/integration/assets.test.ts
```diff
@@ -8,86 +8,70 @@ describe('execute', () => {
...
-    ]
-
-    requests.forEach((req) => {
-      it(`${req.name}`, async () => {
-        const data = await execute(req.testData as AdapterRequest)
-        assertSuccess({ expected: 200, actual: data.statusCode }, data, jobID)
-        expect(data.result).toBeGreaterThan(0)
-        expect(data.data.result).toBeGreaterThan(0)
-      })
+  describe.each([
```
HenryNguyen5, line 11: Is this actually supported by jest? seems like its only supported for `it.each`

packages/core/bootstrap/src/lib/util.ts
justinkaseman: Changed it to `deepType` because it handles every type, not just objects.

packages/sources/coinapi/test/integration/assets.test.ts
justinkaseman: https://jestjs.io/docs/api#describeeachtablename-fn-timeout

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -105,15 +104,15 @@ export class Requester {
...
+    const isObj = deepType(response.data) === 'object'
     const output = isObj
       ? (response as AxiosResponseWithLiftedResult<T>)
       : ({
           ...response,
           data: { payload: response.data },
         } as AxiosResponseWithPayloadAndLiftedResult<T>)
-    output.data.result &&= result
-    output.data.results &&= results
+    if (result) output.data.result = result
```
justinkaseman, line 114: The &&= syntax wasn't working for me. It checks the left side `output.data.result`, when it needs to be checking `result`

packages/core/bootstrap/test/unit/requester.test.ts
```diff
@@ -204,6 +206,31 @@ describe('Requester', () => {
     })
   })
 
+  describe('Requester.withResult', () => {
```
justinkaseman, line 209: Some new unit tests

PR #461, opened 2021-04-27
Tradingeconomics WS refactor

aalu1418:
Tradingeconomics
- [x] Initial code
- [x] Testing - WS works (but tested outside of market hours)
- [x] Documentation

### DIFF
--- a/packages/sources/tradingeconomics/src/config.ts
+++ b/packages/sources/tradingeconomics/src/config.ts
@@ -66,15 +66,13 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
         return getSubscription(base)
       },
       unsubscribe: () => undefined,
-      subsFromMessage: (message) => getSubscription(message?.s),
+      subsFromMessage: message => getSubscription(message?.s),
       isError: (message: any) => Number(message.TYPE) > 400 && Number(message.TYPE) < 900,
       filter: message => {
         return message.topic && message.topic !== 'keepalive'
       },
-      toResponse: (wsResponse: any): AdapterResponse => {
-        wsResponse = { ...wsResponse, result: wsResponse?.price }
-        return Requester.success(undefined, { data: wsResponse })
-      },
+      toResponse: (wsResponse: any): AdapterResponse =>
+        Requester.success(undefined, { data: { result: wsResponse?.price } }),
     }
   }
 }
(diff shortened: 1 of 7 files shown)


packages/sources/tradingeconomics/src/config.ts
```diff
@@ -11,27 +13,65 @@ import { util } from '@chainlink/ea-bootstrap'
...
+      },
+      subscribe: input => {
+        const validator = new Validator(input, customParams)
+        if (validator.error) {
+          return
+        }
+        const base = validator.overrideSymbol(NAME).toUpperCase()
+        return getSubscription(base)
+      },
+      unsubscribe: () => '',
```
krebernisak, line 68: Does the whole connection need to be torn down to unsubscribe? I don't understand these API providers and why they don't handle unsubscribe messages.

krebernisak (commented):
Looks good, but please rebase on top of the latest changes from @RodrigoAD, there will be slight changes to the handler.

packages/sources/tradingeconomics/src/config.ts
aalu1418: Yeah...I couldn't find an unsubscribe command in their SDK:
https://github.com/tradingeconomics/tradingeconomics/tree/master/nodejs/stream-nodejs

or in the one that Jonas made either :(

krebernisak: Ok, then make this `undefined` when Rodrigo makes the underlying changes.

packages/sources/tradingeconomics/src/config.ts
```diff
@@ -11,27 +13,68 @@ import { util } from '@chainlink/ea-bootstrap'
...
+      },
+      unsubscribe: () => undefined,
+      subsFromMessage: (message) => getSubscription(message?.s),
+      isError: (message: any) => Number(message.TYPE) > 400 && Number(message.TYPE) < 900,
+      filter: message => {
+        return message.topic && message.topic !== 'keepalive'
+      },
+      toResponse: (wsResponse: any): AdapterResponse => {
+        wsResponse = { ...wsResponse, result: wsResponse?.price }
+        return Requester.success(undefined, { data: wsResponse })
```
RodrigoAD, line 76: Same as `Intrinio` PR

RodrigoAD:
LGTM. Feel free to merge if acceptance testing is ok

PR #466, opened 2021-04-28
Intrinio WS refactor

aalu1418:
- [x] Code refactor
- [x] Tested
- [x] Testcases + documentation

Includes a new feature to websocket bootstrap: an `init` function that can be called for async calls (needed for intrinio)

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/index.ts
+++ b/packages/core/bootstrap/src/lib/ws/index.ts
@@ -17,6 +17,9 @@ export const withWebSockets = (store: Store<RootState>, makeWsHandler?: MakeWSHa
   // TODO: Warmer
   const wsHandler = makeWsHandler()
   const wsConfig = getWSConfig()
+  if (wsHandler.init !== undefined) {
+    await wsHandler.init(wsHandler)
+  }
   store.dispatch(connect({ config: wsConfig, wsHandler }))
 
   const subscriptionMsg = wsHandler.subscribe(input)
--- a/packages/core/bootstrap/src/lib/ws/index.ts
+++ b/packages/core/bootstrap/src/lib/ws/index.ts
@@ -15,11 +15,8 @@ export const withWebSockets = (store: Store<RootState>, makeWsHandler?: MakeWSHa
 ) => {
   if (!makeWsHandler) return await execute(input)
   // TODO: Warmer
-  const wsHandler = makeWsHandler()
+  const wsHandler = await makeWsHandler()
   const wsConfig = getWSConfig()
-  if (wsHandler.init !== undefined) {
-    await wsHandler.init(wsHandler)
-  }
   store.dispatch(connect({ config: wsConfig, wsHandler }))
 
   const subscriptionMsg = wsHandler.subscribe(input)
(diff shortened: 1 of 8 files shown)


packages/sources/intrinio/src/config.ts
```diff
@@ -1,36 +1,65 @@
...
-    symbols: util.getRequiredEnv('SYMBOLS', prefix),
-    provider: provider.toLowerCase(),
+  return () => {
+    const defaultConfig = config || makeConfig()
+
+    const ws = new IntrinioRealtime({
+      api_key: defaultConfig.apiKey,
+      provider: 'iex',
+    })
+
```
RodrigoAD, line 43: Instead of the init method, we could ask for the token here. As this will be called on every request, we'll probably want to return a previous token if it's not expired. If it expired, we fetch it again

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -17,6 +17,9 @@ export const withWebSockets = (store: Store<RootState>) => (makeWsHandler?: Make
   // TODO: Warmer
   const wsHandler = makeWsHandler()
   const wsConfig = getWSConfig()
+  if (wsHandler.init !== undefined) {
```
RodrigoAD, line 20: And instead of this, `const wsHandler = await makeWsHandler()` would actually make the initialization

krebernisak: Yes, I agree. Let's make `type WSHandlerFactory = () => Promise<WSSubscriptionHandler>` for cases like this. cc @RodrigoAD

packages/sources/intrinio/src/config.ts
krebernisak: We should make `makeWSHandler` async, fetch the token, and return the handler.

krebernisak (commented):
Looks good (with a few suggestions), but please rebase on top of the latest changes from @RodrigoAD, there will be slight changes to the handler.

packages/sources/intrinio/src/config.ts
```diff
@@ -1,36 +1,61 @@
...
+        url: ws._makeSocketUrl(),
+      },
+      subscribe: input => ws._makeJoinMessage(getBase(input)),
+      unsubscribe: input => ws._makeLeaveMessage(getBase(input)),
+      subsFromMessage: message => ws._makeJoinMessage(message.payload.ticker),
+      isError: (message: any) => Number(message.TYPE) > 400 && Number(message.TYPE) < 900,
+      filter: message => message.event == 'quote' && message.payload?.type == 'last',
+      toResponse: (wsResponse: any): AdapterResponse => {
+        wsResponse = { ...wsResponse, result: wsResponse?.payload?.price }
+        return Requester.success(undefined, { data: wsResponse })
```
RodrigoAD, line 57: - Better give `'1'` as job id instead of `undefined`
- Is it necessary to return the full WS message?

aalu1418: Does Requester.success default to `'1'` anyway?

I don't think it's necessary since the it's returning 
```
{
  topic: 'iex:securities:AAPL',
  payload: {
    type: 'last',
    timestamp: 1619704191.36749,
    ticker: 'AAPL',
    size: 50,
    price: 135.01
  },
  event: 'quote'
}
```
and we only need the price. Sounds like it would be good to just return the price to reduce the size of the cache?

RodrigoAD: Right, return the price only then

RodrigoAD (approved):
LGTM. Feel free to merge if acceptance testing is ok

PR #474, opened 2021-04-29
fix: Intrinio needs all handlers, not just server

justinkaseman:
### Description
Intrinio EA needs to have handlers to work with AWS Lambda.

It is a WS adapter, so connections need to be cleaned up.

### DIFF
--- a/intrinio/src/adapter.ts
+++ b/intrinio/src/adapter.ts
@@ -39,11 +39,16 @@ const subscribe = (assets: string[], config: Config) => {
         break
     }
   })
+  return client
 }
 
-export const startService = (config: Config): void => {
+export const startService = (config: Config) => {
   const symbols = config.symbols.toUpperCase().split(',')
-  subscribe(symbols, config)
+  return subscribe(symbols, config)
+}
+
+export const stopService = (client: any): void => {
+  client.destroy()
 }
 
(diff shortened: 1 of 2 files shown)


intrinio/src/adapter.ts
```diff
@@ -39,11 +39,16 @@ const subscribe = (assets: string[], config: Config) => {
...
 
-export const startService = (config: Config): void => {
+export const startService = (config: Config) => {
   const symbols = config.symbols.toUpperCase().split(',')
-  subscribe(symbols, config)
+  return subscribe(symbols, config)
+}
+
+export const stopService = (client: any): void => {
+  client.destroy()
```
HenryNguyen5, line 51: Why do we destroy the client after every response?

justinkaseman: It's a hack right now to make a WS adapter work in Lambda.

Otherwise the WS connections aren't cleaned up https://github.com/intrinio/intrinio-realtime-node-sdk

Once EAEE is released this hack will disappear.

I'm realizing I shouldn't have tagged you for review here because if we deploy this using Fargate we won't need this hack. Messaging Josh now.

justinkaseman:
Outdated PR. Closing.

PR #476, opened 2021-04-30
fix: burst connection handling

RodrigoAD:


packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -24,10 +27,23 @@ export const connectionsReducer = createReducer<ConnectionsState>(
...
       state.active[key] = action.payload.config
+      state.connecting[key] = 0
       // Increment num of active connections
       state.total++
     })
 
+    builder.addCase(actions.connect, (state, action) => {
+      // Add connection
+      const { key } = action.payload.config.connectionInfo
+      state.connecting[key] = !isNaN(Number(state.connecting[key])) ? state.connecting[key] + 1 : 0
```
krebernisak, line 38: Should start with 1 not 0:

```suggestion
      state.connecting[key] = !isNaN(Number(state.connecting[key])) ? state.connecting[key] + 1 : 1
```

krebernisak (approved):

PR #485, opened 2021-05-03
fix: unsubscribe due to unresponsive only on active subscriptions

RodrigoAD:


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -237,7 +237,12 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
 
             const timeout$ = of(unsubscribe(action), subscribe(action)).pipe(
               delay(config.subscriptionUnresponsiveTTL),
-              tap(() => logger.info('WS: unsubscribe -> subscribe (unresponsive channel)')),
+              withLatestFrom(state$),
```
RodrigoAD, line 240: Needed the last state

krebernisak (approved):

PR #487, opened 2021-05-04
Feature/coins calls

justinkaseman:
### Description
To further reduce API credits moves `coin list` requests to their own endpoint and runs them through the core framework so that caching, metrics, and rate-limit usage is utilized.

Makes this change to the following EAs that TA composite uses:
- coingecko
- coinpaprika

### DIFF
--- a/packages/sources/coingecko/src/endpoint/coins.ts
+++ b/packages/sources/coingecko/src/endpoint/coins.ts
@@ -3,17 +3,14 @@ import { Config, ExecuteWithConfig } from '@chainlink/types'
 
 export const NAME = 'coins'
 
-export type CoinsResponse = { id: string; symbol: string; name: string }[]
-
-const customError = (data: any) => {
-  if (Object.keys(data).length === 0) return true
-  return false
+export interface CoinsResponse {
+  id: string
+  symbol: string
+  name: string
 }
 
-const customParams = {}
-
 export const execute: ExecuteWithConfig<Config> = async (request, config) => {
-  const validator = new Validator(request, customParams)
+  const validator = new Validator(request)
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
@@ -22,6 +19,6 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     ...config.api,
     url,
   }
-  const response = await Requester.request<CoinsResponse>(options, customError)
+  const response = await Requester.request<CoinsResponse[]>(options)
   return Requester.success(jobRunID, response, true)
 }
--- a/packages/sources/coingecko/src/endpoint/coins.ts
+++ b/packages/sources/coingecko/src/endpoint/coins.ts
@@ -1,0 +1,27 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig } from '@chainlink/types'
+
+export const NAME = 'coins'
+
+export type CoinsResponse = { id: string; symbol: string; name: string }[]
+
+const customError = (data: any) => {
+  if (Object.keys(data).length === 0) return true
+  return false
+}
+
+const customParams = {}
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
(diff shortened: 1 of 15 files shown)


packages/sources/coingecko/src/util.ts
```diff
@@ -1,29 +1,31 @@
...
-}
-
-export const getCoinIds = async (config: Config): Promise<CoinsListResponse> => {
-  const url = '/coins/list'
-  const options = {
-    ...config.api,
-    url,
-  }
-  return (await Requester.request(options, customError)).data as CoinsListResponse
+export const getCoinIds = (id: string): Promise<CoinsResponse> => {
```
justinkaseman, line 5: Just threw these in the 2 EAs that needed them, but we could probably abstract this to be a utility method exposed by `ea-bootstrap` where `execute` and and `options` are passed in as arguments.

packages/sources/coinpaprika/src/util.ts
```diff
@@ -0,0 +1,35 @@
...
+    const execute = makeExecute()
+    const executeWithMiddleware = executeSync(execute)
+    const options = {
+      data: {
+        endpoint: 'coins',
+      },
+      method: 'post',
+      id,
+    }
+    console.log(options)
```
HenryNguyen5, line 16: Should we be using our logger here?

packages/sources/coinpaprika/src/util.ts
```diff
@@ -0,0 +1,35 @@
...
+        resolve(data.data)
+      })
+    } catch (error) {
+      reject(error)
+    }
+  })
+}
+
+export const getSymbolToId = (symbol: string, coinList: CoinsResponse): string => {
+  coinList.sort((a: { rank: number }, b: { rank: number }) => (a.rank > b.rank ? 1 : -1))
```
HenryNguyen5, line 28: The type shouldnt need to be specified

packages/sources/coinpaprika/src/util.ts
```diff
@@ -0,0 +1,35 @@
...
+  const coin = coinList.find(
+    (x: { symbol: string; rank: number }) =>
```
HenryNguyen5, line 30: The type shouldnt need to be specified

packages/sources/coinpaprika/src/util.ts
```diff
@@ -0,0 +1,35 @@
...
+      x.symbol.toLowerCase() === symbol.toLowerCase() && x.rank !== 0,
```
HenryNguyen5, line 31: Should we have a unit test for this?

packages/sources/coinpaprika/src/util.ts
```diff
@@ -0,0 +1,35 @@
...
+    const options = {
+      data: {
+        endpoint: 'coins',
+      },
+      method: 'post',
+      id,
+    }
+    console.log(options)
+    try {
+      executeWithMiddleware(options, (_, data) => {
```
HenryNguyen5, line 18: Should we throw if an error is returned by the callback?

HenryNguyen5 (changes requested):

packages/sources/coingecko/src/endpoint/coins.ts
```diff
@@ -0,0 +1,27 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig } from '@chainlink/types'
+
+export const NAME = 'coins'
+
+export type CoinsResponse = { id: string; symbol: string; name: string }[]
+
+const customError = (data: any) => {
```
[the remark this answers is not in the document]
justinkaseman: Removed this check, copy/pasting is bad.

packages/sources/coinpaprika/src/util.ts
justinkaseman: It will get logged by Requester. Removing this dev console.log.

HenryNguyen5 (dismissed):

HenryNguyen5 (approved):

PR #488, opened 2021-05-04
perf: added metrics at unresponsive subscriptions

RodrigoAD:


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -235,7 +238,7 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
               connectionInfo: { key: connectionKey, url },
             }
 
-            const timeout$ = of(unsubscribe(action), subscribe(action)).pipe(
+            const timeout$ = of(subscriptionError({ ...action, message: 'WS: unsubscribe -> subscribe (unresponsive channel)' }), unsubscribe(action), subscribe(action)).pipe(
```
krebernisak, line 241: Let's not confuse "message" as we call data passing through the channel a message. Either rename as `errorMsg` or `reason/type`, something different.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -316,6 +329,10 @@ export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
           ws_subscription_active.labels(subscriptionLabels(action.payload)).inc()
           logger.info('WS: subscribed', { payload: action.payload })
           break
+        case subscriptionError.type:
+          ws_subscription_errors.labels(subscriptionErrorLabels(action.payload)).inc()
+          logger.info('WS: subscription error', { payload: action.payload })
```
krebernisak, line 334: Let's use `logger.error` here

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -285,14 +291,21 @@ export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
...
         connection_url: payload.connectionInfo.url,
         feed_id: getFeedId({ ...payload.input }),
         subscription_key: getSubsId(payload.subscriptionMsg),
       })
+      const subscriptionErrorLabels = (payload: WSSubscriptionErrorPayload) => ({
+        connection_key: payload.connectionInfo.key,
+        connection_url: payload.connectionInfo.url,
+        feed_id: payload.input && getFeedId({ ...payload.input }),
+        message: payload.reason,
+        subscription_key: payload.subscriptionMsg && getSubsId(payload.subscriptionMsg),
```
krebernisak, line 307: Not sure if this will work correctly in all cases, will it?
Why the event wouldn't send  `payload.input`  or `payload.subscriptionMsg` ?

krebernisak (approved):

packages/core/bootstrap/src/lib/ws/epics.ts
RodrigoAD: [In this case](https://github.com/smartcontractkit/external-adapters-js/pull/488/files#diff-42dc22535608bc02b8332c4f51fbd00a68974be1dabacdc9f044aaffe01f16f1R129) 
For now we can't identify the subscription from that messags

krebernisak: Ok, got it!

But this is going to set it as false, rather do something like this:
```suggestion
        feed_id: payload.input ? getFeedId({ ...payload.input }) : 'N/A',
        message: payload.reason,
        subscription_key: payload.subscriptionMsg ? getSubsId(payload.subscriptionMsg) : 'N/A',
```

krebernisak (approved):

PR #493, opened 2021-05-05
fix: unsubscription metric happens only if the subs was active

RodrigoAD:
- **Fix 1**: Invalid subscriptions could remain on subscribing state forever. Unresponsive timeout will take them in account now.
- **Fix 2**: Subscriptions that never were active were included into the unsubscription metric. `wasEverActive` check is included on the state, getting track of the timeline of the subscription. (Thanks @aalu1418)

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/epics.ts
+++ b/packages/core/bootstrap/src/lib/ws/epics.ts
@@ -1,7 +1,7 @@
 import { Execute } from '@chainlink/types'
 import { AnyAction } from 'redux'
 import { combineEpics, createEpicMiddleware, Epic } from 'redux-observable'
-import { merge, Subject, of, race, Observable } from 'rxjs'
+import { merge, Subject, of, race, Observable, EMPTY } from 'rxjs'
 import {
   catchError,
   delay,
@@ -122,7 +122,7 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
               () => wsHandler.unsubscribe(payload.input),
               (message) => {
                 /**
-                 * If the error in on the subscription, next requests will try to subscribe
+                 * If the error happens on the subscription, it will be on subscribing state and eventually unresponsiveTimeout will take care of it (unsubs/subs)
                  * If the error happens during a subscription, and is only eventual, can be ignored
                  * If the error happens during a subscription, and the subscription stop receiving messages, the unresponsiveTimeout will take care of it (unsubs/subs)
                  */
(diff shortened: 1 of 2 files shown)


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -244,9 +244,10 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
             const timeout$ = of(subscriptionError({ ...action, reason: 'WS: unsubscribe -> subscribe (unresponsive channel)' }), unsubscribe(action), subscribe(action)).pipe(
               delay(config.subscriptionUnresponsiveTTL),
               withLatestFrom(state$),
-              map(([action, state]) => ({ ...action, isActive: state.ws.subscriptions[subscriptionKey]?.active })),
+              map(([action, state]) => ({ ...action, isActive: state.ws.subscriptions[subscriptionKey]?.active || state.ws.subscriptions[subscriptionKey]?.subscribing > 0 })),
```
krebernisak, line 247: I'd prefer if we were a little bit more explicit here. Maybe something like this:

```
mergeMap(([action, state]) => {
  const isActive = !!state.ws.subscriptions[subscriptionKey]?.active 
  const isSubscribing = !!(state.ws.subscriptions[subscriptionKey]?.subscribing > 0)
  return isActive || isSubscribing ? of(action) || empty()
}
```

We don't need the filter op in this case, as we made a filter ourselves.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -244,9 +244,10 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
...
               // Filters by active subscription. 
               // The timeout could think we don't receive messages because of unresponsiveness, and it's actually unsubscribed
+              // Subscribing is considered active as we want to trigger an unsubscription from a hung subscribing channel
               filter(({ isActive }) => isActive),
               tap(() => logger.info('WS: unsubscribe -> subscribe (unresponsive channel)', { payload: action.subscriptionMsg })),
```
krebernisak, line 252: Also, this `logger.info` will happen 3x (for every event), while we need it to happen only once. Do we need it, as I think we already log on `subscriptionError` in metrics, and it carries a `reason` string.

packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -88,6 +92,8 @@ export const subscriptionsReducer = createReducer<SubscriptionsState>(
       const isSubscribing = state[key]?.subscribing
       state[key] = {
         active: false,
+        wasEverActive: false,
```
krebernisak, line 95: Not sure if we should do it like this. For example, there could be a bug in our events pipeline that would trigger `subscribe` on an already subscribed channel.

So we shouldn't assume it's `unsubscribed=false` on the `subscribe` event.
In the same event, the subscription is active meaning `wasEverActive: true`, we just have a bug :)

That's why I think it's better here to just update the `subscribing` and `input` data, and don't update the rest.

We could then add a WARN log here if we find a mismatch, an invariant in the state which will signal we have a bug.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -244,11 +244,14 @@ export const connectEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
...
-              map(([action, state]) => ({ ...action, isActive: state.ws.subscriptions[subscriptionKey]?.active })),
               // Filters by active subscription. 
               // The timeout could think we don't receive messages because of unresponsiveness, and it's actually unsubscribed
-              filter(({ isActive }) => isActive),
-              tap(() => logger.info('WS: unsubscribe -> subscribe (unresponsive channel)', { payload: action.subscriptionMsg })),
+              // isSubscribing is considered too as we want to trigger an unsubscription from a hung channel
+              mergeMap(([action, state]) => {
+                const isActive = !!state.ws.subscriptions[subscriptionKey]?.active 
+                const isSubscribing = !!(state.ws.subscriptions[subscriptionKey]?.subscribing > 0)
+                return isActive || isSubscribing ? of(action) : EMPTY
```
RodrigoAD, line 253: `empty` seems deprecated. [empty()](https://rxjs-dev.firebaseapp.com/api/index/function/empty) vs. [EMPTY](https://rxjs-dev.firebaseapp.com/api/index/const/EMPTY)

aalu1418 (dismissed):
LGTM!

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -332,12 +335,15 @@ export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
           break
         case subscriptionError.type:
           ws_subscription_errors.labels(subscriptionErrorLabels(action.payload)).inc()
-          logger.error('WS: subscription error', { payload: action.payload })
+          logger.error(action.payload.reason, { payload: action.payload })
```
krebernisak, line 338: Hmm, I'd rather we revert this so its consistent with other logging, and the `reason` should be in the payload, right?

RodrigoAD: Yep ok

krebernisak (approved):

PR #496, opened 2021-05-05
Fix extraneous connections in usage of Intrinio SDK

aalu1418:
The Intrinio SDK initializes a WS connection which is unused, but also not terminated.

Fix: explicitly call `destroy()` to close out SDK WS connection.

Note: Tested with API key and it works, but can not verify that the first connection is closed (based on Intrinio's dashboard calculation)


justinkaseman (approved):
Yep yep

krebernisak (approved):

PR #499, opened 2021-05-06
Update CVI

vladi-coti:
improved the time resolution from days to seconds and smooth-out the effect of added strikes


boxhock (changes requested):
Thanks for submitting this PR! Left a few comments

vladi-coti:
Fixed requested changed

packages/composites/crypto-volatility-index/test/integration/adapter.test.ts
```diff
@@ -2,28 +2,32 @@ import { Requester } from '@chainlink/ea-bootstrap'
 import { assertError, assertSuccess } from '@chainlink/ea-test-helpers'
 import { AdapterRequest } from '@chainlink/types'
 import { execute } from '../../src/adapter'
+jest.setTimeout(10000)
 
 describe('execute', () => {
   const jobID = '1'
-  const contractAddress = '0x0BD102ef50a6a133B38Bf3Bd3d40cE36cc1aB5A8'
+  const contractAddress = '0x1B58B67B2b2Df71b4b0fb6691271E83A0fa36aC5'
+  process.env.COINGECKO_DATA_PROVIDER_URL = 'http://localhost:3000'
```
boxhock, line 10: We should not hard-code this env var, instead we should have it set when running the test.

```suggestion
```

packages/composites/crypto-volatility-index/test/integration/adapter.test.ts
```diff
@@ -2,28 +2,32 @@ import { Requester } from '@chainlink/ea-bootstrap'
 import { assertError, assertSuccess } from '@chainlink/ea-test-helpers'
 import { AdapterRequest } from '@chainlink/types'
 import { execute } from '../../src/adapter'
+jest.setTimeout(10000)
```
boxhock, line 5: This should also be set when running the test, instead of hard-coding here

```suggestion
```

boxhock (changes requested):
Just a couple more things I noticed in the test. Besides this we should be good, and I'll merge once fixed.

boxhock:
@vladi-coti looks like you'll have to rebase as well, as I can't merge if the branch is out of date with the base branch

boxhock (approved):

justinkaseman:
Hey @vladi-coti, we are currently hitting a scenario where `https://www.deribit.com/api/v2/public/get_book_summary_by_currency` only returns one `expiration` timestamp that passes
```
      olderThanHour(instrumentName, hourAgo, instruments) &&
        moment.unix(expiration).weekday() == 5
```
making this check fail

```
  if (!e1) throw new Error('Could not find an expiration date before a full month')
  if (!e2) throw new Error('Could not find an expiration date after a full month')
```

How would you like to proceed?

PR #500, opened 2021-05-06
feat: warmer and ws play together

RodrigoAD:
Warmer and WS modules keep completely separate, but they need to be aware of each other (warmer needs to be aware of WS in reality)

- If both Warmer and WS are available, WS should always have preference over Warmer
- Warmer needs to know what's the current WS state, to decide if subscribe or not
  - Warmer will be active if there is no WS subscription active, and will be checking on every request if the WS subs is active
  - Warmer will unsubscribe at the moment it finds a WS subscription

- [ ] Acceptance Testing

### DIFF
--- a/packages/core/bootstrap/src/index.ts
+++ b/packages/core/bootstrap/src/index.ts
@@ -123,7 +123,7 @@ export const withDebug: Middleware = async (execute) => async (input: AdapterReq
 }
 
 // Init all middleware, and return a wrapped execute fn
-const withMiddleware = async (execute: Execute, middleware: Middleware[]) => {
+export const withMiddleware = async (execute: Execute, middleware: Middleware[]) => {
   // Init and wrap middleware one by one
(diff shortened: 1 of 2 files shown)


aalu1418 (commented):
Tested on Coinbase adapter with these env params:
```
CACHE_ENABLED=true
EXPERIMENTAL_WARMUP_ENABLED=true
LOG_LEVEL=debug
WS_SUBSCRIPTION_TTL=10000
```

Looks good to me!
* Warmer still works as usual on non WS EA (tested with coingecko)
* Warmer works as expected with WS EA (tested with coinbase, after WS timeout it switched to warmer, and turned off when new request comes that triggers WS)

aalu1418 (approved):

HenryNguyen5 (approved):

krebernisak (approved):

PR #502, opened 2021-05-06
WS: Handle state + metrics when TTL parameter is too short

aalu1418:
Edge case that was noticed while acceptance testing warmer + WS.

If the `WS_SUBSCRIPTION_TTL` parameter is too short (like 10 ms instead of 10000 ms) for a connection & subscription to be established. It will still Unsubscribe + Disconnect which subtracts 1 from `total` and in metrics, and leave an active `connecting` state that blocks further subscription attempts even though it is disconnected.

Fix: introduce a `wasEverConnected` parameter similar to `wasEverActive` for subscriptions to prevent negative values, and resetting the `connecting` parameter.

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/reducer.ts
+++ b/packages/core/bootstrap/src/lib/ws/reducer.ts
@@ -14,11 +14,14 @@ export interface ConnectionsState {
   connecting: {
     [key: string]: number // Filter if its more than one
   }
+  wasEverConnected: {
+    [key: string]: boolean
+  }
   /** Number of active connections */
   total: number
 }
 
-const initConnectionsState: ConnectionsState = { active: {}, connecting: {}, total: 0 }
+const initConnectionsState: ConnectionsState = { active: {}, connecting: {}, wasEverConnected: {}, total: 0 }
 
 export const connectionsReducer = createReducer<ConnectionsState>(
   initConnectionsState,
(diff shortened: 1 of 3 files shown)


packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -14,11 +14,14 @@ export interface ConnectionsState {
   connecting: {
     [key: string]: number // Filter if its more than one
   }
+  wasEverConnected: {
```
RodrigoAD, line 17: Connections state is growing a lot, I think it'd be better to have the same structure as subscriptions, so we have all the connection info in one place
```js
state = {
[key: string]: {
    active: boolean
    wasEverConnected?: boolean
    connecting: number
  }
}
```
What do you guys think?

krebernisak (dismissed):
Looks good! Great find again, @aalu1418 💯

packages/core/bootstrap/src/lib/ws/reducer.ts
krebernisak: I agree, let's structure them similarly to subscriptions state, but let's do it in the new PR.

aalu1418:
Rebased to current develop!

krebernisak (approved):

Issue #503, opened 2021-05-07
WS Metrics expose API key in the dashboard


krebernisak:
An example of this is a `cryptocompare-ws` adapter in the "Active websocket subscriptions" dashboard.

aalu1418:
Redacting sensitive parameters through switching to a new package: https://github.com/smartcontractkit/external-adapters-js/pull/524

PR #508, opened 2021-05-07
Redact api key parameters from metrics & logging

aalu1418:
WS URL params can contain api keys which would become exposed in the metrics endpoint.

~Solution: filter out any params passed via the URL to redact api keys passed with a variety of parameters~
* Doesn't affect the tracked states in the WS

See below for latest update on redacting sensitive parameters from URLs/data objects in logging + metrics

Addresses: https://github.com/smartcontractkit/external-adapters-js/issues/503


krebernisak (dismissed):
We also use the logger to log events, so we should make sure to redact the payload there.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -315,6 +316,8 @@ export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
         subscription_key: payload.subscriptionKey,
       })
 
+      action.payload = redact(action.payload) // use redact function in util file before payload is passed onward
```
HenryNguyen5, line 319: Redaction should be done at the logging level only, we really shouldnt modify any internal code for it

HenryNguyen5 (commented):
I think we should heavily consider a different logger that would be easier to use, and more performant, like https://github.com/pinojs/pino/blob/master/docs/redaction.md

This would result in not having to write our own redaction code, and having a faster, easier to maintain codebase

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -225,6 +227,52 @@ const permutations = (collection: any, n: any) => {
...
+/**
+ * @description
+ * Redacts secrets from JSON object (https://github.com/winstonjs/winston/issues/1079#issuecomment-825115754)
+ *
+ * @param obj The object to redact secrets from
+ *
+ * @returns redacted object
+ */
+export const redact = (obj: any): any => {
+  const copyObj = (obj: any): any => JSON.parse(JSON.stringify(obj))
```
RodrigoAD, line 243: We could have some data loss during the `stringify`. Values as `undefined` will be lost, or Dates will be stringified.
You can use lodash, is already used is other places, with `_.cloneDeep`

packages/core/bootstrap/src/lib/ws/epics.ts
krebernisak: The problem is API keys currently end up in Prometheus because of the URL label.

packages/core/bootstrap/src/lib/external-adapter/logger.ts
```diff
@@ -1,23 +1,30 @@
...
 })
 
+// https://github.com/winstonjs/winston/issues/1079#issuecomment-729050088
+const redactData = format(info => {
+  const result = redact(info)
+
+  const levelSym = Symbol.for('level')
+  const splatSym = Symbol.for('splat')
+
+  result[levelSym] = info[(levelSym as unknown) as string]
```
RodrigoAD, line 17: With `_.cloneDeep` you shouldn't need this I think

packages/core/bootstrap/src/lib/ws/epics.ts
RodrigoAD: Forget my comments if we implement ☝️

Issue #506, opened 2021-05-07
WS: Restructure connections state


aalu1418:
Original comment here: https://github.com/smartcontractkit/external-adapters-js/pull/502

Goal: refactor `connections` state to be similar to `subscriptions` state

aalu1418:
Closed with this PR: https://github.com/smartcontractkit/external-adapters-js/pull/509

PR #509, opened 2021-05-07
Refactor connections state to be similar to subscriptions state

aalu1418:
Addresses issue: https://github.com/smartcontractkit/external-adapters-js/issues/506

Original suggestion mentioned here: https://github.com/smartcontractkit/external-adapters-js/pull/502

Summary:
* Reconfigure connections state to match subscriptions state that tracks each connection individually
* Re-added total parameter for subscriptions and connections (updated by a reducer that runs after other reducers are run)


packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -2,61 +2,54 @@ import hash from 'object-hash'
...
+      state[action.payload.connectionInfo.key].active = false
     })
 
     builder.addCase(actions.disconnected, (state, action) => {
       // Remove connection
       const { key } = action.payload.config.connectionInfo
-      delete state.active[key]
-      state.connecting[key] = 0 // turn off connecting
-      // Decrement num of active connections
-      if (state.wasEverConnected[key]) state.total--
```
krebernisak, line 59: We don't use it now, but we'll need to track the number of connections & subscriptions to manage the rate limits.

RodrigoAD: I think keeping count of the subs/connections should be on the consumer side (would be safer I think too)
```js
const state = store.getState()
const activeSubscriptions = Object.values(state.ws.subscriptions).filter((s) => s.active).length
const activeConnections = Object.values(state.ws.connections).filter((c) => c.active).length
```

krebernisak: It's a design choice, I'm good either way.

Reducers are here to prepare the data for consumers, but consumers can also make additional one-off aggregations if necessary. But if this starts to repeat, and it's not just one-off aggregation but starts repeating in multiple consumers, then the better move would be to bring it to the reducer directly.

Why safer though?

RodrigoAD: For maintenance, as we are replicating the same information in two places, so we would need to maintain it in this two places. I guess not doing it properly could lead into some incongruity

aalu1418: Would turning the Subscriptions and/or Connections into a class that extends the interface solve the concerns? That way you can build in the functions for filtering + returning length for tracking totals without tracking the same number in two places.

Not sure if changing this would break underlying functionality...

packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -1,65 +1,65 @@
...
+    state.all[key].connecting = 0 // turn off connecting
+  })
+
+  builder.addMatcher(
+    isAnyOf(actions.connect, actions.connected, actions.connectionError, actions.disconnected),
+    state => {
+      state.total = Object.values(state.all).filter(s => s?.active).length
+    },
+  )
+})
```
aalu1418, line 62: Refactored the state, and added an additional reducer that is triggered after any of the reducer events (reducers are used in the order they are listed) to calculate the total connections.

If this looks good, I can also refactor the subscriptions state to be similar

packages/core/bootstrap/src/lib/ws/reducer.ts
aalu1418: See below for a potential solution that calculates the total connections state using another reducer, which provides consumers the value without needing to actively manage an additional state.

packages/core/bootstrap/src/lib/ws/reducer.ts
RodrigoAD: This looks good to me

krebernisak (approved):

PR #515, opened 2021-05-10
[Feature] Tiingo crypto endpoint

justinkaseman:
### Description
The previously built crypto endpoint using top of book data does not contain VWAP.

Adds the correct Tiingo endpoint to use: `prices`. This endpoint does a VWAP of all the exchanges on the current day and across base tokens.

### Story
https://app.clubhouse.io/chainlinklabs/story/9004/add-support-for-crypto-data-to-tiingo-adapter

### DIFF
--- a/btc.com/README.md
+++ b/btc.com/README.md
@@ -1,29 +1,36 @@
-# Chainlink External Adapters to query address balance from BTC.com
+# Chainlink External Adapters for BTC.com
 
-## Configuration
+### Input Parameters
 
-The adapter takes the following environment variables:
+| Required? |   Name   |              Description              |                              Options                              | Defaults to |
+| :-------: | :------: | :-----------------------------------: | :---------------------------------------------------------------: | :---------: |
+|           | endpoint | The Chainlink unified endpoint to use | [balance](#Balance), [height](#Height), [difficulty](#Difficulty) |  `balance`  |
 
-- `API_TIMEOUT`: Optional timeout param, defaults to `30000`
+---
 
-## Input Params
+## BTC.com Address Endpoint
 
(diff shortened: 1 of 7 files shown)


tiingo/src/endpoint/crypto/prices.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const jobRunID = validator.validated.id
+  const base = validator.validated.data.base.toLowerCase()
+  const quote = validator.validated.data.quote.toLowerCase()
+  const field = validator.validated.data.field || 'fxClose'
+  const url = '/tiingo/crypto/prices'
+
+  const options = {
+    ...config.api,
+    params: {
```
HenryNguyen5, line 49: do we need to do any deep merging here? or `config.api` is only top level objects?

HenryNguyen5 (approved):

tiingo/src/endpoint/crypto/prices.ts
justinkaseman: Won't need any deep merging here. 

Here is what `config.api` looks like:

```
api: {
      withCredentials: !!apiKey,
      baseURL: getEnv(ENV_API_ENDPOINT, prefix),
      baseWsURL: getEnv(ENV_WS_API_ENDPOINT, prefix),
      timeout: parseInt(timeout || '') || DEFAULT_API_TIMEOUT,
      headers: {
        common: {
          'Cache-Control': 'no-cache, no-store, must-revalidate',
          Pragma: 'no-cache',
          'Content-Type': 'application/json',
          Accept: 'application/json',
        },
      },
    },
```

PR #516, opened 2021-05-10
Adapter for expert car broker

boxhock:


packages/sources/expert-car-broker/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@chainlink/expert-car-broker-adapter",
+  "version": "0.0.4",
```
justinkaseman, line 3:
```suggestion
  "version": "0.0.1",
```

boxhock: @justinkaseman might want to change the example adapter, because this was generated from the yarn command

boxhock: Added a commit for it

PR #517, opened 2021-05-10
docs: additional docs for EAEE & EAs directories

justinkaseman:


packages/core/bootstrap/README.md
```diff
@@ -1,28 +1,127 @@
...
+|           | `CACHE_KEY_IGNORED_PROPS` |                                                                                Keys to ignore while deriving the cache key, delimited by `,`. The key set will be added to the default ignored keys                                                                                 |                    | `['id', 'maxAge', 'meta']` |
+|           |      `CACHE_MAX_AGE`      | Maximum age in ms. Items are not pro-actively pruned out as they age, but if you try to get an item that is too old, it'll drop it and return undefined instead of giving it to you. If set to `0` the default will be used, and if set to `< 0` entries will not persist in cache. |                    |    `30000` (30 seconds)    |
+
+### Cache key
+
+The cache key of a stored request is derived by hashing the input object, using the SHA1 hash function, while by default ignoring keys `['id', 'maxAge']`. So for example these few requests will derive the same key:
```
HenryNguyen5, line 57: Doesnt RL add more ignores now?

packages/core/bootstrap/README.md
```diff
@@ -1,28 +1,127 @@
...
+
+- `{"id": 1, "data": {"base":"LINK", "quote": "USD"}}`
+- `{"id": 2, "data": {"base":"LINK", "quote": "USD", "maxAge": 10000}}`
+- `{"id": 3, "data": {"base":"LINK", "quote": "USD"}}`
+
+The `maxAge` input argument can be used to set per item `maxAge` parameter. If not set, or set to `0`, the cache level `maxAge` option will be used. Every time the `maxAge` input argument changes, the item will be cached with the new `maxAge` parameter. To avoid hitting the cache for a specific item set `maxAge: -1` (any value `< 0`).
```
HenryNguyen5, line 63: I dont understand how someone would set the `maxAge` parameter, unless this is developer-only documentation.

packages/core/bootstrap/README.md
```diff
@@ -47,99 +146,72 @@ Each provider is defined within [limits.json](./src/lib/provider-limits/limits.j
...
+|           | `REQUEST_COALESCING_INTERVAL_COEFFICIENT` |                                                                  A coefficient as the base multiplier for exponential back-off interval function.                                                                  |         |     `2`     |
+|           |     `REQUEST_COALESCING_ENTROPY_MAX`      | Amount of random delay (entropy) in milliseconds that will be added to requests. Avoids issue where the request coalescing key won't be set before multiple other instances in a burst try to access the same key. |         |     `0`     |
 
-```bash
-docker run -p 6379:6379 --name ea-redis -d redis redis-server --requirepass SUPER_SECRET
-```
+## Metrics
 
-For **ElastiCache Redis** deployments: if encryption in transit is used, to make a connection `CACHE_REDIS_URL` needs to be set with `rediss://...` protocol.
+A metrics server can be exposed which returns prometheus compatible data on the `/metrics` endpoint on the specified port.
```
HenryNguyen5, line 185:
```suggestion
A metrics server can be exposed which returns prometheus compatible data on the `$BASE_URL/metrics` endpoint on the specified port.
```

packages/core/bootstrap/README.md
```diff
@@ -47,99 +146,72 @@ Each provider is defined within [limits.json](./src/lib/provider-limits/limits.j
...
+Note that this feature is ONLY available when running this application as an http server.
```
HenryNguyen5, line 186:
```suggestion
```
We only run as a http server now

HenryNguyen5 (changes requested):

packages/core/bootstrap/README.md
justinkaseman: Yep, good catch. Changed.

packages/core/bootstrap/README.md
justinkaseman: It could be added as a query string to the bridge URL.

HenryNguyen5: ah i see

PR #527, opened 2021-05-11
augur encoding change

robert-e-davidson3:
Originally we expected to always create all three markets at once. We expected to sometimes need to wait for the lines to come in so we could make the spread and over-under markets. However, the lines can come much later so instead we will create head-to-head markets immediately and pass flags to indicate if spread or over-under markets should be created.

QUESTION: My understanding is that TheRundown returns fake scores of something like `0.001` to indicate a lack of data. Do they do so within lines as well? We don't want to create markets where the expected spread or totalScore is zero when they shouldn't be.

This required a contract change so I redeployed.


boxhock:
Cherry picked this and made some changes in #529. Closing this PR

PR #531, opened 2021-05-12
fix(augur): treat `0.0001` as a special case

boxhock:


justinkaseman (approved):

packages/composites/augur/src/methods/createMarkets.ts
```diff
@@ -83,6 +83,8 @@ export const execute: ExecuteWithConfig<Config> = async (input, config) => {
   return Requester.success(input.id, {})
 }
 
+const transformSpecialNone = (val?: number) => val === 0.0001 ? undefined : val
```
justinkaseman, line 86: Could we add a JSdoc comment about when this happens?

boxhock: Yep, definitely

justinkaseman (approved):

PR #532, opened 2021-05-13
feat(tiingo): add crypto WS handler

justinkaseman:
### Description
Adds Tiingo crypto WS handler

### Story
https://app.clubhouse.io/chainlinklabs/story/9025/add-ws-support-for-crypto-data-to-tiingo-adapter

### DIFF
--- a/packages/sources/tiingo/src/adapter.ts
+++ b/packages/sources/tiingo/src/adapter.ts
@@ -86,8 +86,8 @@ const customParams = {
 }
 
 export const makeWSHandler = (config?: Config): MakeWSHandler => {
-  const defaultConfig = config || makeConfig()
   const getSubscription = (pair: string | undefined, subscribe = true) => {
+    const defaultConfig = config || makeConfig()
     if (!pair) return
     return {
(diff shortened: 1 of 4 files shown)


packages/sources/tiingo/src/adapter.ts
```diff
@@ -36,3 +42,83 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
...
+    }
+  }
+  const getPair = (input: AdapterRequest) => {
+    const validator = new Validator(input, customParams)
+    if (validator.error) return
+    const base = validator.validated.data.base.toLowerCase()
+    const quote = validator.validated.data.quote.toLowerCase()
+    return `${base}${quote}`
+  }
+  return () => {
```
RodrigoAD, line 108: Include the `const defaultConfig = config || makeConfig()` here better. This is useful when other adapters rely on this one

justinkaseman: Not sure that I understand how this would be any different than using it as a closure as `makeConfig` doesn't change, but I moved it so we can get this merged.

boxhock (approved):

PR #536, opened 2021-05-13
Added env var to enable WS connections

aalu1418:
Added environment variable `WS_ENABLED` that needs to be set for EAs to use WS connections (if available).

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/index.ts
+++ b/packages/core/bootstrap/src/lib/ws/index.ts
@@ -14,10 +14,10 @@ export const withWebSockets = (
   store: Store<RootState>,
   makeWsHandler?: MakeWSHandler,
 ): Middleware => async (execute) => async (input: AdapterRequest) => {
-  if (!makeWsHandler) return await execute(input)
+  const wsConfig = getWSConfig()
+  if (!makeWsHandler || !wsConfig.enabled) return await execute(input) // ignore middleware if conditions are met
 
   const wsHandler = await makeWsHandler()
-  const wsConfig = getWSConfig()
   store.dispatch(connect({ config: wsConfig, wsHandler }))
 
   const subscriptionMsg = wsHandler.subscribe(input)
--- a/packages/core/bootstrap/src/lib/ws/config.ts
+++ b/packages/core/bootstrap/src/lib/ws/config.ts
@@ -1,7 +1,7 @@
 import { getEnv, parseBool } from '../util'
 import { WSConfig } from './types'
 
-const ENV_WS_ENABLED = 'EXPERIMENTAL_WS_ENABLED'
+const ENV_WS_ENABLED = 'WS_ENABLED'
 
 // WSConnectionInfo
 const ENV_WS_CONNECTION_KEY = 'WS_CONNECTION_KEY'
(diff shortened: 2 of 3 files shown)


packages/core/bootstrap/src/index.ts
```diff
@@ -139,10 +139,11 @@ const executeSync = (execute: Execute, makeWsHandler?: MakeWSHandler): ExecuteSy
     withLogger,
     skipOnError(withCache),
     cacheWarmer.withCacheWarmer(storeSlice('cacheWarmer'), warmerMiddleware, { store: storeSlice('ws'), makeWSHandler: makeWsHandler })(execute),
-    ws.withWebSockets(storeSlice('ws'), makeWsHandler),
     rateLimit.withRateLimit(storeSlice('rateLimit')),
     withStatusCode,
-  ].concat(metrics.METRICS_ENABLED ? [withMetrics, withDebug] : [withDebug])
+  ]
+    .concat(metrics.METRICS_ENABLED ? [withMetrics, withDebug] : [withDebug])
+    .concat(ws.WS_ENABLED ? [ws.withWebSockets(storeSlice('ws'), makeWsHandler)] : [])
```
RodrigoAD, line 146: You can check it inside the ws middleware better, so it would be like:
```js
if (!makeWsHandler || !WS_ENABLED) return await execute(input) // Middleware is ignored
// Middleware continues
```

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -3,13 +3,16 @@ import { Store } from 'redux'
...
 import { connect, subscribe } from './actions'
+import { parseBool } from '../util'
 
 export * as types from './types'
 export * as config from './config'
 export * as actions from './actions'
 export * as reducer from './reducer'
 export * as epics from './epics'
 
+export const WS_ENABLED = parseBool(process.env.WS_ENABLED)
```
RodrigoAD, line 14: You can include it inside `ws/config.ts` and export it through `getWSConfig`

boxhock:
Has this been requested by anyone? Any thoughts on explicitly disabling it instead? Right now I can't think of any cases where node ops _don't_ want to use WS.

packages/core/bootstrap/src/index.ts
krebernisak: The order of these middlewares matters, so my suggestion is to move it back in its place and make the check inside the middleware. Also agree that the enable flag should be set internally in WS config.

krebernisak (changes requested):

justinkaseman:
> Has this been requested by anyone? Any thoughts on explicitly disabling it instead? Right now I can't think of any cases where node ops _don't_ want to use WS.

I requested this as I would like a safety hatch in case this rollout of WS adapters goes bad. 

We should keep WS on by default, but allow people to disable.

boxhock:
@justinkaseman perfect, thanks that makes sense

aalu1418:
Updates:
* Rebased with changes from over the weekend
* Moved the check inside of `ws/index` as suggested
* Moved/modified old `WS_ENABLED` parameter in `ws/config`
  * Removed `experimental` from environment variable
  * Moved enable toggle within `wsConfig` (not used anywhere else as far as I could tell)

krebernisak (approved):

PR #550, opened 2021-05-18
fix(bootstrap): rate limiter's cache warmer request id

justinkaseman:


packages/core/bootstrap/src/lib/rate-limit/index.ts
```diff
@@ -28,10 +28,10 @@ export const computeThroughput = (
   id: string,
 ): number => {
   // All observed in interval
-  const observedRequests = selectObserved(state, interval).filter((h) => !h.isWarmup)
+  const observedRequests = selectObserved(state, interval).filter((h) => !h.isCacheHit)
```
RodrigoAD, line 31: I don't think this is the issue. Now you are just counting the requests that are going to the provider, but you want to take all of them in account.
I don't think you want to count warmup requests either

RodrigoAD:
Pushed the fix, please try again to check that is solved.
Basically, the result from hashing the input to get the `participantId` from a direct request and from cache warming was different, as with cache warming `request.data` had the field `rateLimitMaxAge`, that wasn't being ignored by the hashing function. 
This was duplicating the same participant, making two different entries on the participants state. So, warming participant was getting a lot of weight on the capacity, leaving no room for the direct requests

PR #559, opened 2021-05-20
fix: rate limit: removed sec and day intervals

RodrigoAD:
`SEC` and `DAY` intervals were not being used on Rate Limiting, and were consuming too much memory

### DIFF
--- a/packages/core/bootstrap/src/lib/rate-limit/reducer.ts
+++ b/packages/core/bootstrap/src/lib/rate-limit/reducer.ts
@@ -4,17 +4,13 @@ import { WARMUP_REQUEST_ID } from '../cache-warmer/config'
 import { successfulRequestObserved } from './actions'
 
 export enum IntervalNames {
-  SEC = 'SEC',
   MINUTE = 'MINUTE',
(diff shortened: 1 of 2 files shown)


packages/core/bootstrap/src/lib/rate-limit/reducer.ts
```diff
@@ -4,22 +4,18 @@ import { WARMUP_REQUEST_ID } from '../cache-warmer/config'
...
-  DAY = 'DAY',
 }
 
 export const Intervals: { [key: string]: number } = {
-  [IntervalNames.SEC]: 1000,
   [IntervalNames.MINUTE]: 60 * 1000,
   [IntervalNames.HOUR]: 60 * 60 * 1000,
-  [IntervalNames.DAY]: 24 * 60 * 60 * 1000,
 }
 export interface Heartbeat {
```
HenryNguyen5, line 15: Can we shorten the rest of the keys here too?

HenryNguyen5 (changes requested):

HenryNguyen5 (approved):

PR #561, opened 2021-05-20
AUGUR: Skip markets with TBD teams. Fix nonce tracking

robert-e-davidson3:


robert-e-davidson3:
Updated with two changes:
1. Query the rundown for today, yesterday, and 2 days ago. Was skipping to 3 days ago.
2. Do not resolve markets immediately. I set this to 2 hours but since TheRundown only gives us game start times, this will need to be higher. I'm not sure how long games last.

PR #562, opened 2021-05-21
Augur changes

boxhock:


pgebheim:
@boxhock I believe this also needs the patch from d9fc73ade96eec7b5eeb252a58d376dabfb321e1

Which is a bug fix for skipping over a day

boxhock:
@pgebheim This commit should have been cherry-picked in b387a2bec0d6b818ec0cc3090515f8a7be7dfb64. Has there been made changes outside the history of this PR?

pgebheim:
Oh you're correct I missed that in the history.

pgebheim:
Also, there is an additional API now available to just get the unresolved market ids from the contract.

https://github.com/AugurProject/turbo/blob/3fb7d91ed5147826d54c0908d158fb41236e08d8/packages/smart/contracts/turbo/AbstractMarketFactory.sol#L258

Perhaps this makes it easier to decided which markets to check against the api

PR #563, opened 2021-05-21
Augur Sportsdataio

boxhock:


packages/composites/augur/src/dataProviders/sportsdataio.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (diffTime / 1000 < startBuffer) continue
+    if (diffTime / (1000 * 3600 * 24) > daysInAdvance) continue
+
+    // TODO: Doesn't do anything right now
+    /*const [headToHeadMarket, spreadMarket, totalScoreMarket]: [ethers.BigNumber, ethers.BigNumber, ethers.BigNumber]
+      = await contract.getEventMarkets(event.GlobalGameID)*/
+
+    const start = Date.parse(event.Date) / 1000
+    // TODO: Missing homeSpread
+    // TODO: Missing totalScore
```
bryanjowers, line 86: from Yaser: 
here is how we want to handle the value for totalScore. When Status = Final, we can add value:
`let’s please add it as AwayScore + HomeScore when Status = Final`

bryanjowers:
lets get this reviewed and merged

packages/composites/augur/src/dataProviders/sportsdataio.ts
boxhock: This has been resolved offline (this case will never happen as we don't create markets when Status = Final)

thodges-gh (approved):

PR #565, opened 2021-05-21
fix: avoid annoying logs if info is on cache

RodrigoAD:


packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -115,6 +115,10 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
     if (!data || !data.data) return
     if (isNaN(data.rateLimitMaxAge as number)) return
     const maxAge = Number(data.rateLimitMaxAge)
+    return maxAge
+  }
+
+  const _logRL = (maxAge: number, data: AdapterRequest) => {
```
HenryNguyen5, line 121: The return value of _logRL isnt being used now

HenryNguyen5 (changes requested):

PR #581, opened 2021-05-27
Refactored example adapter

silaslenihan:
Refactored the example adapter to be more in line with the changes we made in the Bob adapter

### DIFF
--- a/packages/examples/source/src/adapter.ts
+++ b/packages/examples/source/src/adapter.ts
@@ -1,5 +1,5 @@
 import { Requester, Validator, AdapterError } from '@chainlink/ea-bootstrap'
-import { Config, ExecuteWithConfig, ExecuteFactory } from '@chainlink/types'
+import { Config, Execute, ExecuteFactory, AdapterRequest } from '@chainlink/types'
 import { makeConfig, DEFAULT_ENDPOINT } from './config'
 import { example } from './endpoint'
 
@@ -7,7 +7,7 @@ const inputParams = {
   endpoint: false,
 }
 
-export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+export const execute = async (request: AdapterRequest, config: Config): Promise<AdapterResponse> => {
   const validator = new Validator(request, inputParams)
   if (validator.error) throw validator.error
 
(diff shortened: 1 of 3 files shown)


packages/examples/source/src/adapter.ts
```diff
@@ -30,6 +30,6 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   }
 }
 
-export const makeExecute: ExecuteFactory<Config> = (config) => {
-  return async (request) => execute(request, config || makeConfig())
+export const makeExecute = (config?: Config): Execute => {
```
justinkaseman, line 33: Similar to what we were talking about with `ExecuteWithConfig`, what is the goal of this change?

packages/examples/source/src/endpoint/example.ts
```diff
@@ -1,5 +1,5 @@
 import { Requester, Validator } from '@chainlink/ea-bootstrap'
-import { Config, ExecuteWithConfig } from '@chainlink/types'
+import { Config,  ExecuteWithConfig} from '@chainlink/types'
```
justinkaseman, line 2: Looks like Prettier hasn't run to format the code.

What code editor do you use?

silaslenihan: I use VS code

packages/examples/source/src/adapter.ts
silaslenihan: I used that makeExecute in the Bob adapter and I think it was more flexible with the config input which I needed in for that adapter. I can revert it back if you think I should though.

justinkaseman: Yeah, we still want to maintain consistency across adapters with a shared type. Instead what we should do is change it at a source. 

In `packages/core/types/@chainlink/index.d.ts` you can see all of our repository level types.
You ran into trouble with `api` not being on Config, but the type requiring it right? The change would be to make `api` optional on the Config.

```
  export type Config = {
    apiKey?: string
    network?: string
    returnRejectedPromiseOnError?: boolean
    verbose?: boolean
    api?: RequestConfig
  }
```

and then to ensure this doesn't break anything by trying to compile (`yarn setup` from root) and lint again (`yarn lint` from root).

Then if this looks fine I would change the Bob adapter to use our shared types for consistency.

packages/examples/source/src/endpoint/example.ts
justinkaseman: What default formatter do you have set?

If you open the command palette and type `Format Document With` and choose `Configure Default Formatter` at the bottom, what does it say?

packages/examples/source/src/adapter.ts
silaslenihan: Ok great, I will make this change. Thanks

PR #583, opened 2021-05-28
[Feat] Smart Cache Warming

justinkaseman:
### Description
An enhancement to see large reductions in rate limits on HTTP calls for providers with batch support:
- coinapi
- coingecko
- coinmarketcap
- cryptocompare
- nomics

In the common scenario where one EA is serving ~40 feeds the cache-warming middleware could be smarter.
If multiple requests are being warmed that could be retrieved from a single batch request it will join them to reduce API credits usage.

### Steps
- [x] Caching breaks batch requests into individual cached requests
- [x] Cache Warmer breaks batch requests into individual subscriptions
- [x] If a batch request comes in, stop polling children, but maintain as placeholders
- [x] Cache Warmer subscriptions are marked batchable
- [x] Cache Warmer combines batchable subscriptions
-------
- [x] Unit tests
- [x] Handle failures and timeouts of parent subscription
--------
- [ ] Delete soak test ECR

### Stories
https://app.clubhouse.io/chainlinklabs/story/8462/ea-core-framework-smarter-caching-part-1
https://app.clubhouse.io/chainlinklabs/story/8461/ea-core-framework-smart-cache-warming


packages/core/bootstrap/src/lib/cache-warmer/reducer.ts
```diff
@@ -26,25 +26,135 @@ export interface SubscriptionData {
...
+    }
+  })
+
+  builder.addCase(actions.warmupLeaveGroup, (state, { payload }) => {
+    if (!payload)
+      console.error('subscriptionsReducer warmupLeaveGroup - no payload', JSON.stringify(state))
+
+    const batchWarmer = state[payload.parent]
+    if (!batchWarmer) {
+      console.error(
```
HenryNguyen5, line 117: We should be using the logger everywhere instead of `console`

HenryNguyen5 (changes requested):

packages/core/bootstrap/src/lib/cache-warmer/reducer.ts
justinkaseman: Oops, meant to do that. Changing.

packages/core/bootstrap/src/lib/cache-warmer/reducer.ts
```diff
@@ -26,25 +26,135 @@ export interface SubscriptionData {
...
+  builder.addCase(actions.warmupJoinGroup, (state, { payload }) => {
+    if (!payload)
+      console.error('subscriptionsReducer warmupJoinGroup - no payload', JSON.stringify(state))
+
+    const batchWarmer = state[payload.parent]
+    if (!batchWarmer) {
+      console.error(
+        'subscriptionsReducer warmupJoinGroup - no batch warmer exists',
+        JSON.stringify(state),
+      )
```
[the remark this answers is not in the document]
justinkaseman: I think we can safely remove these logs now. Or do you think we should keep them in?

HenryNguyen5 (approved):

PR #585, opened 2021-06-01
[Feat] Minimum TTL

justinkaseman:
### Description
- Adds minimum TTL of 30 secs
- Raises default maximum TTL to 2 minutes
- Adds feed Ids to TTL warnings
- Increases websocket subscription timeouts to 2 minutes


packages/core/bootstrap/package.json
```diff
@@ -12,7 +12,7 @@
     "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
     "prepack": "yarn build",
     "build": "tsc -b",
-    "dev:remotedev": "./remotedev.sh",
+    "dev:remotedev": "sh ./remotedev.sh",
```
HenryNguyen5, line 15: Don't need this, add a shebang to the shell script itself then it'll know what shell to execute

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -114,13 +116,26 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
   const _getRateLimitMaxAge = (data: AdapterRequest): number | undefined => {
     if (!data || !data.data) return
     if (isNaN(data.rateLimitMaxAge as number)) return
+    const feedId = data?.debug?.feedId
     const maxAge = Number(data.rateLimitMaxAge)
     if (maxAge && maxAge > ERROR_MAX_AGE) {
-      logger.warn(`Cache: Key TTL ERROR_MAX_AGE: Max Age is getting max values: ${maxAge} ms`, data)
+      logger.debug(
```
HenryNguyen5, line 122: Should these not be warn?

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -141,9 +156,10 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
     const coalescingKey = _getCoalescingKey(key)
     const endMetrics = metrics.beginObserveCacheExecutionDuration({
       participantId: key,
-      feedId: data.debug?.feedId,
+      feedId: data.debug?.feedId || 'unknown',
```
HenryNguyen5, line 159: I think we should keep it as undefined, i feel like there's a couple other places that are marked as `undefined`, would be a good ticket to make it so that we unify all unknown values in metrics

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -141,9 +156,10 @@ export const withCache: Middleware = async (execute, options = defaultOptions())
...
     })
     let maxAge = _getRequestMaxAge(data) || _getDefaultMaxAge(data)
+    if (maxAge < options.minimumAge) maxAge = options.minimumAge
```
HenryNguyen5, line 162: This causes the `if (maxAge < 0) {` check in `_cacheOnSuccess` to never execute, is this what we want? I know that when data is being from from WS to cache, the maxAge is set to `-1`, we may want to check that interaction with this code change

HenryNguyen5 (changes requested):

packages/core/bootstrap/src/lib/cache/index.ts
justinkaseman: It seems like there are scenarios where these will get thrown on every request just because of imbalanced feeds. It freaks nops out to see them.

I'll change them back though for now, let's see if other changes we make reduce it.

packages/core/bootstrap/src/lib/cache/index.ts
justinkaseman: Moved inside `_cacheOnSuccess`

packages/core/bootstrap/src/lib/cache-warmer/index.ts
```diff
@@ -19,53 +20,51 @@ interface WSInput {
...
 
-      // Could happen that a subscription is still loading. If that's the case, warmer will open a subscription. If the WS becomes active, on next requests warmer will be unsubscribed
-      const isActiveWSSubscription =
-        ws.store.getState().subscriptions.all[wsSubscriptionKey]?.active
-      // If there is a WS subscription active, warmup subscription (if exists) should be removed, and not play for the moment
-      const isActiveCWSubsciption = warmerStore.getState().subscriptions[cacheWarmerKey]
+  if (wsConfig.enabled && ws.makeWSHandler) {
+    // If WS is available, and there is an active subscription, warmer should not be active
+    const wsHandler = await ws.makeWSHandler()
+    console.log(input)
```
HenryNguyen5, line 46: We should remove this

packages/core/bootstrap/src/lib/cache-warmer/index.ts
```diff
@@ -19,53 +20,51 @@ interface WSInput {
...
-      const wsSubscriptionKey = getSubsId(wsHandler.subscribe(input))
-      const cacheWarmerKey = getSubscriptionKey(warmupSubscribedPayload)
+  const wsConfig = getWSConfig()
+  const warmupSubscribedPayload = {
+    id: input.id,
+    // We need to initilialize the middleware on every beat to open a connection with the cache
+    // Wrapping `rawExecute` as `execute` is already wrapped with the default middleware. Warmer doesn't need every default middleware
+    executeFn: async (input: AdapterRequest) =>
+      await (await withMiddleware(rawExecute, middleware))(input),
+    data: omit(input, ['debug', 'rateLimitMaxAge']),
```
HenryNguyen5, line 40: Why do we omit this? Shouldn't the hashing function do this for us?

HenryNguyen5 (changes requested):

packages/core/bootstrap/src/lib/cache-warmer/index.ts
justinkaseman: I suspect that it `object-hash` only ignores keys that are at the top level, not nested keys.
I was seeing duplicate entries when `rateLimitMaxAge` changed within debug.

HenryNguyen5: Shouldn't we omit top level `debug` from `object-hash` anyway?

justinkaseman: Top level `debug` is already ignored. `excludeKeys` defaults to 
```
[
  'id',
  'maxAge',
  'meta',
  'debug',
  'rateLimitMaxAge',
]
```

PR #589, opened 2021-06-02
fix(augur): update documentation

boxhock:


HenryNguyen5 (changes requested):
Can you add this to https://github.com/smartcontractkit/external-adapters-js/blob/93aca8f445439e8d353f5773b9cc462f39304065/packages/composites/augur/schemas/env.json too?

boxhock:
@HenryNguyen5 please see if it makes sense what I did. Not sure if that's the preferred way to do prefixed refs.

HenryNguyen5 (approved):

PR #593, opened 2021-06-03
Refactored gas price for amberdata

silaslenihan:
Moved `gasprice` endpoint from `amberdata-gasprice` to `amberdata` and created new endpoint in `adapter.ts` in `amberdata`. Deleted `amberdata-gasprice`


justinkaseman:
The `gasprice` tests will need to be updated.
Two things changed about them:
1. `endpoint` renamed to `blockchain`
2. The `endpoint` of `gasprice` needs to be supplied because there are multiple endpoints on the `amberdata` adapter

We are still working on having E2E testing run in CI/CD, but until then it's always a good idea to run tests yourself before pushing.
In this scenario you would add an API key to your environment `export API_KEY=the staging key` and then run tests for Amberdata using `yarn test packages/sources/amberdata` from the repo's root.

justinkaseman:
Another tip:
When you are making small tweaks like those latter 2 commits one way that you can keep the commit history clean is by staging the changes and then running `git commit --amend`. This will add the staged files to your most recent commit. You will have to force push your branch because it has modified history. Useful for when working on a branch alone (otherwise force pushing should be done with caution).

justinkaseman:
@silaslenihan This looks ready to go to me.

Last thing to do would be to bump the version up on `amberdata` in its package.json and rebase.

thodges-gh (approved):

PR #597, opened 2021-06-07
Make metrics base url optional

HenryNguyen5:
Our current infrastructure setup uses service discovery to find instances, whereas the baseurl is dynamically keyed. Having a static metrics endpoint rather than have it depend on the baseurl makes more sense.


packages/core/bootstrap/README.md
```diff
@@ -10,31 +10,32 @@ Detailed here is optional configuration that can be provided to any EA through e
...
+|           |       `BASE_URL`       | Set a base url that is used for setting up routes on the external adapter. Ex. Typically a external adapter is served on the root, so you would make requests to `/`, setting `BASE_URL` to `/coingecko` would instead have requests made to `/coingecko`. Useful when multiple external adapters are being hosted under the same domain, and path mapping is being used to route between them. |                 |         `/`          |
+|           | `METRICS_USE_BASE_URL` |                                                                                                                                                         Set to "true" to have the internal metrics endpoint use the supplied base url.                                                                                                                                                          |                 |       `false`        |
```
boxhock, line 29: Should we enable this by default? I assume in most operator's cases they will assume the metrics URL is updated together with `BASE_URL`, and that would be useful for them.

Otherwise we need to update our documentation other places to explain that this needs to be set to true first (e.g. https://github.com/smartcontractkit/external-adapters-js/tree/develop/packages/core/bootstrap#metrics)

boxhock (changes requested):

packages/core/bootstrap/README.md
HenryNguyen5: Im not sure that we should make it `TRUE` by default, because node ops will probably want this option disabled as soon as horizontal scaling is introduced.
Once you have multiple instances you'll probably want service discovery to automatically track instance creation, but having an endpoint that isnt static (ex. in our infra, the base_url is based on the adapter name) makes it really difficult.

I do agree with the inconsistency, we could change the `BASE_URL` wording to specify that it applies only on external ports, where the monitoring port is considered an internal port

boxhock: Cool, if we can update the documentation around it then that's fine (specifically that this env var needs to be set to true for the statement on the link I sent to be accurate).

boxhock (approved):

PR #619, opened 2021-06-15
feat: add Linear Finance composite adapter

boxhock:


packages/composites/linear-finance/README.md
```diff
@@ -0,0 +1,387 @@
+# Chainlink Linear Finance Composite Adapter
+
+This adapter calculates an index value from the symbols and weights as defined in the CSV file.
```
thodges-gh, line 3:
```suggestion
This adapter calculates an index value from the symbols and units as defined in the CSV file.
```

thodges-gh (changes requested):
Only one thing in the docs, looks good to me otherwise.

thodges-gh (approved):

PR #622, opened 2021-06-17
[Feat] NCFX External Adapter

justinkaseman:
### Description

NCFX WS adapter


packages/sources/ncfx/src/adapter.ts
```diff
@@ -0,0 +1,85 @@
...
+      isError: (message: any) => Number(message.TYPE) > 400 && Number(message.TYPE) < 900,
+      filter: () => {
+        return true
+      },
+      toResponse: (message: any, input: any) => {
+        console.log(message)
+        if(Array.isArray(message) && message.length > 0) {
+          const { data: { base, quote } } = input 
+          const pair = `${base}/${quote}`
+          const pairMessage = message.find(({ currencyPair }) => currencyPair === pair)
```
cds95, line 71: Not sure how we should handle when `pairMessage` is null.

packages/sources/ncfx/src/adapter.ts
```diff
@@ -0,0 +1,85 @@
...
+    const defaultConfig = config || makeConfig()
+    return {
+      connection: {
+        url: `${defaultConfig.api.baseWebsocketURL}/cryptodata`,
+      },
+      subscribe: (input) => getSubscription('subscribe', getPair(input)),
+      unsubscribe: (input) => getSubscription('unsubscribe', getPair(input)),
+      subsFromMessage: (message, subscriptionMsg) => {
+        if(Array.isArray(message) && message.length > 0) {
+          const pairMessage = message.find(({ currencyPair }) => currencyPair === subscriptionMsg.ccy)
```
cds95, line 57: Not sure how we should handle when `pairMessage` is null.

packages/sources/ncfx/src/adapter.ts
```diff
@@ -0,0 +1,85 @@
...
+          return getSubscription('subscribe', `${pairMessage.currencyPair}`)
+        }
+        return getSubscription('subscribe', `${message}`)
+      },
+      isError: (message: any) => Number(message.TYPE) > 400 && Number(message.TYPE) < 900,
+      filter: () => {
+        return true
+      },
+      toResponse: (message: any, input: any) => {
+        console.log(message)
```
justinkaseman, line 57:
```suggestion
```

packages/sources/ncfx/src/adapter.ts
justinkaseman: It will throw on the next line as it tries to walk to result path.
I'd probably just add a log here like:
```
if (!pairMessage) logger.warn(`${pair} not found in message`)
```

packages/sources/ncfx/src/config.ts
```diff
@@ -0,0 +1,20 @@
+import { Requester, util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const NAME = 'NCFX' // This should be filled in with a name corresponding to the data provider using UPPERCASE and _underscores_.
```
justinkaseman, line 4:
```suggestion
export const NAME = 'NCFX'
```

packages/scripts/src/new/lib.ts
```diff
@@ -16,7 +16,7 @@ function writeJson(data: any) {
   const files = Object.keys(data)
 
   // write to each file
-  files.forEach(f => {
+  files.forEach((f) => {
```
justinkaseman, line 19: Let's revert all of the changes to this file. (Prettier might re-format it again, which is okay).
We don't have a github action build matrix anymore.

cds95 (dismissed):

cds95 (dismissed):

cds95 (approved):

PR #626, opened 2021-06-19
[OLD] Skip markets with TBD teams

pgebheim:
AUGUR: resolve markets with delay

only increment nonce on tx success; remove resolution buffer code

augur: resolve markets going back until the day before yesterday

augur: use new event resolution list


pgebheim:
Closing but here for reference

PR #628, opened 2021-06-22
Augur NCAA-FB support

boxhock:


packages/composites/augur/src/dataProviders/sportsdataio.ts
```diff
@@ -5,40 +5,137 @@ import { BigNumber, ethers } from 'ethers'
...
+    case 'ncaa-fb': {
+      return (filtered as CFBGames[]).map(event => ({
+        Date: event.DateTime || event.Day,
+        GameID: event.GlobalGameID,
+        AwayTeamID: event.GlobalAwayTeamID,
+        HomeTeamID: event.GlobalHomeTeamID,
+        Status: event.Status,
+        PointSpread: event.PointSpread
+      }))
+    }
```
thodges-gh, line 95: Should we have a default to error?

boxhock: We could throw an error here, but the sport should already be validated to be in `SPORTS_SUPPORTED` on L156. I can add in just a default handling just in case.

boxhock:
Closing in favor of #742

PR #631, opened 2021-06-23
Cpc rainfall adapter

cds95:


cds95:
Closing as it has been more than 6 months and the API is still down.

PR #635, opened 2021-06-23
[Feature] Dynamic test fixtures

justinkaseman:
#### Problem:
We want to test the health of composite adapters, but the fixture used to make this request has an input parameter that is coupled with the environment variables that are set.

### Solution:
Add the ability to load a JS file for test payloads that will handle this logic.

### Story
https://app.clubhouse.io/chainlinklabs/story/11099/composite-adapter-liveness-checks


packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
...
+]
+
+const getEnvName = (name, prefix = '') => {
+  const envName = prefix ? `${prefix}_${name}` : name
+  if (!isEnvNameValid(envName))
+    throw Error(`Invalid environment var name: ${envName}. Only '/^[_a-z0-9]+$/i' is supported.`)
+  return envName
+}
+
+const isEnvNameValid = (name) => /^[_a-z0-9]+$/i.test(name)
```
HenryNguyen5, line 23: Why do we need this? environment variables have their own naming requirements that come from the shell they're created in

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
...
+
+const getEnv = (name, prefix = '') => process.env[getEnvName(name, prefix)]
```
HenryNguyen5, line 25: Do we need prefix? Looks like it's unused for now

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
+const basePayload = {
```
HenryNguyen5, line 1: I would move this in the `generateTestPayload` function

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
...
+
+function searchEnvironment() {
+  for (const [environmentVariable, output] of testAndReplace) {
+    const isSetEnvVar = getEnv(environmentVariable)
+    if (isSetEnvVar) return output
+  }
+}
+
+function generateTestPayload() {
+  const payload = { ...basePayload }
```
HenryNguyen5, line 35: I would just in-line the `basePayload` here, so we dont need to create an unneeded object 
```ts
const payload = {
  request: {
    source: searchEnvironment()
  }
}
```

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const isEnvNameValid = (name) => /^[_a-z0-9]+$/i.test(name)
+
+const getEnv = (name, prefix = '') => process.env[getEnvName(name, prefix)]
+
+function searchEnvironment() {
+  for (const [environmentVariable, output] of testAndReplace) {
+    const isSetEnvVar = getEnv(environmentVariable)
+    if (isSetEnvVar) return output
```
HenryNguyen5, line 30: Should `output` be named `defaultValue`?

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,40 @@
+const basePayload = {
+  request: {},
+}
+
+const testAndReplace = [
+  ['AMBERDATA_DATA_PROVIDER_URL', 'amberdata'],
```
HenryNguyen5, line 6: I would probably make these objects so that if nops need to make their own payloads, they'll have an easier time creating them
```js
const environmentVariables = [{envKey: "AMBERDATA_DATA_PROVIDER_URL", defaultValue: "amberdata" }]
```

HenryNguyen5 (changes requested):

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,28 @@
...
+  for (const {envKey, defaultValue} of environmentVariables) {
+    const isSetEnvVar = process.env[envKey]
+    if (isSetEnvVar) return defaultValue
+  }
+}
+
+function generateTestPayload() {
+  const payload = {
+    request: {
+        source = searchEnvironment()
```
HenryNguyen5:
```suggestion
        source : searchEnvironment()
```

packages/composites/apy-finance/test-payload.js
```diff
@@ -0,0 +1,28 @@
...
+  {envKey: 'COINPAPRIKA_DATA_PROVIDER_URL', defaultValue: 'coinpaprika'},
+  {envKey: 'CRYPTOCOMPARE_DATA_PROVIDER_URL', defaultValue: 'cryptocompare'},
+  {envKey: 'KAIKO_DATA_PROVIDER_URL', defaultValue: 'kaiko'},
+  {envKey: 'NOMICS_DATA_PROVIDER_URL', defaultValue: 'nomics'},
+]
+
+function searchEnvironment() {
+  for (const {envKey, defaultValue} of environmentVariables) {
+    const isSetEnvVar = process.env[envKey]
+    if (isSetEnvVar) return defaultValue
```
HenryNguyen5, line 15: We should warn log here that we're using a default var

HenryNguyen5 (changes requested):

packages/composites/apy-finance/test-payload.js
justinkaseman: Oh it's not actually a default var per say. It's if there _is_ an environment variable set we want to return the value.

Like it needs to return `coingecko` if we find `COINGECKO_DATA_PROVIDER_URL`, no mater what `COINGECKO_DATA_PROVIDER_URL` may be.

packages/composites/apy-finance/test-payload.js
justinkaseman: Beat you :racehorse:

packages/composites/apy-finance/test-payload.js
HenryNguyen5: oh interesting, alright

HenryNguyen5 (dismissed):

packages/composites/apy-finance/test-payload.js
justinkaseman: Maybe we should just call it `value`?

HenryNguyen5: Yeah

packages/composites/outlier-detection/test-payload.js
```diff
@@ -0,0 +1,32 @@
...
+    const isSetEnvVar = process.env[envKey]
+    if (isSetEnvVar) return value
+  }
+}
+
+function generateTestPayload() {
+  const payload = {
+    request: {
+      contract: '0x0dEaf87519D434DCF74551B2E907aF18D2304946',
+      multiply: 1,
```
boxhock, line 24:
```suggestion
      multiply: 1e8,
```

packages/composites/market-closure/test-payload.js
```diff
@@ -0,0 +1,25 @@
...
+    const isSetEnvVar = process.env[envKey]
+    if (isSetEnvVar) return value
+  }
+}
+
+function generateTestPayload() {
+  const payload = {
+    request: {
+      contract: '0x5c4939a2ab3A2a9f93A518d81d4f8D0Bc6a68980',
+      multiply: 1,
```
boxhock, line 17:
```suggestion
      multiply: 1e8,
```

boxhock (changes requested):
Added correct `multiply` values. Looks good otherwise :+1:

boxhock (approved):

PR #636, opened 2021-06-24
chore/synth index integration testing

HenryNguyen5:
- Update pnpify config
- Add type fest
- Cleanup types
- Test toFixedMax
- Add supertest
- Add jest-runner settings to vscode
- Return server object on init
- Fix typeguard logic
- Add 'base' var to not-found message
- Remove unneeded exports
- Migrate e2e tests to their own package
- Strip out e2e tests
- Add synth-index X coingecko tests


packages/composites/synth-index/src/index.ts
```diff
@@ -1,7 +1,5 @@
 import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute } from './adapter'
-import { makeConfig } from './config'
 
-const NAME = 'SYNTH-INDEX'
-
-export = { NAME, makeConfig, makeExecute, ...expose(makeExecute()) }
```
justinkaseman, line 7: This is losing the `makeConfig` and `makeExecute` exports?

HenryNguyen5: Yeah, i dropped them since it looks like they're not being used

.github/workflows/main.yml
```diff
@@ -36,7 +47,7 @@ jobs:
       - run: yarn install
       - run: yarn setup
       - run: yarn lint
-      - run: yarn test unit
+      - run: yarn test:unit
```
justinkaseman, line 50: Should we also separate this out to its own check?

HenryNguyen5: sure, yeah we could

HenryNguyen5: actually, probably not, seems like a waste of a runner, integration tests can get hefty hence their own runner

package.json
```diff
@@ -14,6 +14,8 @@
     "lint:fix": "yarn lint --fix",
     "setup": "tsc -b ./packages/tsconfig.json ",
     "test": "jest",
+    "test:unit": "jest unit",
+    "test:integration": "jest synth-index",
```
justinkaseman, line 18: In the future this will become just:
```
"test:integration": "jest integration",
```
?

HenryNguyen5: yeah

justinkaseman (approved):

Issue #257, opened 2021-01-20
TS export = disables us to export types


krebernisak:
We should migrate for the way we currently export form adapters:

```javascript
// only values can be exported here
export = { NAME, /* types */, makeExecute, makeConfig, ...expose(util.wrapExecute(makeExecute())) }
```

... to something like this:

```javascript
const handlers = expose(util.wrapExecute(makeExecute()))

export { NAME, types, makeExecute, makeConfig, handlers }
```

GMSteuart:
@krebernisak there is a much simpler route than all these exports to begin with

PR #637, opened 2021-06-24
update the way variables are exported from tiingo adapter

cds95:
This PR is to update the way variables are exported from the Tiingo adapter so that we can export types from the adapter.

### DIFF
--- a/packages/sources/tiingo/src/index.ts
+++ b/packages/sources/tiingo/src/index.ts
@@ -2,9 +2,5 @@ import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute, makeWSHandler, endpointSelector } from './adapter'
 import { NAME, makeConfig } from './config'
 
-export = {
-  NAME,
-  makeExecute,
-  makeConfig,
-  ...expose(NAME, makeExecute(), makeWSHandler(), endpointSelector),
-}
+const { server } = expose(makeExecute(), makeWSHandler())
+export { NAME, makeExecute, makeConfig, server }
--- a/packages/sources/tiingo/src/index.ts
+++ b/packages/sources/tiingo/src/index.ts
@@ -1,7 +1,7 @@
 import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute, makeWSHandler, endpointSelector } from './adapter'
 import { NAME, makeConfig } from './config'
-import { ResponseSchema } from './endpoint/crypto/top'
+import * as types from "./endpoint"
 
 const { server } = expose(makeExecute(), makeWSHandler())
-export { NAME, makeExecute, makeConfig, server, ResponseSchema }
+export { NAME, makeExecute, makeConfig, server, types }
--- a/packages/sources/tiingo/src/index.ts
+++ b/packages/sources/tiingo/src/index.ts
@@ -1,6 +1,7 @@
 import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute, makeWSHandler, endpointSelector } from './adapter'
 import { NAME, makeConfig } from './config'
+import { ResponseSchema } from './endpoint/crypto/top'
 
 const { server } = expose(makeExecute(), makeWSHandler())
-export { NAME, makeExecute, makeConfig, server }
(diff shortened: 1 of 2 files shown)


packages/sources/tiingo/src/index.ts
```diff
@@ -1,5 +1,7 @@
 import { expose } from '@chainlink/ea-bootstrap'
 import { makeExecute, makeWSHandler } from './adapter'
 import { NAME, makeConfig } from './config'
+import { ResponseSchema } from './endpoint/crypto/top'
```
thodges-gh, line 4: What if we want to export a different ResponseSchema from a different endpoint?

boxhock: Can we do something like this so we can support ResponseSchema from multiple different endpoints?

```
import * as CryptoTop from './endpoint/crypto/top'
```

cds95: do we then want to export everything from `CryptoTop`?

justinkaseman: Even better might be to come up with a data structure for exporting types. We want this across all adapters that will be exported like:

```
export { NAME, makeExecute, makeConfig, types }
```

Something like:
```
{
endpoints: {
  [name: string]: {
 ResponseSchema
}
},
config
}
```

justinkaseman: @HenryNguyen5 curious on your thoughts on how to wrangle types

HenryNguyen5: I would structure the directory tree in the following way:
``` ts
// endpoints.ts
export * as foo from './foo'
export * as bar from './'bar'
```

```ts
// endpoints/foo/index.ts
import {ResponseSchema} from '../types'
export * as baz from './baz'
export const x: ResponseSchema = {...}
export const y: ResponseSchema = {...}
```

```ts
// endpoints/foo/baz/index.ts
import {ResponseSchema} from '../../types'
export const z: ResponseSchema = {...}
```

```ts
// endpoints/bar/index.ts
import {ResponseSchema} from '../types'
export const x: ResponseSchema = {...}
export const y: ResponseSchema = {...}
```

Then we can import them as the following:
``` ts
// index.ts
import * as endpoints from './endpoints'
endpoints.foo.x()
endpoints.foo.baz.z()
endpoints.bar.y()
```

cds95: @HenryNguyen5 are you suggesting that we define the `ResponseSchema` interface in a types file and then export it from there?  The `ResponseSchema` interface is different for each endpoint though.  Should each endpoint have it's own types file in it's directory?

HenryNguyen5: Ah, didnt realize that.
Yeah, each endpoint can define its own types, such as `PriceResponse`, `CryptoTopResponse`, rather than a generic `ResponseSchema`

cds95: Alright sounds good.  Then in the adapter's index file we just import everything from the `endpoints` and just export that as `types`?

HenryNguyen5: You can re-export the types from within their relevant endpoint files, then they'd be bound to that module scope.
```ts
// endpoints/foo
export * from './types'
import * as responses from './types'
export const crypto: responses.CryptoResponse = () => {}
```
```ts
// endpoints/foo/types.ts
export interface CryptoResponse {}
```
```ts
// index
import * as endpoints from './endpoints'
type CryptoResponse = endpoints.foo.CryptoResponse
const cryptoResponse = await endpoints.foo.crypto(af
```

justinkaseman (commented):
Revisiting this - what was the purpose behind this PR? Types are getting exported, but not used.

Also, why is all of the endpoint code exported as types?
Shouldn't it be `import type`?

cds95:
@justinkaseman ah yeah I think we can rename it from `types` to `endpoints`.  This was the original issue that motivated this PR https://github.com/smartcontractkit/external-adapters-js/issues/257.

HenryNguyen5 (approved):

PR #641, opened 2021-06-28
Price endpoint refactor

silaslenihan:


package.json
```diff
@@ -26,6 +26,7 @@
     "@babel/core": "^7.13.14",
     "@babel/preset-env": "^7.13.12",
     "@babel/preset-typescript": "^7.13.0",
+    "@chainlink/external-adapter": "^0.2.6",
```
justinkaseman, line 29:
```suggestion
```

packages/sources/coinlore/src/endpoint/global.ts
```diff
@@ -17,6 +17,10 @@ const customParams = {
 export const execute: ExecuteWithConfig<Config> = async (request, config) => {
   const validator = new Validator(request, customParams)
   if (validator.error) throw validator.error
+  const endpoint = validator.validated.data.endpoint || config.defaultEndpoint
+  if (endpoint.toLowerCase() === 'globalmarketcap') {
```
justinkaseman, line 21: Something is strange here.

`path` and `field` are carrying the same purpose - to the path to key into the object.

We need to consolidate these across all adapters by going with one or the other. Let's talk about it in standup tomorrow and do this in a new branch & PR.

PR #646, opened 2021-06-29
fix token one-off overrides

boxhock:


packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -117,6 +168,29 @@ export class Validator {
...
+    // Search through `presetIncludes` to find matching override for adapter and to/from pairing.
+    for (const pair of presetIncludes) {
+      if(pair.from === from && pair.to === to) {
+        for (let i = 0; i < pair.includes.length; i++) {
+          if (pair.includes[i].adapters.includes(adapter.toUpperCase())) {
+            return pair.includes[i]
+          }
+        }
+      }
+    }
```
boxhock, line 190: Instead of doing a lot of nesting, we should be able to use some lambda functions to make it a bit more readable. Below is just an (untested) suggestion:

```suggestion
    // Search through `presetIncludes` to find matching override for adapter and to/from pairing.
    const pairs = presetIncludes.filter(pair => pair.from === from && pair.to === to)
    for (const pair of pairs) {
      const matchingIncludes = pair.includes
        .find(include => include.adapters.length === 0 || include.adapters.includes(adapter.toLowerCase()))
      if (matchingIncludes) return matchingIncludes
    }
```

packages/core/types/@chainlink/index.d.ts
```diff
@@ -147,6 +147,16 @@ declare module '@chainlink/types' {
...
+
+  // Includes is an alternative symbol mapping that can be used to represent
+  // the original request, such as wrapped tokens on DEXes.
+  export type Includes = {
+    from: string, // From symbol
+    to: string, // To symbol
+    adapters?: string[], // Array of adapters this applies to
+    inverse?: boolean // If the inverse should be calculated instead
+    tokens?: boolean // If the token addresses should be used instead
+  }
```
boxhock, line 159: Can we add this to the README above where `overrideIncludes` is documented? Would make it easier to understand what data format we're dealing with without going into the code.

packages/sources/amberdata/test/unit/price.test.ts
```diff
@@ -20,6 +20,10 @@ describe('price endpoint', () => {
         name: 'quote not supplied',
         testData: { id: jobID, data: { base: 'LINK' } },
       },
+      // {
+      //   name: 'overrides test',
+      //   testData: { id: jobID, data: { from: 'DIGG', to: 'BTC' } },
+      // },
```
boxhock, line 26: Is this something we should uncomment or remove?

packages/sources/kaiko/src/adapter.ts
```diff
@@ -16,33 +16,20 @@ const customParams = {
...
+  to.toLowerCase() === 'eth'
+    ? directUrl(from, to)
+    : `/spot_exchange_rate/${from.toLowerCase()}/${to.toLowerCase()}`
+
+const directUrl = (from: string, to: string) =>
+  `/spot_direct_exchange_rate/${from.toLowerCase()}/${to.toLowerCase()}`
+
 export const execute: ExecuteWithConfig<Config> = async (request, config) => {
-  const validator = new Validator(request, customParams)
+  const validator = new Validator(request, customParams, { name: "kaiko" })
```
boxhock, line 28: What does the "name" option here do?
Regardless, if we need the adapter name we should use the name that's exported in `index.ts` or `config.ts`.

packages/sources/kaiko/test/unit/adapter.test.ts
```diff
@@ -20,6 +20,10 @@ describe('execute', () => {
         name: 'quote not supplied',
         testData: { id: jobID, data: { base: 'ETH' } },
       },
+      // {
+      //   name: 'overrides test',
+      //   testData: { id: jobID, data: { from: 'DIGG', to: 'BTC' } },
+      // },
```
boxhock, line 26: Same here, should this be included?

packages/sources/amberdata/test/unit/price.test.ts
silaslenihan: We can delete them, I was just using them for convenience. I'll delete them in the next commit

packages/sources/kaiko/test/unit/adapter.test.ts
silaslenihan: See above comment

packages/sources/kaiko/src/adapter.ts
silaslenihan: Yeah, this code is unnecessary it seems, will delete.

packages/core/bootstrap/src/lib/external-adapter/README.md
```diff
@@ -67,6 +67,55 @@ const base = validator.validated.data.base
...
+
+```typescript
+type Includes = {
+    from: string, // From symbol
+    to: string, // To symbol
+    adapters?: string[], // Array of adapters this applies to
+    inverse?: boolean // If the inverse should be calculated instead
+    tokens?: boolean // If the token addresses should be used instead
+}
+```
```
boxhock, line 103: I think we should change the order so we make the explanation for overrideIncludes a bit clearer

edit: removed suggested code change due to github formatting

boxhock: Github completely messed up the formatting here :sweat_smile: 
My point was that we should first document the Includes type, then we document overrideIncludes

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -118,6 +169,29 @@ export class Validator {
...
+    )[0]
+    // Search through `presetIncludes` to find matching override for adapter and to/from pairing.
+    const pairs = presetIncludes.filter(pair => pair.from === from && pair.to === to)
+    for (const pair of pairs) {
+      const matchingIncludes = pair.includes
+      .find(include => include.adapters.length === 0 || include.adapters.includes(adapter.toUpperCase()))
+      if (matchingIncludes) {
+        console.log(matchingIncludes)
+        return matchingIncludes
+      }
```
boxhock, line 190:
```suggestion
      const matchingIncludes = pair.includes
        .find(include => include.adapters.length === 0 || include.adapters.includes(adapter.toUpperCase()))
      if (matchingIncludes) {
        return matchingIncludes
      }
```

boxhock: Just fixing some formatting and removing the console.log

packages/sources/kaiko/src/adapter.ts
```diff
@@ -63,21 +52,55 @@ export const execute: ExecuteWithConfig<Config> = async (request, config) => {
     params,
     timeout: 10000,
   }
-
   const response = await Requester.request(requestConfig, customError)
-  let result = Requester.validateResultNumber(
+  console.log(response.data)
```
boxhock, line 56:
```suggestion
```

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -118,6 +169,29 @@ export class Validator {
...
+    return this.validated.tokenOverrides.get(network.toLowerCase())?.get(symbol.toLowerCase())
+  }
+
+  overrideIncludes = (adapter: string, from: string, to: string, includes: Includes[]): Includes | undefined => {
+    includes.filter(include =>
+      (include.from.toLowerCase() === from.toLowerCase() && include.to.toLowerCase() === to.toLowerCase()) &&
+      (!include.adapters || include.adapters.map(adapter => adapter.toLowerCase()).includes(adapter.toLowerCase()))
+    )[0]
+    // Search through `presetIncludes` to find matching override for adapter and to/from pairing.
+    const pairs = presetIncludes.filter(pair => pair.from === from && pair.to === to)
```
boxhock, line 183: Gotta normalize these

```suggestion
    const pairs = presetIncludes.filter(pair => pair.from.toLowerCase() === from.toLowerCase() && pair.to.toLowerCase() === to.toLowerCase())
```

boxhock (commented):
Did some testing and discovered that we treat symbols case sensitive right now. Besides that it's looking really good!

silaslenihan (approved):

PR #652, opened 2021-06-30
The graph adapter

cds95:
- Implement generic GraphQL adapter
 - Implement The Graph adapter.  The adapter currently only supports pulling price data from the Uniswap subgraph.

Sample input for The Graph adapter

```
{
    "jobRunId": 1,
    "data": {
        "baseCoinTicker": "UNI", // Defaults to USD
        "quoteCoinTicker": "LINK" // Defaults to USD
    }
}
```

Sample response for The Graph adapter

```
{
    "jobRunID": "1",
    "result": "0.9427133439155078024961031037532572",
    "statusCode": 200,
    "data": {
        "result": "0.9427133439155078024961031037532572"
    }
}
```


packages/composites/the-graph/test-payload.js
```diff
@@ -0,0 +1,22 @@
+const environmentVariables = [
```
cds95, line 1: Todo:  Fill these in

packages/sources/graphql/README.md
```diff
@@ -0,0 +1,59 @@
+# Chainlink External Adapter for Graphql
```
cds95, line 1: Todo:  Fill in ReadMe

packages/sources/graphql/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,63 @@
+import { Requester } from '@chainlink/ea-bootstrap'
```
cds95, line 1: Update tests

packages/composites/the-graph/README.md
```diff
@@ -0,0 +1,52 @@
+# Chainlink The-graph Composite Adapter
```
cds95, line 1: Update ReadMe

packages/composites/the-graph/schemas/env.json
```diff
@@ -0,0 +1,21 @@
+{
```
cds95, line 1: Update file

packages/composites/the-graph/src/adapter.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const execute = async (input: AdapterRequest, config: Config): Promise<AdapterResponse> => {
+  const validator = new Validator(input, customParams)
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.jobRunID
+  const method = validator.validated.data.method
+  const provider = new ethers.providers.JsonRpcProvider(config.infuraHost)
+  let response
+  switch (method) {
```
cds95, line 18: Will add more in the future once we have more use cases

packages/composites/the-graph/src/index.ts
```diff
@@ -0,0 +1,7 @@
+import { expose } from '@chainlink/ea-bootstrap'
+import { makeExecute } from './adapter'
+import { makeConfig } from './config'
+
+const NAME = 'EXAMPLE_COMPOSITE'
```
cds95, line 5: Rename

packages/composites/the-graph/src/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    query: string,
+    variables: {
+        [T: string]: string | number
+    },
+    graphqlEndpoint: string
+}
+
+export interface TokenInformation {
+    id: string
+    decimals: number
```
cds95, line 11: decimals is currently not used but might be useful in the future

packages/composites/the-graph/src/methods/prices/dataProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+        "stateMutability": "view",
+        "type": "function"
+    }
+]
+
+export const getUSDPriceInUSDT = async (provider: ethers.providers.Provider): Promise<number> => {
+    const priceFeed = new ethers.Contract(USDT_USD_AGGREGATOR_V3_ADDRESS, aggregatorV3InterfaceABI, provider);
+    const feedResponse = await priceFeed.latestRoundData()
+    return feedResponse.answer.toNumber() / 100000000
+}
```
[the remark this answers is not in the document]
boxhock: While digging a bit deeper into this, I realized how this is used.

We should make this as generic as possible.

In a request, the requester should be able to define if they want to use a reference contract or not. If none is provided, return the value directly from The graph.
If a reference contract is provided, we should just fetch the latest value from that contract, divide it by the amount specified in the request, and multiply the value from The graph by this value.
(Optionally provide a param to define if we should multiply or divide by this value)

This way we are not locked into a specific set of price feeds or quotes.

Happy to explain further offline if that would be helpful.

packages/composites/the-graph/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+  infuraHost: string,
+  dexSubgraphs: {
+    [T: string]: DexSubgraph
+  }
+}
+
+export const INFURA_HOST = process.env.INFURA_HOST || ""
+export const USDT_USD_AGGREGATOR_V3_ADDRESS = process.env.AGGREGATOR_V3_ADDRESS || "0x3E7d1eAB13ad0104d2750B8863b489D65364e32D"
+
+export const PRICE = "price"
```
boxhock, line 15: We can keep the method names inside their respective files in the methods directory.

packages/composites/the-graph/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const USD = "USD"
+export const ETH = "ETH"
+export const WETH = "WETH"
+export const USDT = "USDT"
```
boxhock, line 20: We should be able to get rid of this to make it as generic as possible. See my other comments.

packages/composites/the-graph/src/methods/prices/prices.ts
```diff
@@ -0,0 +1,86 @@
...
+    const token1 = await dexSubgraph.getToken(jobRunID, quoteCoinTicker)
+    const token1PerToken0 = await dexSubgraph.getTokenPairPrice(jobRunID, token0.id, token1.id)
+    if (token1PerToken0) {
+        return token1PerToken0
+    }
+    const ETHPerToken0 = await dexSubgraph.getTokenPairPrice(jobRunID, token0.id, wethTokenAddress)
+    const ETHPerToken1 = await dexSubgraph.getTokenPairPrice(jobRunID, token1.id, wethTokenAddress)
+    validateTokenPrices(jobRunID, ETHPerToken0, ETHPerToken1, baseCoinTicker, quoteCoinTicker)
+    return (ETHPerToken0 as number) / (ETHPerToken1 as number)
+}
```
boxhock, line 75: As mentioned in my other comments, we should be able to make these as generic as possible, so no mention of USD vs non-USD, and we base it solely on the request params that comes in.

packages/composites/the-graph/src/methods/prices/prices.ts
```diff
@@ -0,0 +1,86 @@
...
+    baseCoinTicker: false,
+    quoteCoinTicker: false,
+    dex: false
+}
+
+export const execute = async (input: AdapterRequest, config: Config, provider: ethers.providers.Provider): Promise<AdapterResponse> => {
+    const validator = new Validator(input, customParams)
+    if (validator.error) throw validator.error
+    const jobRunID = validator.validated.id
+    const { baseCoinTicker = USD, quoteCoinTicker = USD, dex = UNISWAP } = validator.validated.data
```
boxhock, line 18: We should make sure we're not case sensitive here, so would recommend normalizing input to upper/lower-case before it's used.

boxhock (changes requested):
Good work for your first adapters from scratch!
It mostly looks good, but I want to steer you in the direction of making it more generic and allowing the request params to change which calculations are made, instead of hard-coding things like USD references inside the adapter.
Happy to talk more offline if I should clarify some things.

packages/composites/the-graph/src/config.ts
cds95: I've left `WETH` as I'm using it as a default value

packages/composites/the-graph/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+import { uniswapSubgraph } from "./methods/prices/dex"
+
+export type Config = DefaultConfig & {
+  infuraHost: string,
+  dexSubgraphs: {
+    [T: string]: DexSubgraph
+  }
+}
+
+export const INFURA_HOST = process.env.INFURA_HOST || ""
```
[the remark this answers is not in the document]
cds95: I've removed the environment variable from this adapter as it looks like the function in the package is already looking for `RPC_URL`.

boxhock (changes requested):
Great job! Left a few comments.

One thing not mentioned in comments is that there seems to be a few files missing newlines at the end. We should add that as well.

We can probably get this in once these comments have been resolved.

packages/composites/the-graph/src/config.ts
```diff
@@ -0,0 +1,23 @@
...
+    [T: string]: DexSubgraph
+  }
+}
+
+export const WETH = "WETH"
+export const UNISWAP = "UNISWAP"
+const UNISWAP_V2_SUBGRAPH_ENDPOINT = "https://api.thegraph.com/subgraphs/name/ianlapham/uniswapv2"
+
+export const makeConfig = (): Config => {
+  const uniswapV2SubgraphEndpoint = util.getEnv("UNISWAP_V2_SUBGRAPH_ENDPOINT") || UNISWAP_V2_SUBGRAPH_ENDPOINT
```
boxhock, line 17:
```suggestion
const DEFAULT_UNISWAP_V2_SUBGRAPH_ENDPOINT = "https://api.thegraph.com/subgraphs/name/ianlapham/uniswapv2"

export const makeConfig = (prefix?: string): Config => {
  const uniswapV2SubgraphEndpoint = util.getEnv("UNISWAP_V2_SUBGRAPH_ENDPOINT", prefix) || DEFAULT_UNISWAP_V2_SUBGRAPH_ENDPOINT
```

packages/composites/the-graph/README.md
```diff
@@ -0,0 +1,117 @@
...
+The Graph adapter is a generic adapter to query information from The Graph.  It currently only supports fetching prices from the Uniswap subgraph.
+
+## Configuration
+
+The adapter takes the following environment variables:
+
+| Required? |   Name    |         Description          | Options | Defaults to |
+| :-------: | :-------: | :--------------------------: | :-----: | :---------: |
+|    ✅     | `RPC_URL` | The RPC Url to connect to |         |             | 
+|           | `UNISWAP_V2_SUBGRAPH_ENDPOINT` | The Uniswap V3 subgraph endpoint |         |     https://api.thegraph.com/subgraphs/name/ianlapham/uniswapv2        |
```
boxhock, line 12: > The Uniswap V3 subgraph endpoint
Is this v2 or v3 subgraph, or are they interchangeable?

boxhock (changes requested):

packages/composites/the-graph/README.md
cds95: V2, the V3 one has different request/response objects.

boxhock (approved):
:tada:

PR #654, opened 2021-06-30
feat: branch

tuvalusoftware:
updating for changes to pass variables directly to createMarket


justinkaseman (changes requested):
Hey @tuvalusoftware , thanks for the PR.

Couple things I want to bring up:
- There are quite a lot of console.logs that have made it into this. We try to keep these out of the codebase. When logging is needed with have a logger that can be imported and used via:
```
import { Logger } from '@chainlink/ea-bootstrap'

  Logger.debug('Hello World')
```
We have four levels of logs to use: `error`, `info`, `debug`, and `trace`

- There is an API key checked in here, is this a publicly available one? Or should this be removed? If so you may want to amend the commit and force push to remove it from history.

- I see that there is some commented out code. Is this code coming back eventually? Or can it be removed?

thodges-gh:
Closing this, can talk offline.

PR #655, opened 2021-06-30
Query bitcoin balances from local node

thodges-gh:


packages/composites/bitcoin-json-rpc/README.md
```diff
@@ -27,7 +26,19 @@ See the [Composite Adapter README](../README.md) for more information on how to
 
 | Required? |    Name    |      Description       | Options | Defaults to  |
 | :-------: | :--------: | :--------------------: | :-----: | :----------: |
-|           | `endpoint` | Parameter to query for |         | `difficulty` |
+|  | `endpoint` | RPC method to call | `difficulty`, `getblockchaininfo` | `getblockchaininfo` |
```
justinkaseman, line 29:
```suggestion
|  | `endpoint` | RPC method to call | `difficulty`, `getblockchaininfo`, `scantxoutset` | `difficulty` |
```
Adding the new endpoint and setting default back to what is in `config.ts`:
```
export const DEFAULT_ENDPOINT = 'difficulty'
```

packages/composites/bitcoin-json-rpc/src/endpoint/scantxoutset.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export const execute: ExecuteWithConfig<Config> = async (request, config) => {
+  const validator = new Validator(request, inputParams)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const scanobjects = validator.validated.data.scanobjects.map((address: string) => {
+    // Addresses must be formatted as addr(39e7mxbeNmRRnjfy1qkphv1TiMcztZ8VuE)
+    if (address.substr(0, 4) == 'addr') return address
```
justinkaseman, line 19: Should we enforce this more comprehensively?

Such as with something like `/^addr\(\w*\)$/g.test(address)`.
Or even test valid character lengths of BTC addresses.

justinkaseman (approved):

PR #656, opened 2021-07-01
fix(bootstrap): payload loader

justinkaseman:
### Description
Amend payload loader to work from Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,6 @@ WORKDIR /home/node/app
 
 COPY --from=builder /home/node/app/$location/bundle ./
 # Wildcards are included to handle cases where this file doesnt exist
-COPY --from=builder /home/node/app/$location/package.json /home/node/app/$location/*test-payload.json* /home/node/app/$location/*test-payload.js* ./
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -16,6 +16,6 @@ WORKDIR /home/node/app
 
 COPY --from=builder /home/node/app/$location/bundle ./
 # Wildcards are included to handle cases where this file doesnt exist
-COPY --from=builder /home/node/app/$location/package.json /home/node/app/$location/*test-payload.json*  ./
+COPY --from=builder /home/node/app/$location/package.json /home/node/app/$location/*test-payload.json* /home/node/app/$location/*test-payload.js* ./
```
boxhock, line 19: Should we get rid of this one if the other one handles the wildcard extension for `.js*`?

```suggestion
COPY --from=builder /home/node/app/$location/package.json /home/node/app/$location/*test-payload.js* ./
```

justinkaseman: Yes :slightly_smiling_face:

boxhock (approved):
lgtm

PR #662, opened 2021-07-02
feat(coinranking): refactor and add marketcap support

boxhock:


packages/sources/coinranking/README.md
```diff
@@ -6,65 +6,85 @@ The adapter takes the following environment variables:
...
+}
+```
+
+### Sample Output
 
 ```json
 {
   "jobRunID": "1",
   "data": {
-    "uuid": "razxDUgYGNAdQ",
```
justinkaseman, line 22: Could we get this code-ified into an interface in `coin.ts` endpoint?

I originally started putting the full verbose response here for easy reference when changing the adapter, but agree that it makes the README very lengthy.

packages/sources/coinranking/src/endpoint/coin.ts
```diff
@@ -0,0 +1,68 @@
...
+  const options = {
+    ...config.api,
+    url,
+  }
+
+  const resultPaths: { [key: string]: string[] } = {
+    [Paths.Price]: ['data', 'coin', 'price'],
+    [Paths.MarketCap]: ['data', 'coin', 'marketCap'],
+  }
+  const response = await Requester.request(options)
```
justinkaseman, line 63: Something I tweaked recently was the ability to consume interfaces here by making `.request` a generic.

Here is an example of an adapter endpoint that uses the patterns:
https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/nomics/src/endpoint/price.ts#L150

So we can have our response typed and then use a new helper `Requester.withResult` to modify the response with the result instead of mutating the response object.

boxhock: Nice, this is really useful

justinkaseman (approved):
Great job!

PR #669, opened 2021-07-08
Fix prefixing for mma jobs

pgebheim:


pgebheim:
@boxhock -- Prefixing was missing for resolution jobs for sdio. Added here.

PR #674, opened 2021-07-09
Crypto API V2

cds95:
This MR implements the Crypto APIs V2 external adapter.


packages/sources/crypto-api-v2/src/endpoint/balance.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!account.coin) {
+    throw new Error(`Account ${account.address} is missing blockchain parameter`)
+  }
+  const coin = BLOCKCHAIN_NAME_MAP[account.coin.toLowerCase()]
+  const options = {
+    ...config.api,
+    url: getBalanceURI(account.address, account.chain as string, coin as string),
+  }
+  const response = await Requester.request(options)
+  const balance = response.data.data.item.confirmedBalance.amount
```
cds95, line 23: We used to do some conversion in the CryptoApi V1 adapter.  We seemed to always assume that the balance is in BTC in the V1 API.  This isn't necessarily true in V2.

packages/core/test-helpers/src/helpers.ts
```diff
@@ -3,7 +3,7 @@ import { AdapterRequest, Execute } from '@chainlink/types'
 
 export function assertError(statusCode: any, data: any, expectedJobId: any) {
   expect(statusCode.actual).toEqual(statusCode.expected)
-  expect(data.jobRunID).toEqual(expectedJobId)
+  expect(data.jobRunID).toEqual(expectedJobId || '1')
```
cds95, line 6: Want to double check this.  My tests were failing because the validator object exported from core set the jobId to 1 if it's null.  This test was failing for the Crypto API V1 adapter as well.

boxhock: I think we want to include the jobRunID on the failing tests, so this change shouldn't be necessary imo

packages/sources/crypto-api-v2/README.md
```diff
@@ -0,0 +1,250 @@
+# Chainlink CryptoAPIs V2 External Adapter
+
+### Environment Variables
+
+| Required? |    Name     |                             Description                             | Options | Defaults to |
+| :-------: | :---------: | :-----------------------------------------------------------------: | :-----: | :---------: |
+|    ✅     |   API_KEY   | An API key that can be obtained from [here](https://my.cryptoapis.io/register) |         |             |
+|           | API_TIMEOUT |                          Timeout parameter                          |         |   `30000`   |
```
boxhock, line 8: Is the default here 3000 or 30000?
Since this is an API key for all adapters, I think we don't have to include it here.

packages/sources/crypto-api-v2/README.md
```diff
@@ -0,0 +1,250 @@
...
+|           |   `dataPath`    |                   Path where to find the addresses array                    |         |  `result`   |
+|           | `confirmations` |                           Confirmations parameter                           |         |      6      |
+|           |   `addresses`   | Array of addresses to query (this may also be under the `result` parameter) |         |             |
+
+Addresses is an an array of objects that contain the following information:
+
+| Required? |   Name    |                 Description                  |                  Options                  | Defaults to |
+| :-------: | :-------: | :------------------------------------------: | :---------------------------------------: | :---------: |
+|    ✅     | `address` |               Address to query               |                                           |             |
+|           |  `coin`   |              Currency to query               | `btc`. `eth`, `bch`, `ltc`, `etc`, `bch`, `dash`, `doge` |    `btc`    |
```
boxhock, line 144:
```suggestion
|           |  `coin`   |              Currency to query               | `btc`, `eth`, `bch`, `ltc`, `etc`, `bch`, `dash`, `doge` |    `btc`    |
```

boxhock: What's the difference here between the chain and coin params? Could we maybe update the description to make it a bit clearer?

packages/sources/crypto-api-v2/src/endpoint/bc_info.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  const url = `/v2/blockchain-data/${BLOCKCHAIN_NAME_MAP[blockchain.toLowerCase()]}/${network.toLowerCase()}/blocks/last`
+
+  const reqConfig = { ...config.api, url }
+
+  const response = await Requester.request(reqConfig)
+  response.data.result = Requester.validateResultNumber(response.data, payloadDataPath)
+
+  return Requester.success(jobRunID, response, config.verbose)
+}
```
boxhock, line 39: missing newline

packages/sources/crypto-api-v2/test-payload.json
```diff
@@ -0,0 +1,7 @@
+{
+    "request": {
+        "blockchain": "BTC",
+        "endpoint": "difficulty",
+        "network": "mainnet"
+    }
+}
```
boxhock, line 7: newline

boxhock (changes requested):
While we're at it, should we add types for the response schema?

You can see Justin's comments on my previous PR here: https://github.com/smartcontractkit/external-adapters-js/pull/662#discussion_r663127534

packages/sources/crypto-api-v2/README.md
cds95: It's 30000.  I just copied this from the CryptoAPI's ReadMe.  Is that ReadMe incorrect?

packages/sources/crypto-api-v2/README.md
cds95: I also copied this directly from the V1 CryptoAPI adapter.  The `coin` parameter tells the adapter which blockchain to query from.  The `chain` parameter tells the adapter whether to query from the mainnet or testnet.  Do you think we should rename the variables instead of just updating the ReadMe?  I had decided against this to keep the adapter's parameters consistent with V1.

packages/sources/crypto-api-v2/README.md
boxhock: Ah it looks like that's correct

packages/sources/crypto-api-v2/README.md
boxhock: Got it. Yeah we should keep the parameter names the same, but I think we can clarify the readme here. What you explained here is a bit clearer than the current description in the table.

packages/sources/crypto-api-v2/src/endpoint/price.ts
```diff
@@ -0,0 +1,42 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+import { ExecuteWithConfig, Config } from '@chainlink/types'
+
+export const Name = 'price'
+
+interface ResponseSchema {
```
boxhock, line 6: We should probably export these as well. Just a good habit so we could import these types in other composite adapters in the future.

packages/sources/crypto-api-v2/src/config.ts
```diff
@@ -0,0 +1,12 @@
+import { Requester } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const DEFAULT_ENDPOINT = 'price'
+export const NAME = "CRYPTO_API_V2"
+
+export const makeConfig = (prefix?: string): Config => {
+  const config = Requester.getDefaultConfig(prefix, true)
```
boxhock, line 8: Just noticed that we are making the API key required here, but we're not marking it as required in the README.

boxhock: We should have the API_KEY env var in the readme regardless, to indicate that this adapter can use it.

packages/sources/crypto-api-v2/schemas/env.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/example-source-adapter.json",
+  "title": "@chainlink/example-source-adapter env var schema",
+  "required": [],
```
boxhock, line 4: If the API key is required, we need to mark it here

packages/sources/crypto-api-v2/schemas/env.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/example-source-adapter.json",
+  "title": "@chainlink/example-source-adapter env var schema",
```
boxhock, line 3: Gotta update these

packages/sources/crypto-api-v2/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@chainlink/crypto-api-v2-adapter",
```
boxhock, line 2: The data provider is called "CryptoAPIs", so the package name (and path) should be "cryptoapis-v2".

```suggestion
  "name": "@chainlink/cryptoapis-v2-adapter",
```

boxhock (changes requested):

packages/sources/crypto-api-v2/package.json
cds95: Can I just change the package name and this file here or are there other places I need to change it too?

boxhock: There are a few other files (like tsconfig) that reference this (or at least the path). You might have to do a search for all references of that string and replace it.

packages/sources/crypto-api-v2/src/config.ts
cds95: should we include the timeout as well then?  It's being used when setting up the default config.

boxhock: I don't think so, because API_TIMEOUT is global and applied to every single adapter. API_KEY is only used on a subset of adapters, and a subset of these have them as required. The README should make it clear if this is supported and/or required.

cds95:
Closing this PR as the yarn.lock and yarn cache files are corrupted.  Created a new MR for the clean version here https://github.com/smartcontractkit/external-adapters-js/pull/683.

PR #680, opened 2021-07-13
DXDao Adapter

cds95:
This PR is to implement the DxDAO adapter to pull the TVL (Total Value Locked) from pair contracts.  The Total Value Locked is calculated by fetching the WETH balance of the adapter, multiplying it by 2 and finally converting it to the USD value if required.


packages/sources/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,65 @@
...
+   return Requester.success(jobRunID, response, config.verbose)
+}
+
+const getTvlAtAddressInETH = async (pairContractAddress: string, wethContractAddress: string): Promise<BigNumber> => {
+   const provider = new ethers.providers.JsonRpcProvider(JSON_RPC_URL)
+   Logger.info(`Fetching TVL for contract '${pairContractAddress}' using WETH contract address ${wethContractAddress}`)
+   const contract = new ethers.Contract(wethContractAddress, dxdWethContractAbi, provider)
+   const { _hex: pairBalanceHex } = await contract.balanceOf(pairContractAddress)
+   const tvlInWei = BigNumber.from(pairBalanceHex).mul(BigNumber.from(2))
+   return tvlInWei.div(BigNumber.from("10000000000000000000"))
```
cds95, line 64: Is there a utility method for this?

packages/sources/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,65 @@
+import { Requester, Validator, Logger } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig } from '@chainlink/types'
+import { getRpcLatestAnswer } from "@chainlink/ea-reference-data-reader"
+import { JSON_RPC_URL } from '../config'
+import { ethers, BigNumber } from 'ethers'
+
+export const NAME = 'TVL'
+
+const customParams = {
+  wethContractAddress: true,
```
cds95, line 10: I had thought about setting this to something more generic but it looks like we may only be querying from the WETH token address to calculate TVL.  Interested to hear any opinions on this.

packages/sources/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,65 @@
...
+  pairContractAddress: true,
+  xdaiEthUsdPriceFeedAddress: true,
```
cds95, line 12: Also thought about naming this something more generic such as `priceFeedModifierAddressOnXDai` though I'm not sure why we would use a feed other than ETH-USD.  Also interested in hearing any thoughts.

packages/sources/dxdao/src/endpoint/tvl.ts
boxhock: Don't know if it will work, but is it possible to do `BigNumber.from("1e19")` (or however many zeros there are)?
In Gauntlet I previously did something like `BigNumber.from(1e6).mul(1e6) ...`, which wasn't pretty.

Oh also, BigNumber methods _should_ accept `number | string | BigNumber` iirc. So I think `tvlInWei.div(BigNumber.from("10000000000000000000")) == tvlInWei.div("10000000000000000000")`

packages/sources/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,64 @@
...
+      Logger.info(`Fetching USD/ETH price from XDai price feed address ${xdaiEthUsdPriceFeedAddress}.`)
+      // Price feed returns with 8 decimal places
+      const USDPerETH = await getRpcLatestAnswer(xdaiEthUsdPriceFeedAddress, 10**8)
+      response.data.result = tvlInEth.mul(USDPerETH).toString()
+   }
+   return Requester.success(jobRunID, response, config.verbose)
+}
+
+const getTvlAtAddressInETH = async (pairContractAddress: string, wethContractAddress: string): Promise<BigNumber> => {
+   const jsonRpcUrl = util.getRequiredEnv('RPC_URL')
```
boxhock, line 57: We'll probably want to add this to the config object, and document this in the README.

See the example adapter I sent on Slack for inspiration.

packages/sources/dxdao/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+      }
+    }
+  }
+}))
+
+jest.mock("@chainlink/ea-reference-data-reader", () => ({
+  ...(jest.requireActual("@chainlink/ea-reference-data-reader")),
+  getRpcLatestAnswer: jest.fn().mockReturnValue(2020)
+}))
```
boxhock, line 26: Nice job with the mocks! :+1: this will be useful in a lot of other adapters too

boxhock (changes requested):
Sent a message on Slack as well, we should consider using Token Allocation for this instead of relying on price feeds.

packages/sources/dxdao/test/integration/adapter.test.ts
cds95: thanks!

packages/composites/dxdao/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Dxdao Endpoint
+
+This endpoint fetches the TVL(Total Value Locked) inside a pair that is deployed on the XDai chain.  The TVL is returned in USD.
+
+### Input Params
+
+| Required? |            Name            |               Description                |       Options       | Defaults to |
+| :-------: | :------------------------: | :--------------------------------------: | :-----------------: | :---------: |
+|    ✅     | `wethContractAddress`  |   The WETH contract address on the XDai Chain    |  |             |
```
boxhock, line 29: Do you think it would make sense to make this an env var?
WETH contract address should be dependent on which RPC_URL you have set (which blockchain you're connected to)

packages/composites/dxdao/README.md
```diff
@@ -0,0 +1,82 @@
...
+|    ✅     | `pairContractAddress` | The pair contract's address on the XDai Chain |   |             |
+|         | `tokenAllocationSource` | Which source to fetch the Ethereum USD price from.  Including this will return the TVL in USD |   |             |
```
boxhock, line 31: Instead of giving this it's own param name, we should just pass on the entire input to TA and refer to the TA readme here: https://github.com/smartcontractkit/external-adapters-js/tree/develop/packages/composites/token-allocation#readme

boxhock (changes requested):
It also looks like a few of the tests aren't passing, see CI

packages/composites/dxdao/README.md
cds95: I think having it as an env var is fine too.  I don't really have a preference on this.

packages/composites/dxdao/README.md
cds95: Wouldn't we have to still construct the input in the code as we are passing in the Ethereum balance at run time?

packages/composites/dxdao/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Dxdao Endpoint
+
+This endpoint fetches the TVL(Total Value Locked) inside a pair that is deployed on the XDai chain.  The TVL is returned in USD.
+
+### Input Params
+
+| Required? |            Name            |               Description                |       Options       | Defaults to |
+| :-------: | :------------------------: | :--------------------------------------: | :-----------------: | :---------: |
+|    ✅     | `pairContractAddress` | The pair contract's address on the XDai Chain |   |             |
```
boxhock, line 30:
```suggestion
---

## Dxdao Endpoint

This endpoint fetches the TVL(Total Value Locked) inside a pair that is deployed on the XDai chain.  The TVL is returned in USD.

### Input Params

| Required? |            Name            |               Description                |       Options       | Defaults to |
| :-------: | :------------------------: | :--------------------------------------: | :-----------------: | :---------: |
|    ✅     | `pairContractAddress` | The pair contract's address on the XDai Chain |   |             |

**Additional environment input params must be set according to the Token Allocation adapter.**
This composite adapter executes the code from the Token Allocation composite adapter. As such the same configuration and input parameters apply to this adapter. See [../token-allocation/README.md](../token-allocation/README.md) for more details.
```

packages/composites/dxdao/schemas/env.json
```diff
@@ -0,0 +1,25 @@
...
+      "format": "uri"
+    },
+    "WETH_CONTRACT_ADDRESS": {
+      "type": "string",
+      "format": "uri"
+    },
+    "OPTION": {
+      "default": true,
+      "type": "boolean"
+    }
```
boxhock, line 18:
```suggestion
```

packages/composites/dxdao/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+### Sample Input for Fetching TVL In USD
+
+```json
+{
+    "id": 1,
+    "data": {
+        "wethContractAddress": "0x6A023CCd1ff6F2045C3309768eAd9E68F978f6e1",
```
boxhock, line 39:
```suggestion
```

packages/composites/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const getTvlAtAddressInWei = async (pairContractAddress: string, wethContractAddress: string, jsonRpcUrl: string): Promise<BigNumber> => {
+   const provider = new ethers.providers.JsonRpcProvider(jsonRpcUrl)
+   Logger.info(`Fetching TVL for contract '${pairContractAddress}' using WETH contract address ${wethContractAddress}`)
+   const contract = new ethers.Contract(wethContractAddress, dxdWethContractAbi, provider)
+   const { _hex: pairBalanceHex } = await contract.balanceOf(pairContractAddress)
+   const tvlInWei = BigNumber.from(pairBalanceHex).mul(2)
+   return tvlInWei
+}
```
boxhock, line 55: missing newline

packages/composites/dxdao/src/endpoint/tvl.ts
```diff
@@ -0,0 +1,55 @@
...
+      decimals: 18
+   }]
+}
+
+const getTvlAtAddressInWei = async (pairContractAddress: string, wethContractAddress: string, jsonRpcUrl: string): Promise<BigNumber> => {
+   const provider = new ethers.providers.JsonRpcProvider(jsonRpcUrl)
+   Logger.info(`Fetching TVL for contract '${pairContractAddress}' using WETH contract address ${wethContractAddress}`)
+   const contract = new ethers.Contract(wethContractAddress, dxdWethContractAbi, provider)
+   const { _hex: pairBalanceHex } = await contract.balanceOf(pairContractAddress)
+   const tvlInWei = BigNumber.from(pairBalanceHex).mul(2)
```
boxhock, line 53: Does `.balanceOf()` not already return a BigNumber?

Would this work?

```suggestion
   const tvlInWei = (await contract.balanceOf(pairContractAddress)).mul(2)
```

packages/composites/dxdao/test-payload.json
```diff
@@ -0,0 +1,10 @@
+{
+    "request": {
+        "id": 1,
+        "data": {
+            "wethContractAddress": "0x6A023CCd1ff6F2045C3309768eAd9E68F978f6e1",
```
boxhock, line 5:
```suggestion
```

boxhock (changes requested):
Left a few suggested changes. Looks like CI is not passing

packages/composites/dxdao/src/endpoint/tvl.ts
cds95: Apparently not.  I had tried this and got an error saying that `.mul` is not a function.

packages/composites/dxdao/schemas/env.json
```diff
@@ -0,0 +1,21 @@
...
+  "required": ["RPC_URL"],
+  "type": "object",
+  "properties": {
+    "RPC_URL": {
+      "type": "string",
+      "format": "uri"
+    },
+    "WETH_CONTRACT_ADDRESS": {
+      "type": "string",
+      "format": "uri"
```
boxhock, line 13:
```suggestion
```

packages/composites/dxdao/src/endpoint/tvl.ts
boxhock: Can you do `... as BigNumber`? I'm pretty sure the `_hex` you're getting is the hex from the BigNumber.

At least the ethers.js documentation says so: https://docs.ethers.io/v5/single-page/#/v5/api/contract/contract/-%23-Contract-functionsCall

boxhock (changes requested):
just a couple small things

boxhock (approved):

justinkaseman (approved):
Great work!

PR #706, opened 2021-07-22
Composite adapter integration tests

cds95:
Add integration tests for token allocation and apy finance composite adapters.

### DIFF
--- a/packages/composites/token-allocation/test/integration/adapter.test.ts
+++ b/packages/composites/token-allocation/test/integration/adapter.test.ts
@@ -1,34 +1,151 @@
-import { Requester } from '@chainlink/ea-bootstrap'
-import { assertError } from '@chainlink/ea-test-helpers'
-import { AdapterRequest } from '@chainlink/types'
-import { makeExecute } from '../../src/adapter'
-import { makeConfig } from '../../src/config'
+import { AdapterRequest, Execute } from '@chainlink/types'
+import * as tokenAllocationAdapter from "../../src/index" 
+import { Requester } from "@chainlink/ea-bootstrap"
+
+const jobRunID = 1
+const sources = {
+    "amberdata": {
+        providerUrlEnvVar: `AMBERDATA_DATA_PROVIDER_URL`
+    }, 
+    "coinapi": {
+        providerUrlEnvVar: `COINAPI_DATA_PROVIDER_URL`
+    }, 
+    "coingecko": {
+        providerUrlEnvVar: `COINGECKO_DATA_PROVIDER_URL`,
+        isBatched: true 
+    }, 
+    "coinmarketcap": {
+        providerUrlEnvVar: `COINMARKETCAP_DATA_PROVIDER_URL`,
+        isBatched: true
+    }, 
+    "coinpaprika": {
+        providerUrlEnvVar: `COINPAPRIKA_DATA_PROVIDER_URL`
+    }, 
+    "cryptocompare": {
+        providerUrlEnvVar: `CRYPTOCOMPARE_DATA_PROVIDER_URL`,
+        isBatched: true
+    }, 
+    "kaiko": {
+        providerUrlEnvVar: `KAIKO_DATA_PROVIDER_URL`
+    }, 
+    "nomics": {
+        providerUrlEnvVar: `NOMICS_DATA_PROVIDER_URL`,
+        isBatched: true
+    }, 
+    "tiingo": {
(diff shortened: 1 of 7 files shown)


packages/composites/crypto-volatility-index/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`execute with isAdaptive false should return success 1`] = `
+Object {
+  "data": Object {
+    "result": NaN,
+  },
+  "debug": undefined,
+  "jobRunID": "1",
+  "result": NaN,
```
cds95, line 10: I spent quite a bit of time trying to configure the mock data to give valid responses but haven't had any luck.  I tried running the adapter end to end but this didn't work either as the API responses from `https://www.deribit.com/api/v2/public` didn't return a response that allowed the adapter to return a 200.  Open to other suggestions here.

packages/composites/crypto-volatility-index/src/sigmaCalculator.ts
```diff
@@ -46,7 +44,7 @@ export class SigmaCalculator {
       const { strikePrice, midPrice, underlyingPrice } = call
       if (strikePrice.gt(underlyingPrice)) {
         if (!midPrice) return
-        const prevMax: Decimal = Decimal.max(...compareCalls.slice(Math.max(0, idx - 10), idx))
+        const prevMax: Decimal = Decimal.max(...[0, ...compareCalls.slice(Math.max(0, idx - 10), idx)])
```
justinkaseman, line 47: Does this fix the `no date found before` thing?

cds95: It doesn't fix it but it fixed another issue I was hitting.  I can't remember the exact issue but I'll repro it tomorrow and get back to you.

cds95: So I removed this and ran the code again and got this error

```
[DecimalError] Invalid argument: undefined
```

I think it's because `[...compareCalls.slice(Math.max(0, idx - 10), idx)]` by itself returns an empty array, which causes `Decimal.max` to error out.

packages/composites/token-allocation/test/integration/adapter.test.ts
```diff
@@ -1,34 +1,151 @@
...
+    }, 
+    "tiingo": {
+        providerUrlEnvVar: `TIINGO_DATA_PROVIDER_URL`
+    }
+}
+Requester.request = (request: any) => {
+    const source = sources[request.baseURL]
+    const { quote } = request.data.data 
+    if(source.isBatched || (request.baseURL === "coinapi" && request.data.data.quote === "USD")) {
+        return {
```
HenryNguyen5, line 43: Can we extract out fixture data?

packages/composites/token-allocation/test/integration/adapter.test.ts
```diff
@@ -1,34 +1,151 @@
...
+        isBatched: true
+    }, 
+    "tiingo": {
+        providerUrlEnvVar: `TIINGO_DATA_PROVIDER_URL`
+    }
+}
+Requester.request = (request: any) => {
+    const source = sources[request.baseURL]
+    const { quote } = request.data.data 
+    if(source.isBatched || (request.baseURL === "coinapi" && request.data.data.quote === "USD")) {
```
HenryNguyen5, line 42: We should use `nock` instead here if possible, that way we dont need to mock out `Requester.request`

HenryNguyen5 (changes requested):

packages/composites/crypto-volatility-index/test/integration/__snapshots__/adapter.test.ts.snap
cds95: @justinkaseman and I noticed an issue when we ran the adapters e2e locally so it looks like even mocking actual responses won't work.

packages/composites/crypto-volatility-index/package.json
```diff
@@ -37,6 +37,8 @@
     "@chainlink/types": "0.0.1",
     "@types/jest": "^26.0.22",
     "@types/node": "^14.14.35",
+    "ethers": "5.0.32",
```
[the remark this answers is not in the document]
cds95: It looks like this is the same version as what other libraries are using.

PR #711, opened 2021-07-23
Layer 2 Sequencer Status Check adapter

RodrigoAD:
Some context behind the Adapter: https://www.notion.so/chainlink/L2-Sequencer-Health-EA-be0f84b92b484e45b1b1f6bf9e83b4b2

Part of the Layer 2 Emergency protocol: https://www.notion.so/chainlink/L2-Emergency-Protocol-Sequencer-offline-c184e0919da441b8b824db35f748db29

### DIFF
--- a/packages/sources/layer2-sequencer-health/README.md
+++ b/packages/sources/layer2-sequencer-health/README.md
@@ -8,6 +8,12 @@ Adapter that checks the Layer 2 Sequencer status
 | Required? |  Name   |                                                        Description                                                         | Options | Defaults to |
 | :-------: | :-----: | :------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
 |           | `DELTA` | Maximum time in miliseconds from last seen block for the sequencer to be considered healthy |         |     180000 (3 min)        |
+|           | `ARBITRUM_RPC_ENDPOINT` | Arbitrum RPC Endpoint |         |     https://arb1.arbitrum.io/rpc       |
+|           | `ARBITRUM_HEALTH_ENDPOINT` | Arbitrum Health Endpoint |         |   |
+|           | `OPTIMISM_RPC_ENDPOINT` | Optimism RPC Endpoint |         |     https://mainnet.optimism.io       |
+|           | `OPTIMISM_HEALTH_ENDPOINT` | Optimism Health Endpoint |         |     https://mainnet-sequencer.optimism.io/health       |
+
+For the adapter to be useful on the desired network, at least one endpoint (RPC or HEALTH) needs to provided
 
 ---
 
@@ -35,7 +41,8 @@ Adapter that checks the Layer 2 Sequencer status
 {
   "jobRunID": "278c97ffadb54a5bbb93cfec5f7b5503",
   "data": {
-    "healthy": true
+    "result": true,
+    "isHealthy": true
   },
   "statusCode": 200
 }
--- a/packages/sources/layer2-sequencer-health/README.md
+++ b/packages/sources/layer2-sequencer-health/README.md
@@ -2,16 +2,15 @@
 
 Adapter that checks the Layer 2 Sequencer status
 
-
 ### Environment Variables
 
-| Required? |  Name   |                                                        Description                                                         | Options | Defaults to |
-| :-------: | :-----: | :------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
-|           | `DELTA` | Maximum time in miliseconds from last seen block for the sequencer to be considered healthy |         |     180000 (3 min)        |
-|           | `ARBITRUM_RPC_ENDPOINT` | Arbitrum RPC Endpoint |         |     https://arb1.arbitrum.io/rpc       |
-|           | `ARBITRUM_HEALTH_ENDPOINT` | Arbitrum Health Endpoint |         |   |
-|           | `OPTIMISM_RPC_ENDPOINT` | Optimism RPC Endpoint |         |     https://mainnet.optimism.io       |
-|           | `OPTIMISM_HEALTH_ENDPOINT` | Optimism Health Endpoint |         |     https://mainnet-sequencer.optimism.io/health       |
+| Required? |            Name            |                                  Description                                   | Options |                 Defaults to                  |
+| :-------: | :------------------------: | :----------------------------------------------------------------------------: | :-----: | :------------------------------------------: |
+|           |          `DELTA`           | Maximum time in miliseconds from last seen block to consider sequencer healthy |         |                120000 (2 min)                |
+|           |  `ARBITRUM_RPC_ENDPOINT`   |                             Arbitrum RPC Endpoint                              |         |         https://arb1.arbitrum.io/rpc         |
+|           | `ARBITRUM_HEALTH_ENDPOINT` |                            Arbitrum Health Endpoint                            |         |                                              |
+|           |  `OPTIMISM_RPC_ENDPOINT`   |                             Optimism RPC Endpoint                              |         |         https://mainnet.optimism.io          |
+|           | `OPTIMISM_HEALTH_ENDPOINT` |                            Optimism Health Endpoint                            |         | https://mainnet-sequencer.optimism.io/health |
 
 For the adapter to be useful on the desired network, at least one endpoint (RPC or HEALTH) needs to provided
 
@@ -19,11 +18,12 @@ For the adapter to be useful on the desired network, at least one endpoint (RPC 
 
 ### Input Parameters
 
-| Required? |   Name   |     Description     |           Options            | Defaults to |
-| :-------: | :------: | :-----------------: | :--------------------------: | :---------: |
-|     ✅     | network | Layer 2 Network to check | arbitrum or optimism |     |
+| Required? |  Name   |       Description        |       Options        | Defaults to |
+| :-------: | :-----: | :----------------------: | :------------------: | :---------: |
+|    ✅     | network | Layer 2 Network to check | arbitrum or optimism |             |
 
 ---
+
 ### Sample Input
 
 ```json
--- a/packages/sources/layer2-sequencer-health/README.md
+++ b/packages/sources/layer2-sequencer-health/README.md
@@ -1,4 +1,4 @@
-# Chainlink External Adapter for Example
+# Chainlink External Adapter for Layer 2 Sequencer Health Check
 
 Adapter that checks the Layer 2 Sequencer status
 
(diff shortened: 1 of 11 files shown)


packages/sources/layer2-sequencer-health/schemas/env.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/example-source-adapter.json",
+  "title": "@chainlink/example-source-adapter env var schema",
+  "required": [],
+  "type": "object",
+  "properties": {},
```
boxhock, line 6:
```suggestion
  "properties": {
    "DELTA": {
      "type": "number",
      "default": 180000
    }
  },
```

packages/sources/layer2-sequencer-health/src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+export const HEALTH_ENDPOINTS = {
+  [Networks.Arbitrum]: {
+    endpoint: undefined,
+    responsePath: [],
+  },
+  [Networks.Optimism]: {
+    endpoint: 'https://mainnet-sequencer.optimism.io/health',
+    responsePath: ['healthy'],
+  },
+}
```
boxhock, line 28: Are these endpoints considered completely static? Should we add the ability to override these with ENV vars?

Specifically thinking about `https://arb1.arbitrum.io/rpc`. "arb1" implies there could be an "arb2". Can node ops switch if "arb1" goes down?

Could also be useful if they want to use testnet endpoints.

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -0,0 +1,30 @@
...
+    method: 'POST',
+    url: RPC_ENDPOINTS[network],
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    data: {
+      jsonrpc: '2.0',
+      method: 'eth_blockNumber',
+      params: [],
+      id: 83,
```
boxhock, line 15: Is this an arbitrary ID?

boxhock (changes requested):
Looks good for the most part. Left just a few comments

packages/sources/layer2-sequencer-health/src/config.ts
RodrigoAD: That sounds better, I'll change it

packages/sources/layer2-sequencer-health/src/network.ts
RodrigoAD: It is, I'll remove it

boxhock (dismissed):
LGTM

RodrigoAD:
Thanks for the review @boxhock. I'll leave it pending as it needs some review over the functionality

packages/sources/layer2-sequencer-health/README.md
```diff
@@ -0,0 +1,49 @@
+# Chainlink External Adapter for Example
+
+Adapter that checks the Layer 2 Sequencer status
+
+
+### Environment Variables
+
+| Required? |  Name   |                                                        Description                                                         | Options | Defaults to |
+| :-------: | :-----: | :------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|           | `DELTA` | Maximum time in miliseconds from last seen block for the sequencer to be considered healthy |         |     180000 (3 min)        |
```
se3000, line 10: Do we have any data around this? I'd like to make it smaller if possible, 1 or 2 minutes.

packages/sources/layer2-sequencer-health/src/adapter.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  // #2 Option: Check block height (getNetworkStatus)
+  let isHealthyFromNetwork
+  try {
+    isHealthyFromNetwork = await getNetworkStatus(network, config.delta)
+  } catch (e) {
+    Logger.error(`Network health check failed: ${e.message}`)
+  }
+
+  const isHealthy = !!isHealthyFromSequencer || !!isHealthyFromNetwork
```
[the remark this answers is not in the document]
se3000: I would prefer to err on the side of caution: if any of the three methods fails, that is an indication that there is a problem with the health of the network and we should raise the flag.

packages/sources/layer2-sequencer-health/src/adapter.ts
```diff
@@ -0,0 +1,81 @@
...
+  const _updateLastSeenBlock = (block: number): void => {
+    lastSeenBlock = {
+      block,
+      timestamp: Date.now(),
+    }
+  }
+
+  return async (delta: number): Promise<boolean> => {
+    const block = await requestBlockHeight(network)
+    if (!_isValidBlock(block)) throw new Error('Block found is previous to last seen')
```
se3000, line 27: Shouldn't this return false instead of erroring? Seems like we'd want to raise a flag.

packages/sources/layer2-sequencer-health/src/adapter.ts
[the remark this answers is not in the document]
se3000: Also, based on the first `!!` it looks like the checks only really check for the presence of the object, but any body could pass, maybe even a non 200 status code would be accepted(idk this library). We should check the body to make sure that the sequencer health check actually returned the body we expected.

se3000 (changes requested):

packages/sources/layer2-sequencer-health/src/adapter.ts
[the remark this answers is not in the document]
RodrigoAD: Sure, that's being checked already: https://github.com/smartcontractkit/external-adapters-js/pull/711/files#diff-bc93b36b44ed77b8fc2e088f0db7b6fd451afbbf181c9c107741585db8215a57R19

`Requester` throws on non 200 responses. If block is not on the response, it also errors

packages/sources/layer2-sequencer-health/src/adapter.ts
RodrigoAD: Any error should be considered as a `false` on the consumer

packages/sources/layer2-sequencer-health/src/adapter.ts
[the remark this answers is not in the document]
krebernisak: If the health endpoint is available and it errors or responds as non-healthy (we'd need different response parsing strategies for different networks), then we don't continue but return non-healthy right away.

Only if the first check is ok, we continue to the next one.

packages/sources/layer2-sequencer-health/src/config.ts
```diff
@@ -0,0 +1,39 @@
+import { Requester, util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const NAME = 'LAYER2_SEQUENCER_HEALTH'
+
+// 3 minutes
+export const DEFAULT_DELTA_TIME = 3 * 60 * 1000
+
+export enum Networks {
```
krebernisak, line 9: Reads better if renamed as just `Network` type.

packages/sources/layer2-sequencer-health/src/config.ts
```diff
@@ -0,0 +1,39 @@
+import { Requester, util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const NAME = 'LAYER2_SEQUENCER_HEALTH'
```
krebernisak, line 4:
```suggestion
export const NAME = 'L2_SEQUENCER_HEALTH'
```

packages/sources/layer2-sequencer-health/src/adapter.ts
```diff
@@ -0,0 +1,81 @@
...
+  try {
+    isHealthyFromSequencer = await getSequencerHealth(network)
+  } catch (e) {
+    Logger.error(`Direct sequencer check failed: ${e.message}`)
+  }
+
+  // #2 Option: Check block height (getNetworkStatus)
+  let isHealthyFromNetwork
+  try {
+    isHealthyFromNetwork = await getNetworkStatus(network, config.delta)
```
krebernisak, line 67: Consider renaming this method as `getL2NetworkStatus` and add a mocked third check `getL1RollupStatus` that always returns `true` as healthy, for now.

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -0,0 +1,35 @@
...
+  const hexBlock = response?.data?.result
+  if (!hexBlock) {
+    throw new Error(`Block number not found on network: ${network}`)
+  }
+  return BigNumber.from(hexBlock).toNumber()
+}
+
+export const getSequencerHealth = async (network: Networks): Promise<boolean> => {
+  if (!HEALTH_ENDPOINTS[network].endpoint) {
+    throw new Error(`Health endpoint not available for network: ${network}`)
```
krebernisak, line 29: If the health endpoint is not set we need to continue with the second check, not error here.

packages/sources/layer2-sequencer-health/src/config.ts
```diff
@@ -0,0 +1,39 @@
...
+    responsePath: ['healthy'],
+  },
+}
+
+export interface ExtendedConfig extends Config {
+  delta: number
+}
+
+export const makeConfig = (prefix?: string): ExtendedConfig => {
+  const config = Requester.getDefaultConfig(prefix)
```
krebernisak, line 36: Consider adding a check if the cache is accidentally enabled via env var, and error early if this is the case. This adapter should not be cached.

RodrigoAD: Nice point

packages/sources/layer2-sequencer-health/README.md
RodrigoAD: Really depends on the Network workload, more transactions, more blocks. Assuming that our L2 feeds will be updated quite frequently, I think we can lower it to 1 or 2

packages/sources/layer2-sequencer-health/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@chainlink/layer2-sequencer-health",
```
boxhock, line 2:
```suggestion
  "name": "@chainlink/layer2-sequencer-health-adapter",
```

packages/sources/layer2-sequencer-health/schemas/env.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/example-source-adapter.json",
+  "title": "@chainlink/example-source-adapter env var schema",
```
boxhock, line 3:
```suggestion
  "$id": "https://external-adapters.chainlinklabs.com/schemas/layer2-sequencer-health-adapter.json",
  "title": "@chainlink/layer2-sequencer-health-adapter env var schema",
```

boxhock (changes requested):

boxhock (approved):

PR #715, opened 2021-07-27
Refactor coinmarketcap crypto tests

cds95:


packages/sources/coinmarketcap/src/endpoint/globalMarketCap.ts
```diff
@@ -27,6 +27,7 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   response.data.result = Requester.validateResultNumber(response.data, [
     'data',
     'quote',
+    'quote',
```
cds95, line 30: Wanted to double check this.  I had to add this to get the adapter to work.  Is there an existing bug with this adapter?

packages/composites/token-allocation/src/config.ts
```diff
@@ -1,7 +1,7 @@
 import Amberdata from '@chainlink/amberdata-adapter'
 import CoinApi from '@chainlink/coinapi-adapter'
 import CoinGecko from '@chainlink/coingecko-adapter'
-import CoinMarketCap from '@chainlink/coinmarketcap-adapter'
+import * as CoinMarketCap from '@chainlink/coinmarketcap-adapter'
```
cds95, line 4: Had to update this as the variables are being exported differently from the coinmarketcap adapter.  Open to suggestions on a better way to do this.

packages/sources/coinmarketcap/src/endpoint/globalMarketCap.ts
justinkaseman: It's working for me from the current `develop` branch :thinking:

packages/composites/token-allocation/src/config.ts
justinkaseman: Coinmarketcap looks the same as the other adapters.

What error was it hitting?

packages/sources/coinmarketcap/src/endpoint/globalMarketCap.ts
cds95: I must have been looking at another endpoint's response.  Confirmed that this change is not needed.

packages/composites/token-allocation/src/config.ts
cds95: I had to change the way variables were exported to get the tests to work.

```
const server = expose(NAME, makeExecute(), undefined, endpointSelector).server

export { NAME, makeConfig, makeExecute, server }
```

justinkaseman (approved):

packages/composites/token-allocation/src/config.ts
HenryNguyen5: This looks like the correct way of importing to me, before with the `export =` syntax, TS was treating it as a "default export" which was one object that contained keys referencing the exports. Now with the `export {}` syntax, TS treats it as an ES module which has named exports like `NAME, makeConfig, makeExecute`. This is what results in the two different import styles we have now.

HenryNguyen5 (approved):
Looks great, nice job!

PR #717, opened 2021-07-27
Wootrade

bogos:


packages/sources/wootrade/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,37 @@
...
+  const limit = '1'
+  const url = `/v1/public/market_trades/`
+
+  const params = {
+    symbol,
+    limit
+  }
+
+  const options = { ...config.api, params, url }
+  const response = await Requester.request(options, customError)
```
boxhock, line 32: `Requester.request` also supports taking a type for the response. Could we add the type of the response schema, and also export it? Here's an example: https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/coinmetrics/src/endpoint/price.ts

boxhock (changes requested):

packages/sources/wootrade/README.md
```diff
@@ -0,0 +1,55 @@
+# Chainlink External Adapter for Wootrade
+
+Adapter using the public Wootrade market API for both HTTP(s) and WS.
+
+### Environment Variables
+
+| Required? |  Name   |                                                        Description                                                         | Options | Defaults to |
+| :-------: | :-----: | :------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|     ✅      | API_KEY | An key to use the wootrade WS API |         |             |
```
boxhock, line 9: Do you only need an API key to use the WS? Or HTTP as well?

Also, where do you sign up for an API key? I tried at https://x.woo.network/, but it's not letting me submit the signup form :confused:

bogos: Only for ws, the documentation is here: https://kronosresearch.github.io/wootrade-documents/#introduction

boxhock:
Closing in favor of #751

PR #722, opened 2021-07-29
Add batch requesting support for Forex EAs

silaslenihan:
This PR Includes changes for the following adapters: (and their from to to abilities)
Polygon: from: many to: many 
OpenExchangeRates: from: one to:many
1Forge: from: many to: many
Tradermade: from: many to: many
MetalsAPI: from: one to: many
CurrencyLayer: from: one to: many (pending acceptance testing w/ Valid API Key)
Fixer: from: one to: many (pending acceptance testing w/ Valid API Key)

---------------

1forge - added /quotes endpoint

currencylayer - added /live endpoint
Exchange rate data is refreshed every 60 minutes for Free & Basic customers, every 10 minutes for the Professional Plan, and every 60 seconds for Business customers. 

fixer - added /latest endpoint 
Depending on your subscription plan, the API's latest endpoint will return real-time exchange rate data updated every 60 minutes, every 10 minutes or every 60 seconds.

metalsapi - added /latest endpoint
Depending on your subscription plan, the API's latest endpoint will return real-time exchange rate data updated every 60 minutes, every 10 minutes or every 60 seconds.

polygon - added /snapshot endpoint
uses 1 min sliding window close price

openexchangerate - batching added to existing endpoint
tradermade - batching added to existing endpoint


packages/sources/polygon/src/endpoint/tickers.ts
```diff
@@ -0,0 +1,78 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+//import { warmupStopped } from '@chainlink/ea-bootstrap/src/lib/cache-warmer/actions'
```
justinkaseman, line 2:
```suggestion
```

packages/sources/polygon/src/endpoint/tickers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const handleBatchedRequest = (
+  jobRunID: string,
+  request: AdapterRequest,
+  response: AxiosResponse<any>,
+  resultPath: string[],
+) => {
+  const payload: [AdapterRequest, number][] = []
+  for (const pair of response.data.tickers) {
+    const base = pair.ticker.substring(2,5)
```
justinkaseman, line 23: Does this handle tickers that are more than `3` characters?

packages/sources/tradermade/src/adapter.ts
```diff
@@ -26,10 +55,22 @@ export const execute: ExecuteWithConfig<Config> = async (input, _, config) => {
   const options = { ...config.api, params }
 
   const response = await Requester.request(options)
+  if (Array.isArray(symbol) || Array.isArray(to)) return handleBatchedRequest(jobRunID, request, response, 'mid')
+
   response.data.result = Requester.validateResultNumber(response.data, ['quotes', 0, 'mid'])
-  return Requester.success(jobRunID, response)
+  return Requester.success(jobRunID, response, config.api.verbose, ['from', 'to'])
```
justinkaseman, line 61: Needs the update for a separately exported `batchablePropertyPath`

packages/sources/openexchangerates/src/endpoint/forex.ts
```diff
@@ -30,7 +52,9 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   }
 
   const response = await Requester.request(options)
+  if (Array.isArray(to)) return handleBatchedRequest(jobRunID, request, response, 'rates', to)
```
justinkaseman, line 55: Both `base` and `quote` are batchable. This conditional should check for either being an array.

packages/sources/openexchangerates/src/endpoint/forex.ts
```diff
@@ -30,7 +52,9 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
+
   response.data.result = Requester.validateResultNumber(response.data, ['rates', to])
 
-  return Requester.success(jobRunID, response, config.verbose)
+  return Requester.success(jobRunID, response, config.verbose, ['base', 'quote'])
```
justinkaseman, line 59: Needs the change for `batchablePropertyPath` exported.

packages/sources/openexchangerates/README.md
```diff
@@ -23,195 +23,124 @@
...
 | :-------: | :------------------------: | :-------------------------------------------------------: | :----------------------------------------------------------------------------------: | :---------: |
 |    ✅     | `base`, `from`, or `coin`  |            The symbol of the currency to query            |                                                                                      |             |
-|    ✅     | `quote`, `to`, or `market` |         The symbol of the currency to convert to          |                                                                                      |             |
+|    ✅     | `quote`, `to`, or `market` |         The symbol of the currency or currenicies to convert to          |                                                                                      |             |
 |    🟡     |        `overrides`         | If base provided is found in overrides, that will be used | [Format](../../core/bootstrap/src/lib/external-adapter/overrides/presetSymbols.json) |             |
 
-### Output
+## Singular Currnecy Pair Example
```
justinkaseman, line 29:
```suggestion
## Singular Currency Pair Example
```

packages/sources/metalsapi/src/endpoint/latest.ts
```diff
@@ -0,0 +1,59 @@
...
+  const params = {
+    access_key: config.apiKey,
+    from
+  }
+
+  const reqConfig = { ...config.api, params, url }
+
+
+  const response = await Requester.request(reqConfig, customError)
+  if (Array.isArray(to)) return handleBatchedRequest(jobRunID, request, response, 'rates', to)
```
justinkaseman, line 55: Both base and quote are batchable. This conditional should check for either being an array.

packages/sources/metalsapi/src/config.ts
```diff
@@ -4,7 +4,7 @@ import { Config } from '@chainlink/types'
 export const NAME = 'METALSAPI'
 
 export const DEFAULT_BASE_URL = 'https://metals-api.com/api/'
-export const DEFAULT_ENDPOINT = 'convert'
+export const DEFAULT_ENDPOINT = 'latest'
```
justinkaseman, line 7:
```suggestion
export const DEFAULT_ENDPOINT = 'convert'
```
We should leave this as is for backwards compatibility.

packages/sources/fixer/src/endpoint/latest.ts
```diff
@@ -0,0 +1,62 @@
...
+    access_key: config.apiKey,
+    base: from,
+    symbols: to.join()
+  }
+
+  const reqConfig = { ...config.api, params, url }
+
+
+  const response = await Requester.request(reqConfig, customError)
+  if (Array.isArray(to)) return handleBatchedRequest(jobRunID, request, response, 'rates', to)
```
justinkaseman, line 58: If base and quote are batchable this conditional should check for either being an array.

packages/sources/currencylayer/src/endpoint/live.ts
```diff
@@ -0,0 +1,63 @@
...
+    access_key: config.apiKey,
+    source: from,
+    currencies: to.join()
+  }
+
+  const reqConfig = { ...config.api, params, url }
+
+
+  const response = await Requester.request(reqConfig, customError)
+  if (Array.isArray(to)) return handleBatchedRequest(jobRunID, request, response, 'quotes', to)
```
justinkaseman, line 59: If base and quote are both batchable this conditional should check for either being an array.

packages/sources/currencylayer/src/endpoint/live.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const jobRunID = validator.validated.id
+  const from = validator.overrideSymbol(AdapterName)
+  const to = validator.validated.data.quote
+  const url = `live`
+
+  const params = {
+    access_key: config.apiKey,
+    source: from,
+    currencies: to.join()
```
justinkaseman, line 52: What if `to` is just a string?

Individual feed requests come in as: `to: 'USD'`.

We need to handle both arrays and strings.

packages/sources/1forge/src/endpoint/quotes.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+  const url = `/quotes`
+  const from = validator.overrideSymbol(AdapterName)
+  const to = validator.validated.data.quote
+  const pairArray = []
+
+  for (const fromCurrency of formatArray(from)) {
+    for (const toCurrency of formatArray(to)) {
+      pairArray.push(`${fromCurrency.toUpperCase()}/${toCurrency.toUpperCase()}`)
```
justinkaseman, line 48: What if to is just a string?

Individual feed requests come in as: to: 'USD'.

We need to handle both arrays and strings.

packages/sources/polygon/src/endpoint/tickers.ts
silaslenihan: No it wouldn't, I didn't know that there could be ones that are longer than 3 characters...

packages/sources/1forge/src/endpoint/quotes.ts
silaslenihan: The `formatArray` function makes it so both strings and arrays are handled. If its an array, it returns the array, if its a string it makes it into an array w/ one string.

packages/sources/openexchangerates/src/endpoint/forex.ts
silaslenihan: Only quote / to is batchable in this case

packages/sources/metalsapi/src/endpoint/latest.ts
silaslenihan: Same situation here - only quote / to are batchable

packages/sources/currencylayer/src/endpoint/live.ts
silaslenihan: Same situation here - only quote / to are batchable

packages/sources/currencylayer/src/endpoint/live.ts
silaslenihan: Fixed this in most recent commit

packages/sources/1forge/src/endpoint/quotes.ts
```diff
@@ -0,0 +1,79 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+//import { warmupStopped } from '@chainlink/ea-bootstrap/src/lib/cache-warmer/actions'
```
justinkaseman, line 2:
```suggestion
```

packages/sources/1forge/src/endpoint/quotes.ts
justinkaseman: Nice. 
One suggestion for something like this would be to throw it in the core framework's `utils` to re-use it across adapters.

packages/sources/currencylayer/src/endpoint/live.ts
justinkaseman: In that case line 6 should be:
```
export const batchablePropertyPath = ['quote']
```

packages/sources/fixer/src/endpoint/latest.ts
justinkaseman: Repeating other comment:

Line 6 should be:
```
export const batchablePropertyPath = ['quote']
```

packages/sources/metalsapi/src/config.ts
justinkaseman: Change didn't take :open_mouth:

packages/sources/1forge/src/endpoint/quotes.ts
silaslenihan: Did this in most recent commit

justinkaseman (dismissed):
You're awesome, thank you.

HenryNguyen5:
@justinkaseman what's left to do here?

justinkaseman:
> @justinkaseman what's left to do here?

There is a bug or two in the code that I see at a glance, so first I'll need to fix those.

But we are also waiting for the data team to do some discovery on if these API endpoints are the data that we need. Some of them changed to a batchable API endpoint.

boxhock:
Hey @justinkaseman, I see this PR has been open for some time. Did the data team get back on this stuff?

packages/sources/polygon/src/endpoint/tickers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const handleBatchedRequest = (
+  jobRunID: string,
+  request: AdapterRequest,
+  response: AxiosResponse<any>,
+  resultPath: string[],
+) => {
+  const payload: [AdapterRequest, number][] = []
+  for (const pair of response.data.tickers) {
+    const base = pair.ticker.substring(2, 5)
```
justinkaseman, line 28: Still going to possibly cause unforeseen issues. Tickers can be longer than 3.

justinkaseman (approved):
Long time coming :smile: 
Nice work!

PR #744, opened 2021-08-03
Fix for allowing multiple capped batch requests

cds95:


packages/core/bootstrap/test/unit/cache-warmer.test.ts
```diff
@@ -52,6 +52,7 @@ describe('side effect tests', () => {
   }
   const adapterRequest1: AdapterRequest = { data: {}, id: '0' }
   const adapterRequest2: AdapterRequest = { data: { foo: 'bar' }, id: '0' }
+  const adapterRequest3: AdapterRequest = { foo: 'bar', id: '0' }
```
justinkaseman, line 55: What is this new request for?

We expect request data to come in within `data`.

packages/sources/coinmarketcap/src/endpoint/crypto.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '@chainlink/types'
 
 export const supportedEndpoints = ['crypto', 'price', 'marketcap']
-export const batchablePropertyPath = ['base', 'convert']
+export const batchablePropertyPath = [{ name: 'base' }, { name: 'convert' }]
```
justinkaseman, line 12: Since we know the limit here let's add it:
```suggestion
export const batchablePropertyPath = [{ name: 'base' }, { name: 'convert', limit: 120 }]
```

packages/core/bootstrap/src/lib/cache-warmer/util.ts
```diff
@@ -12,3 +15,106 @@ export function getSubscriptionKey(
     conf.hashOpts,
   )
 }
+
+export function splitIntoBatches(requestData: SubscriptionData): {
+  [path: string]: string[]
```
justinkaseman, line 20: Could clean up this file a lot and make it easier to grok by making a couple types that are re-used.

justinkaseman (commented):
Left a couple comments. It would be good to add a test that shows the `limit` being respected.

Maybe by using a stub for `executeFn` that throws if over the limit.

packages/core/bootstrap/test/unit/cache-warmer.test.ts
cds95: So the parent and child warmers have slightly different shapes stored in Redux.  The parent stores it as 

```
{
  origin: {
    base: ...,
    quote: ...
  }
}
```

while the child stores it like this

```
{
   origin: {
      data: {
        base: ...,
        quote: ...
    }
}
```
```

File: packages/core/bootstrap/src/lib/cache-warmer/util.ts (hunk 3 above)
Comment (cds95) [reply packages/core/bootstrap/src/lib/cache-warmer/util.ts:20]: so just create a type for `{ [path: string]: string[] }`?

**Comment by cds95** [comment]
The code doesn't throw an error if there is a limit though.  It'll still succeed by making multiple calls.

File: packages/core/bootstrap/src/lib/cache-warmer/util.ts (hunk 3 above)
Comment (justinkaseman) [reply packages/core/bootstrap/src/lib/cache-warmer/util.ts:20]: There are a lot of types in there that seem to get re-used a lot.

Seems like smallest pieces is something like:

`type BatchPaths = string[]`

which gets used in 
```
interface BatchRequestChunk {
[path: string]:  BatchPaths[]
}
```

which gets used in 

```
interface BatchesByPath {
[path: string]:  BatchRequestChunk[]
}
```

File: packages/core/bootstrap/test/unit/cache-warmer.test.ts (hunk 1 above)
Comment (justinkaseman) [reply packages/core/bootstrap/test/unit/cache-warmer.test.ts:55]: I'm not following. Did you mean that flipped around?

It gets used in the code @ 
https://github.com/smartcontractkit/external-adapters-js/pull/744/files/4056e37cbfd55348b25499946bf5e39d6fd97aea#diff-0b96592c3d7aac6f7a84adba3fd48d9219a737782f051a9875244e668fbcb9efR424
which is setting a child state.

**Comment by justinkaseman** [comment]
> The code doesn't throw an error if there is a limit though. It'll still succeed by making multiple calls.

Right.

So you would have a mock `executeFn` that throws when for example `base` goes over 5.
In the first test you don't provide a limit and send a request with 6 `base`s. You expect it to throw.
In the second test you provide a limit and send a request with 6 `base`s. You expect it succeed by making multiple calls. If using a Jest mock you could also test how many times `executeFn` was called. You expect it to be 2 in this example.

File: packages/core/bootstrap/src/lib/cache-warmer/epics.ts [hunk 4]
```diff
@@ -174,18 +174,44 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
...
     // fetch our required state to make a request to warm up an adapter
     withLatestFrom(state$),
-    map(([action, state]) => ({
-      requestData: state.cacheWarmer.subscriptions[action.payload.key],
-      key: action.payload.key,
-    })),
+    map(([action, state]) => {
+      const requestData = state.cacheWarmer.subscriptions[action.payload.key]
+      let batchablePropertyPath
+      if (requestData && requestData.childLastSeenById && Object.keys(requestData.childLastSeenById).length > 0) {
```
justinkaseman, line 180:
```suggestion
      if (requestData?.childLastSeenById && Object.keys(requestData.childLastSeenById).length > 0) {
```
optional chaining makes this a bit more concise.

also could use lodash.isEmpty as a more robust way to check for an empty object

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -174,18 +174,44 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
     filter(warmupRequested.match),
     // fetch our required state to make a request to warm up an adapter
     withLatestFrom(state$),
-    map(([action, state]) => ({
-      requestData: state.cacheWarmer.subscriptions[action.payload.key],
-      key: action.payload.key,
-    })),
+    map(([action, state]) => {
+      const requestData = state.cacheWarmer.subscriptions[action.payload.key]
```
justinkaseman, line 178: We could using a nullish coalescing operator to avoid doing `Object.keys` twice.

```suggestion
      const requestData = state.cacheWarmer.subscriptions[action.payload.key] ?? {}
      const childKeys = Object.keys(requestData.childLastSeenById ?? {})
```

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -174,18 +174,44 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
...
+      let batchablePropertyPath
```
justinkaseman, line 179: Actually it may just be easier to store `batchablePropertyPath` on the parent/batch warmer as well and skip doing this completely.

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -174,18 +174,44 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
...
+          const batches = splitIntoBatches(requestData)
+          let result = null 
+          for (const batch of Object.values(batches)) {
+            const data = {
+              ...requestData.origin,
+              ...batch,
+              resultPath: getBatchRequestResultPath(requestData.batchablePropertyPath, subscriptions, batch),
+              maxAge: -1
+            }
+            const batchResult = await requestData.executeFn({
```
justinkaseman, line 207: We should send these in parallel rather than awaiting each response with something like `Promise.all`.

packages/core/bootstrap/src/lib/cache-warmer/reducer.ts
```diff
@@ -50,9 +55,12 @@ export interface SubscriptionState {
 export const subscriptionsReducer = createReducer<SubscriptionState>({}, (builder) => {
   builder.addCase(actions.warmupSubscribed, (state, { payload }) => {
     const key = payload.key || getSubscriptionKey(payload)
-
+    const { data, ...rest } = payload.data
```
justinkaseman, line 58: Should be able to remove this now.

packages/core/bootstrap/test/unit/cache-warmer.test.ts
justinkaseman: Should be able to change this now.

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
cds95: We don't have `batchablePropertyPath` when we save the parent in Redux though.

packages/core/bootstrap/src/lib/cache-warmer/reducer.ts
cds95: Wait don't we need this as we are sending payload like

```
{
  data: {
     data: {}
  }
}
```

or are we updating the action's payload?

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -174,18 +175,40 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
     filter(warmupRequested.match),
     // fetch our required state to make a request to warm up an adapter
     withLatestFrom(state$),
-    map(([action, state]) => ({
-      requestData: state.cacheWarmer.subscriptions[action.payload.key],
-      key: action.payload.key,
-    })),
+    map(([action, state]) => {
+      const requestData = state.cacheWarmer.subscriptions[action.payload.key]
+      return {
```
justinkaseman, line 180: Could take out some syntax now:

```
map(([action, state]) => ({
        requestData: state.cacheWarmer.subscriptions[action.payload.key]
        key: action.payload.key,
        subscriptions: state.cacheWarmer.subscriptions 
    }),
```

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -174,18 +175,40 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
...
-          data: { ...requestData.origin, maxAge: -1 },
-        }),
+        (async () => {
+          const batches = splitIntoBatches(requestData)
+          const requests = []
+          for (const batch of Object.values(batches)) {
+            const data = {
+              ...requestData.origin,
+              ...batch,
+              resultPath: getBatchRequestResultPath(requestData.batchablePropertyPath, subscriptions, batch),
```
justinkaseman, line 197: The result path should fill itself based on the `endpoint`.

If you take this out does it still work?

cds95: We'll need to set it to `null` or `undefined` as it's being saved in the parent warmer.  The issue is that the path being saved in the parent warmer might not work with the data being sent in a batched request.

justinkaseman: Yeah, let's filter it out as start or add to the parent warmer.

There's a bigger gap here in having the system handle `resultPath` better. There are some changes that need to be made to the cache. But it's for another PR.

justinkaseman (approved):
Nice work!

PR #750, opened 2021-08-05
[Perf] Flatten Burst Limit state

justinkaseman:


packages/core/bootstrap/src/lib/burst-limit/reducer.ts
```diff
@@ -3,10 +3,12 @@ import { makeId } from '../rate-limit'
 import * as actions from './actions'
 
 export enum IntervalNames {
+  // SECOND = 'SECOND',
```
cds95, line 6: Can this be removed?

packages/core/bootstrap/src/lib/burst-limit/reducer.ts
```diff
@@ -3,10 +3,12 @@ import { makeId } from '../rate-limit'
...
   MINUTE = 'MINUTE',
 }
 
 export const Intervals: { [key: string]: number } = {
+  // [IntervalNames.SECOND]: 1000,
```
cds95, line 11: Same here

packages/core/types/@chainlink/index.d.ts
```diff
@@ -1,11 +1,12 @@
...
 declare module '@chainlink/types' {
-  import type { CacheOptions } from '@chainlink/ea-bootstrap'
+  import type { Cache, RateLimit } from '@chainlink/ea-bootstrap'
   import { cacheWarmer } from "@chainlink/ea-bootstrap"
 
   export interface AdapterContext {
     name?: string
-    cache?: CacheOptions
+    cache?: Cache.CacheOptions
+    rateLimit?: RateLimit.config.Config
```
cds95, line 9: So now each adapter defines it's own `rateLimit` configuration?

packages/core/bootstrap/src/lib/burst-limit/reducer.ts
justinkaseman: I'm about to enable them within the next hour, just want to see a before / after cpu & memory usage of tracking by second.

packages/core/types/@chainlink/index.d.ts
justinkaseman: Not defines it per say, but rather initializes it once on startup and provides it through context rather than re-initializing it on every request.

This was done because of an annoying log message, but is the better way to do it because the config can't change between requests.

cds95 (approved):

PR #751, opened 2021-08-05
New.wootrade

bogos:


packages/sources/wootrade/README.md
```diff
@@ -0,0 +1,55 @@
+# Chainlink External Adapter for Wootrade
+
+Adapter using the public Wootrade market API for both HTTP(s) and WS.
+
+### Environment Variables
+
+| Required? |  Name   |                                                        Description                                                         | Options | Defaults to |
+| :-------: | :-----: | :------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|     ✅      | API_KEY | An key to use the wootrade WS API |         |             |
```
boxhock, line 9: Can we remove that the API_KEY is required here, and rather have a WebSocket section below where we explain that it's required?

packages/sources/wootrade/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+| Required? |   Name   |     Description     |           Options            | Defaults to |
+| :-------: | :------: | :-----------------: | :--------------------------: | :---------: |
+|           | endpoint | The endpoint to use | [crypto](#Crypto-Endpoint) |   ticker   |
+
+---
+
+## Crypto Endpoint
+
+An example endpoint description
```
boxhock, line 23: Remove/modify

packages/sources/wootrade/schemas/env.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/wootrade-adapter.json",
+  "title": "@chainlink/wootrade-adapter env var schema",
+  "required": ["API_KEY"],
```
boxhock, line 4: I think we should just remove it as required here, since the (HTTP) adapter can work without it.

packages/sources/wootrade/src/config.ts
```diff
@@ -0,0 +1,16 @@
+import { Requester } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const NAME = 'WOOTRADE'
+
+export const DEFAULT_ENDPOINT = 'crypto'
+export const DEFAULT_BASE_URL = 'https://api.staging.woo.network'
```
boxhock, line 7: I think we probably should be using the prod endpoint here :sweat_smile: 

```suggestion
export const DEFAULT_BASE_URL = 'https://api.woo.network'
```

boxhock (changes requested):
Thanks for the updated PR! Got a few items below

justinkaseman:
Could you add a file in `packages/sources/wootrade`  called `test-payload.json` containing
```
{
 "requests": [{
   "base": "ETH",
   "quote": "USDT"
 }]
}
```

I'll need to add this pattern to the example adapter.

justinkaseman:
Another thing (should be added to the README, my bad!)

In `packages/core/bootstrap/src/lib/provider-limits/limits.json` we need to define Wootrade's API rate limits by adding 
```
  "wootrade": {
    "http": {
 // some tiers defined here
    },
    "ws": {
// some tiers defined here
}
  },
```

packages/sources/wootrade/README.md
bogos: removed

packages/sources/wootrade/schemas/env.json
bogos: removed

packages/sources/wootrade/src/config.ts
bogos: fixed

packages/core/bootstrap/src/lib/provider-limits/limits.json
```diff
@@ -1027,5 +1027,15 @@
...
     "http": {},
     "ws": {}
+  },
+  "wootrade": {
+    "http": {},
+    "ws": {     
+      "private": {
+        "connections": 1,
+        "subscriptions": 1,
+        "note": "you need to request the application_id directly with wootrade"
```
boxhock, line 1037: Are these limits actually 1, or are there no limits?
Can we do -1 if there are no limits, and add a note that docs doesn't mention limits?

boxhock (changes requested):
quick comment, then I think we're good

packages/core/bootstrap/src/lib/provider-limits/limits.json
bogos: remove :) there is not information about ws limits

boxhock (approved):
Looks good :slightly_smiling_face:

packages/sources/wootrade/README.md
```diff
@@ -0,0 +1,53 @@
+# Chainlink External Adapter for Wootrade
+
+Adapter using the public Wootrade market API for both HTTP(s) and WS.
+
+### Input Parameters
+
+| Required? |   Name   |     Description     |           Options            | Defaults to |
+| :-------: | :------: | :-----------------: | :--------------------------: | :---------: |
+|           | endpoint | The endpoint to use | [crypto](#Crypto-Endpoint) |   ticker   |
```
boxhock, line 9:
```suggestion
|           | endpoint | The endpoint to use | [crypto](#Crypto-Endpoint) |   crypto   |
```

boxhock (changes requested):

boxhock (approved):

bryanjowers:
nice!

PR #763, opened 2021-08-06
[Feat] Per second burst rate limiting

justinkaseman:


packages/core/bootstrap/src/lib/burst-limit/index.ts
```diff
@@ -13,37 +13,77 @@ import { AdapterError, logger } from '../external-adapter'
...
+
+    const observedRequestsInSecond = selectTotalNumberOfRequestsFor(requests, IntervalNames.SECOND)
+
+    if (observedRequestsInSecond > burstCapacity1s) {
+      logger.debug(
+        `Per Second Burst rate limit cap of ${burstCapacity1s} reached. ${observedRequestsInSecond} requests sent in the last minute. Waiting 1 second. Retry number: ${retry}`,
+      )
+      await delay(1000)
+    }
+    return true
```
cds95, line 37: Should this be in the `if` block?

packages/core/bootstrap/src/lib/rate-limit/config.ts
```diff
@@ -30,19 +31,27 @@ export function get(context: AdapterContext): Config {
...
   if (enabled) {
     const provider = getEnv('RATE_LIMIT_API_PROVIDER') || context.name?.toLowerCase() || ''
     const tier = getEnv('RATE_LIMIT_API_TIER') || ''
     try {
-      const providerConfig = getBurstLimit(provider, tier)
-      burstCapacity = Number(providerConfig)
+      const limit = getHTTPLimit(provider, tier, 'rateLimit1s')
+      burstCapacity1s = Number(limit)
+    } catch {
+      // Ignore
```
cds95, line 43: Should we at least add some logs here for debugging?

cds95 (changes requested):

packages/core/bootstrap/src/lib/burst-limit/index.ts
justinkaseman: No. This function is checking if we are above the limit on requests we sent in the last second.
If so (if block) delay a second and check again.
If not quit out by returning true.

packages/core/bootstrap/src/lib/rate-limit/config.ts
justinkaseman: The functions in their give their own more granular logs.

packages/core/bootstrap/src/lib/burst-limit/index.ts
cds95: but then won't this mean the `for` loop never goes past the first iteration as we'll always return `true` in the first iteration?

justinkaseman: Ah shoot, you are right. Missing an `else` there.

cds95 (approved):

PR #776, opened 2021-08-10
Tiingo crypto batch support

boxhock:
With the way Tiingo's API currently works, we should not be using batching on their endpoint. This PR will be useful if the change their API so that we can use their batching endpoint.


boxhock:
Closing this PR as this change set will become outdated by the time this can be used

PR #777, opened 2021-08-10
Make the cache warmer more configurable

boxhock:


packages/core/bootstrap/README.md
```diff
@@ -160,11 +160,12 @@ Being:
...
+| Required? |             Name              |                                                                                      Description                                                                                      | Options |      Defaults to      |
+| :-------: | :---------------------------: | :-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: | :-----: | :-------------------: |
+|           | `EXPERIMENTAL_WARMUP_ENABLED` |                                                                        Enable the cache warmer functionality.                                                                         |         |        `false`        |
+|           | `WARMUP_UNHEALTHY_THRESHOLD`  |                                  The number of times a warmup execution can fail before we drop a warmup subscription for a particular cache key.to.                                  |         |          `3`          |
```
boxhock, line 166: Added the ability to disable `WARMUP_UNHEALTHY_THRESHOLD`.

This is useful when we know that we do want to continue warming this data as long as requests are coming in (controlled by `WARMUP_SUBSCRIPTION_TTL`), and we should continue trying to warm the cache even if the data provider goes down for some amount of time.

packages/core/bootstrap/src/lib/cache-warmer/config.ts
```diff
@@ -34,7 +35,7 @@ export function get(): Config {
   return {
     hashOpts: getHashOpts(),
     unhealthyThreshold: Number(process.env.WARMUP_UNHEALTHY_THRESHOLD) || 3,
-    warmupInterval: Number(process.env.CACHE_MIN_AGE) || MINIMUM_AGE,
+    warmupInterval: Number(process.env.WARMUP_INTERVAL),
```
boxhock, line 38: This variable seems to not have been used at all prior to this PR. Hence, making it return `WARMUP_INTERVAL` or undefined should be fine.

boxhock:
PR is broken

PR #781, opened 2021-08-11
Finage Batching discovery #769

bogos:
Batching calls was added in the 'stock' endpoint. Is the only one that allows this feature. [Multiple Stock](https://finage.co.uk/docs/api/multiple-stock-last-quotes)

### DIFF
--- a/packages/sources/finage/src/adapter.ts
+++ b/packages/sources/finage/src/adapter.ts
@@ -1,4 +1,7 @@
-import { ExecuteWithConfig, ExecuteFactory, Config } from '@chainlink/types'
+import { ExecuteWithConfig, ExecuteFactory, Config, 
+  // AdapterRequest,
+  AxiosResponse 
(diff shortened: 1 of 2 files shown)


packages/sources/finage/src/adapter.ts
```diff
@@ -1,4 +1,7 @@
-import { ExecuteWithConfig, ExecuteFactory, Config } from '@chainlink/types'
+import { ExecuteWithConfig, ExecuteFactory, Config, 
+  // AdapterRequest,
```
thodges-gh, line 2: Please remove commented-out code

thodges-gh (changes requested):
Just one small thing

packages/sources/finage/src/adapter.ts
bogos: done!

thodges-gh (approved):

PR #787, opened 2021-08-12
Ws for batched requests

cds95:


HenryNguyen5:
We've got some file conflicts here @cds95, care to schedule some pair review time as this touches some sensitive code?

PR #788, opened 2021-08-12
Cache warmer interval refactor and tests

boxhock:


packages/core/bootstrap/test/unit/cache-warmer.test.ts
```diff
@@ -387,6 +387,46 @@ describe('side effect tests', () => {
...
+          cacheWarmer: {
+            subscriptions: {
+              [key1]: { isDuplicate: false },
+              [key2]: { isDuplicate: false },
+            },
+          },
+        })
+
+        const config = { ...epicDependencies.config, warmupInterval: 8000 }
+        const output$ = warmupSubscriber(action$, state$, { config })
```
justinkaseman, line 417: Instead of passing through the config here could we change this to instead set the environment variable in the test?

It will not only keep the `Subscriber` cleaner, but will also mimic how users will interact with it.

boxhock: I was mainly copying how the other epics got their configs passed in, as they also got them like this. Should they remain as-is? Feels a bit weird if we're inconsistent with them.

HenryNguyen5: This looks like the proper way to inject dependencies into an epic to me, if we need users to configure the `warmupInterval`, then that configured value should be passed into the epic as a dependency when the epic is initialized

PR #789, opened 2021-08-13
Dxfeed.batch

bogos:


packages/sources/dxfeed/src/endpoint/price.ts
```diff
@@ -19,9 +25,12 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
-  const symbol = (validator.overrideSymbol(AdapterName) as string).toUpperCase()
-  const events = quoteEventSymbols[symbol] ? 'Quote' : 'Trade'
+  const base = validator.validated.data.base
+  const symbol = Array.isArray(base)
+    ? base.map((symbol) => symbol.toUpperCase()).join(',')
```
cds95, line 30: Just a style thing, can we remove the parentheses around `symbol`?  So `base.map(symbol => symbol.toUpperCase()).join(',')`.

cds95 (changes requested):
Just one minor comment

packages/sources/dxfeed/src/endpoint/price.ts
bogos: done! :)

cds95 (dismissed):

cds95 (approved):

packages/sources/dxfeed/test/integration/adapter.test.ts
```diff
@@ -24,6 +24,23 @@ describe('execute', () => {
...
+        name: 'supports multiple symbols',
+        testData: { id: jobID, data: { base: ['TSLA', 'AAPL'] } },
+      },
+    ]
+
+    requests.forEach((req) => {
+      it(`${req.name}`, async () => {
+        const data = await execute(req.testData as AdapterRequest)
+        assertSuccess({ expected: 200, actual: data.statusCode }, data, jobID)
+        expect(Object.keys(data.result).length).toBeGreaterThan(0)
```
boxhock, line 39: I think this should be checking `data.data.results`. The test is failing as-is right now.

boxhock (changes requested):
Just one thing. I think we're good to merge once this is resolved.

packages/sources/dxfeed/test/integration/adapter.test.ts
bogos: done!

PR #794, opened 2021-08-14
Upvest.gasprice

bogos:


packages/sources/upvest/src/endpoint/gasprice.ts
```diff
@@ -0,0 +1,40 @@
+import { Requester, Validator } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
+
+export const supportedEndpoints = ['gasprice']
+
+interface ResponseSchema  {
```
cds95, line 6: Can we export this?

cds95 (changes requested):

packages/sources/upvest/src/endpoint/gasprice.ts
bogos: done!

boxhock:
@bogos Could you please rebase the PR and make sure the CI is still passing? Besides this it looks good

PR #797, opened 2021-08-16
feat(coinpaprika): crypto batch support

boxhock:


justinkaseman (commented):
Since this is switching the API endpoint would there be value in keeping the previous one around, but still switching the `supportedEndpoints` over so nops use this new one?

packages/sources/coinpaprika/src/endpoint/crypto.ts
```diff
@@ -27,38 +59,102 @@ export const inputParameters: InputParameters = {
...
+    throw new Error(`unable to find coin: ${coin || symbol}`)
+  }
 
-  return Requester.success(jobRunID, response, config.verbose)
+  const result = Requester.validateResultNumber(coinData, [
+    'quotes',
+    requestedQuotes.toUpperCase(),
+    resultPath,
+  ])
+  return Requester.success(jobRunID, Requester.withResult(response, result), config.verbose)
```
justinkaseman, line 159:
```suggestion
  return Requester.success(jobRunID, Requester.withResult(response, result), config.verbose, batchablePropertyPath)
```

Also need `batchablePropertyPath` on unbatched requests so that it knows they should be part of the batch warmer.

justinkaseman (changes requested):

packages/sources/coinpaprika/test/unit/util.test.ts
```diff
@@ -1,106 +1,53 @@
...
+import { ResponseSchema } from '../../src/endpoint/crypto'
 
 describe('utils tests', () => {
-  describe('getSymbolToId', () => {
-    const mockCoinList = [
-      {
-        id: 'btc-bitcoin',
-        name: 'Bitcoin',
-        symbol: 'BTC',
+  describe('getCoin', () => {
```
justinkaseman, line 5: Nice use of unit tests.

boxhock:
@justinkaseman Yeah good idea, I'll make the old one a separate endpoint

justinkaseman (approved):
:clap:

PR #806, opened 2021-08-18
feat: add new medianizer composite adapter

boxhock:


packages/tsconfig.json
```diff
@@ -245,6 +248,9 @@
     {
       "path": "./sources/kaiko"
     },
+    {
```
cds95, line 251: Is this intentional?

cds95 (commented):
This looks good to me.  Are there any tests for this adapter?

packages/tsconfig.json
boxhock: Yeah, it was accidentally not included in the original PR, so we should bring it in ASAP

packages/composites/medianizer/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,198 @@
...
+describe('medianizer', () => {
+  let execute: Execute
+  let config: Config
+  let sandbox: sinon.SinonSandbox
+  let server: sinon.SinonFakeServer
+
+  beforeEach(() => {
+    execute = makeExecute()
+    config = makeConfig()
+    sandbox = createSandbox()
```
HenryNguyen5, line 44: Can we use the same testing pattern that's presented in `packages/composites/synth-index/test/integration`? We use nock which lets you record http calls first, which makes it super easy to generate stubs later on

HenryNguyen5 (changes requested):

packages/composites/medianizer/test/integration/adapter.test.ts
boxhock: Happy to do this in the test improvements of `outlier-detection`, though I don't see it as necessary to get this PR out.

HenryNguyen5 (approved):

PR #810, opened 2021-08-19
refactor(intrinio): programatic WS connection url for auth token

justinkaseman:
### Description
Fixes the issue for Intrinio where when under heavy load the endpoint to generate the auth token to open the websocket connection starts getting rate limited.

This happens because the call to generate the auth token happens on every single request coming in, even if the connection is already open.

EA framework change:
Allows for generating the connection URL from a callback, rather than static string.


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -411,22 +428,18 @@ export const metricsEpic: Epic<AnyAction, AnyAction, any, any> = (action$, state
     tap(([action, state]) => {
       const connectionLabels = (payload: WSConfigPayload) => ({
         key: payload.config.connectionInfo.key,
-        url: censor(payload.wsHandler.connection.url),
```
justinkaseman, line 414: Unsure what to do with these.
I took them out unless someone can think of a better way.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -58,6 +60,16 @@ const deserializer = (message: any) => {
   }
 }
 
+type connectRequestedActionWithState = [
```
HenryNguyen5, line 63:
```suggestion
type ConnectRequestedActionWithState = [
```
Types should always be PascalCase

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -68,6 +80,11 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
       const isConnecting = state.ws.connections.all[connectionKey]?.connecting > 1
       return !isActiveConnection && !isConnecting
     }),
+    concatMap(async (data) => {
+      const url = data[0].payload.wsHandler.connection.url
+      if (typeof url !== 'string') data[0].payload.wsHandler.connection.url = await url()
```
HenryNguyen5, line 85: Mind adding a comment here for when url can be a function? At a glance it reads as-if url should always be a string property, otherwise it would be named `getUrl()`. How do you feel about `getUrl()` always being a function?

packages/core/bootstrap/src/lib/ws/epics.ts
HenryNguyen5: What's wrong with having them in?

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -11,50 +11,49 @@ export * as epics from './epics'
...
 export * as types from './types'
 
-export const withWebSockets = (
-  store: Store<RootState>,
-  makeWsHandler?: MakeWSHandler,
-): Middleware => async (execute, context) => async (input: AdapterRequest) => {
-  const wsConfig = getWSConfig()
-  if (!makeWsHandler || !wsConfig.enabled) return await execute(input, context) // ignore middleware if conditions are met
+export const withWebSockets =
+  (store: Store<RootState>, makeWsHandler?: MakeWSHandler): Middleware =>
```
HenryNguyen5, line 15: Should rebase formatting changes against develop

packages/core/types/@chainlink/index.d.ts
```diff
@@ -170,14 +169,15 @@ declare module '@chainlink/types' {
     warning?: string
   }
 
+  export type WSHandlerConnectionURL = (() => string) | (() => Promise<string>) | string
```
HenryNguyen5, line 172: How do you feel if it was only an async getter? Less conditionals the better.

packages/sources/intrinio/src/adapter.ts
```diff
@@ -48,18 +48,13 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
...
       provider: 'iex',
     })
-
-    await ws._refreshToken()
-    ws.destroy() // close WS connection generated by the Intrinio SDK to prevent unmaintained connections
-
     return {
       connection: {
-        url: ws._makeSocketUrl(),
+        url: async () => await ws._makeSocketUrl(),
```
HenryNguyen5, line 57:
```suggestion
        url: ws._makeSocketUrl,
```

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+      req.on('error', (e) => {
+        console.error('IntrinioRealtime | Could not get auth token: ' + e)
+        reject(e)
+      })
+
+      req.end()
+    })
+  }
+
+  async _makeSocketUrl(): Promise<string> {
```
HenryNguyen5, line 186: Instead of using the `_` unused syntax, we should make use of typescripts `private`, `protected`, and `public` access modifiers. If you really want it to be private, you can use the javascript private fields access modifier `#` to make it "hard" private, where its not accessible from even runtime code.

It looks like we don't need to worry about hard privacy though, but ill let you make the final judgement call on this since you have more context.

See https://www.typescriptlang.org/docs/handbook/2/classes.html#private for more details

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+    } else if (this.options.provider == 'fxcm') {
+      return (
+        'wss://fxcm.intrinio.com/socket/websocket?vsn=1.0.0&token=' + encodeURIComponent(this.token)
+      )
+    }
+
+    return ''
+  }
+
+  _makeHeartbeatMessage(): any {
```
HenryNguyen5, line 213: Can we get rid of the `any` returns?

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+    if (typeof api_key !== 'string') {
+      return false
+    }
+
+    if (api_key === '') {
+      return false
+    }
+
+    return true
+  }
```
HenryNguyen5, line 295:
```suggestion
  _validAPIKey(api_key: string): boolean {
    return typeof api_key === 'string' && api_key !== ''
  }
```

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+  _parseIexTopic(channel: string): string {
+    let topic = null
+    if (channel == '$lobby') {
+      topic = 'iex:lobby'
+    } else if (channel == '$lobby_last_price') {
+      topic = 'iex:lobby:last_price'
+    } else {
+      topic = 'iex:securities:' + channel
+    }
+    return topic
```
HenryNguyen5, line 282:
```suggestion
   switch (channel) {
 case '$lobby': 
  return 'iex:lobby'
case '$lobby_last_price': 
  return 'iex:lobby:last_price'
default:
  return `iex:securities:${channel}`
}
```

Good opportunity to use a switch/case statement here

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+        agent: agent,
+        headers: headers,
+      }
+
+      const req = https.get(options, (res: any) => {
+        if (res.statusCode == 401) {
+          this._throw('Unable to authorize')
+          reject()
+        } else if (res.statusCode != 200) {
+          console.error(
```
HenryNguyen5, line 164: Should we use a proper logger here?

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+        'Content-Type': 'application/json',
+        Authorization: 'Basic ' + Buffer.from(username + ':' + password).toString('base64'),
+      }
+    }
+  }
+
+  _refreshToken() {
+    Logger.debug('Requesting auth token...')
+
+    return new Promise<void>((fulfill, reject) => {
```
HenryNguyen5, line 146:
```suggestion
    return new Promise<void>((resolve, reject) => {
```
Typical terminology for promises is `resolve,reject` i believe

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+        'Content-Type': 'application/json',
+      }
+    } else {
+      const { username, password } = this.options
+
+      return {
+        'Content-Type': 'application/json',
+        Authorization: 'Basic ' + Buffer.from(username + ':' + password).toString('base64'),
+      }
+    }
```
HenryNguyen5, line 140:
```suggestion
      return {
        'Content-Type': 'application/json',
        Authorization: 'Basic ' + Buffer.from(`${options.username}:${options.password}`).toString('base64'),
      }
```
How do you feel about making this more terse?

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+  _makeAPIAuthUrl(auth_url: any): any {
+    let path = auth_url.path
+
+    if (path.includes('?')) {
+      path = path + '&'
+    } else {
+      path = path + '?'
+    }
+
+    auth_url.path = path + 'api_key=' + this.options.api_key
```
HenryNguyen5, line 124:
```suggestion
    return auth_url.path = `${auth_url.path}${auth_url.path.includes('?') ? '&' : '?'}api_key=${this.options.api_key}` 
```
How do you feel about making this more terse?

HenryNguyen5: BTW, we may want to use the URL node lib here instead, so we dont have to manually manipulate strings

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+    }
+  }
+
+  _makeAuthUrl(): { host: string; path: string } {
+    let auth_url = {
+      host: '',
+      path: '',
+    }
+
+    if (this.options.provider == 'iex') {
```
HenryNguyen5, line 86: Good spot for a switch/case statement to reduce redundancy

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
+import * as https from 'https'
+import * as events from 'events'
+import { Logger } from '@chainlink/ea-bootstrap'
+
+const EventEmitter = events.EventEmitter
```
HenryNguyen5, line 5:
```suggestion
import { EventEmitter } from 'events'
import { Logger } from '@chainlink/ea-bootstrap'
```

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+const SELF_HEAL_BACKOFFS = [0, 100, 500, 1000, 2000]
+
+export class IntrinioRealtime extends EventEmitter {
+  options: any
+  token: null | string = null
+  websocket = null
+  ready = false
+  channels = {}
+  joinedChannels = {}
+  afterConnected = null // Promise
```
HenryNguyen5, line 16:
```suggestion
  afterConnected: Promise<type?> | null = null 
```

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
+import * as https from 'https'
+import * as events from 'events'
+import { Logger } from '@chainlink/ea-bootstrap'
+
+const EventEmitter = events.EventEmitter
+
+const SELF_HEAL_BACKOFFS = [0, 100, 500, 1000, 2000]
+
+export class IntrinioRealtime extends EventEmitter {
+  options: any
```
HenryNguyen5, line 10: We should specify these with access modifiers

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+  token: null | string = null
+  websocket = null
+  ready = false
```
HenryNguyen5, line 13: How does this get mutated to `true`?

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
...
+          res.on('data', (data: any) => {
+            this.token = Buffer.from(data, 'base64').toString()
+            Logger.debug('Received auth token!')
+            fulfill()
+          })
+        }
+      })
+
+      req.on('error', (e) => {
+        console.error('IntrinioRealtime | Could not get auth token: ' + e)
```
HenryNguyen5, line 178: Should we use a proper logger here?

packages/sources/intrinio/src/util.ts
```diff
@@ -0,0 +1,296 @@
+import * as https from 'https'
+import * as events from 'events'
+import { Logger } from '@chainlink/ea-bootstrap'
+
+const EventEmitter = events.EventEmitter
+
+const SELF_HEAL_BACKOFFS = [0, 100, 500, 1000, 2000]
+
+export class IntrinioRealtime extends EventEmitter {
```
HenryNguyen5, line 9: We should definitely add integration tests which use something like:
- https://www.npmjs.com/package/@relaycorp/ws-mock
- https://www.npmjs.com/package/mock-socket

HenryNguyen5 (dismissed):

packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -38,24 +38,34 @@ export const withWebSockets =
...
-        connectionInfo: {
-          key: wsConfig.connectionInfo.key,
-          url: wsHandler.connection.url,
-        },
-        subscriptionMsg,
-        input: singleInput,
-        context,
-      }
+    if (isConnected(store, wsConfig.connectionInfo.key)) {
+      await separateBatches(input, async (singleInput: AdapterRequest) => {
```
justinkaseman, line 42: Since this is the `subscribeReadyEpic` repeated could we pull this out of here and only have it in epics?

cds95:
Will address comments to refactor utils file in this ticket https://app.clubhouse.io/chainlinklabs/story/16731/refactor-intrinio-ws-file

packages/core/bootstrap/src/lib/ws/index.ts
cds95: We'll address it in this PR https://app.clubhouse.io/chainlinklabs/story/16732/refactor-ws-middleware-flow

cds95 (approved):

cds95 (approved):

PR #819, opened 2021-08-19
refactor(bootstrap): return actual error code in error response

justinkaseman:


itmustbejj (approved):

HenryNguyen5 (commented):
Should this be in draft? There's only logging messages currently

HenryNguyen5 (approved):

PR #837, opened 2021-08-25
Make deep copy of imported JSON files

HenryNguyen5:


boxhock (approved):
Confirmed it works now :+1:

PR #842, opened 2021-08-26
test(augur): integration tests with hardhat

pgebheim:
This is a large patch which updates the augur composite adapter to support integration testing using Augur Turbo's `hardhat-deploy` infrastructure. These tests are not run automatically as they still currently require being `yarn link`ed to a development repository. When the turbo smart contract npm modules have been published, dependencies can be updated and the tests run automatically during CI.

This patch:
 - Tells jest to ignore any tests where "test/hardhat" is in the path.
 - Adds test:augur yarn script to the top level
 - Updates a slew of dependencies w/ .pnp.js as things needed to get updated to make it all work
 - Adds new tests to packages/composites/augur/test/hardhat
 - Depends on *mocha* and not jest for the above tests

This patch also fixes some bugs in the augur adapter's crypto (pokeMarkets) method.

please review @boxhock


pgebheim:
> Hey @pgebheim, really appreciate the additional tests!
> 
> There are a number of requirements that need to be met before this is in a mergeable state:
> 
>     * All tests should use jest rather than mocha
> 
>     * Tests should be run automatically and not require symlinking
> 
>     * Tests should not need a top level script, they'll get automatically run if in the `test/{unit,integration}` paths
> 
> 
> Please place the PR in the `draft` state in the meanwhile

Hey @HenryNguyen5 --

Unfortunately this isn't viable at the moment. A jest test runner that integrates with hardhat does exists (hardhat-jest-plugin), but it is currently basically untested and is just a proof of concept. Furthermore, even if it did exists you need to run these tests directly with `hardhat` cli because of the extra bootstrapping work it does for test environments (making a local chain, running deployments and snapshotting), so currently these need to be run through `hardhat test` (or `hardhat test:jest` if hardhat-jest-plugin does turn out to work).

Also, we need to wait until Augur smart modules are deployed to NPM in order to enable these tests at all since currently the module doesn't exist to be imported.

I'd like to get this code in and then iterate on the tests in order to make them more integrated with your current environment. It's still massively useful to have these put into place and to have this code merged up to dev, since all of this contains prerequisites for actual running adapters.

PR #867, opened 2021-08-31
Enable WS support for tiingo equities

cds95:
This PR adds a new function property to WSHandler to filter out requests that should not be served using WS.  This is applied to the Tiingo adapter so that only requests to the IEX endpoint are served using WS.  I've intentionally left out tests for now and will add them to the PR once we feel good about how this is handled.

### DIFF
--- a/packages/sources/tiingo/src/adapter.ts
+++ b/packages/sources/tiingo/src/adapter.ts
@@ -59,8 +59,7 @@ interface UpdateMessage extends Message {
 }
 
 const customParams = {
-  base: ['base', 'from', 'coin'],
-  quote: ['quote', 'to', 'market'],
+  base: ['base', 'from', 'coin', 'ticker'],
 }
 
 export const makeWSHandler = (config?: Config): MakeWSHandler | undefined => {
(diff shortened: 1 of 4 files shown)


packages/sources/tiingo/src/config.ts
```diff
@@ -5,7 +5,7 @@ export const NAME = 'TIINGO'
 
 export const DEFAULT_ENDPOINT = 'crypto'
 export const DEFAULT_BASE_URL = 'https://api.tiingo.com'
-export const DEFAULT_WS_API_ENDPOINT = 'wss://api.tiingo.com/crypto'
```
cds95, line 8: We should extend our framework in the future to allow supporting multiple WS endpoints.  This should be ok for now as we want to disable WS for `crypto` anyways.

packages/sources/tiingo/src/adapter.ts
```diff
@@ -59,42 +59,41 @@ interface UpdateMessage extends Message {
...
   }
   return () => {
     const defaultConfig = config || makeConfig()
     return {
       connection: {
         url: defaultConfig.api.baseWsURL || DEFAULT_WS_API_ENDPOINT,
       },
-      subscribe: (input) => getSubscription(getPair(input)),
-      unsubscribe: (input) => getSubscription(getPair(input), false),
+      shouldNotServeInputUsingWS: (input) => input.data.endpoint !== 'iex',
```
boxhock, line 94: We should also accept any of the aliases for iex. In other words, any entry in this array: https://github.com/smartcontractkit/external-adapters-js/blob/4e34ac7a144de024dcc45a62acddbf5f5fd92826/packages/sources/tiingo/src/endpoint/iex.ts#L4

boxhock (changes requested):

boxhock (approved):

PR #869, opened 2021-09-01
Tradermade ws

cds95:
- Add extra endpoint for forex.  `forex` endpoint does the same thing as the `live` endpoint 
 - Enable WS support for `forex` endpoint
 - Modify default config to allow setting `WS_API_KEY` environment variable


packages/sources/tradermade/src/endpoint/forex.ts
```diff
@@ -0,0 +1,9 @@
+import { ExecuteWithConfig, Config } from '@chainlink/types'
+import * as live from './live'
+
+export const supportedEndpoints = ['forex']
+
+export const execute: ExecuteWithConfig<Config> = async (input, context, config) => {
```
cds95, line 6: This is a bit weird but it looks like we only want to limit WS support for `forex` pairs.

boxhock (changes requested):
High-level this looks good. Could you please also add the WS limits in the `limits.json` file? https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/provider-limits/limits.json#L922-L944

packages/sources/tradermade/src/endpoint/forex.ts
boxhock: I'm not sure if I'm missing something, but couldn't we add `forex` in the `supportedEndpoints` array? And shouldn't they be switched around?

Not sure I fully understand when do we want HTTP only vs when do we want to enable WS requests? Isn't it enough with only one endpoint and controlling WS with `WS_ENABLED`?

boxhock (changes requested):

packages/sources/tradermade/src/endpoint/forex.ts
cds95: that's what I had thought a well but it sounded like Bryan wanted equities to keep using http and for only forex to use WS.  I'll confirm with him again during standup.

cds95: Confirmed with Bryan that we can collapse this into a single endpoint.

packages/sources/tradermade/README.md
```diff
@@ -1,32 +1,94 @@
 # Chainlink External Adapter for Tradermade
 
+## Endpoints
+
+- `live` and `forex`. Use `forex` when querying for currency pairs. WebSockets can be enabled when fetching data from the `forex` endpoint but not the `live` endpoint. Adapter defaults to the `live` endpoint.
+
```
boxhock, line 6: Can we update this README to follow the format of how we usually document adapters with different endpoints? That would give us some more room to explain each endpoint in their own respective sections

boxhock (approved):

PR #874, opened 2021-09-03
Update Linear finance

boxhock:


packages/composites/linear-finance/test/unit/csv.test.ts
```diff
@@ -15,12 +15,12 @@ describe('CSV parsing', () => {
     expect(result).toStrictEqual([
       {
         symbol: 'BTC',
-        balance: '10542985000000000',
+        balance: '554978470000000000',
```
HenryNguyen5, line 18: Did these results change due to the xlci changes? If so, we might want to make these snapshot tests, they would be easier to maintain.

HenryNguyen5 (dismissed):

packages/composites/linear-finance/test/unit/csv.test.ts
boxhock: They changed because we changed the function to get "weight" instead of "unit", so unrelated to the index changes

thodges-gh (approved):

PR #880, opened 2021-09-06
Refactor tests

HenryNguyen5:
# Motivation

This PR gets our integration tests in a CI-able state. The previous integration tests that we used to have were actually e2e tests, since we were making live API calls. With having all of our integration tests now run, we're able to start running coverage reporting and having newly written integration tests automatically included in our CI test suite.

# Change Summary
- All un-used integration tests (that were actually e2e) that were mostly copy-paste were deleted. If they ever need to be referenced in the future, they can always be recovered and repaired via Git.
- Integration tests that had adapter-specific logic in them were kept and moved to `e2e`, where they can be run in an ephemeral environment at a later date, or be converted to a proper integration test via `nock`-ifying them
- A number of tests were in the root of the adapter's `test` folder, rather than being nested in `{unit,integration,e2e}`. With the way our test runner is setup, this means that they were never run. They have been organized accordingly.
- Empty test files have been removed
- The apy-finance integration test is failing despite it being a `nock` based test, it has been `.skip`'ed for now to get the suite passing. @cds95 could you take a look at this test?
- A behaviour change has been noticed while enabling more tests, if no `jobId` is sent to the adapter, it'll return a default `jobId` of `"1"` rather than `undefined`. Is this an intended change? Or should it be marked as a regression and changed back to its old behaviour? @justinkaseman @thodges-gh @boxhock 
- The integration test for the cmc adapter has had its snapshots updated. Rate limit failures used to be expected as `500`'s, now they're `429`'s.

# Commit Summary
- Ignore coverage folder
- Default returned jobId to 1 if requestId is undef
- Delete unused source adapter integration tests
- Convert agoric test to e2e
- Convert harmony test to e2e
- Remove unused target adapter integration tests
- Move dydx-stark test to unit
- Rmove empty dns-query test
- Move intrinio test to unit
- Move json-rpc tests to e2e
- Remove redundant example test
- Move bitcoin-json-rpc tests to e2e
- Move bob integration tests to e2e
- Move CVI tests to unit
- Delete empty composite adapter tests
- Fix import paths on moved unit tests
- Moved google-weather tests to e2e
- Restructure market-closure tests
- Moved root outlier-detection test to integration
- Removed unused the-graph integration test
- Fix failing cmc tests
- Move balance adapter tests to e2e
- Import missing nock dep in apy-finance
- Move paypal integration tests to e2e
- Fix json-rpc documentation for running tests
- Get rid of failing example integration tests
- Add readme around composite integration tests
- Remove failing outlier-detection integration test
- Skip failing apy-finance integration test
- Run all integration tests in CI
- Configure coverage for jest


justinkaseman (approved):

packages/composites/apy-finance/test/integration/adapter.test.ts
```diff
@@ -39,7 +39,7 @@ describe('execute', () => {
       },
     }
 
-    it('should return success', async () => {
+    it.skip('should return success', async () => {
```
tateexon, line 42: Do we have a ticket to come back to this skipped test?

thodges-gh:
> A behaviour change has been noticed while enabling more tests, if no jobId is sent to the adapter, it'll return a default jobId of "1" rather than undefined. Is this an intended change? Or should it be marked as a regression and changed back to its old behaviour?

If no `id` is supplied in the request, we don't need to return a `jobRunId` in the response. It used to be that the node required the `jobRunId` to be provided on response, but at some point that changed.

packages/composites/apy-finance/test/integration/adapter.test.ts
thodges-gh: I just created one: 16514

thodges-gh (approved):

tateexon (approved):

PR #888, opened 2021-09-08
feat: add changesets dependency

justinkaseman:
### Description
Add dependency for changesets to manage our monorepo.
This handles version bumps and changelog creation during the PR level, making releases a lot easier.

### Status
- [x] Add dependency
- [ ] Clean up old monorepo management dependencies
- [ ] Modify github check


HenryNguyen5:
We should get rid of our other release dependencies and git hooks, no?

PR #889, opened 2021-09-08
Fix docker-compose local dev environment scripts

HenryNguyen5:
# Summary
This PR enables bridge networking mode for our docker-compose dev environment. MacOS and Windows users can now use our docker-compose script to spin up their own local environment to experiment with.

The context injected into the generated docker-compose file was static, which is now dynamic so when compose files are merged, they have the right context to resolve the dockerfiles they refer to.

During testing, a bug arised with how the baseUrl is evaluated with the coingecko adapter, since docker-compose fills in missing environment variables with an empty string, rather than leaving them undefined. The coingecko adapter now uses falsey evaluation, which considers empty strings also false.

- Use bridge networking mode
- Set default env vars
- Set context on docker-compose file generation
- Use falsey evaluation when determining baseUrl


tateexon (approved):
The compose generation works and spins up on my mac now!

PR #917, opened 2021-09-17
feat(bootstrap): add cache warmer metrics & more helpful logs

justinkaseman:


packages/core/bootstrap/src/lib/cache-warmer/actions.ts
```diff
@@ -90,6 +91,7 @@ interface WarmupFailedPayload {
   /**
    * State lookup key
    */
+  id: string
```
HenryNguyen5, line 94: How is the `id` different than `key`?

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -158,6 +161,10 @@ export const warmupSubscriber: Epic<AnyAction, AnyAction, any, EpicDependencies>
     }),
     // on a subscribe action being dispatched, spin up a long lived interval if one doesnt exist yet
     mergeMap(([{ payload, key }]) => {
+      payload.childLastSeenById
```
HenryNguyen5, line 164: I would move this outside of the `mergeMap` into a `tap` function since it causes side-effects

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
```diff
@@ -227,7 +240,15 @@ export const warmupRequestHandler: Epic<AnyAction, AnyAction, any> = (action$, s
             }),
       ).pipe(
         mapTo(warmupFulfilled({ key })),
-        catchError((error: unknown) => of(warmupFailed({ error: error as Error, key }))),
+        catchError((error: unknown) =>
+          of(
+            warmupFailed({
+              id: getFeedId({ id: '0', data: requestData?.origin }),
```
HenryNguyen5, line 246: Should `id` always be 0?

packages/core/bootstrap/src/lib/cache-warmer/metrics.ts
```diff
@@ -0,0 +1,11 @@
+import * as client from 'prom-client'
+
+export const cache_warmer_count = new client.Gauge({
```
HenryNguyen5, line 3: I would merge both of these into one metric, with a `isBatched` boolean style label to differentiate between the two.

HenryNguyen5 (changes requested):

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
justinkaseman: This is just a dummy ID to conform it to the expected Adapter Input type.

packages/core/bootstrap/src/lib/cache-warmer/actions.ts
justinkaseman: Key is the key into the redux store for that subscription.
ID is the label of the feed. So like `BTC/ETH`. Should we call it something else?

packages/core/bootstrap/src/lib/cache-warmer/epics.ts
justinkaseman: Changed it to reflect warmer id

packages/core/bootstrap/src/lib/cache-warmer/actions.ts
justinkaseman: Changed name to `feedLabel`

HenryNguyen5 (approved):

PR #918, opened 2021-09-17
Onchain gas EA POC

cds95:
This PR implements a new adapter to fetch gas prices from on chain.  It does this by subscribing to the RPC's WS endpoint to listen to new heads and determines the gas price by taking the median gas price of it's transactions.


cds95:
Missing integration and e2e tests as our current testing framework does not allow us to write tests for WS only adapters.  I've created a ticket for us to handle this in the future. https://app.shortcut.com/chainlinklabs/story/17228/implement-testing-framework-for-ws-only-adapters.  I think it should be ok to leave this out for now as the adapter is still a POC at this point.  Open to discussion on this though.

packages/core/types/@chainlink/index.d.ts
```diff
@@ -208,6 +209,7 @@ declare module '@chainlink/types' {
       | undefined
     // Optional flag to ensure adapter only uses WS and doesn't send HTTP requests
     noHttp?: boolean
+    toSaveFromFirstMessage?: (message: any) => any
```
cds95, line 212: I'm open to a better name for this.  This function is called if we ever need to save anything from the first message received from the WS endpoint.

packages/sources/onchain-gas/src/adapter.ts
```diff
@@ -0,0 +1,69 @@
...
+          params: ['newHeads'],
+        }
+      },
+      isError: () => false,
+      filter: (message) => message.method === 'eth_subscription',
+      toResponse: async (message: any, input: AdapterRequest) => {
+        const validator = new Validator(input, endpoints.gas.inputParameters)
+        if (validator.error) throw validator.error
+        const hexedBlockNum: string = message.params.result.number
+        const medianGasPrices = await endpoints.gas.getTransactionsInPastBlocks(
```
thodges-gh, line 58: One thing that we might want to account for is I know the node used to have an issue when the result contained an array. So we might want the response data to look like:
```json
{
    "data": {
        "values": [101768070829],
        "result": 101768070829
    },
    "result": 101768070829
}
```

This also means when there are multiple blocks queried, we'll want another param to specify which individual value we come to for the `result`.

packages/sources/onchain-gas/src/config.ts
```diff
@@ -0,0 +1,16 @@
+import { Requester, util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
+export const NAME = 'ONCHAIN_GAS' // This should be filled in with a name corresponding to the data provider using UPPERCASE and _underscores_.
```
thodges-gh, line 4: :scissors: comment

packages/sources/onchain-gas/src/endpoint/gas.ts
```diff
@@ -0,0 +1,81 @@
+import { Validator, Logger, Requester } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
+import { DEAFULT_NUM_BLOCKS, MAX_BLOCKS_TO_QUERY } from '../config'
+
+// This should be filled in with a lowercase name corresponding to the API endpoint
```
thodges-gh, line 5: :scissors: comment

thodges-gh (changes requested):

packages/sources/onchain-gas/README.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+## Onchain-gas Endpoint
+
+### Input Params
+
+| Required? |    Name     |                           Description                            | Options | Defaults to |
+| :-------: | :---------: | :--------------------------------------------------------------: | :-----: | :---------: |
+|           | `numBlocks` |     The number of blocks to use to determine the gas price.      |         |      1      |
+|           | `blockIdx`  | The index of the block used for the result field in the response |         |      1      |
```
thodges-gh, line 28:
```suggestion
|           | `blockIdx`  | The index of the block used for the result field in the response |         |      0      |
```

thodges-gh (changes requested):
Just one thing to correct.

thodges-gh (approved):

PR #955, opened 2021-09-27
adding override for BCH on coingecko

brigeo:
Adding override for Bitcoin cash to use id bitcoin-cash instead of BCH.


justinkaseman (approved):
LGTM

PR #964, opened 2021-09-29
bugfix for onchain gas adapter

cds95:
Fix issue with onchain gas adapter when the first response from the WS RPC endpoint does not have the `params` field with the subscription ID.

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/epics.ts
+++ b/packages/core/bootstrap/src/lib/ws/epics.ts
@@ -435,13 +435,19 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
           const subscription = state.ws.subscriptions.all[key]
           return subscription && !subscription.subscriptionParams
         }),
-        mergeMap(async ([action]) => {
-          return saveFirstMessageReceived({
-            subscriptionKey: action.payload.subscriptionKey,
-            message: wsHandler.toSaveFromFirstMessage
-              ? wsHandler.toSaveFromFirstMessage(action.payload.message)
-              : {},
-          })
+        mergeMap(([action]) => {
+          const toSave = wsHandler.toSaveFromFirstMessage
+            ? wsHandler.toSaveFromFirstMessage(action.payload.message)
+            : undefined
+          if (!toSave) {
+            return EMPTY
+          }
+          return of(
+            saveFirstMessageReceived({
+              subscriptionKey: action.payload.subscriptionKey,
+              message: toSave,
+            }),
+          )
         }),
       )
 
--- a/packages/sources/onchain-gas/src/adapter.ts
+++ b/packages/sources/onchain-gas/src/adapter.ts
@@ -28,9 +28,14 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
       connection: {
         url: defaultConfig.api.baseURL,
       },
-      toSaveFromFirstMessage: (message: any) => ({
-        subscriptionId: message.params.subscription,
-      }),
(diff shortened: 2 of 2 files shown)


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -435,13 +435,19 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
...
         }),
-        mergeMap(async ([action]) => {
-          return saveFirstMessageReceived({
-            subscriptionKey: action.payload.subscriptionKey,
-            message: wsHandler.toSaveFromFirstMessage
-              ? wsHandler.toSaveFromFirstMessage(action.payload.message)
-              : {},
-          })
+        mergeMap(([action]) => {
+          const toSave = wsHandler.toSaveFromFirstMessage
```
justinkaseman, line 439:
```suggestion
          const toSave = wsHandler.toSaveFromFirstMessage && wsHandler.toSaveFromFirstMessage(action.payload.message)
```

Might read a little nicer

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -435,13 +435,19 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
...
+            ? wsHandler.toSaveFromFirstMessage(action.payload.message)
+            : undefined
+          if (!toSave) {
+            return EMPTY
+          }
+          return of(
```
justinkaseman, line 445:
```suggestion
          return toSave ? of( saveFirstMessageReceived({
              subscriptionKey: action.payload.subscriptionKey,
              message: toSave,
            }),) : EMPTY
```
Might read a little nicer

packages/sources/onchain-gas/src/adapter.ts
```diff
@@ -28,9 +28,14 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
       connection: {
         url: defaultConfig.api.baseURL,
       },
-      toSaveFromFirstMessage: (message: any) => ({
-        subscriptionId: message.params.subscription,
-      }),
+      toSaveFromFirstMessage: (message: any) => {
+        if (message.method !== 'eth_subscription' || !message.params) {
```
justinkaseman, line 32:
```suggestion
        if (message.method !== 'eth_subscription' || !message.params) return null
```
Do you prefer brackets and such?
Null is nice as a return because it denotes we chose to return it.

justinkaseman (approved):

packages/sources/onchain-gas/src/adapter.ts
```diff
@@ -28,9 +28,14 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
       connection: {
         url: defaultConfig.api.baseURL,
       },
-      toSaveFromFirstMessage: (message: any) => ({
-        subscriptionId: message.params.subscription,
-      }),
+      toSaveFromFirstMessage: (message: any) => {
```
justinkaseman, line 31: Could we add a comment explaining a little bit about why this is used and when it happens?

cds95: There's a comment in the `index.d.ts` file.

packages/sources/onchain-gas/src/adapter.ts
cds95: I don't have a preference on whether or not to use brackets.  I think I'm just more used to writing code with it.

justinkaseman (approved):

PR #966, opened 2021-09-29
feat(bootstrap): add rate limiting Express middlewares

justinkaseman:
### Description
Adds the following packages as Express middleware:
- [Express Rate Limit](https://www.npmjs.com/package/express-slow-down)
- [Express Slow Down](https://www.npmjs.com/package/express-rate-limit)

This will rate limit requests to the EA's server before hitting the request pipeline.

Default values allow `250` requests every 5 seconds, which is `3000` requests a minute. At `80%` of the limit (200 req within 5 seconds), the requests will hit a `500 ms` delay.


packages/core/bootstrap/src/lib/server.ts
```diff
@@ -124,3 +130,29 @@ function setupMetricsServer() {
...
+  // TODO: utilize context for Redis
+
+  const limiter = rateLimit({
+    windowMs: 5 * 100, // 5 seconds
+    max: 250, // limit each IP to 250 requests per windowMs
+    keyGenerator: () => '*', // use one key for all incoming requests
+  })
+  app.use(limiter)
+
+  const speedLimiter = slowDown({
```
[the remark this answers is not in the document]
HenryNguyen5: Also, these values should always be smaller than the rate limiter, no? Otherwise, the rate limiter will always take precedence and `429` requests before the slow down limiter can do anything. Maybe we should configure the `delayAfter` value to be proportional (80%? 70%?) to the `max` value of the rate limiter. This would give us the following variable set:
```ts
const windowMs = 1000 * 60 // define window as per minute, since this is a common interval
const max = parseInt(process.env.EA_RATE_LIMIT_MAX) || 1500 // default to 1500 req/min max
const delayAfter = max * ( Number(process.env.EA_SLOW_DOWN_AFTER_FACTOR) || 0.8) // we start slowing down requests when we reach 80% of our max limit for the current interval
const delayMs = parseInt(process.env.EA_SLOW_DOWN_DELAY_MS) || 250 // default to slowing down each request by 250ms
```

HenryNguyen5 (changes requested):

justinkaseman:
Reverting back to a 5 second window as our problem is burst requests from retries, not requests per minute.

packages/core/bootstrap/schemas/env.json
```diff
@@ -29,6 +29,15 @@
...
       "type": "boolean"
     },
+    "SERVER_RATE_LIMIT_MAX": {
+      "type": "string"
+    },
+    "SERVER_SLOW_DOWN_AFTER_FACTOR": {
+      "type": "string"
+    },
+    "SERVER_SLOW_DOWN_DELAY_MS": {
+      "type": "string"
```
boxhock, line 39: Should all of these be strings or int?

packages/core/bootstrap/src/lib/server.ts
```diff
@@ -124,3 +127,33 @@ function setupMetricsServer() {
 
   metricsApp.listen(metricsPort, () => logger.info(`Monitoring listening on port ${metricsPort}!`))
 }
+
+const windowMs = 1000 * 5
+const max = parseInt(process.env.SERVER_RATE_LIMIT_MAX || '250') // default to 250 req / 5 seconds max
+const delayAfter = max * (Number(process.env.SERVER_SLOW_DOWN_AFTER_FACTOR) || 0.8) // we start slowing down requests when we reach 80% of our max limit for the current interval
```
boxhock, line 133: Any reason we `parseInt()` the others and `Number()` here? Can we be consistent with them?

packages/core/bootstrap/src/lib/server.ts
```diff
@@ -124,3 +127,33 @@ function setupMetricsServer() {
...
+const delayMs = parseInt(process.env.SERVER_SLOW_DOWN_DELAY_MS || '500') // default to slowing down each request by 500ms
+
+function initExpressMiddleware(app: express.Express) {
+  app.set('trust proxy', 1)
```
boxhock, line 137: Did we not set this previously? What's the reason we're setting it now?

Should this be configurable by the operator, so they can disable this if they're not actually behind a proxy?

boxhock (changes requested):

packages/core/bootstrap/src/lib/server.ts
boxhock: Ah, is it because of decimals?

packages/core/bootstrap/src/lib/server.ts
justinkaseman: Did not set it previously. I added it because it is shown in the example for these middleware packages.

Based on some quick researching I didn't see a downside to leaving it on for everyone.
https://stackoverflow.com/questions/23413401/what-does-trust-proxy-actually-do-in-express-js-and-do-i-need-to-use-it

We could make it configurable if there is a risk, but otherwise the less environment variables to set the better.

packages/core/bootstrap/src/lib/server.ts
justinkaseman: Yes decimals ⬆️

packages/core/bootstrap/schemas/env.json
justinkaseman: Currently we don't type check them. When parsed from the environment they are always strings, so for that reason and consistency I set them as such.

I could see the argument for saying int here, though `SERVER_SLOW_DOWN_AFTER_FACTOR` is a float.

packages/core/bootstrap/src/lib/server.ts
boxhock: I see. Under the assumption that node ops have their networking set correctly, I guess it shouldn't be an issue.

If the adapters are reachable from the public however, this would allow attackers to spoof the requester IP address and essentially limit the legitimate requesters. But I think node ops are in bigger trouble if the adapters were public anyways.

boxhock (approved):

packages/core/bootstrap/src/lib/server.ts
justinkaseman: This scenario of spoofing IPs to get them rate limited won't actually even be relevant because of:
```
    keyGenerator: () => '*', // use one key for all incoming requests
    ```

PR #968, opened 2021-09-30
Feature/17539: add ability to spin up/down adapters in the qa cluster

tateexon:
## Description

Add scripts to spin up and down ephemeral adapters in the qa cluster. Currently coingecko and coinpaprika work with these changes.

## Changes

- Added scripts to start and stop adapters in the qa cluster. The scripts can be used for testing local changes or in CI.
- Added the cl-adapter helm chart temporarily. Once we have an internal helm repository to pull from then this can go away.

## Steps to Test

1. Run `ADAPTER=coingecko RELEASE_TAG=test yarn qa:adapter:start` with your kubectx set to the qa cluster to start an ephemeral adapter.
2. Run `ADAPTER=coingecko RELEASE_TAG=test yarn qa:adapter:stop` with your kubectx set to the qa cluster to stop an ephemeral adapter.

### DIFF
--- a/ephemeral/cl-adapter/templates/ingress.yaml
+++ b/ephemeral/cl-adapter/templates/ingress.yaml
@@ -1,0 +1,26 @@
+{{- $name := .Values.name -}}
+{{- $svcPort := .Values.service.port -}}
+apiVersion: networking.k8s.io/v1beta1
+kind: Ingress
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{- include "cl-adapter.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/rewrite-target: /$1
+spec:
+  rules:
+    - http:
+        paths:
+          - path: /{{ $name }}
+            pathType: Prefix
+            backend:
+              serviceName: adapter-{{ $name }}
+              servicePort: {{ $svcPort }}
+          - path: /{{ $name }}/(.*)
+            pathType: Prefix
+            backend:
+              serviceName: adapter-{{ $name }}
+              servicePort: {{ $svcPort }}
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,3 +8,5 @@ test-payload.json
 .yarn
 .vscode
 .pnp.cjs
+
+ephemeral/cl-adapter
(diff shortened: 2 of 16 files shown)


ephemeral/cl-adapter/templates/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  rules:
+    - http:
+        paths:
+          - path: /{{ $name }}
+            pathType: Prefix
+            backend:
+              serviceName: adapter-{{ $name }}
+              servicePort: {{ $svcPort }}
+          - path: /{{ $name }}/(.*)
```
HenryNguyen5, line 22: I'm guessing we need this path route so that any subroutes are still passed to the root level (which in this case is the EA?)

HenryNguyen5 (commented):
Looks great, I'm wondering if we should be writing these in TS though, since all of our other scripts and tooling are in TS. If we ever need more complex testing or coordination, then it would help having it in TS too. I did notice some duplication when it comes to env var checks also.

It's something to consider especially since we're trying to minimize tech debt.

Another thing that I'm curious about is how you feel about a single pod vs a job. I'm not sure how you're planning on running the k6 tests, but we could make a job (maybe even a cron one?) that spawns a pod with two containers, a test container and a container under test.

Going to re-review this after feedback on the above comments

tateexon:
@HenryNguyen5 
I can definitely switch to typescript although these scripts are more or less built for CI and are basically a kubectl call and 2 helm calls (of which the helm calls will be replaced in the future with the integrations-framework cli). I see an official node lib for kubectl but I do not see one for helm (and there likely won't be one for the integrations-framework cli at first either) so I will need to wrap those calls. What complex testing/coordination do you foresee?

As for the difference between a pod and a job, I was just reusing what is in infra-k8s to keep it as similar as possible to what we have on staging, the fewer differences there are the more likely we won't see test only issues. I am going to recommend not putting k6 in the same pod for a couple of reasons.
- If we keep the vpa in the deployed app then we risk having the k6 test shut down if the pod needs to scale. The vpa evicts pods on a scaling action.
- It is bad practice to put your load test on the same machine as the running application due to potential resource competition. In k8s this is not guaranteed even in separate pods but at least the app can jump to a different node if k6 decides to be a resource hog.
- We may want to run a k6 test against several adapters at once. Which pod would we put k6 on or would we put all the adapters and k6 in the same pod?

HenryNguyen5:
Great points, I definitely agree with you on them :D

thodges-gh (approved):

PR #969, opened 2021-09-30
Set token index

bogos:


packages/composites/defi-pulse/README.md
```diff
@@ -2,8 +2,14 @@
 
 The adapter combines real-time DPI allocation data from on-chain with off-chain price sources in order to calculate an accurate price for the DPI.
 
+## IMPORTANT
+
+This EA will be deprecated, set-token-index will be used for future reference. [Go to Set Token Index](../set-token-index/README.md)
```
boxhock, line 7: I like this message. Can we also get a WARN log level message in the `execute()` function saying the same thing?

boxhock (changes requested):

packages/composites/defi-pulse/README.md
bogos: done! :)

boxhock (approved):

PR #970, opened 2021-10-01
Bugfix for duplicate ws cache writes

cds95:
We've been seeing issues where adapters running with WS turned on have been maxing out their CPU usage.  This seems to have been because our adapters wrote the same WS message to the cache multiples times instead of just once.  The fix for this is to move the logic to cache messages into a separate epic.

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/actions.ts
+++ b/packages/core/bootstrap/src/lib/ws/actions.ts
@@ -9,6 +9,7 @@ export interface WSConfigPayload {
   config: WSConfig
   // TODO: wsHandler should not be sent as an event
   wsHandler: WSHandler
+  shouldNotRetryConnecting?: boolean
 }
 
 export interface WSConfigDetailedPayload extends WSConfigPayload {
@@ -101,6 +102,7 @@ export interface WSSubscriptionPayload {
 export interface WSSubscriptionErrorPayload extends WSErrorPayload {
   subscriptionMsg?: any
   input?: AdapterRequest
+  error?: any
 }
 
 export const subscribeRequested = createAction(
--- a/packages/core/bootstrap/src/lib/ws/actions.ts
+++ b/packages/core/bootstrap/src/lib/ws/actions.ts
@@ -102,7 +102,7 @@ export interface WSSubscriptionPayload {
 export interface WSSubscriptionErrorPayload extends WSErrorPayload {
   subscriptionMsg?: any
   input?: AdapterRequest
-  error?: any
+  error?: unknown
 }
 
 export const subscribeRequested = createAction(
--- a/packages/core/bootstrap/src/lib/ws/actions.ts
+++ b/packages/core/bootstrap/src/lib/ws/actions.ts
@@ -131,9 +131,11 @@ export interface WSMessagePayload {
   input: AdapterRequest
   context: AdapterContext
   connectionInfo: WSConnectionInfo
+  wsHandler: WSHandlerOverride
 }
 
 export const messageReceived = createAction('WS/MESSAGE_RECEIVED', asAction<WSMessagePayload>())
+export const writeToCache = createAction('WS/WRITE_MESSAGE_TO_CACHE', asAction<WSMessagePayload>())
 
 /** OVERRIDES */
 export interface WSHandlerOverride extends WSHandler {
--- a/packages/core/bootstrap/src/lib/ws/actions.ts
+++ b/packages/core/bootstrap/src/lib/ws/actions.ts
@@ -137,7 +137,6 @@ export interface WSMessagePayload {
 }
 
 export const messageReceived = createAction('WS/MESSAGE_RECEIVED', asAction<WSMessagePayload>())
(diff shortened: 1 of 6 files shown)


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -76,6 +77,62 @@ type ConnectRequestedActionWithState = [
   },
 ]
 
+export const writeMessageToCacheEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
```
justinkaseman, line 80: Nice! We need even more of this. (not for this PR, but great start)

Could we move this so that the order of reading this file is closer to their execution order?
(e.g. put this below `connectEpic`)

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -297,28 +365,23 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
...
                   return of(
                     subscribeFulfilled(payload),
-                    messageReceived({
-                      message,
-                      subscriptionKey,
-                      input: payload.input,
-                      context: payload.context,
-                      connectionInfo: payload.connectionInfo,
-                    }),
+                    writeToCache(actionPayload),
```
justinkaseman, line 380: Do we need a whole new action for `writeToCache`? Could we just use `messageRecieved` in the epic instead?

`messageRecieved` currently has no use besides in the metrics epic.

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -379,7 +442,10 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
       )
 
       // All received messages
```
justinkaseman, line 444:
```suggestion
      // All received messages using the same connection key
```

packages/core/bootstrap/src/lib/ws/epics.ts
cds95: Yeah I think we can do that

packages/core/bootstrap/src/lib/ws/actions.ts
```diff
@@ -101,6 +102,7 @@ export interface WSSubscriptionPayload {
 export interface WSSubscriptionErrorPayload extends WSErrorPayload {
   subscriptionMsg?: any
   input?: AdapterRequest
+  error?: any
```
justinkaseman, line 105:
```suggestion
  error?: unknown
```

packages/sources/coinmetrics/src/adapter.ts
```diff
@@ -78,6 +78,10 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
           url,
         }
       },
+      shouldRetryConnection: (closeContext) => {
```
justinkaseman, line 81: Do we know the shape of the error message that we can document here as an interface?

packages/core/types/@chainlink/index.d.ts
```diff
@@ -239,6 +239,8 @@ declare module '@chainlink/types' {
     heartbeatIntervalInMS?: number
     // Filters out messages that are not expected from sending a message constructed by one of the onConnect hooks
     isOnConnectChainMessage?: (message: any) => boolean
+    // Should try open connection again after error
+    shouldRetryConnection?: (errorMessage: any) => boolean
```
justinkaseman, line 243:
```suggestion
    shouldRetryConnection?: (errorMessage: unknown) => boolean
```

packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -638,8 +652,10 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
           action$.pipe(
             // TODO: not seeing unsubscribe events because of this
             filter(disconnectFulfilled.match),
-            tap((action) => logger.info('WS: Disconnected', connectionMeta(action.payload))),
             filter((a) => a.payload.config.connectionInfo.key === connectionKey),
+            tap((action) => {
+              logger.info('WS: Disconnected Fulfilled', connectionMeta(action.payload))
```
justinkaseman, line 596:
```suggestion
              logger.debug('WS: Disconnected Fulfilled', connectionMeta(action.payload))
```

cds95: I think `debug` would be better as there are times when we actually want to disconnect from a WS endpoint.

packages/sources/coinmetrics/src/adapter.ts
```diff
@@ -78,7 +79,7 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
           url,
         }
       },
-      shouldRetryConnection: (closeContext) => {
+      shouldRetryConnection: (closeContext: WebsocketErrorMessageSchema) => {
```
justinkaseman, line 82: Won't even need the type definition here because `shouldRetryConnection` is already typed. Yay TypeScirpt.

justinkaseman (approved):

PR #909, opened 2021-09-14
feat(coinmetrics): add "burned" endpoint

boxhock:
......

## Changes

- Added "burned" endpoint

## Steps to Test

1. Request in README

## Quality Assurance

- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `<ADAPTER_PACKAGE>/schemas/env.json` and `<ADAPTER_PACKAGE>/README.md`
- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.
- [x] The branch naming follows git flow (`feature/x`, `chore/x`, `release/x`, `hotfix/x`, `fix/x`) or is created from Clubhouse/Shortcut
- [x] This is related to a maximum of one Clubhouse/Shortcut story or GitHub issue
- [x] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)

PR #973, opened 2021-10-04
fix(coinmetrics): fixes burned endpoint to calc EIP-1559 burned ETH

vnavascues:
Fixes the `burned` endpoint from feature #909 to calculate the ETH burned since EIP-1559.

Breaking changes:
- Renamed endpoint as `burned-eth`

Changes:
- Removed custom params: `assets`, `frequency` and `pageSize` (all their values are now hard coded query parameters)
- Added custom params: `startDate` and `endDate` (both of them optional)
- Added pagination support 
- Calculate the cumulative burned ETH within a time range


packages/sources/coinmetrics/README.md
```diff
@@ -57,50 +57,41 @@ Endpoint to get the reference price of the asset.
 }
 ```
 
-## Burned Endpoint
+## Burned ETH Endpoint
```
boxhock, line 60: We should not be replacing this endpoint. Rather, we should create a new endpoint that utilizes the existing one.

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
-import { Requester, Validator } from '@chainlink/ea-bootstrap'
+import { Requester, Validator, AdapterError } from '@chainlink/ea-bootstrap'
 import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
-import { NAME as AdapterName } from '../config'
 import { ethers } from 'ethers'
+import { Decimal } from 'decimal.js'
+Decimal.set({ precision: 100 })
```
boxhock, line 5: Let's not introduce decimal.js unless we have to. Ethers' BigNumber library is our preferred BN library, and it can parse out the amounts we're dealing with here.

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
...
+
+export interface AssetMetrics {
+  asset: string
+  time: string
+  FeeTotNtv: string
+  IssTotNtv: string
+  RevNtv: string
+}
+
+interface ResponseData {
```
boxhock, line 26:
```suggestion
export interface ResponseData {
```

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
...
+      statusCode: 400,
+    })
+  }
+
+  if (Date.parse(startDate) > Date.parse(endDate)) {
+    throw new AdapterError({
+      jobRunID,
+      message: `Invalid date range: 'startDate': ${startDate} cannot be greater than 'endDate': ${endDate}.`,
+      statusCode: 400,
+    })
```
boxhock, line 117: How many of these validations are necessary in the adapter vs get caught by the data provider?

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
...
   }
 
-  const options = { ...config.api, params, url }
+  const options = { ...config.api, params, url: URL }
+
+  let totalBurnedETH = new Decimal(0)
+  let response
+  /*eslint no-constant-condition: ["error", { "checkLoops": false }]*/
+  while (true) {
+    response = await Requester.request(options)
```
boxhock, line 136:
```suggestion
    response = await Requester.request<ResponseData>(options)
```

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
...
 
-  const response = await Requester.request<ResponseSchema>(options)
-  const FeeTotNtv = ethers.utils.parseEther(response.data.data[0].FeeTotNtv)
-  const RevNtv = ethers.utils.parseEther(response.data.data[0].RevNtv)
-  const IssTotNtv = ethers.utils.parseEther(response.data.data[0].IssTotNtv)
-  const result = FeeTotNtv.sub(RevNtv.sub(IssTotNtv))
+    const responseData: ResponseData = response.data
```
boxhock, line 138: This is no longer necessary with the change above

```suggestion
    const responseData = response.data
```

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -1,63 +1,164 @@
...
+    if (!nextPageToken || assetMetricsList.length < PAGE_SIZE) break
+    options.params.next_page_token = nextPageToken
+  }
 
   return Requester.success(
     jobRunID,
-    Requester.withResult(response, ethers.utils.formatEther(result)),
+    Requester.withResult(
+      response,
+      ethers.utils.parseUnits(totalBurnedETH.toString(), 'ether').toString(),
```
boxhock, line 160: The data provider returns the units in ether, so we should also return the units in ether.

packages/sources/coinmetrics/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const PAGE_SIZE = 10_000
```
boxhock, line 1: Seems like this does not need it's own file

boxhock: If anything, this should be set as a default in the config file, and be modifiable by env vars.

packages/sources/coinmetrics/src/endpoint/burned.ts
vnavascues: All of them are caught by the data provider, however:

1. Any meaningful error message is hidden by how the `Requester` [processes failed requests](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/external-adapter/requester.ts#L52).

Below an example of the error response for a given datetime with bad format:
```typescript
{
    "error": {
        "type": "bad_parameter",
        "message": "Bad parameter 'start_time'. Incorrect time format '2021-1013'. Supported formats are 'yyyy-MM-dd', 'yyyyMMdd', 'yyyy-MM-ddTHH:mm:ss', 'yyyy-MM-ddTHHmmss', 'yyyy-MM-ddTHH:mm:ss.SSS', 'yyyy-MM-ddTHHmmss.SSS', 'yyyy-MM-ddTHH:mm:ss.SSSSSS', 'yyyy-MM-ddTHHmmss.SSSSSS', 'yyyy-MM-ddTHH:mm:ss.SSSSSSSSS', 'yyyy-MM-ddTHHmmss.SSSSSSSSS'."
    }
}
```

This error data is in `error.response.data`. As far as I'm aware the only way to improve the vague `Could not reach endpoint: bad request 400` would be something like this in `Requester.request()`:

```typescript
      try {
        response = await axios(config)
      } catch (error) {
        let message: string
        if (error.response) {
          message =
            `Request failed with status code ${error.response.status}. ` +
            `Response data: ${JSON.stringify(error.response.data)}`
        } else {
          message = JSON.stringify(error.message)
        }
        // Request error
        if (n === 1) {
          logger.error(`Could not reach endpoint: ${message}`)
          throw new AdapterError({ message, cause: error })
        }
        return await _delayRetry(`Caught error. Retrying: ${message}`)
      }
```

2. API `start_time` and `end_time` also accept datetime formats (see error message above). However, `FeeTotNtv, RevNtv, IssTotNtv` are requested by `1d` frequency (set at midnight). I think any valid format other than 'yyyy-MM-dd' and 'yyyyMMdd' is misleading, as the API won't complain and the requester may think the time part is being taken into account.

All in all I just wanted to provide a better understanding for both the user and node operator.

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -34,30 +60,48 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
+  while (true) {
+    response = await Requester.request<ResponseSchema>(options)
 
-  const response = await Requester.request<ResponseSchema>(options)
-  const FeeTotNtv = ethers.utils.parseEther(response.data.data[0].FeeTotNtv)
-  const RevNtv = ethers.utils.parseEther(response.data.data[0].RevNtv)
-  const IssTotNtv = ethers.utils.parseEther(response.data.data[0].IssTotNtv)
-  const result = FeeTotNtv.sub(RevNtv.sub(IssTotNtv))
+    const responseData = response.data
+    const { data: assetMetricsList } = responseData
```
boxhock, line 85:
```suggestion
    const { data: assetMetricsList } = response.data
```

packages/sources/coinmetrics/src/endpoint/burned.ts
```diff
@@ -34,30 +60,48 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
   }
 
-  const options = { ...config.api, params, url }
+  const options = { ...config.api, params, url: URL }
+
+  let totalBurnedTKN = BigNumber.from('0')
+  let response
+  /*eslint no-constant-condition: ["error", { "checkLoops": false }]*/
+  while (true) {
+    response = await Requester.request<ResponseSchema>(options)
```
boxhock, line 82: I don't think we should be doing this loop here, it's better fit for the total burned endpoint.

boxhock:
Hey @vnavascues, the concern with the proposed solution is that we create two extra endpoints just to hard code some params that we could just have set in the request.

From your comments it seems like the functionality of `total-burned` is all we need? And to keep the backwards compatibility with the current `burned` endpoint, all we need is to default to the params you set in your proposed solution for this endpoint.

I realize this is closer to the functionality of your initial commit, but I believe this will be sufficient with all input params exposed. Please let me know if you see any issues with this approach.

vnavascues:
> Hey @vnavascues, the concern with the proposed solution is that we create two extra endpoints just to hard code some params that we could just have set in the request.
> 
> From your comments it seems like the functionality of `total-burned` is all we need? And to keep the backwards compatibility with the current `burned` endpoint, all we need is to default to the params you set in your proposed solution for this endpoint.
> 
> I realize this is closer to the functionality of your initial commit, but I believe this will be sufficient with all input params exposed. Please let me know if you see any issues with this approach.

First of all sorry for not getting back to this sooner, I appreciate your feedback here and in the accuweather one.

I understand the concerns about creating endpoints that preset params to other endpoints. At the same time I also see the benefits of this pattern; making more straight to request very specific data (e.g. ETH burned since EIP-1559). Anyway, I'm keen on learning the do's and don'ts of implementing EAs and adopting your patterns.

Discarding these preset endpoints, my first commit `total-burned` is all we need as you said, having these inputs: `asset`, `frequency`, `pageSize`, `startDate` and `endDate`. However, my proposal above has a mistake re `burned` backward compatibility!

In the context of `total-burned` being the only endpoint, both `startDate` and `endDate` have to be **yesterday's date (as YYYY-MM-DD)** to mimic the `burned` endpoint. Therefore, if we still want to keep the `burned` endpoint it should execute `total-burned` presetting `startDate` and `endDate`. Nonetheless 2 endpoints instead of 3.

boxhock:
@vnavascues Got it, I think `total-burned + burned` would be a good solution then :+1:

vnavascues:
> @vnavascues Got it, I think `total-burned + burned` would be a good solution then +1

Great, I'll crack on with it. Thanks @boxhock

packages/sources/coinmetrics/README.md
```diff
@@ -86,21 +88,119 @@ Endpoint to calculate the number of burned coins/tokens for an asset.
...
-        "IssTotNtv": "13388.1875",
-        "RevNtv": "15282.160506520757527274"
+        "time": "2021-11-04T00:00:00.000000000Z",
+        "FeeTotNtv": "14465.059425601977193289",
+        "IssTotNtv": "13175.1875",
+        "RevNtv": "14887.21684537933981053"
+      }
+    ],
+    "next_page_token": "0.MjAyMS0xMS0wNFQwMDowMDowMFo",
+    "next_page_url": "https://api.coinmetrics.io/v4/timeseries/asset-metrics?assets=eth&metrics=FeeTotNtv,RevNtv,IssTotNtv&frequency=1d&page_size=1&api_key=hGeSkKBRGhDLIyskqnGE&next_page_token=0.MjAyMS0xMS0wNFQwMDowMDowMFo",
```
boxhock, line 104:
```suggestion
    "next_page_url": "https://api.coinmetrics.io/v4/timeseries/asset-metrics?assets=eth&metrics=FeeTotNtv,RevNtv,IssTotNtv&frequency=1d&page_size=1&next_page_token=0.MjAyMS0xMS0wNFQwMDowMDowMFo",
```

boxhock (changes requested):
Looks good to me! :smile: 

As we're just cutting a release atm, you'll have to rebase. Could you also please update the version in package.json? It will probably be `1.0.0` after the rebase, so you can change it to `1.1.0`.

boxhock:
@vnavascues Looks like CI wasn't too happy about the version update. You might have to update this in the `core/legos/package.json` reference to this adapter as well.

vnavascues:
@boxhock updated `core/legos/package.json` and generated again the `yarn.lock`.

boxhock (approved):
Looks good! :+1:

PR #975, opened 2021-10-05
Dxfeed WS batch requests fix

cds95:
This PR addresses the issue when a DXFeed adapter with WS turned on receives batch requests before it's on connect chain has finished running.  The issue was that subscription requests that were sent before the on connect chain completed never got a response from the server.  The fix is to prevent sending these requests until after the on connect chain has finished running.  

Another issue this PR fixes is the heartbeat mechanism to the WS server.  The code previously sent the heartbeat message in 30s intervals.  This didn't work as DXFeed requires the client to response to the server's heartbeat message instead of sending a message every 30s.

### DIFF
--- a/packages/core/bootstrap/src/lib/ws/epics.ts
+++ b/packages/core/bootstrap/src/lib/ws/epics.ts
@@ -467,6 +467,26 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
         }),
       )
 
+      const respondWithHeartbeat$ = message$.pipe(
+        filter(
+          (action) =>
+            !!wsHandler.shouldReplyToServerHeartbeat &&
+            wsHandler.shouldReplyToServerHeartbeat(action.payload.message),
+        ),
+        withLatestFrom(state$),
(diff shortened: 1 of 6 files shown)


packages/core/bootstrap/src/lib/ws/epics.ts
```diff
@@ -461,6 +473,26 @@ export const connectEpic: Epic<AnyAction, AnyAction, { ws: RootState }, any> = (
...
+        mergeMap(([action, state]) => {
+          const { connectionInfo, message } = action.payload
+          const key = connectionInfo.key
+          const { requestId, connectionParams } = state.ws.connections.all[key]
+          if (wsHandler.heartbeatReplyMessage) {
+            wsSubject.next(wsHandler.heartbeatReplyMessage(message, requestId, connectionParams))
+          }
+          return of(action)
+        }),
+        tap(() => logger.info('Responded to server heartbeat')),
```
justinkaseman, line 492: We probably want this on `logger.debug` otherwise it could be a lot of noise.

PR #976, opened 2021-10-05
integrate uniswap router adapter

drbh:
wip: adding integration tests next

## Steps to Test

Test like other adapter as specified on top level readme

```bash
yarn test uniswap-v2/test/unit
```

## Quality Assurance


drbh:
## Considerations

Slippage impacts the output price if there is not sufficient liquidity - this can cause very low/high prices if the pool is small. How do we guard against this? Is there a specific amount of liquidity that is needed to allow reporting?

Would we like to handle multihop paths? Right now we are only showing the price of a single swap - however sometimes better prices are provided if there is multiple hops in the transaction.

PR #977, opened 2021-10-05
Feature/dex uniswap v2

drbh:
wip: adding integration tests before moving draft PR to PR

## Steps to Test

Test like other adapter as specified on top level readme

```bash
yarn test uniswap-v2/test/unit
```

## Quality Assurance

## Additional Considerations

Slippage impacts the output price if there is not sufficient liquidity - this can cause very low/high prices if the pool is small. How do we guard against this? Is there a specific amount of liquidity that is needed to allow reporting?

Would we like to handle multihop paths? Right now we are only showing the price of a single swap - however sometimes better prices are provided if there is multiple hops in the transaction.

### DIFF
--- a/packages/sources/uniswap-v2/README.md
+++ b/packages/sources/uniswap-v2/README.md
@@ -1,0 +1,72 @@
+# Chainlink External Adapter for UniswapV2.fi
+
+This adapter allows querying UniswapV2.fi contracts
+
+### Environment Variables
+
+| Required? |         Name         |                                                            Description                                                             | Options |                 Defaults to                  |
+| :-------: | :------------------: | :--------------------------------------------------------------------------------------------------------------------------------: | :-----: | :------------------------------------------: |
+|    ✅     |       RPC_URL        |                           An http(s) RPC URL to a blockchain node that can read the UniswapV2 contracts                            |         |                                              |
+|           |   ADDRESS_PROVIDER   |            The address of the UniswapV2 address provider contract. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET            |         | `0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D` |
+|           | EXCHANGE_PROVIDER_ID | The index of the exchange provider contract in the address provider contract. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET |         |                     `2`                      |
+|           |  BLOCKCHAIN_NETWORK  |             The network to get pre-defined token addresses from. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET              |         |                  `ethereum`                  |
+
+---
+
+### Input Parameters
+
+| Required? |   Name   |     Description     |          Options           | Defaults to |
+| :-------: | :------: | :-----------------: | :------------------------: | :---------: |
+|           | endpoint | The endpoint to use | [crypto](#Crypto-Endpoint) |   crypto    |
+
+---
+
+## Crypto Endpoint
+
+Gets the exchange rate between two tokens
+
+### Input Params
+
+| Required? |            Name            |                                                     Description                                                      | Options | Defaults to |
+| :-------: | :------------------------: | :------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|    ✅     | `base`, `from`, or `coin`  |                                    The symbol or address of the currency to query                                    |         |             |
+|           |       `fromAddress`        |          Optional param to pre-define the address to convert from. If set, it takes precedence over `from`           |         |             |
+|           |       `fromDecimals`       | Optional param to pre-define the number of decimals in the `from` token. Setting this will make the query run faster |         |             |
+|    ✅     | `quote`, `to`, or `market` |                                 The symbol or address of the currency to convert to                                  |         |             |
+|           |        `toAddress`         |            Optional param to pre-define the address to convert to. If set, it takes precedence over `to`             |         |             |
+|           |        `toDecimals`        |  Optional param to pre-define the number of decimals in the `to` token. Setting this will make the query run faster  |         |             |
+|           |          `amount`          |               The exchange amount to get the rate of. The amount is in full units, e.g. 1 USDC, 1 ETH                |         |     `1`     |
+|           |        `resultPath`        |                                                 The result to fetch                                                  |         |   `rate`    |
+
+### Sample Input
+
+```json
+{
+  "id": "1",
+  "data": {
+    "from": "USDC",
+    "to": "USDT"
+  }
+}
+```
+
+### Sample Output
+
+```json
+{
+  "jobRunID": "1",
+  "result": 0.999465,
+  "statusCode": 200,
+  "data": {
+    "pool": "0xA5407eAE9Ba41422680e2e00537571bcC53efBfD",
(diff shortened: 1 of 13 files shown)


packages/sources/uniswap-v2/src/config.ts
```diff
@@ -0,0 +1,29 @@
...
+  uniswapRouter: string
+}
+
+export const makeConfig: ConfigFactory<Config> = (prefix: string | undefined) => {
+  return {
+    ...Requester.getDefaultConfig(prefix),
+    defaultEndpoint: DEFAULT_ENDPOINT,
+    provider: new ethers.providers.JsonRpcProvider(util.getRequiredEnv(ENV_RPC_URL, prefix)),
+    network: util.getEnv(ENV_BLOCKCHAIN_NETWORK, prefix) || DEFAULT_BLOCKCHAIN_NETWORK,
+    uniswapRouter: UNISWAP_ROUTER_CONTRACT,
```
thodges-gh, line 27: Let's allow the environment to override this so that we can use it on different networks and router contracts.

thodges-gh (changes requested):

packages/sources/uniswap-v2/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,122 @@
...
+ *     1. Check if the number of decimals was provided in the request.
+ *     2. Query the contract at the address found above to see how many decimals it's set to.
+ * @param validator The validation class to use
+ * @param direction Used to get the params in the request
+ * - `{direction}` is the symbol of the token (to find pre-set token details)
+ * - `{direction}Address` is the token address as set in the request
+ * - `{direction}Decimals` is the number of decimals for the token as set in the request
+ * @param config Configuration to extract token decimals from
+ */
+const getTokenDetails = async (
```
boxhock, line 92: Is this the exact same function as on the Curve adapter? If so, would it make sense to collect these in a utility function?

packages/sources/uniswap-v2/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This adapter allows querying UniswapV2.fi contracts
+
+### Environment Variables
+
+| Required? |        Name        |                                                Description                                                | Options |                 Defaults to                  |
+| :-------: | :----------------: | :-------------------------------------------------------------------------------------------------------: | :-----: | :------------------------------------------: |
+|    ✅     |      RPC_URL       |               An http(s) RPC URL to a blockchain node that can read the UniswapV2 contracts               |         |                                              |
+|           | BLOCKCHAIN_NETWORK |    The uniswap router address to get price from. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET     |         | `0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D` |
+|           |  ROUTER_CONTRACT   | The network to get pre-defined token addresses from. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET |         |                  `ethereum`                  |
```
boxhock: Looks like the names of these two are switched

boxhock (changes requested):

packages/sources/uniswap-v2/README.md
boxhock: Nvm fixed in f054d38d92a2e5aefa10e6ddaa45f79f39d75a27 :sweat_smile:

drbh: woops, yea I just noticed that and pushed a typo commit

packages/sources/uniswap-v2/src/endpoint/crypto.ts
drbh: yes it is, and I agree that it would be useful for other DEX adaptors. Where is the best place to move this function?

packages/sources/uniswap-v2/src/config.ts
drbh: added ability to overwrite ROUTER_CONTRACT via env var in recent commit

packages/sources/uniswap-v2/src/endpoint/crypto.ts
boxhock: I'm leaning towards the validator: https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/external-adapter/validator.ts#L175

@justinkaseman any opinions?

drbh: that looks like a logical place to me 👍

drbh: the only potential issue with moving it into `validator.ts` is that we'll need to import `ethers` and add the `erc20.json` abi into the validator. Do we want to couple web3 logic with the Validator?

Possible changes needed in `validator.ts`
```js
import { ethers, BigNumber } from 'ethers'
import { Decimal } from 'decimal.js'

...


  getDecimals = async (address: string, config: Config): Promise<number> =>
  new ethers.Contract(address, erc20ABI, config.provider).decimals()

  /**
   * getTokenDetails will find the address and number of decimal for a token.
   *
   * The order of operations is as follows:
   *  - address:
   *     1. Check if the address was provided in the request.
   *     2. If not, check the symbol in the request to see if we have pre-set the address for this symbol/network.
   *     3. If not, we assume the symbol param was actually the address.
   *  - decimals:
   *     1. Check if the number of decimals was provided in the request.
   *     2. Query the contract at the address found above to see how many decimals it's set to.
   * @param validator The validation class to use
   * @param direction Used to get the params in the request
   * - `{direction}` is the symbol of the token (to find pre-set token details)
   * - `{direction}Address` is the token address as set in the request
   * - `{direction}Decimals` is the number of decimals for the token as set in the request
   * @param config Configuration to extract token decimals from
   */
  getTokenDetails = async (
    adapter: string,
    validator: Validator,
    direction: string,
    config: Config,
  ): Promise<{ address: string; decimals: number }> => {
    const symbol = this.overrideSymbol(
      adapter,
      this.validated.data[direction],
    ) as string
    const address =
      this.validated.data[`${direction}Address`] ||
      this.overrideToken(symbol, config.network) ||
      symbol
    const decimals =
      this.validated.data[`${direction}Decimals`] || (await this.getDecimals(address, config))

    return { address, decimals }
  }
```

boxhock: I see, that also has a dependency on `Config` (which we can move out, but we still need the ethers provider).

This is growing in scope, so I'm OK with just leaving this as-is for now and I'll create a story for us to put more thought into this later.

I'll do a more thorough review tomorrow with A/T

boxhock (changes requested):
Just two small comments, otherwise LGTM. Good job :slightly_smiling_face:

boxhock (approved):

packages/sources/uniswap-v2/README.md
```diff
@@ -0,0 +1,70 @@
+# Chainlink External Adapter for UniswapV2.fi
+
+This adapter allows querying UniswapV2.fi contracts
```
boxhock, line 3:
```suggestion
# Chainlink External Adapter for Uniswap V2

This adapter allows querying Uniswap V2 contracts
```

thodges-gh (approved):

boxhock (approved):

PR #979, opened 2021-10-05
refactor(token-allocation): add extra logs to assist with debugging

justinkaseman:
## Description
Adds additional logs for debugging Token Allocations, which is used by many composite adapters.

Handles two cases:
1. Source adapter request errors
2. Null or 0 value returned for a pair

Story #17460


boxhock (approved):
Love this :+1:

PR #985, opened 2021-10-07
Feature/dex uniswap v3

drbh:
## Steps to Test

Test like other adapters as specified on top level readme

```bash
yarn test uniswap-v3/test/unit
yarn test uniswap-v3/test/integration

export RPC_URL=<WEB3 ENDPOINT>
yarn test uniswap-v3/test/e2e
```

## Quality Assurance

## Additional Considerations

Would we like to handle multihop paths? Right now we are only showing the price of a single swap - however sometimes better prices are provided if there is multiple hops in the transaction. If we'd like to support this functionality we'll have to use a different method on the quoter contract.

### DIFF
--- a/packages/sources/uniswap-v3/README.md
+++ b/packages/sources/uniswap-v3/README.md
@@ -27,16 +27,17 @@ Gets the exchange rate between two tokens
 
 ### Input Params
 
-| Required? |            Name            |                                                     Description                                                      | Options | Defaults to |
-| :-------: | :------------------------: | :------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
-|    ✅     | `base`, `from`, or `coin`  |                                    The symbol or address of the currency to query                                    |         |             |
-|           |       `fromAddress`        |          Optional param to pre-define the address to convert from. If set, it takes precedence over `from`           |         |             |
-|           |       `fromDecimals`       | Optional param to pre-define the number of decimals in the `from` token. Setting this will make the query run faster |         |             |
-|    ✅     | `quote`, `to`, or `market` |                                 The symbol or address of the currency to convert to                                  |         |             |
-|           |        `toAddress`         |            Optional param to pre-define the address to convert to. If set, it takes precedence over `to`             |         |             |
-|           |        `toDecimals`        |  Optional param to pre-define the number of decimals in the `to` token. Setting this will make the query run faster  |         |             |
-|           |          `amount`          |               The exchange amount to get the rate of. The amount is in full units, e.g. 1 USDC, 1 ETH                |         |     `1`     |
-|           |        `resultPath`        |                                                 The result to fetch                                                  |         |   `rate`    |
+| Required? |            Name            |                                                     Description                                                      | Options |    Defaults to     |
+| :-------: | :------------------------: | :------------------------------------------------------------------------------------------------------------------: | :-----: | :----------------: |
+|    ✅     | `base`, `from`, or `coin`  |                                    The symbol or address of the currency to query                                    |         |                    |
+|           |       `fromAddress`        |          Optional param to pre-define the address to convert from. If set, it takes precedence over `from`           |         |                    |
+|           |       `fromDecimals`       | Optional param to pre-define the number of decimals in the `from` token. Setting this will make the query run faster |         |                    |
+|    ✅     | `quote`, `to`, or `market` |                                 The symbol or address of the currency to convert to                                  |         |                    |
+|           |        `toAddress`         |            Optional param to pre-define the address to convert to. If set, it takes precedence over `to`             |         |                    |
+|           |        `toDecimals`        |  Optional param to pre-define the number of decimals in the `to` token. Setting this will make the query run faster  |         |                    |
+|           |          `amount`          |               The exchange amount to get the rate of. The amount is in full units, e.g. 1 USDC, 1 ETH                |         |        `1`         |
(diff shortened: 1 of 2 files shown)


packages/sources/uniswap-v3/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,156 @@
...
+  )
+
+  // build final data payload
+  encoded = methodSignature + encoded.substring(2)
+
+  // call to get expected price
+  const res = await config.provider.call({
+    to: config.uniswapQuoter,
+    data: encoded,
+  })
```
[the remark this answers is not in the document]
drbh: 🧠 wow! that fantastic, it worked as a drop in replacement without issue! Thanks @boxhock!

packages/sources/uniswap-v3/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,136 @@
...
+  // execute non view function as a call
+  const result = await quoterContract.callStatic.quoteExactInputSingle(
+    from,
+    to,
+    config.feeAmount,
+    amount,
+    sqrtPriceLimitX96,
+  )
+
+  return [result]
```
boxhock, line 135: Since we only have a single value, can we just return that instead?

```suggestion
  return await quoterContract.callStatic.quoteExactInputSingle(
    from,
    to,
    config.feeAmount,
    amount,
    sqrtPriceLimitX96,
  )
```

packages/sources/uniswap-v3/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This adapter allows querying Uniswap V3 contracts
+
+### Environment Variables
+
+| Required? |        Name        |                                                 Description                                                 | Options |                 Defaults to                  |
+| :-------: | :----------------: | :---------------------------------------------------------------------------------------------------------: | :-----: | :------------------------------------------: |
+|    ✅     |      RPC_URL       |               An http(s) RPC URL to a blockchain node that can read the Uniswap V3 contracts                |         |                                              |
+|           |  QUOTER_CONTRACT   | The address of the Uniswap V3 address quoter contract. NOTE: THIS SHOULD NOT BE CHANGED ON ETHEREUM MAINNET |         | `0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6` |
+|           |        FEE         |                                          The Uniswap V3 fee amount                                          |         |                    `500`                     |
```
boxhock, line 11: To my knowledge, isn't the fee based on the pool? In other words, should we make this a request parameter instead? I guess we can leave the DEFAULT_FEE as an env var.

kevincharm: Mostly sounds good. However we should take into account that the default fee pools (0.05%, 0.3%, 1%) don't all exist for every pool - they must each be created: https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Factory.sol#L35
Example, this token only has 2 fee tiers: https://info.uniswap.org/#/tokens/0x0ace32f6e87ac1457a5385f8eb0208f37263b415

boxhock: @drbh I don't see any iteration of the fee tiers in the commits (nor the configuration of the different tiers as discussed above). Are you sure it got pushed up?

packages/sources/uniswap-v3/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,147 @@
...
+  const quoterContract = new ethers.Contract(config.uniswapQuoter, quoterABI, config.provider)
+
+  // encode inputs for quoteExactInputSingle
+  const sqrtPriceLimitX96 = 0
+  let bestPrice = BigNumber.from(0)
+
+  // iterate over fee tiers
+  for (const fee of feeTiers) {
+    // execute non view function as a call
+    const price = await quoterContract.callStatic.quoteExactInputSingle(
```
boxhock, line 133: What happens if the fee tier does not exist for this from/to? Do we have to wrap this in try/catch, or will it just return a price of 0?

drbh: good point. `quoteExactInputSingle` does throw an error when the pool/fee combination does not exists. try/catch has been added.

packages/sources/uniswap-v3/src/endpoint/crypto.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (validator.error) throw validator.error
+  const jobRunID = validator.validated.id
+  const { address: from, decimals: fromDecimals } = await getTokenDetails(validator, 'from', config)
+  const { address: to, decimals: toDecimals } = await getTokenDetails(validator, 'to', config)
+  const inputAmount = validator.validated.data.amount || 1
+  const amount = BigNumber.from(inputAmount).mul(BigNumber.from(10).pow(fromDecimals))
+  const resultPath = validator.validated.data.resultPath
+
+  const feeTiers = validator.validated.data.feeTiers || config.feeTiers
+  const output = await getBestRate(from, to, amount, feeTiers, config)
```
boxhock, line 49: Can we throw an error if the output is 0? Or in some way catch it if no pool exists for the from/to (or the requested fee tiers do not exists)?

Basically, we want to avoid returning the price as 0 if the pool/fee tiers do not exist.

boxhock (changes requested):
Sorry for the back and forth. I think we're good once these are resolved.

PR #991, opened 2021-10-07
fix(bootstrap): cache Warmer correctly formats include overrides

justinkaseman:


packages/core/bootstrap/src/lib/ws/index.ts
```diff
@@ -70,6 +74,7 @@ const isConnected = (store: Store<RootState>, connectionKey: string): boolean =>
   const isActiveConnection = connectionState.active
   const isConnecting = connectionState.connecting > 1
   const hasOnConnectChainCompleted = connectionState.isOnConnectChainComplete
+  console.log(isActiveConnection, !isConnecting, hasOnConnectChainCompleted)
```
cds95, line 77: Do we still need this?

justinkaseman: No

cds95 (approved):

PR #993, opened 2021-10-08
Fix DXFeed connection issues

cds95:
- Fix DXFeed WS connection issue by making sure heartbeat message is always sent when the server expects a response from it's WS message
 -  In order to keep the code simple, the adapter will only start using WS once the `onConnectChain` has completed.  This means it'll take a bit longer for it to start using WS.  I tested this and it'll start using WS after about 30s.


packages/core/bootstrap/src/lib/ws/reducer.ts
```diff
@@ -209,6 +203,7 @@ export const subscriptionsReducer = createReducer<SubscriptionsState>(
 
       state.all[key].active = false
       state.all[key].unsubscribed = true
+      state.all[key].subscribing = 0
```
cds95, line 206: Want to double check this.  Should this be set to 0 if the adapter successfully unsubscribes?

PR #1005, opened 2021-10-14
update AP election response to set winner info in fields

cds95:
This PR updates the response from the AP Election data provider to include the winner's first name, last name and vote count as separate fields in the response's data payload.


boxhock (approved):
Looks good :+1:

PR #1007, opened 2021-10-14
NBA Sports Data

aiagentdev-1990:
This PR adds support for fetching NBA player data to the SportsDataIO provider.  It does the following:

 - Add new `nba` sport to the `sportdataio` adapter
 - Add `player-stats` endpoint to NBA sport
 - Add integration tests to verify output from NBA sport

### DIFF
--- a/packages/sources/sportsdataio/src/sport/nba/endpoint/player-stats.ts
+++ b/packages/sources/sportsdataio/src/sport/nba/endpoint/player-stats.ts
@@ -12,6 +12,7 @@ const customParams = {
 }
 
 export interface ResponseSchema {
+  epochS: number
   StatID: number
   TeamID: number
   PlayerID: number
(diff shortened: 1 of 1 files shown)


packages/sources/sportsdataio/src/sport/nba/endpoint/player-stats.ts
```diff
@@ -0,0 +1,173 @@
...
+  return Requester.success(
+    jobRunID,
+    Requester.withResult(response, packResponse(response.data)),
+    config.verbose,
+  )
+}
+
+const packResponse = (response: ResponseSchema): string => {
+  const dataTypes = [
+    'uint16',
```
cds95, line 124: I had to change this to `uint16` as `Season` did not fit into `uint8` specified in the spec https://www.notion.so/chainlink/LaMelo-Ball-NFT-ether-cards-0d4f5ca7f40a4b689daa799d9065e158.

PR #1008, opened 2021-10-15
[Feat] Nomics volume endpoint

justinkaseman:
### Description
Nomics requires the result path to be nested to get `volume`.
This PR changes the result path types to support arrays.


momentmaker (approved):
looks great!

PR #1016, opened 2021-10-20
feat: view-function adapter

### LINKED ISSUE #1011: view-function EA
These projects have similar requests that can be handled with a single adapter that only needs to have a different job spec. This will also save us time in the future since it's likely for this to become a common type of request. We can call this a "view-function" adapter.

Request params:

- Function signature (string)
-- Should be formated [how Ethers specifies the short-hand for ABIs](https://docs.ethers.io/v5/single-page/#/v5/getting-started/-%23-getting-started--contracts)

- Contract address (string)

Config:
- RPC_URL

Returns:
- The hex of the response from the function call

karen-stepanyan:
fix #1011

## Changes

- new EA adaper

## Steps to Test

`yarn test packages/sources/view-function/test/unit/`

## Quality Assurance


packages/sources/view-function/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Sample Output
+
+```json
+{
+  "jobRunID": 1,
+  "result": "0x0000000000000000000000000000000000000000000000000000000055534454",
+  "statusCode": 200,
+  "data": {
+    "result": "0x0000000000000000000000000000000000000000000000000000000055534454"
```
thodges-gh, line 53: The encoded result of a string should be larger than this. For example:
```
0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000045553445400000000000000000000000000000000000000000000000000000000
```
We should be replicating the encoding for any type rather than converting to hex.

packages/sources/view-function/src/endpoint/function.ts
```diff
@@ -0,0 +1,64 @@
...
+  const fnSignature = validator.validated.data.signature
+  const params = validator.validated.data.inputParams || []
+
+  if (typeof address !== 'string' || !address.length) {
+    throw new AdapterError({
+      jobRunID,
+      message: `Input, at 'address' or 'contract' path, must be a non-empty string.`,
+      statusCode: 400,
+    })
+  }
```
thodges-gh, line 29: Isn't this already handled by the validator?

packages/sources/view-function/src/endpoint/function.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (typeof fnSignature !== 'string' || !fnSignature.length) {
+    throw new AdapterError({
+      jobRunID,
+      message: `Input, at 'signature' or 'function' path, must be a non-empty string.`,
+      statusCode: 400,
+    })
+  }
```
thodges-gh, line 37: Isn't this already handled by the validator?

packages/sources/view-function/src/endpoint/function.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const contract = new Contract(address, [fnSignature], config.provider)
+  const fnName = Object.keys(contract.functions)[1]
+  let executionResult = await contract[fnName](...params)
```
thodges-gh, line 41: Can you check if using [a provider call](https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-call) would be better? We really just want to return the hex of the function since that's pre-formatted for its return types.

thodges-gh (changes requested):

packages/sources/view-function/src/endpoint/function.ts
karen-stepanyan: Validator throws error only if the field is `undefined`. `null` and `empty strings` are not validated

boxhock: Created a ticket for it: https://app.shortcut.com/chainlinklabs/story/19390/ea-validator-should-check-if-request-param-strings-are-empty

thodges-gh (approved):

PR #1018, opened 2021-10-20
BEA upgraded EA

### LINKED ISSUE #1013: BEA upgraded EA
The CPI feed using BEA EA (which is a consumer price index feed) uses a very old adapter, pulling from a public API that only updates once per month. 

The current EA is written in Golang, and we'd like to migrate over to EAv2 using typescript.

---

Existing adapter: [https://github.com/smartcontractkit/bea-adapter](https://github.com/smartcontractkit/bea-adapter)

@boxhock  also has context on this. When we originally wrote the adapter, we got an error from the API that our connection type wasn't allowed (using nodejs), but it worked in Golang so that's what we went with. There's no way of knowing if that restriction is still in place (or if it can be worked around) without attempting to make the adapter again. Suggest you validate the API connection using node as soon as possible before completing the adapter.

Assuming the connection works, make a new adapter for `bea-adapter` within this repo and add an endpoint for cpi to perform this calculation.

The adapter itself is very simple. You'll reach out to [`https://apps.bea.gov/api/data/`](https://apps.bea.gov/api/data/) with the following query parameters:

- "DataSetName":  "NIPA",
"TableName":    "T20804",
"ResultFormat": "json",
"method":       "getData",
"Frequency":    "M",
"Year":         `$(currentYear),$(currentYear-1)`

All the query params can be hard coded except for the year, which should dynamically calculate the current year (as 4-digit), and the previous year, separated by a comma. The response schema is in the existing adapter and would just need to be ported to TypeScript. The result calculation can also be ported to TypeScript using the existing adapter as a base.

bogos:
## Description

BEA upgraded EA

## Changes

- Add BEA adapter

## Steps to Test
```
export API_KEY=<API_KEY>
yarn test bea/test/e2e
```

## Quality Assurance


packages/sources/bea/src/endpoint/price.ts
```diff
@@ -0,0 +1,132 @@
...
+  const options = { ...config.api, params, url }
+  const response = await Requester.request<ResponseSchema>(options, customError)
+
+  if (response.data.BEAAPI.Error) {
+    throw new AdapterError({
+      jobRunID,
+      message: response.data.BEAAPI.Error.ErrorDetail.Description,
+      statusCode: 400,
+    })
+  }
```
thodges-gh, line 105: This should be handled in `customError`.

thodges-gh (changes requested):
One thing to check out and I'd like @boxhock to look over it since he made the original adapter. I verified that I receive the same result as the golang adapter.

packages/sources/bea/src/endpoint/price.ts
```diff
@@ -0,0 +1,132 @@
...
+  response.data.BEAAPI.Results.Data.forEach((element: DataSchema) => {
+    if (element.SeriesCode === series) {
+      const date = element.TimePeriod.split('M')
+      const year = date[0]
+      const month = date[1]
+      values.push({
+        value: Number(element.DataValue),
+        year: year,
+        month: month,
+      })
```
boxhock, line 117: I believe these values can come out of order in some cases. We should ensure this array is sorted by year/month before we fetch the N latest values.

packages/sources/bea/src/endpoint/price.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  let sum = 0
+  let count = 0
+  for (let i = 0; values.length > 0 && count < last; i++) {
+    const item = values.pop() as ValueSchema
+    sum += item.value
+    count += 1
+  }
+
+  response.data.result = sum / last
```
boxhock, line 129: Since we allow `last` to be set by the request params, this could be a value higher than `values.length`. To get the right average, we should divide this by `count` instead.

packages/sources/bea/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Input Parameters
+
+| Required? |   Name   |     Description     |        Options         | Defaults to |
+| :-------: | :------: | :-----------------: | :--------------------: | :---------: |
+|           | endpoint | The endpoint to use | [price](#Bea-Endpoint) |    price    |
+
+---
+
+## Bea Endpoint
```
boxhock, line 19: The endpoint name here doesn't match what the endpoint name is in the code.

Anyways, can we call this the AVG endpoint? And also add the description that we will get the average of the last N values.

boxhock (changes requested):

packages/sources/bea/test/e2e/price.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    ]
+
+    requests.forEach((req) => {
+      it(`${req.name}`, async () => {
+        const adapterResponse = await execute(req.testData as AdapterRequest)
+        assertSuccess(adapterResponse.statusCode, adapterResponse, jobID)
+      })
+    })
+  })
+})
```
boxhock, line 43: Could we please add integration tests with a mock endpoint as well?

That would give a good opportunity to test that the sorting works as expected as well, by purposely giving the response items out of order.

thodges-gh (approved):

PR #1032, opened 2021-10-25
feat(accuweather): get current conditions by coordinates

vnavascues:
## Description

AccuWeather EA. It provides 3 endpoints:

- `location`: requests [Locations API](https://apidev.accuweather.com/developers/locationsAPIguide), in particular gets a location information by geoposition.
- `current-conditions`: requests [Current Conditions](https://apidev.accuweather.com/developers/currentConditionsAPIGuide), gets the current weather conditions of a location by its location Key.
- `location-current-condition`: calls first `location` and then `current-conditions`.

By default each endpoint returns the data encoded. The reason behind is to avoid "stack too deep" in the fulfillment function (declaration and body) due to the amount of data (from location info, and weather conditions).

Data conversions (e.g. decimals to integers), and tuple encoding are done in the EA instead of in the TOML job. The reason behind is that I was unable to encode as tuple using `ethabiencode`.

## Steps to Test

1. Set `API_ENDPOINT=http://apidev.accuweather.com` env var in case of having AccuWeather API dev credentials.
2. [The TOML jobspec](https://gist.github.com/vnavascues/e783090b495239dce5efc7fe0e614388)
3. [A consumer contract](https://gist.github.com/vnavascues/d8c25cc6509b3af921d3566335935d6e)

## Quality Assurance


packages/sources/accuweather/src/endpoint/current-conditions.ts
```diff
@@ -0,0 +1,262 @@
...
+  currentConditionsResult.precipitationPastHour = Math.trunc(
+    currentConditionsResult.precipitationPastHour * 100,
+  )
+  currentConditionsResult.pressure = Math.trunc(currentConditionsResult.pressure * 100)
+  currentConditionsResult.temperature = Math.trunc(currentConditionsResult.temperature * 10)
+  currentConditionsResult.windSpeed = Math.trunc(currentConditionsResult.windSpeed * 10)
+
+  // NB: currently all conditions are expected to be numbers
+  Object.entries(currentConditionsResult).forEach(([key, value]) => {
+    if (Number.isNaN(value))
```
austinborn, line 162: Will any of these numbers NOT be integers? If they're all supposed to be ints, we could make this a stronger check (`Number.isInteger(value)`)

packages/sources/accuweather/src/endpoint/location-current-conditions.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  const locationResult: LocationResult = locationResponse.data.result
+  // No location found for the given coordiantes
+  if (locationResult.locationFound === false) {
+    let result: LocationCurrentConditionsResult | LocationCurrentConditionsResultEncoded
+    if (encodeResult) {
+      result = noEndpointResultEncoded
+    } else {
+      result = noEndpointResult
+    }
```
austinborn, line 65:
```suggestion
    const result: LocationCurrentConditionsResult | LocationCurrentConditionsResultEncoded = (
      encodeResult
        ? noEndpointResultEncoded
        : noEndpointResult
    )
```

packages/sources/accuweather/test/integration/location-current-conditions.ts
```diff
@@ -0,0 +1,180 @@
...
+        expect(response.body).toMatchSnapshot()
+      })
+    })
+  })
+}
+
+// no location found as JSON
+// no location found encoded
+// found as JSON
+// found encoded
```
austinborn, line 180: Are these suggestions for future tests? Maybe you can move these to the top and make it explicit?

austinborn (commented):
Thank you for contributing! From a code clarity and correctness standpoint, it looks great to me. With the result data to be recorded on-chain, it looks like a lot to be saving on-chain. I'd like someone with more familiarity to comment on that aspect of adapters (@boxhock? @kevincharm?) before approving.

packages/sources/accuweather/src/endpoint/current-conditions.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+  let currentConditionsResult: CurrentConditionsResult
+  try {
+    currentConditionsResult = getCurrentConditionsResult(units, currentConditionsList)
+  } catch (error) {
+    throw new AdapterError({
+      jobRunID,
+      message: `Unprocessable response by location key: ${locationKey}. ${error}.`,
+      statusCode: 400,
+    })
```
boxhock, line 236: We shouldn't need to throw `AdapterError` here.
We'll catch it later and add relevant context, so we should just throw a standard `Error` here

packages/sources/accuweather/src/endpoint/current-conditions.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+  let result: CurrentConditionsResult | string
+  if (encodeResult) {
+    try {
+      result = encodeCurrentConditionsResult(currentConditionsResult)
+    } catch (error) {
+      throw new AdapterError({
```
boxhock, line 243: same here

boxhock: And all other `AdapterError`s being thrown

boxhock (changes requested):
Just nit-pick really

boxhock:
@austinborn that's a valid concern. Luckily, the new TOML job specs on the node gives us more flexibility to select and encode only the data we want if the EA returns the full response payload (`API_VERBOSE=true`). The EA currently also allows for encoding a tuple with values, which is a good (right now, at least, but should be improved in the future when core has better support for this) way to return "a lot" of data.

packages/sources/accuweather/src/endpoint/current-conditions.ts
vnavascues: This is a very good one, thanks!

packages/sources/accuweather/test/integration/location-current-conditions.ts
vnavascues: Oh, just comments to be removed.

vnavascues:
@austinborn @boxhock thanks for reviewing it. I do understand the concerns about the amount of data written on-chain. Given the hackaton context I wanted to provide an easy-to-setup (jobspec-wise) and easy-to-consume (contract-wise) experience where no one has to worry about getting a stack too deep. As Jonas said with `API_VERBOSE=true` you get much more granularity.

I also considered implementing a couple of optional input params that would allow the consumer to define the data to be requested/written on-chain, i.e. `locationFields` and `currentConditionFields`. For instance:

```solidity
        req.add("locationFields", "locationKey");
        req.add("currentConditionFields", "temperature,pressure,weatherIcon");
```

The consumer would have to define the desired tuples in the contract:

```solidity
struct Location {
  uint256 locationKey;
}
struct CurrentConditions {
  int16 temperature;
  uint24 pressure;
  uint8 weatherIcon;
}
```
Duplicated/nonexistent fields would not be accepted, and field order matters (still tuple encoding).

packages/sources/accuweather/src/endpoint/current-conditions.ts
vnavascues: Sorry, by catching it later you mean in the `location-current-conditions` endpoint?

Due to `location` and `current-conditions` are exposed endpoints too, I throw `AdapterError` in each one.
 
I get that the `location` endpoint is not very "appealing" and probably no one will ever set a job for it. On the contrary, I see value/utility on the `current-conditions` once you know the location key.

What do you think?

boxhock: So we can just throw a regular error here:
```suggestion
    throw new Error(`Unprocessable response by location key: ${locationKey}. ${error}.`)
```

The [EA framework will catch this](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/index.ts#L213-L227) and give the appropriate feed ID, job run ID, etc. The only thing you gain from the current implementation is that you can set an appropriate status code.

vnavascues: done!

packages/sources/accuweather/src/endpoint/current-conditions.ts
vnavascues: done!

boxhock (approved):
Nice work :+1:

PR #1054, opened 2021-11-01
add logs for endpoint, feedid, reason on errors

### LINKED ISSUE #1023: Improve error logging within our EA framework
The goal is to improve the data we log when EAs throw an error

Currently, data providers can return detailed error messages on bad requests, but they are not logged in our EA framework. It's usually something like:

```Could not reach endpoint: bad request 400```

We should also log the response from the data provider in case it contains useful information. The goal is to standardize and return the valuable info here.

Additional data points:
1. which endpoint threw the error
2. the raw error passed through
3. the feed id
4. the job id

### PR DESCRIPTION
**Pull request by karen-stepanyan**
## Changes

Added more info when Requester throws errors. (endpoint, jobid, feedid, raw error)

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/index.ts
+++ b/packages/core/bootstrap/src/index.ts
@@ -1,6 +1,7 @@
 import {
   AdapterMetricsMeta,
   AdapterRequest,
+  AdapterRequestData,
   AdapterContext,
   Execute,
   ExecuteSync,
@@ -79,13 +80,39 @@ const withStatusCode: Middleware = async (execute, context) => async (input) => 
 
 // Log adapter input & output data
 const withLogger: Middleware = async (execute, context) => async (input: AdapterRequest) => {
+  interface IError {
+    jobRunID: string
+    params: AdapterRequestData
+    message: string
+    feedID: string
+    endpoit?: string
+    errorResponse?: any
+    rawError?: any
+  }
   Logger.debug('Input: ', { input })
   try {
     const result = await execute(input, context)
     Logger.debug(`Output: [${result.statusCode}]: `, { output: result.data })
     return result
   } catch (error) {
-    Logger.error(error.toString(), { stack: error.stack })
+    const feedID = metrics.util.getFeedId(input)
+    const errorLog: IError = {
+      message: error.toString(),
+      jobRunID: input.id,
+      params: input.data,
+      feedID,
+    }
+    if (error.endpoint) {
+      errorLog.endpoit = error.endpoint
+    }
+    if (error.errorResponse) {
+      errorLog.errorResponse = error.errorResponse
+    }
+    if (Logger.level === 'trace') {
+      errorLog.rawError = error.rawError
+    }
+
+    Logger.error({ ...errorLog, stack: error.stack })
     throw error
   }
 }
@@ -195,7 +222,8 @@ export const executeSync: ExecuteSync = async (
 
     return callback(result.statusCode, result)
   } catch (error) {
-    return callback(error.statusCode || 500, Requester.errored(data.id, error))
+    const feedID = metrics.util.getFeedId(data)
+    return callback(error.statusCode || 500, Requester.errored(data.id, error, undefined, feedID))
   }
 }
 
--- a/packages/core/bootstrap/src/index.ts
+++ b/packages/core/bootstrap/src/index.ts
@@ -85,9 +85,10 @@ const withLogger: Middleware = async (execute, context) => async (input: Adapter
     params: AdapterRequestData
     message: string
     feedID: string
-    endpoit?: string
+    endpoint?: string
     errorResponse?: any
     rawError?: any
+    stack: any
   }
   Logger.debug('Input: ', { input })
   try {
@@ -101,18 +102,16 @@ const withLogger: Middleware = async (execute, context) => async (input: Adapter
       jobRunID: input.id,
       params: input.data,
       feedID,
+      stack: error.stack,
+      endpoint: error.endpoint,
+      errorResponse: error.errorResponse,
     }
-    if (error.endpoint) {
-      errorLog.endpoit = error.endpoint
-    }
-    if (error.errorResponse) {
-      errorLog.errorResponse = error.errorResponse
-    }
+
     if (Logger.level === 'trace') {
       errorLog.rawError = error.rawError
     }
 
-    Logger.error({ ...errorLog, stack: error.stack })
+    Logger.error(errorLog)
     throw error
   }
 }
--- a/packages/core/bootstrap/src/lib/external-adapter/errors.ts
+++ b/packages/core/bootstrap/src/lib/external-adapter/errors.ts
@@ -18,9 +18,9 @@ export class AdapterError extends Error {
     name = 'AdapterError',
     message = 'An error occurred.',
     cause,
-    endpoint = '',
-    errorResponse = null,
-    feedID = '',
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: packages/core/bootstrap/src/index.ts [hunk 1]
```diff
@@ -79,13 +80,39 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
 
 // Log adapter input & output data
 const withLogger: Middleware = async (execute, context) => async (input: AdapterRequest) => {
+  interface IError {
+    jobRunID: string
+    params: AdapterRequestData
+    message: string
+    feedID: string
+    endpoit?: string
```
austinborn, line 88:
```suggestion
    endpoint?: string
```

packages/core/bootstrap/src/index.ts
```diff
@@ -79,13 +80,39 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
...
-    Logger.error(error.toString(), { stack: error.stack })
+    const feedID = metrics.util.getFeedId(input)
+    const errorLog: IError = {
+      message: error.toString(),
+      jobRunID: input.id,
+      params: input.data,
+      feedID,
+    }
+    if (error.endpoint) {
+      errorLog.endpoit = error.endpoint
```
austinborn, line 106:
```suggestion
      errorLog.endpoint = error.endpoint
```

packages/core/bootstrap/src/lib/external-adapter/errors.ts
```diff
@@ -24,11 +30,20 @@ export class AdapterError extends Error {
...
+    this.feedID = feedID
   }
 
   toJSONResponse(): AdapterErrorResponse {
     const showDebugInfo = process.env.NODE_ENV === 'development' || process.env.DEBUG === 'true'
-    const errorBasic = { name: this.name, message: this.message }
+    const errorBasic = {
+      name: this.name,
+      message: this.message,
+      endpoint: this.endpoint || undefined,
```
austinborn, line 43: It would be better if we can remove `|| undefined` and print exactly what `this.endpoint` is (same for the others).

packages/core/bootstrap/src/index.ts
```diff
@@ -79,13 +80,39 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
...
+    }
+    if (error.errorResponse) {
+      errorLog.errorResponse = error.errorResponse
+    }
```
austinborn, line 110: I think we should just log whatever the value is, even if it's falsy. Conditionally logging certain values provides a less complete picture when trying to debug with the logs.

packages/core/bootstrap/src/index.ts
```diff
@@ -79,13 +80,39 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
...
+    if (Logger.level === 'trace') {
+      errorLog.rawError = error.rawError
+    }
+
+    Logger.error({ ...errorLog, stack: error.stack })
```
austinborn, line 115: Can we just add `stack: error.stack` to the `const errorLog: IError = ` declaration? Then this line could be 
```
Logger.error(errorLog)

File: packages/core/bootstrap/src/lib/external-adapter/requester.ts [hunk 6]
```diff
@@ -155,9 +159,10 @@ export class Requester {
         statusCode,
         message: error.message,
         cause: error,
+        feedID: feedID,
```
austinborn, line 162:
```suggestion
        feedID,
```

austinborn (changes requested):

packages/core/bootstrap/src/index.ts
karen-stepanyan: In many cases either endpoint or errorResponse will be undefined. Not sure if we intentionally should include this in logs

austinborn (approved):
Thanks for the changes, lgtm

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -46,17 +46,18 @@ export class Requester {
       }
 
       let response: AxiosResponse<T>
+      const endpoint = `${config.baseURL}/${config.url}`
```
boxhock, line 49: Should we use `path.join()` instead to avoid weird paths if duplicate "/"s?

justinkaseman, line 49: The endpoint should come from the request parameter `endpoint`.
`.url` won't be set in most places.

packages/core/bootstrap/src/index.ts
```diff
@@ -79,13 +80,38 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
 
 // Log adapter input & output data
 const withLogger: Middleware = async (execute, context) => async (input: AdapterRequest) => {
+  interface IError {
```
justinkaseman, line 83: Could we move this to `@chainlink/types`?

packages/core/bootstrap/src/lib/external-adapter/requester.ts
justinkaseman: Though this might still be useful information. I would keep this, but rename it `url`.

packages/core/bootstrap/src/index.ts
```diff
@@ -79,13 +80,38 @@ const withStatusCode: Middleware = async (execute, context) => async (input) =>
...
     return result
   } catch (error) {
-    Logger.error(error.toString(), { stack: error.stack })
+    const feedID = metrics.util.getFeedId(input)
+    const errorLog: IError = {
+      message: error.toString(),
+      jobRunID: input.id,
+      params: input.data,
+      feedID,
+      stack: error.stack,
```
justinkaseman, line 105: Similar to log level trace below, stack is very verbose. 
We should only include it on `.level === 'debug'`.

And the `rawError` also includes stack trace right? So don't include `stack` when level is trace.

packages/core/bootstrap/src/index.ts
```diff
@@ -195,7 +221,8 @@ export const executeSync: ExecuteSync = async (
 
     return callback(result.statusCode, result)
   } catch (error) {
-    return callback(error.statusCode || 500, Requester.errored(data.id, error))
+    const feedID = metrics.util.getFeedId(data)
+    return callback(error.statusCode || 500, Requester.errored(data.id, error, undefined, feedID))
```
justinkaseman, line 225: The third argument into `Requester.errored` should be `error.statusCode`.

packages/core/bootstrap/src/lib/external-adapter/requester.ts
karen-stepanyan: We don't have access to endpoint parameter.

karen-stepanyan: resolved in #1058

karen-stepanyan: resolved in #1058

packages/core/bootstrap/src/index.ts
karen-stepanyan: resolved in #1058

packages/core/bootstrap/src/index.ts
karen-stepanyan: resolved in #1058

PR #1057, opened 2021-11-02
Wrapped.com

### LINKED ISSUE #1019: Wrapped.com EA
This is a new composite EA for wrapped.com that simply retrieves addresses associated with a token ticket symbol and fetches the PoR for those addresses . We are using the new ETH PoR EA recently built

the endpoint: http://chainlink.wrappedeng.com/deposits

Workflow: 
1. this EA will take 'token symbol' as input
2. this EA output will be the addresses fetched from the API for that 'token symbol'

lets start with cUSD: 
```"cUSD":["0xbF120a30a122FE06a2935dBb5C53f127aC270d2E","0x9A10359F42978c33620D84444EF1a8f2Fb8286FD"]}
"cUSD":["0xbF120a30a122FE06a2935dBb5C53f127aC270d2E","0x9A10359F42978c33620D84444EF1a8f2Fb8286FD"]}
```

EAs: 
- there needs to be a souce adapter for wrapped.com API

bogos:
## Description

Wrapped source and proof-of-reserve support 

## Changes

- Add Wrapped adapters

## Steps to Test
```
yarn test wrapped

```

## Quality Assurance

- [] This is related to a maximum of one Clubhouse/Shortcut story or GitHub issue


packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,46 @@
...
+export const execute: ExecuteWithConfig<Config> = async (request, _, config) => {
+  const validator = new Validator(request, inputParameters)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const symbol = validator.validated.data.symbol || 'ETH'
+  const url = `/deposits`
+
+  const options = { ...config.api, url }
+  const response = await Requester.request(options, customError)
```
justinkaseman, line 25: If you pass in the response structure as generic here the response will be typed, which can be helpful for logic that happens before the result is returned. 

```suggestion
  const response = await Requester.request<Addresses[]>(options, customError)
```

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,46 @@
+import { Requester, Validator, AdapterError } from '@chainlink/ea-bootstrap'
+import { Config, ExecuteWithConfig, InputParameters, AxiosResponse } from '@chainlink/types'
+
+export const supportedEndpoints = ['deposits']
+
+export type Addresses = {
+  address: string
+}
+
```
justinkaseman, line 9:
```suggestion
export interface ResponseSchema {
 [token: string]: Addresses[]
}
```

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const symbol = validator.validated.data.symbol || 'ETH'
+  const url = `/deposits`
+
+  const options = { ...config.api, url }
+  const response = await Requester.request(options, customError)
+  const addresses = response.data[symbol]
+  const keys = Object.keys(response.data).join()
```
justinkaseman, line 27: Could `resposne.data` ever not be an object?
We could protect against this using a `customError` function passed into `Requester.request`.

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,46 @@
...
+      message: `Input, at 'symbol' path, must be one of the following values: ${keys}`,
+      statusCode: 400,
+    })
+  }
+
+  const result = addresses.map((element: string) => {
+    return { address: element }
+  })
+
+  response.data.result = result as Addresses[]
```
justinkaseman, line 41: You won't need this because you are using `Requester.withResult`

packages/sources/wrapped/test/integration/fixtures.ts
```diff
@@ -0,0 +1,106 @@
+import nock from 'nock'
+
+export const mockDataProviderResponses = (): nock =>
```
justinkaseman, line 3: Nice!

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!addresses) {
+    throw new AdapterError({
+      jobRunID,
+      message: `Input, at 'symbol' path, must be one of the following values: ${keys}`,
+      statusCode: 400,
+    })
+  }
+
+  const result = addresses.map((element: string) => {
+    return { address: element }
```
justinkaseman, line 38: What blockchain and network is this planned to be used with?

The default that Proof of Reserves uses is `btc` `mainnet`. If this differs we will need to specify a `coin` and `network`.

bogos: for now we do not have support with all networks, we are using the `wrapped` with eth by default in conjunction with the` eth_balance` adapter to get the balances

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw new AdapterError({
+      jobRunID,
+      message: `Input, at 'symbol' path, must be one of the following values: ${keys}`,
+      statusCode: 400,
+    })
+  }
+
+  const result = addresses.map((value: Addresses) => {
+    return { address: value }
+  })
```
austinborn, line 45:
```suggestion
  const result = addresses.map((address: Addresses) => ({ address }))
```

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,51 @@
...
+  const jobRunID = validator.validated.id
+  const symbol = validator.validated.data.symbol || 'ETH'
+  const url = `/deposits`
+
+  const options = { ...config.api, url }
+  const response = await Requester.request<ResponseSchema>(options, customError)
+  const addresses = response.data[symbol]
+  const keys = Object.keys(response.data).join()
+
+  if (!addresses) {
```
austinborn, line 35: Unless `keys` is used outside of the `if` statement, I'd just put it inside
```suggestion

  if (!addresses) {
    const keys = Object.keys(response.data).join()
```

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const supportedEndpoints = ['deposits']
+
+export interface ResponseSchema {
+  [token: string]: Addresses[]
+}
+
+export type Addresses = {
+  address: string
+}
```
austinborn, line 12: Could we just call this `Address` or `AddressObject`? It's a little confusing that you have the type `Addresses` which has a single address, then multiple places with the type `Addresses[]`.

austinborn (approved):

packages/core/legos/package.json
```diff
@@ -140,6 +140,7 @@
     "@chainlink/view-function-adapter": "*",
     "@chainlink/wbtc-address-set-adapter": "*",
     "@chainlink/wootrade-adapter": "*",
+    "@chainlink/wrapped-adapter": "*",
```
boxhock, line 143:
```suggestion
    "@chainlink/wrapped-adapter": "1.0.0",
```

packages/sources/wrapped/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@chainlink/wrapped-adapter",
+  "version": "0.0.1",
```
boxhock, line 3:
```suggestion
  "version": "1.0.0",
```

boxhock (changes requested):
Needs a rebase, after that the following changes will be relevant. Otherwise LGTM!

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,52 @@
...
+export const inputParameters: InputParameters = {
+  symbol: false,
+}
+
+export const execute: ExecuteWithConfig<Config> = async (request, _, config) => {
+  const validator = new Validator(request, inputParameters)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const symbol = validator.validated.data.symbol || 'ETH'
```
boxhock, line 27: Unless explicitly required otherwise, I'd like to require the symbol to be defined

```suggestion
  symbol: true,
}

export const execute: ExecuteWithConfig<Config> = async (request, _, config) => {
  const validator = new Validator(request, inputParameters)
  if (validator.error) throw validator.error

  const jobRunID = validator.validated.id
  const symbol = validator.validated.data.symbol
```

packages/sources/wrapped/test/unit/adapter.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  const execute = makeExecute()
+
+  describe('validation error', () => {
+    const requests = [
+      { name: 'empty body', testData: {} },
+      { name: 'empty data', testData: { data: {} } },
+      {
+        name: 'symbol not supplied',
+        testData: { id: jobID, data: { symbol: '' } },
+      },
```
boxhock, line 17: Something to watch out for: before my comment above, this adapter would not require any input params, making this test incorrect (as we experienced with uscpi-one :wink:  )

boxhock: If we add the required field above, this test will be fine.

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,52 @@
...
+      jobRunID,
+      message: `Input, at 'symbol' path, must be one of the following values: ${keys}`,
+      statusCode: 400,
+    })
+  }
+
+  const result = {
+    coin: symbol,
+    addresses: addresses.map((address: Address) => ({ address })),
+  }
```
boxhock, line 46: It doesn't look like this matches what the README shows.

To address part of @justinkaseman's request, I think something like this would be better suited:

```suggestion
  const result = addresses.map((address: Address) => ({ address, coin: symbol }))
```

boxhock (changes requested):

packages/sources/wrapped/src/endpoint/deposits.ts
bogos: I think that since it is filtered by symbol, there is no need to specify the symbol in each address. Putting it as a separate attribute in the answer works best.

```  
"result": {
        "coin": "ETH",
        "addresses": [
            {
                "address": "0x6E4C739B4d66E7104739Fc0F235E9c9FFD1F8da4"
            },
            {
                "address": "0x67a53ADbA557E6129D3Eb1aab090E23D22711DB2"
            },
            {
                "address": "0x57Fc3f242aD907E9518645Ba4D0802F7a65B3677"
            }
        ]
    },
```
instead of returning: 
```  
"result": {
        "addresses": [
            {
                "address": "0x6E4C739B4d66E7104739Fc0F235E9c9FFD1F8da4",
                "coin: "ETH"
            },
            {
                "address": "0x67a53ADbA557E6129D3Eb1aab090E23D22711DB2",
                "coin: "ETH"
            },
            {
                "address": "0x57Fc3f242aD907E9518645Ba4D0802F7a65B3677",
                "coin: "ETH"
            }
        ]
    },
```

boxhock: I had a chat with Justin and Dimitri about this yesterday. We should make the adapter return an array of this interface:
```
{
  address: string // e.g. "0x123abc..."
  network: string // e.g. "ethereum"
  chainId: string | number // e.g "1" - or in cases of no chainId: "mainnet"
}
```

So in the case of your example above, we should expect to get this back:

```
"result": [
            {
                "address": "0x6E4C739B4d66E7104739Fc0F235E9c9FFD1F8da4",
                "network": "ethereum",
                "chainId": 1
            },
            {
                "address": "0x67a53ADbA557E6129D3Eb1aab090E23D22711DB2",
                "network": "ethereum",
                "chainId": 1
            },
            {
                "address": "0x57Fc3f242aD907E9518645Ba4D0802F7a65B3677",
                "network": "ethereum",
                "chainId": 1
            }
        ],
```

Since the Wrapped.com data source doesn't provide enough information itself to determine the network and chainId, the adapter should take `network` and `chainId` as required input params.

packages/sources/wrapped/README.md
```diff
@@ -0,0 +1,69 @@
...
+        {
+          "address": "0x67a53ADbA557E6129D3Eb1aab090E23D22711DB2"
+        },
+        {
+          "address": "0x57Fc3f242aD907E9518645Ba4D0802F7a65B3677"
+        }
+      ]
+    }
+  }
+}
```
boxhock, line 68: Needs updating to new interface

packages/sources/wrapped/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+---
+
+## Deposits Endpoint
+
+### Input Params
+
+| Required? |   Name   |             Description             |       Options       | Defaults to |
+| :-------: | :------: | :---------------------------------: | :-----------------: | :---------: |
+|    ✅     | `symbol` | The symbol of the currency to query | `BTC`, `ETH`, `LTC` |             |
```
boxhock, line 19: We probably also want to include `network` and `chainId` in the request params

boxhock (changes requested):
We've also introduced changesets since this PR was started, so that will have to be included as well!

boxhock (changes requested):
Reminder that we'll need changeset as well before merging

austinborn (approved):
LGTM!

packages/sources/wrapped/src/endpoint/deposits.ts
```diff
@@ -0,0 +1,82 @@
...
+  const network = validator.validated.data.network || networks[symbol]
+  const url = `/deposits`
+
+  const options = { ...config.api, url }
+  const response = await Requester.request<ResponseSchema>(options, customError)
+  const addresses = response.data[symbol]
+
+  if (!addresses) {
+    const keys = Object.keys(response.data).join()
+    throw new AdapterError({
```
cds95, line 65: we should add an integration test for the case when the request contains an invalid symbol.

bryanjowers:
please prioritize to get merged.

packages/sources/wrapped/src/endpoint/deposits.ts
bogos: Since we nock the response from this endpoint http://chainlink.wrappedeng.com/deposits, which is a list of symbols with their respective addresses. What we do is only filter the symbol we need, I don't think that could be necessary to nock the error because the query is always made to the same endpoint without any extra parameter.

cds95 (approved):

boxhock (approved):

PR #1058, opened 2021-11-02
fix type issues, add error url

karen-stepanyan:
## Changes

fix types, rename 'endpoint' to 'url', update integration test snapshots

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/lib/external-adapter/errors.ts
+++ b/packages/core/bootstrap/src/lib/external-adapter/errors.ts
@@ -7,9 +7,9 @@ export class AdapterError extends Error {
   name: string
   message: string
   cause: any
-  url: string
+  url?: string
   errorResponse: any
-  feedID: string
+  feedID?: string
 
   constructor({
     jobRunID = '1',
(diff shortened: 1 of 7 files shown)


packages/core/bootstrap/src/lib/external-adapter/errors.ts
```diff
@@ -30,7 +30,7 @@ export class AdapterError extends Error {
     this.name = name
     this.message = message
     this.cause = cause
-    this.endpoint = endpoint!
+    this.url = url!
```
justinkaseman, line 33: I don't think we should use non-null assertions like this.
Since `url` isn't passed in on every `AdapterError` it is actually functioning as optional.
We should either make it so in the static property types, or ensure it is actually non-null everywhere (which may be a lot of work).

packages/core/types/@chainlink/index.d.ts
```diff
@@ -83,6 +83,17 @@ declare module '@chainlink/types' {
     cause: string
   }
 
+  export type AdapterErrorFull = {
```
justinkaseman, line 86:
```suggestion
  export type AdapterErrorLog = {
```
Since we only use this for logging purposes could we name it as such?

PR #1065, opened 2021-11-03
Expand Sportsdataio to fetch MLB schedule and score data

cds95:
This PR expands the SportsDataIO adapter to fetch MLB data.  It does this by taking in a generic request payload that can be easily expanded to other sports.


packages/sources/sportsdataio/src/sport/mlb/endpoint/schedule.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const games = await getGamesByDate(date, config)
+  const encodedGames = encodeGames(games)
+  const respData = {
+    data: {
+      encodedGames,
+      games,
+      result: encodedGames.length,
+    },
+    result: encodedGames.length,
```
boxhock, line 29: Why are we setting `result` to `encodedGames.length`?
Would it not be better to have a different key like `count` to make this more explicit?

packages/sources/sportsdataio/test/integration/mlb/mlb.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        .set('Accept', '*/*')
+        .set('Content-Type', 'application/json')
+        .expect('Content-Type', /json/)
+        .expect(400)
+      expect(response.body).toMatchSnapshot()
+    })
+  })
+
+  describe('fetch mlb game by game id', () => {
+    mockGamesResponse(MOCK_KEY)
```
boxhock, line 90: Nice, I like this way of handling the API key

boxhock (commented):
Looks good, just had a comment

packages/sources/sportsdataio/src/sport/mlb/endpoint/schedule.ts
cds95: We could.  I didn't know what else to set `result` to.  Any suggestions?

boxhock: Ah, in that case I'd prefer just `games` or `encodedGames`

austinborn (approved):
Good pending @boxhock 's suggestion

packages/sources/sportsdataio/src/sport/mlb/endpoint/schedule.ts
cds95: Can `result` be of any type or must it be a primitive value?

boxhock: It should be possible for it to be of any type, it just limits what the CL node can do with it without transforming it in any way. Considering this will be used by TOML job specs, I assume, then that shouldn't be a problem.

cds95: alright I'll make this change

boxhock (approved):

PR #1066, opened 2021-11-04
fix amberdata import in por adapter

cds95:
Bug

A recent change in the way the Amberdata adapter exported variables broke the PoR adapter.  This was because the imported value in `PoR` would be undefined hence causing an exception.  We should be careful in the future when we update how variables are exported from an adapter.


austinborn (approved):
Thanks for catching that. The same import change will likely need to happen when adding integration tests for the other EAs as well.

cds95:
yup that's right

PR #1070, opened 2021-11-04
reduce sportsdataio response when fetching mlb data

cds95:
Reduce the size of the response when fetching MLB data from SportsDataIO by only returning a subset of fields for each game.


packages/sources/sportsdataio/src/sport/mlb/endpoint/schedule.ts
```diff
@@ -22,7 +22,6 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   const encodedGames = encodeGames(games)
   const respData = {
     data: {
-      encodedGames,
```
cds95, line 25: Removed as it was redundant.

austinborn (approved):

PR #1094, opened 2021-11-09
refactor: stop umbrella release tags pipelines

justinkaseman:


.github/workflows/release.yml
```diff
@@ -73,16 +74,20 @@ jobs:
           aws-region: us-east-1
           aws-ecr-private: 'false'
       # If we're on master, publish again with a forced 'latest' tag
```
HenryNguyen5, line 75: Delete this line

.github/workflows/release.yml
```diff
@@ -54,7 +54,8 @@ jobs:
       - uses: actions/checkout@v2
 
       # Handle publishing to our public ECR repo
-      - name: Configure AWS Credentials for SDLC Public ECR
+      - if: github.ref == 'refs/heads/master'
```
HenryNguyen5, line 57: This check should be moved to the following step, since we want to limit publishing to just master

HenryNguyen5 (changes requested):

.github/workflows/release.yml
justinkaseman: Oops, yeah I had it on the wrong step :face_with_head_bandage:

.github/workflows/release.yml
```diff
@@ -62,7 +62,8 @@ jobs:
           aws-region: ${{ secrets.AWS_REGION }}
           role-to-assume: ${{ secrets.AWS_PUBLICECR_ROLE_ARN }}
           role-duration-seconds: 1200
-      - name: Publish adapter to public ECR
+      - if: github.ref == 'refs/heads/master'
+        name: Publish adapter to public ECR
         uses: ./.github/actions/release/publish-artifacts
         with:
           branch: ${{ fromJSON('[undefined, "develop"]')[github.ref == 'refs/heads/develop'] }}
```
HenryNguyen5, line 69: If we're going to permanently remove `develop-latest`, we could simplify the branch + latest logic so that it's
```
branch: undefined
latest: undefined
```

justinkaseman: Like so?

HenryNguyen5 (approved):

PR #1100, opened 2021-11-10
Durin adapter

cds95:
This PR implements the MVP for the Durin adapter.  It currently only supports a single endpoint namely `optimism-gateway`, which has the same functionality as the example gateway server implemented here https://github.com/smartcontractkit/ccip-read/tree/optimism-experimental/examples/optimism-gateway.


packages/sources/durin/README.md
```diff
@@ -0,0 +1,230 @@
...
+
+The optimism global endpoint reads the latest proof from an Optimism as the L2 chain and returns the proof to the caller.
+Currently this endpoint has the same functionality as the server in this example https://github.com/smartcontractkit/ccip-read/tree/6d4deb917781f3becda39b9ebad6f21e037af1a6/examples/optimism-gateway.
+
+### Input Params
+
+| Required? |  Name  |                                              Description                                               | Options | Defaults to |
+| :-------: | :----: | :----------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|    ✅     | `abi`  | The ABI of the originally called L1 contract. In this example it is the OptimismResolverStub contract. |         |             |
+|    ✅     |  `to`  |                        The **L1** address of the originall called L1 contract.                         |         |
```
boxhock, line 33:
```suggestion
|    ✅     |  `to`  |                        The **L1** address of the original called L1 contract.                         |         |
```

packages/sources/durin/schemas/env.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/durin.json",
+  "title": "@chainlink/durin env var schema",
+  "required": [],
```
boxhock, line 4:
```suggestion
  "required": ["RPC_URL"],
```

packages/sources/durin/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@chainlink/durin-adapter",
+  "version": "0.0.1",
```
boxhock, line 3: We probably ought to update the example adapter, but I think we should start adapters on 1.0.0. @justinkaseman does this sound right?

```suggestion
  "version": "1.0.0",
```

packages/sources/durin/src/config.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface Config extends BaseConfig {
+  l2RpcUrl?: string
+  addressManagerContract?: string
+}
+
+export const makeConfig = (prefix?: string): Config => {
+  const config = Requester.getDefaultConfig(prefix)
+  config.defaultEndpoint = DEFAULT_ENDPOINT
+  config.rpcUrl = util.getRequiredEnv('RPC_URL')
```
boxhock, line 16: Looks like this is unused

```suggestion
```

boxhock (changes requested):
Do we have any integration tests for this adapter?
Reached out on Slack re: new release process

packages/sources/durin/package.json
austinborn: My vote would be 1.0.0 as well

packages/sources/durin/schemas/env.json
```diff
@@ -0,0 +1,25 @@
...
+  "type": "object",
+  "properties": {
+    "RPC_URL": {
+      "required": true,
```
austinborn, line 8: Are these required fields used anywhere else? Long-term I would like to see a single source of truth, either these `required` attributes or the `required` array, but not both

packages/sources/durin/src/endpoint/optimism-gateway/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  const hash = (el: Buffer | string): Buffer => {
+    return Buffer.from(ethers.utils.keccak256(el).slice(2), 'hex')
+  }
+  const leaves = elements.map((element) => {
+    return Buffer.from(element.slice(2), 'hex')
+  })
+  const tree = new MerkleTree(leaves, hash)
+  return tree.getProof(leaves[index], index).map((element) => {
+    return element.data
+  })
```
austinborn, line 171:
```suggestion
  return tree.getProof(leaves[index], index).map(({ data }) => data)
```

packages/sources/durin/src/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { Fragment, Interface, JsonFragment } from '@ethersproject/abi'
+
+export const toInterface = (
+  abi: string | readonly (string | Fragment | JsonFragment)[] | Interface,
+) => {
+  if (Interface.isInterface(abi)) {
+    return abi
+  }
+  return new Interface(abi)
```
austinborn, line 9:
```suggestion
  return Interface.isInterface(abi) ? abi : new Interface(abi)
```

packages/sources/durin/src/endpoint/optimism-gateway/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        stateRoots,
+      }
+    }
+  }
+  throw Error('No state root batches found')
+}
+
+const getElementsToConstructProof = (stateBatchHeader: StateBatchHeader): string[] => {
+  const elements = []
+  for (let i = 0; i < Math.pow(2, Math.ceil(Math.log2(stateBatchHeader.stateRoots.length))); i++) {
```
austinborn, line 151:
```suggestion
  const maxElements = Math.pow(2, Math.ceil(Math.log2(stateBatchHeader.stateRoots.length)))
  for (let i = 0; i < maxElements; i++) {
```

austinborn (commented):
Looks great, but I would like to see integration tests as well

cds95:
@boxhock @austinborn I spoke to @thodges-gh and we are going to hold off on adding integration tests for this for now as it's still a MVP.  Currently the only test I've added runs makes a call to the original optimism-gateway server and compares it's output with the adapter's output.  I've created a ticket to add integration tests in the future.

https://app.shortcut.com/chainlinklabs/story/21369/add-integration-tests-to-durin-adapter

packages/sources/durin/schemas/env.json
cds95: I'm actually not sure about this.  @justinkaseman do you know the answer?

packages/sources/durin/test/unit/adapter.test.ts
```diff
@@ -0,0 +1,398 @@
...
+import { assertError } from '@chainlink/ea-test-helpers'
+import { AdapterRequest } from '@chainlink/types'
+import { makeExecute } from '../../src/adapter'
+
+describe('execute', () => {
+  const jobID = '1'
+  const execute = makeExecute()
+  process.env.RPC_URL = 'http://localhost:9545'
+
+  describe('validation error', () => {
```
tateexon, line 11: Looks like we only have tests for validation errors. Can we add a test that would pass validation and return a 200?

cds95: I don't think any of our adapters have unit tests that pass validation and return a 200 right now.  These are typically done in the integration tests which we are holding off for now.

austinborn (approved):

boxhock (approved):

PR #1129, opened 2021-11-17
Por adapters refactor

### LINKED ISSUE #1118: Gemini/Lotus adapters should use the common address list format
Currently, the Gemini and Lotus source adapters both expect list of addresses for PoR to be an array of address-strings:

```
["f2eaaj6w4evrdscw4s4o5c3df7ph725tbs3yvg6gi", "f225ey7bq53ur6sgrkxgf74hl2ftxkajupatwnmay"]
```

PoR adapters should use a slightly different format with an array of objects with the "address" key:

```
[
    {
      "address": "0x6E4C739B4d66E7104739Fc0F235E9c9FFD1F8da4"
    },
    {
      "address": "0x67a53ADbA557E6129D3Eb1aab090E23D22711DB2"
    },
    {
      "address": "0x57Fc3f242aD907E9518645Ba4D0802F7a65B3677"
    }
  ]
```

karen-stepanyan:
Closes #1118 

## Changes

- Changed the response shape for Gemini EA so that now it returns an array of objects with 'address, chainId, network' keys, whereas before it returned just an array of strings
- Changed the request  for Gemini EA so that now it optionally accepts 'network' and 'chainId'
- Changed the response shape for RenVM EA so that now it returns an array of objects with 'address, chainId, network, coin' keys
- Changed the response shape for wBTC EA so that now the response also contains 'network' and 'chainId'
- Changed the response shape for Chain-reserve-wallet EA so that now it returns an array of objects with 'address, chainId, network' keys, whereas before it returned just an array of strings
- Changed the request  for Chain-reserve-wallet EA so that now it expects 'network' 
- Changed the request  for Chain-reserve-wallet EA so that now expected 'chainId' is a string, not a number
- Changed the expected request shape for Lotus EA, so that now it expects an array of objects for 'addresses' key, rather array of strings
- Changed the expected request shape for Ada-balance EA, so that now it expects an array of objects for 'addresses' key, rather array of strings
- Changed integration tests to match with new changes

## Quality Assurance


bryanjowers:
comment from @justinkaseman 
```
It's important to include network and coin in the object structure too for balance adapters.
Otherwise it defaults to BTC mainnet.

We can sync on this for more context.
```

boxhock:
> comment from @justinkaseman
> 
> ```
> It's important to include network and coin in the object structure too for balance adapters.
> Otherwise it defaults to BTC mainnet.
> 
> We can sync on this for more context.
> ```

@justinkaseman let's sync tomorrow if you think this is an issue. For a number of address set adapters it might be impossible to get the network used - so this might be something we should add to job spec / PoR composite instead.

boxhock:
High level looks good, needs a changeset

packages/sources/chain-reserve-wallet/src/endpoint/wallet.ts
```diff
@@ -2,6 +2,19 @@ import { Requester, Validator } from '@chainlink/ea-bootstrap'
...
+const chainIds = ['mainnet', 'testnet']
+
+const networkChainMap: { [key: string]: number } = {
+  cardano_testnet: 0,
+  doge_testnet: 1,
+  cardano_mainnet: 2,
+  doge_mainnet: 3,
+}
+
+const isNetwork = (maybeNetwork: any) => networks.indexOf(maybeNetwork) !== -1
```
cds95, line 15: should this be of type `string`?

packages/sources/gemini/src/endpoint/reserves.ts
```diff
@@ -11,6 +11,8 @@ export interface ResponseSchema {
 
 export const inputParameters: InputParameters = {
   token: ['token', 'asset', 'coin'],
+  network: false,
```
cds95, line 14: do we have tests to make sure that the adapter works when these fields are missing from the request params?

cds95 (changes requested):
Looks good overall.  Just left a few comments.

packages/sources/chain-reserve-wallet/src/endpoint/wallet.ts
karen-stepanyan: it should

packages/sources/gemini/src/endpoint/reserves.ts
karen-stepanyan: the integration test example works without 'network' and 'chainId'

cds95 (dismissed):

boxhock:
@cds95 regarding your comment, are you coordinating with the engops team to update request params?

kevincharm:
Tested this branch with a `bitcoin-json-rpc` EA connected to a Bitcoin node. Looks good 👍 

WBTC
```sh
curl \
		-d '{"id":"2","data":{"protocol": "wbtc", "indexer": "bitcoin_json_rpc"}}' \
		http://localhost:8003
{"jobRunID":"1","result":"25670225851726","statusCode":200,"data":{"result":"25670225851726"}}
```

RenBTC
```sh
curl \
		-d '{"id":"2","data":{"protocol": "renvm", "indexer": "bitcoin_json_rpc", "network": "mainnet"}}' \
		http://localhost:8003
{"jobRunID":"1","result":"2073570708716","statusCode":200,"data":{"result":"2073570708716"}}
```

boxhock (changes requested):
Have not A/T'ed Filecoin yet, but just noticed this

packages/sources/lotus/test/e2e/balance.test.ts
```diff
@@ -11,20 +11,25 @@ describe('execute', () => {
...
       },
       {
         name: 'multiple addresses supplied',
         testData: {
           id: jobID,
           data: {
             addresses: [
-              'f2eaaj6w4evrdscw4s4o5c3df7ph725tbs3yvg6gi',
-              'f225ey7bq53ur6sgrkxgf74hl2ftxkajupatwnmay',
+              { address: 'f2eaaj6w4evrdscw4s4o5c3df7ph725tbs3yvg6gi' },
```
cds95, line 31: Similar to how the change to the `chain-reserve-wallet` response affects the `ada-balance` adapter, will changing this response also affect the downstream adapter in the PoR flow?  I'm looking at `balance.ts` in the PoR adapter directory and it looks like we just pass in the response from this adapter to when constructing the request payload to the second adapter if we are not pulling wBTC balances.  Do we have to do a similar object mapping like we did to the `ada-balance` adapter?

```
{
          id: input.jobRunID,
          data: {
            result: input.data.result,
            dataPath: 'result',
            endpoint: 'balance',
            confirmations,
          },
        }
 ```

File: packages/sources/gemini/README.md [hunk 4]
```diff
@@ -41,20 +43,30 @@ This adapter takes no environment variables.
...
-    "addresses": [
-      "f2eaaj6w4evrdscw4s4o5c3df7ph725tbs3yvg6gi",
-      "f225ey7bq53ur6sgrkxgf74hl2ftxkajupatwnmay"
-    ],
-    "ethereum_supply": 33427.594125,
-    "currency": "EFIL",
     "result": [
-      "f2eaaj6w4evrdscw4s4o5c3df7ph725tbs3yvg6gi",
-      "f225ey7bq53ur6sgrkxgf74hl2ftxkajupatwnmay"
+      {
```
cds95, line 60: Same situation here as Lotus.

packages/sources/ada-balance/src/endpoint/balance.ts
```diff
@@ -23,11 +23,17 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
+    throw new AdapterError({
+      jobRunID,
+      message: `Input, at 'addresses' or 'result' path, must be a non-empty array.`,
+      statusCode: 400,
+    })
   }
+
   const result = await getAddressBalances(
-    addresses,
+    addresses.map((address) => address.address),
```
cds95, line 36: Actually looking at this again it might be better to just leave this as it is and map the address objects in the PoR adapter to addresses in the `balance.ts` file in PoR.  That way we don't have to worry about breaking downstream adapters.

boxhock (approved):
Verified that the updated adapters work with Filecoin PoR as-is without any param changes. 

@cds95 will you make sure it's clear to engops that all adapters in a PoR feed will have to be updated for things to work? In other words, we need to make sure node ops don't update the proof-of-reserves composite but not the Gemini adapter, for instance.

PR #1143, opened 2021-11-18
Feature/changesets

HenryNguyen5:
The CD pipeline has been refactored to play along with the strengths of
the recent atlassian/changesets integration. The previous behaviour of
the pipeline in summary is as follows:
- Merging into develop
  - a new "develop-latest" tagged image for every single adapter would
  be pushed for both the private and public ECR repos, causing image
  spam since not all changes required a new image.
- Merging into master
  - a new "latest" tagged image for every single adapter would be pushed
  for both the private and public ECR repos, again, causing image spam
  since no diffing logic was performed.
  - a versioned release based on the current version in each EA was
  pushed to public and private ECR, but no checks were made to see if we
  were overriding an existing version.

The lack of version diff checking caused a lot of redunant image spam in
the ECR repos, and consumers of the EA's would not know if a newly
published image included relevant changes, was a noop from the
previous pushed image, or was an unstable internal release not
intended for public consumption.

The new pipeline behaves as follows:
- Merging into develop
  - Only the private ECR repository is taken into account. Each
  adapter's package.json version is checked against what's already in
  ECR. We only push a new image if a different version is found within
  ECR.
- Merging into master
  - Both the public and private ECR repos are versioned checked in the
  same fashion as described in the "merging into develop" point to see
  if a new image should be pushed.
  - Both the public and private ECR repos have a latest version pushed
  only if the previous step was successful in pushing an image.

The new logic in publishing enforces that only new versions are
published. When combined with atlassian/changesets, this also means that
new versions must include a relevant changelog. This reduces the image
spam we get, and allows for external consumers to know what each version
brings them in terms of improvements / bugfixes, while also only exposes
images that we deem stable from staging testing to the general public.


justinkaseman:
Only change is we should have `:latest` off for now.

Otherwise great stuff.

justinkaseman (approved):

PR #1181, opened 2021-11-24
Add new inputParameter config system to Validator

austinborn:
## Description

This change refactors the input parameter configuration methods for the Validator, while maintaining backwards compatibility with old system.

## Changes
- Add new inputConfig setup with the following type:
```export type InputParameter = {
    aliases?: InputParameterAliases
    description?: string
    type?: 'bigint' | 'boolean' | 'list' | 'number' | 'object' | 'string'
    required?: RequiredInputParameter
    options?: (number | string)[]
    default?: boolean | number | string
    dependsOn?: string[]
    exclusive?: string[]
  }
```
- Refactor some parts of the Validator code to be more efficient or readable

## Steps to Test

TBD (TODO)

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/lib/external-adapter/validator.ts
+++ b/packages/core/bootstrap/src/lib/external-adapter/validator.ts
@@ -5,7 +5,10 @@ import {
   APIEndpoint,
   Includes,
   IncludePair,
+  InputParameter,
+  InputParameters,
   Config,
+  InputParameterAliases,
 } from '@chainlink/types'
 import { merge } from 'lodash'
 import { isArray, isObject } from '../util'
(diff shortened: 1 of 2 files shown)


austinborn:
Note: need to test new Validator to confirm backwards compatibility and proper functioning. Ideally add unit tests to cover new verification.

boxhock:
Just want to note that we'll need a changeset for this PR

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -18,121 +21,84 @@ import { Requester } from './requester'
...
+  }
   error: AdapterError | undefined
   errored: AdapterErrorResponse | undefined
+  shouldLogError: boolean
 
-  constructor(input = {}, customParams = {}, options = {}, shouldLogError = true) {
+  constructor(input = {}, inputConfigs = {}, options = {}, shouldLogError = true) {
     this.input = { ...input }
-    this.customParams = { ...customParams }
+    this.inputConfigs = { ...inputConfigs }
```
cds95, line 45: Can this just be `this.inputConfigs = inputConfigs`?

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -18,121 +21,84 @@ import { Requester } from './requester'
...
-          customParams: this.customParams,
-        },
-        shouldLogError,
-      )
+      this.parseError(e)
     }
   }
 
-  parseError(error: any, context: any, shouldLogError: boolean) {
+  parseError(error: any): void {
```
cds95, line 101: Does `error` not have a type?

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -265,46 +208,82 @@ export class Validator {
...
+  validateObjectParam(key: string): void {
+    const inputConfig = this.inputConfigs[key] as InputParameter
+
+    const usedKey = this.getUsedKey(key, inputConfig.aliases ?? [])
+
+    const param = usedKey
+      ? this.input.data[usedKey as string] ?? inputConfig.default
+      : inputConfig.default
+
+    if (inputConfig.required && (param === undefined || param === null))
```
cds95, line 224: Validate empty strings here as well

cds95 (changes requested):
Looks good overall!  Just some minor comments.

packages/core/bootstrap/src/lib/external-adapter/validator.ts
austinborn: Looks like I should be using the `Error` type

packages/core/bootstrap/src/lib/external-adapter/validator.ts
austinborn: I'm guessing this is a catch-all to avoid changing the original objects passed into the validator (same with the other inputs). It doesn't look like `inputConfigs` is ever modified, so it isn't necessary, although it might still be good to keep as a safeguard for future changes. Thoughts?

packages/core/types/@chainlink/index.d.ts
```diff
@@ -146,8 +146,18 @@ declare module '@chainlink/types' {
 
   export type RequiredInputParameter = boolean
   export type InputParameterAliases = string[]
+  export type InputParameter = {
+    aliases?: InputParameterAliases
+    description?: string
+    type?: 'bigint' | 'boolean' | 'array' | 'number' | 'object' | 'string'
+    required?: RequiredInputParameter
```
boxhock, line 153: Why are we making this a separate type instead of just calling it a `boolean`?

austinborn: I was originally trying to make use of the existing types, but tbh I think it's cleaner to just remove those. I'll add an update for it now

packages/core/types/@chainlink/index.d.ts
```diff
@@ -146,8 +146,18 @@ declare module '@chainlink/types' {
...
+    options?: any[]
+    default?: any
+    dependsOn?: string[]
+    exclusive?: string[]
```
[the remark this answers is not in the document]
austinborn: `exclusive` is the opposite of `dependsOn`. It's a list of other parameters that cannot be present when this input param is. I'll add some comments here to clarify

packages/core/bootstrap/src/lib/external-adapter/validator.ts
cds95: I see. Hmm in that case then I think it’s fine to keep it as it is.

cds95 (approved):

boxhock (commented):
Did we have any documentation for this previously that needs updating?

Otherwise LGTM, really like the improvement

austinborn:
> Did we have any documentation for this previously that needs updating?
> 
> Otherwise LGTM, really like the improvement

I didn't see any documentation in the README or main markdown docs, but I did add comments by the type file itself, and the example source adapter's `inputParameters` object. If you'd like, I can add more details in the root README or the example adapter README.

boxhock:
@austinborn ah I think I just found a file: packages/core/bootstrap/src/lib/external-adapter/README.md

cds95 (approved):
LGTM

PR #1247, opened 2021-12-01
Historical average adapter

boxhock:
## Changes

- Added historical average adapter
- Modified CMC's historical endpoint
- Exported types from CMC adapter

## Steps to Test

See integration/e2e tests and examples from historical average adapter

1. Run CMC adapter
2. Run historical-average adapter
3. Send test requests from README or run e2e tests

## Quality Assurance


packages/composites/historical-average/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@chainlink/historical-average-adapter",
+  "version": "0.0.0",
```
austinborn, line 3: Should we start with 1.0.0?
```suggestion
  "version": "1.0.0",
```

packages/composites/historical-average/src/adapter.ts
```diff
@@ -0,0 +1,128 @@
...
+export const getFromToDates = (
+  from?: string,
+  to?: string,
+  days?: string | number,
+): { fromDate: Date; toDate: Date } => {
+  if (from && to) {
+    return {
+      fromDate: new Date(from),
+      toDate: new Date(to),
+    }
```
austinborn, line 100: Can you add validation that `from` is before `to`?

packages/composites/historical-average/src/adapter.ts
```diff
@@ -0,0 +1,128 @@
...
+  } else if (!days) {
```
austinborn, line 101: Is it worth considering the edge case where `days = 0`?

packages/composites/historical-average/src/dataProvider.ts
```diff
@@ -0,0 +1,61 @@
...
+    quote: string,
+    fromDate: Date,
+    toDate: Date,
+    interval: string,
+  ): Promise<ResponsePayload> => {
+    try {
+      switch (source.toUpperCase()) {
+        case cmc.NAME:
+          return getCoinMarketCapPrice(jobRunID, base, quote, fromDate, toDate, interval, apiConfig)
+      }
```
austinborn, line 23: Can you add a `default` here?

austinborn (changes requested):

packages/composites/historical-average/package.json
boxhock: The changeset should bump this to 1.0.0

packages/composites/historical-average/src/dataProvider.ts
boxhock: I purposefully didn't put it here because this is happening inside a try catch, so throwing an error in default here would just make it get caught and wrapped in a misleading error.

Do you think it's worth refactoring to support `default` here?

packages/composites/historical-average/package.json
austinborn: Got it

packages/composites/historical-average/src/dataProvider.ts
austinborn: Oh, I see the `throw new Error` on line 31. Yeah it's a little funny but I think it's fine. IMO not worth refactoring 👍

austinborn (approved):
Awesome, thanks!

PR #1253, opened 2021-12-02
Hotfix/catch errors

boxhock:
## Changes

- Catch errors caught from `execute()`s in adapters that cache responses internally

## Steps to Test

1. Run coingecko/coinpaprika without cache
2. Send many requests until it 429s
3. Now it will return an error and close the socket, where it would keep it open previously

## Quality Assurance

- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `<ADAPTER_PACKAGE>/schemas/env.json` and `<ADAPTER_PACKAGE>/README.md`
- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.
- [x] The branch naming follows git flow (`feature/x`, `chore/x`, `release/x`, `hotfix/x`, `fix/x`) or is created from Clubhouse/Shortcut
- [x] This is related to a maximum of one Clubhouse/Shortcut story or GitHub issue
- [x] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)

PR #1254, opened 2021-12-02
Hotfix/catch errors

boxhock:
Master version of #1253


boxhock:
Need to re-do this PR.

PR #1262, opened 2021-12-03
BEth price feed

cds95:
## Description

Implements the adapter to support the BEth price feed

## Changes

- Create a new price composite adapter.  This adapter is supposed to a generic price adapter that fetches an intermediary token price using the token allocation adapter and then calls a handler for a specific token
-  Add handler for `BEth`

## Steps to Test

1. Run price adapter
2. Send request to adapter and verify that it returns a 200
3. Try step 2 with the quote as `BTC` and `ETH`.

Sample request

```
{
  "id": "1",
  "data": {
    "from": "BETH",
    "to": "USD",
    "source": "coingecko"
  }
}
```

## Quality Assurance

- [] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.


packages/composites/price/src/endpoint/price/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const { from, to, quoteDecimals } = validator.validated.data
+  const fromUpperCase = from.toUpperCase()
+
+  let taDecimals: number
+  let priceExecute: PriceExecute
+  let intermediaryTokenSymbol: string
+  switch (fromUpperCase) {
+    case beth.FROM:
+      taDecimals = beth.INTERMEDIARY_TOKEN_DECIMALS
```
cds95, line 41: I go back and forth between hardcoding this and passing it as a request param.  I decided to hardcode it so that the caller doesn't have to think about it.  Happy to get other opinions.

thodges-gh: Since the price endpoint doesn't require a decimals param on any other adapter, I'm good with hard-coding it. The job naturally handles the required amount of decimals to go to a feed anyway, we only need to be sure that we have enough precision.

thodges-gh (approved):

thodges-gh (changes requested):
I just realized the adapter itself is called "price". Can we rename this to anchor? Price is very generic and this adapter is very specific.

packages/composites/price/CHANGELOG.md
```diff
@@ -0,0 +1,40 @@
+# @chainlink/example-composite-adapter
```
thodges-gh, line 1: This should probably go to a more top-level directory.

boxhock: We've recently started versioning each EA package individually, so we'll need changelogs for each package

packages/composites/price/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@chainlink/price-adapter",
+  "version": "1.0.4",
```
boxhock, line 3: We can start this on version 1.0.0 and empty out the changelog

packages/composites/price/schemas/env.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$id": "https://external-adapters.chainlinklabs.com/schemas/example-composite-adapter.json",
+  "title": "@chainlink/example-composite-adapter env var schema",
```
boxhock, line 3: needs updating

packages/composites/price/src/config.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const makeConfig = (prefix?: string): Config => {
+  const config = Requester.getDefaultConfig(prefix, true)
+  config.rpcUrl = util.getRequiredEnv('RPC_URL', prefix)
+  config.defaultEndpoint = DEFAULT_ENDPOINT
+
+  return {
+    ...config,
+    anchorVaultContractAddress: util.getRequiredEnv('ANCHOR_VAULT_CONTRACT_ADDRESS', prefix),
```
boxhock, line 18: We can either have this EA be Anchor-specific (which we'd need to update the name of this EA) or make the configuration here more generic.

boxhock (changes requested):

packages/composites/anchor/src/endpoint/price/bluna.ts
```diff
@@ -0,0 +1,45 @@
...
+  const viewFunctionAdapterRequest: AdapterRequest = {
+    id: input.id,
+    data: {
+      address: validator.validated.data.terraBLunaContractAddress,
+      query: {
+        state: {},
+      },
+    },
+  }
+  const viewFunctionAdapterResponse = await _execute(viewFunctionAdapterRequest, context)
```
justinkaseman, line 33: Could we type this response?

packages/composites/anchor/src/endpoint/price/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      priceExecute = beth.execute
+      intermediaryTokenSymbol = beth.INTERMEDIARY_TOKEN
+      break
+    case bluna.FROM:
+      taDecimals = bluna.INTERMEDIARY_TOKEN_DECIMALS
+      priceExecute = bluna.execute
+      intermediaryTokenSymbol = bluna.INTERMEDIARY_TOKEN
+      break
+    default:
+      throw Error(`Invalid from symbol ${fromUpperCase}`)
```
justinkaseman, line 52: It might help the user experience to show what _is_ supported too.

cds95: good idea.  I'll update it.

packages/composites/anchor/src/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { makeConfig } from './config'
+
+/**
+ * NOTE: This adapter could probably be made more general in the future.  All it is really doing is fetching an intermediary token price
+ * using the Token Allocation adapter and handling the price after that.
+ */
+
+const NAME = 'ANCHOR'
+
+const { server } = expose(NAME, makeExecute())
```
justinkaseman, line 12: Until we can simplify it, this needs the endpointSelector reference.

```
const { server } = expose(NAME, makeExecute(), undefined, endpointSelector)
```

packages/composites/anchor/src/endpoint/price/beth.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const execute: PriceExecute = async (input, _, config, taAdapterResponse) => {
+  const rpcUrl = config.rpcUrl
+  const provider = new ethers.providers.JsonRpcProvider(rpcUrl)
+  const anchorVaultContract = new ethers.Contract(
+    config.anchorVaultContractAddress,
+    anchorVaultAbi,
+    provider,
+  )
+  const bEthExchangeRateBigNum = await anchorVaultContract.get_rate()
```
justinkaseman, line 26: Could we just use `terra-view-function` adapter for this?

cds95: we can't as the `anchorVaultContract` isn't on Terra

justinkaseman: Ah I gotcha. Yeah that would be good for the "more general" implementation in the future.

packages/composites/anchor/src/endpoint/price/bluna.ts
cds95: isn't it already implicitly typed from the function call's response?

Screen Shot 2021-12-06 at 1 32 01 PM

justinkaseman (approved):

packages/composites/anchor/src/endpoint/price/bluna.ts
justinkaseman: AdapterResponse has an any type on `data`. 

I'll handle this in my types OKR though to make it a generic type.

PR #1263, opened 2021-12-03
Upgrade Redis to V4

austinborn:
## Description
This PR upgrades Redis from V3 to V4, which had some moderate changes, mostly to the `createClient` interface. You can read about the V3 to V4 notes [here](https://github.com/redis/node-redis/blob/master/docs/v3-to-v4.md), and the new `createClient` config [here](https://github.com/redis/node-redis/blob/master/docs/client-configuration.md).

## Changes
- Update `createClient` options interface
- Add new type imports (hacky solution for now, but hopefully this can be fixed in the future)
- Simplify the new `reconnectStrategy` from `retryStrategy` since the new version only allows the number of retries as an input argument, without any other data
- `await client.connect()` to explicitly connect the client to the Redis server

## Steps to Test

The only tests I've run so far are confirming that the Redis server can be used without failing outright. No stress testing or confirming specific Redis configuration params.
1. `yarn && yarn setup`
2. `cd packages/sources/<adapter-to-test> && yarn start`
3. In separate terminal, run ` docker run -p 6379:6379 --name ea-redis -d redis redis-server`
4. Run a query for the adapter twice to confirm it caches the first time, and uses the cached result the second time

Ex. 
`yarn && yarn setup && cd packages/sources/coinbase && yarn start`, then in a separate window,
`curl -X POST http://localhost:8080 -H "Content-Type: application/json" -d '{"jobRunID": "1", "data": {"from": "BTC","to": "USD"}}'`
**First response:**
`{"jobRunID":"1","result":50792.59,"debug":{"staleness":0,"performance":0.192739083,"providerCost":1},"statusCode":200,"data":{"data":{"base":"BTC","currency":"USD","amount":"50792.59"},"result":50792.59}}`
**Second response (after running the curl command again):**
`{"result":50792.59,"maxAge":30000,"debug":{"cacheHit":true,"staleness":0.854,"performance":0.009729041,"providerCost":0},"statusCode":200,"data":{"data":{"base":"BTC","currency":"USD","amount":"50792.59"},"result":50792.59}}`

## Quality Assurance


packages/core/bootstrap/src/lib/cache/redis/redis.ts
```diff
@@ -43,32 +59,9 @@ export const redactOptions = (opts: RedisOptions) => {
...
-  if (options.total_retry_time > 1000 * 60 * 60) {
-    // End reconnecting after a specific timeout and flush all commands
-    // with a individual error
-    logger.warn('Redis retry strategy exhausted.', options)
-    return new Error('Retry time exhausted')
-  }
-  if (options.attempt > 10) {
-    logger.warn(`Redis retry attempt #${options.attempt}`, options)
-    return undefined
-  }
```
boxhock, line 64: Looks like all of these cases would stop the retry strategy previously, but is not included in your changes. Should we include these as well?

boxhock: Oooh I read your PR description now.
> Simplify the new reconnectStrategy from retryStrategy since the new version only allows the number of retries as an input argument, without any other data

Is there a way to limit the number of commands in the queue btw?

austinborn: If you have any ideas to keep some of this logic, lmk. I certainly don't want to get rid of it if there is an alternative.

austinborn: It does look like we can limit the queue though. I might dig into the `createClient` options interface and see what other configuration settings there are:
```
export interface RedisClientOptions<M extends RedisModules, S extends RedisScripts> extends RedisPlugins<M, S> {
    url?: string;
    socket?: RedisSocketOptions;
    username?: string;
    password?: string;
    database?: number;
    commandsQueueMaxLength?: number;
    readonly?: boolean;
    legacyMode?: boolean;
    isolationPoolOptions?: PoolOptions;
}

Comment (austinborn) [reply packages/core/bootstrap/src/lib/cache/redis/redis.ts:64]: If this is correct, there are a lot of options for the `isolationPoolOptions` as well:
https://github.com/coopernurse/node-pool/blob/1c5cb79dcbea27c4b1839bd75bfc41274adb8b94/lib/PoolOptions.js

File: packages/core/bootstrap/README.md [hunk 2]
```diff
@@ -88,14 +88,17 @@ For example, if the `CACHE_KEY_IGNORED_PROPS=timestamp` is set, these requests w
...
-|           | `CACHE_REDIS_PASSWORD` |                                                      The password required for redis auth.                                                      |         |   `null`    |
-|           | `CACHE_REDIS_TIMEOUT`  |                                      The timeout in ms if connection to Redis errors or is not responding.                                      |         |    `500`    |
+| Required? |                 Name                 |                                                                   Description                                                                   | Options | Defaults to |
+| :-------: | :----------------------------------: | :---------------------------------------------------------------------------------------------------------------------------------------------: | :-----: | :---------: |
+|           |   `CACHE_REDIS_CONNECTION_TIMEOUT`   |                                                     Timeout per long lived connection (ms)                                                      |         |   `15000`   |
```
boxhock, line 93: Will this make it reconnect every 15s, or how is this different from `CACHE_REDIS_TIMEOUT`?

packages/core/bootstrap/src/lib/server.ts
```diff
@@ -65,14 +64,16 @@ export const initHandler =
...
     app.get(join(baseUrl, 'health'), async (_, res) => {
-      if (cacheOptions.enabled && cacheOptions.cacheImplOptions.type === 'redis') {
-        logger.debug('Checking if redis connection initialized')
-        const cache = context.cache.instance as redis.RedisCache
-        if (!cache.client.connected) {
-          res.status(500).send({ message: 'Redis not connected', version })
-          return
-        }
-      }
+      // TODO add health check for Redis connection (the following is deprecated)
```
boxhock, line 67: Do we have a Shortcut story we can link to here?

packages/core/bootstrap/src/lib/cache/redis/redis.ts
```diff
@@ -43,118 +60,36 @@ export const redactOptions = (opts: RedisOptions) => {
...
 export class RedisCache {
   options: RedisOptions
-  client: RedisClient
-  _auth: any
-  _get: any
-  _set: any
-  _del: any
-  _quit: any
-  _pttl: any
+  client: any //TODO use RedisClientType with some modifications?
```
boxhock, line 65: Shortcut story?

boxhock:
Mostly looks good to me. We'll want to soak test this properly in staging before proceeding with it. We'll have a new release go out today probably, but I think we can merge this after that.

packages/core/bootstrap/README.md
austinborn: The `CACHE_REDIS_TIMEOUT` is a config on our side to determine when any given request is considered "timed-out". This `CACHE_REDIS_CONNECTION_TIMEOUT` is used as `connectTimeout` in the client to determine when to destroy the socket connection due to inactivity. [Here](https://github.com/redis/node-redis/blob/ac808ea781c4cdcc13e4e7a68777532d646b56c9/packages/client/lib/client/socket.ts#L154-L155) is where the client uses `connectTimeout`, and [here](https://nodejs.org/api/net.html#socketsettimeouttimeout-callback) is an explanation of `socket.setTimeout`.

austinborn: I can add more details in the README to make this clear

austinborn: 70f5781

packages/core/bootstrap/src/lib/server.ts
austinborn: e499557

packages/core/bootstrap/src/lib/cache/redis/redis.ts
austinborn: e499557

packages/core/bootstrap/README.md
boxhock: Thanks for the explanation and readme clarification. Should be good now :+1:

boxhock (approved):

PR #1266, opened 2021-12-03
Snowflake Covid adapter

alejoberardino:
## Description
Adds an external adapter to query data from [Snowflake](https://www.snowflake.com/). Currently only contains an endpoint to query US confirmed Covid cases per County to serve as a POC. Some notes:
- Other adapters like the BigQuery one take it as an input parameter, this approach increases the capabilities of the endpoint but makes it more difficult for the end consumer, so for this initial implementation the the query is hardcoded in the adapter.
- The input parameters are `state` and `county`, to get a unique record from the query.
- The table used from the dataset is a "Dashboard" table that contains an updated sum aggregate of cases, so no date parameter is needed. This simplifies the input, but removes the possibility of fetching historical data; in any case it's a minor change required to alter the approach. 

## Steps to Test

1. Setup an account in Snowflake (or reach out to me for credentials to skip to the last step)
2. Follow docs to [assign a key pair to a user](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#using-key-pair-authentication) and [finding your account identifier](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html)
3. Set the env vars with the data from the previous step
2. Run `yarn test flightaware/test/unit`

## Quality Assurance


packages/sources/snowflake/src/endpoint/covid-cases.ts
```diff
@@ -0,0 +1,64 @@
...
+    data: {
+      database: config.api.database,
+      schema: config.api.schema,
+      statement: `
+        select confirmed
+        from JHU_DASHBOARD_COVID_19_GLOBAL
+        where country_region = 'United States'
+        and province_state = :1
+        and county = :2
+        limit 1
```
thodges-gh, line 38: Since we're doing a `limit 1` do we also need to have an `order by` to ensure we get the correct first result?

packages/sources/snowflake/src/util.ts
```diff
@@ -0,0 +1,39 @@
+import crypto from 'crypto'
+import jwt from 'jsonwebtoken'
+
+const LIFETIME = 120 // seconds
```
thodges-gh, line 4: I believe this is milliseconds if we're applying it to `Date.now()`.

packages/sources/snowflake/test-payload.json
```diff
@@ -0,0 +1,8 @@
+{
+  "requests": [
+    {
+      "state": "Alabama",
+      "county": "Autauga"
+    }
+  ]
+}
```
thodges-gh, line 8: nit: include the final newline character.

thodges-gh (changes requested):

packages/sources/snowflake/src/endpoint/covid-cases.ts
alejoberardino: The combination of `province_state` and `county` is unique, so the limit wasn't technically necessary, removed in new rev.

packages/sources/snowflake/CHANGELOG.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Updated dependencies
+  - @chainlink/ea-bootstrap@1.0.1
+  - @chainlink/ea-test-helpers@1.0.0
+
+## 1.0.0
+
+### Major Changes
+
+- EAv2 Release. Start of individual EA versioning.
```
boxhock, line 44: We can empty this out

```suggestion
## 1.0.0

### Major Changes

- Initial EA release
```

packages/sources/snowflake/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@chainlink/snowflake-adapter",
+  "version": "1.0.4",
```
boxhock, line 3: Let's start it on v1.0.0

```suggestion
  "version": "1.0.0",
```

packages/sources/snowflake/README.md
```diff
@@ -0,0 +1,95 @@
...
+| :-------: | :---------: | :-----------------------------------------------------------------------: | :-----: | :----------------------: |
+|    ✅     | PRIVATE_KEY | Private PKCS PEM Key matching assigned public key for a user in Snowflake |         |                          |
+|    ✅     |   ACCOUNT   |                Unique identifier for the Snowflake account                |         |                          |
+|    ✅     | DB_USERNAME |                         Name of the database user                         |         |                          |
+|           |  DATABASE   |                    Name of the database to connect to                     |         | COVID19_BY_STARSCHEMA_DM |
+|           |   SCHEMA    |             Database schema where the main tables are located             |         |          PUBLIC          |
+|           |  PROVIDER   |           Shortcut for the selected cloud provider (e.g. 'aws')           |         |                          |
+|           |   REGION    |        Shortcut for the designated cloud region (e.g. 'us-east-2')        |         |                          |
```
boxhock, line 16: To be more explicit, do you think it would make sense to prefix these with `CLOUD_`?

packages/sources/snowflake/src/endpoint/covid-cases.ts
```diff
@@ -0,0 +1,63 @@
...
+  const jobRunID = validator.validated.id
+  const state = validator.validated.data.state
+  const county = validator.validated.data.county
+
+  const jwtToken = buildSnowflakeJWT({
+    privateKey: config.api.privateKey,
+    qualifiedUsername: config.api.qualifiedUsername,
+  })
+
+  const response = await Requester.request({
```
boxhock, line 25: If you could create a TypeScript interface for the `ResponseSchema`, you can add the type here like this:

```suggestion
  const response = await Requester.request<ResponseSchema>({
```

That adds some more type safety and we can export that type out of this adapter if needed

packages/sources/snowflake/src/endpoint/covid-cases.ts
```diff
@@ -0,0 +1,63 @@
...
+    baseURL: config.api.baseURL,
+    url: '/statements',
+    method: 'POST',
+    data: {
+      database: config.api.database,
+      schema: config.api.schema,
+      statement: `
+        select confirmed
+        from JHU_DASHBOARD_COVID_19_GLOBAL
+        where country_region = 'United States'
```
boxhock, line 35: Is the requirement that it needs to be in the US?
I would maybe prefer that we take this as an input as well - just in case we ever want a different country at some point we'd not have to do any code changes to support it.

packages/sources/snowflake/src/config.ts
```diff
@@ -0,0 +1,40 @@
...
+  const baseURL = `https://${fullAccount}.snowflakecomputing.com/api`
+
+  config.api = {
+    account,
+    baseURL,
+    database,
+    privateKey,
+    qualifiedUsername,
+    schema,
+    username,
```
boxhock, line 34: For all of these extra params, I would create a new Config type that extends the Config type from `@chainlink/types`, so that we can define these more clearly.

packages/sources/snowflake/src/index.ts
```diff
@@ -0,0 +1,6 @@
+import { expose } from '@chainlink/ea-bootstrap'
+import { makeExecute, endpointSelector } from './adapter'
+import { makeConfig, NAME } from './config'
+
+const { server } = expose(NAME, makeExecute(), undefined, endpointSelector)
+export { NAME, makeExecute, makeConfig, server }
```
boxhock, line 6: This will be useful once you add the ResponseSchema interface

```suggestion
import { makeConfig, NAME } from './config'
import * as types from './endpoint'

const { server } = expose(NAME, makeExecute(), undefined, endpointSelector)
export { NAME, makeExecute, makeConfig, server, types }
```

boxhock (changes requested):
Congrats on your first PR! :smile: 

Had a few changes to request, but I think PR looks really good otherwise

packages/sources/snowflake/src/endpoint/covid-cases.ts
alejoberardino: Some tables in the dataset technically have info for more countries, but after taking a look at the data it seems like it's not very complete for non US regions (for example, the data Argentina in the JHU table is not available by county). The decision was to go for US only, one source only (JHU dashboard), since this is more of a PoC. Ideally with normalized complete data we'd change this to accept country, and potentially take an aggregate of multiple sources in the dataset to give a result, but since the data won't be available I prefer to hardcode the country to set expectations

boxhock (approved):
Nice work :+1:

PR #1279, opened 2021-12-07
Provide fixed feed_id for cache warmer requests

alejoberardino:
## Description

The `feed_id` label for metrics currently takes all symbols provided in the `base` parameter and joins them together to form its value. Additionally, the order for symbols is not guaranteed. This means very high cardinality, which [for services like Prometheus is not ideal](https://www.robustperception.io/cardinality-is-key). This PR changes the behavior to normalize the order of the parameters when present, and foregoing this behavior when the request comes from the cache warmer, to reduce the number of values seen in metrics.

## Changes

- `getFeedId` method now returns fixed `CACHE_WARMER` value when requests are triggered by the cache warmer.
- Symbols are sorted before joining to avoid different label values for the same set.

## Steps to Test

1. Run an adapter e.g. CoinGecko with `EXPERIMENTAL_METRICS_ENABLED=true CACHE_ENABLED=true`
2. Send requests for price information for different tickers
3. Call the metrics endpoint and verify that there are no `feed_id` values for the batch warmer consolidation of requests (something like `"feed_id": "[BTC|ETH|...]/USD"`), being instead replaced and condensed to one `CACHE_WARMER`

## Quality Assurance

### DIFF
--- a/.changeset/purple-fishes-beg.md
+++ b/.changeset/purple-fishes-beg.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
+---
+
+Provide fixed feed_id for cache warmer requests
--- a/packages/core/bootstrap/src/lib/metrics/util.ts
+++ b/packages/core/bootstrap/src/lib/metrics/util.ts
@@ -37,6 +37,33 @@ export function normalizeStatusCode(status?: number): string {
 export const MAX_FEED_ID_LENGTH = 300
 
 /**
+ * Fixed label value for cache warmer feed_id
+ */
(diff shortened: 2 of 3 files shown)


boxhock:
@alejoberardino Will give a proper review later, but just adding a note now while I remember: You'll have to create a changeset for this PR with `yarn changeset`.

This will ensure the EA bootstrap package has a version bump for the next release, and all EAs also get bumped.

boxhock:
This PR looks good to me. Just needs a changeset and @justinkaseman to review as well.

austinborn (approved):

justinkaseman (approved):
LGTM!

PR #1283, opened 2021-12-08
fix/Fixed Cache Warmer metric issue

silaslenihan:
## Description
Fixes the error with the cache warmer metrics causing them to go into the negatives

## Changes
- Changes logic in `epics.ts` to correctly decrement on the correct events
- Alters the way cache intervals are inputted

## Steps to Test

1. Run an adapter locally with EXPIERIMENTAL_METRICS_ENABLED = true
2. Run Prometheus metrics
3. Keep track of the `cache_warmer_get_count` metric in Prometheus, ensuring it never dips below 0 after losing connection or failing in some other way
4. Once we have this live, we know it will be working because the metrics will never dip below 0 like they have been doing.
## Quality Assurance

### DIFF
--- a/.changeset/tasty-avocados-sparkle.md
+++ b/.changeset/tasty-avocados-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
+---
+
+This PR fixes an issue with the cache warmer metrics that was causing them to go into the negatives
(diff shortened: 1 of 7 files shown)


austinborn (approved):
The `$` in variable names is strange to me, but this PR didn't introduce it so I'm not concerned about it right now. Otherwise, LGTM

austinborn (changes requested):
Be sure to add changeset with `yarn changeset`

justinkaseman (approved):

PR #1292, opened 2021-12-10
Double the default max allowed entries in the local cache

justinkaseman:
## Description
The local cache has a setting for the max number of entries in the cache.
Nops are at a scale where their caching breaks because this cap is hit.
......

## Changes
- Double local cache MAX_ITEMS

## Steps to Test
-

## Quality Assurance

### DIFF
--- a/.changeset/soft-oranges-collect.md
+++ b/.changeset/soft-oranges-collect.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
+---
+
+Increase default local cache MAX_ITEMS to 1000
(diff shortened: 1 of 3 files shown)


alejoberardino:
Do we know the usual memory usage for the local cache? I don't think it's likely this will fill the heap, but if cached responses are large enough, we might run into that limit, specially since node processes usually have a low default cap set

justinkaseman:
> 

It's not something that we have thorough testing on yet. The original 500 number just came from the `lru-cache` dependencies documentation example. I'll make a new story to investigate.

alejoberardino (approved):

PR #1297, opened 2021-12-12
[Fix] Redis request coalescing

justinkaseman:
## Description
Previously our redis client was able to use the `set` command with a value of a boolean
Our move to node-redis v4 is closer to the actual redis command (https://redis.io/commands/set) which only allows strings to be set.
This caused a breaking to our implementation of request coalescing causing the following error:

```
ReplyError: ERR Protocol error: expected '$', got '*'
```

......

## Changes

- Sets `true` as a string value for the request coalescing marker.

## Steps to Test

Run and query an adapter with request coalescing on.

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/lib/cache/index.ts
+++ b/packages/core/bootstrap/src/lib/cache/index.ts
@@ -183,10 +183,9 @@ export class AdapterCache {
       feedId: adapterRequest.metricsMeta?.feedId || 'N/A',
     })
 
-    const cachedAdapterResponse =
-      this.options.requestCoalescing.enabled && this.options.cacheImplOptions.type === 'local'
-        ? await this.getWithCoalescing(key)
-        : await this.cache.getResponse(key)
+    const cachedAdapterResponse = this.options.requestCoalescing.enabled
+      ? await this.getWithCoalescing(key)
+      : await this.cache.getResponse(key)
 
     if (cachedAdapterResponse) {
       const maxAgeOverride = getMaxAgeOverride(adapterRequest)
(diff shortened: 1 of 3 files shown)


austinborn (dismissed):

packages/core/bootstrap/src/lib/cache/redis/redis.ts
```diff
@@ -105,25 +104,20 @@ export class RedisCache {
...
 
-  // TODO: We should have seperate services for response entries, and coalescing support
-  async setFlightMarker(key: string, maxAge: number) {
-    return this.contextualTimeout(this.client.set(key, true, { PX: maxAge }), 'setFlightMarker', {
-      key,
-      maxAge,
-    })
+  async setFlightMarker(): Promise<void> {
+    // NOTE: node-redis v4 supports auto-pipelining
+    // See https://github.com/redis/node-redis/issues/492
```
boxhock, line 109: Auto-pipelining != request coalescing (in our context at least), or am I missing anything?

I'm afraid removing this will break prod feeds if this was merged (DM me for more details).

Can we just disable request coalescing by default instead?

boxhock (changes requested):

packages/core/bootstrap/src/lib/cache/index.ts
```diff
@@ -183,9 +183,10 @@ export class AdapterCache {
       feedId: adapterRequest.metricsMeta?.feedId || 'N/A',
     })
 
-    const cachedAdapterResponse = this.options.requestCoalescing.enabled
-      ? await this.getWithCoalescing(key)
-      : await this.cache.getResponse(key)
+    const cachedAdapterResponse =
+      this.options.requestCoalescing.enabled && this.options.cacheImplOptions.type === 'local'
```
boxhock, line 187: This is still disabling it for Redis

```suggestion
    const cachedAdapterResponse = this.options.requestCoalescing.enabled
```

boxhock (changes requested):
LGTM besides this one thing

PR #1299, opened 2021-12-14
fix(agoric): don't scale Agoric fees, avoid 'number' requiredFee

michaelfig:
## Description

At Agoric we have modified the bridged $LINK to use 18 decimal places, so the best way to represent these fee amounts is to use JSON strings. We can and must remove the scaling hack that we needed before to translate between 6 decimals and 18 decimals.

## Changes

- Don't scale Agoric fees from the old 6-decimal-place value; just use 18 decimal places for $LINK on both chains
- Use JSON strings to represent $LINK fee amounts for high precision

## Steps to Test

Follow the steps in https://github.com/Agoric/dapp-oracle#readme to test, following the Chainlink instructions, which requires also smartcontractkit/external-initiator#160

## Quality Assurance

### DIFF
--- a/.changeset/odd-clouds-occur.md
+++ b/.changeset/odd-clouds-occur.md
@@ -1,0 +1,9 @@
+---
+'@chainlink/agoric-adapter': major
+---
+
+Agoric \$LINK is now 18 decimal places and serialised as a string
+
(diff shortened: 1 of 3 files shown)


austinborn:
@michaelfig Thank you for your contribution. Closing this branch since the change was introduced in #1391, but it should make it in the next release to `master`.

PR #1300, opened 2021-12-14
Change DP error status code surfacing

alejoberardino:
## Description

A common occurrence with EAs is that the Data Provider sources fail, and we currently surface their status code as the response from the EA itself. Since EAs are not a completely clean proxy by design, the end result is that it's not possible to discern in metrics whether the failing point is the EA or the DP.

This PR changes the behavior of the core bootstrap so that a failing request to a DP results in a `200` response from the EA, providing a new field `providerStatusCode` to give complete information to the requester. The reasoning behind the new field was to avoid the confusing perspective of a single `statusCode` in the response, that used to match the HTTP status but would no longer do so, as well as leaving the existing behavior intact. It's true that the `statusCode` field will remain somewhat redundant, which is a change that could be made in the future to remove it altogether.

**Note**: the first commit includes the changes to the core package, the other two are fixes for all tests that failed from those changes.

## Changes

-  Change the behavior of the core requester to surface response error status code as a new field, with main status code changing to 200

## Steps to Test

1. Run any adapter of choice
2. Make sure that whatever input you provide will cause an error in the API (not in the adapter)
3. The response should have a `statusCode` of 200, and a `providerStatusCode` of whatever the original API failure code was

## Quality Assurance


packages/sources/sochain/src/endpoint/balance.ts
```diff
@@ -24,10 +24,14 @@ const getBalance: balance.GetBalance = async (account, config) => {
...
   const balance = ethers.utils.parseUnits(response.data.data.confirmed_balance, 8).toString()
 
-  return {
+  const result = {
     payload: response.data,
     result: [{ ...account, balance }],
+    status: response.status,
   }
+
+  console.log(result)
```
boxhock, line 33:
```suggestion
```

packages/composites/circuit-breaker/test/integration/adapter.test.ts
```diff
@@ -121,13 +121,20 @@ describe('execute', () => {
...
+    const expectedProviderStatusCodes = [0, 500]
+
+    request.forEach((req, i) => {
       it(`${req.name}`, async () => {
         try {
           await execute(req.input, {})
         } catch (error) {
           const errorResp = Requester.errored(jobID, error)
-          assertError({ expected: 500, actual: errorResp.statusCode }, errorResp, jobID)
+          console.log(errorResp)
```
boxhock, line 132:
```suggestion
```

boxhock (changes requested):
Does this PR also change the metrics to differentiate between EA vs DP status codes?

We'll need a changeset for this as well

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -70,7 +71,13 @@ export class Requester {
         if (n === 1) {
           const message = `Could not retrieve valid data: ${JSON.stringify(response.data)}`
           const cause = response.data.error || 'customError'
-          throw new AdapterError({ message, cause, url })
+          throw new AdapterError({
+            statusCode: 200,
+            providerStatusCode: response.data.error?.code || response.status,
```
austinborn, line 76: If `response.data.error?.code` is `0`, do we want to override with `response.status`? I'd suggest using `??` in place of `||` for number overrides

austinborn (approved):

austinborn (changes requested):
Waiting for verified author

boxhock (dismissed):

austinborn (approved):

PR #1308, opened 2021-12-14
Change status code in metrics

alejoberardino:
## Description

A common occurrence with EAs is that the Data Provider sources fail, and we currently surface their status code as the response from the EA itself. Since EAs are not a completely clean proxy by design, the end result is that it's not possible to discern in metrics whether the failing point is the EA or the DP.

This PR changes only the metrics published in the optional `/metrics` endpoint, so that a failing request to a DP results in metric labels of `status_code="2XX",provider_status_code="5XX"` for example. 

## Changes

- Changed metric labels calculation to modify cases with failing DP requests only.

## Steps to Test

1. Run an adapter of choice with `EXPERIMENTAL_METRICS_ENABLED=true`
2. Make a request so that the DP request succeeds
3. Make another so that the DP request fails
4. You should see entries for `http_total_requests` in the `/metrics` endpoint with `status_code="2XX",provider_status_code="2XX"` and `status_code="2XX",provider_status_code="5XX"` respectively

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/index.ts
+++ b/packages/core/bootstrap/src/index.ts
@@ -124,9 +124,14 @@ const withMetrics: Middleware = async (execute, context) => async (input: Adapte
     }
     const end = metrics.httpRequestDurationSeconds.startTimer()
 
-    return (statusCode?: number, type?: metrics.HttpRequestType) => {
-      labels.type = type
-      labels.status_code = metrics.util.normalizeStatusCode(statusCode)
+    return (props: {
+      providerStatusCode?: number
+      statusCode?: number
+      type?: metrics.HttpRequestType
+    }) => {
+      labels.type = props.type
+      labels.status_code = metrics.util.normalizeStatusCode(props.statusCode)
+      labels.provider_status_code = metrics.util.normalizeStatusCode(props.providerStatusCode)
       end()
       metrics.httpRequestsTotal.labels(labels).inc()
     }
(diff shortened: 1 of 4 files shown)


.changeset/hot-days-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
```
boxhock, line 2: I think we might want to make this a `minor` bump, since it adds metrics labels

```suggestion
'@chainlink/ea-bootstrap': minor
```

packages/core/bootstrap/src/index.ts
```diff
@@ -135,15 +140,20 @@ const withMetrics: Middleware = async (execute, context) => async (input: Adapte
...
   try {
     const result = await execute({ ...input, metricsMeta }, context)
-    record(
-      result.statusCode,
-      result.data.maxAge || (result as any).maxAge
-        ? metrics.HttpRequestType.CACHE_HIT
-        : metrics.HttpRequestType.DATA_PROVIDER_HIT,
-    )
+    record({
+      statusCode: result.statusCode,
```
justinkaseman, line 144: Should we also record the `providerStatusCode` as 200 here?

alejoberardino: The normalization that happens within the record function will now default that to 200, so it will be added there

justinkaseman (approved):
LGTM, nice work

PR #1316, opened 2021-12-17
Refactor/1forge types

karen-stepanyan:
## Changes

- Adds forex endpoint ResponseSchema

## Quality Assurance

### DIFF
--- a/.changeset/chatty-cobras-exist.md
+++ b/.changeset/chatty-cobras-exist.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/1forge-adapter': patch
+---
+
+added response types
--- a/packages/sources/1forge/src/endpoint/forex.ts
+++ b/packages/sources/1forge/src/endpoint/forex.ts
@@ -10,6 +10,12 @@ export const inputParameters: InputParameters = {
   quantity: false,
 }
 
+interface ResponseSchema {
+  value: string
+  text: string
+  timestamp: number
+}
+
 export const execute: ExecuteWithConfig<Config> = async (request, _, config) => {
   const validator = new Validator(request, inputParameters)
   if (validator.error) throw validator.error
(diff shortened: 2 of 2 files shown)


packages/sources/ncfx/src/endpoint/forex.ts
```diff
@@ -4,8 +4,18 @@ import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
 export const supportedEndpoints = ['forex']
 
 export const inputParameters: InputParameters = {
-  base: ['base', 'from', 'coin'],
-  quote: ['quote', 'to', 'market'],
+  base: {
+    aliases: ['from', 'coin'],
+    description: 'The symbol of the currency to query',
+    options: ['BTC', 'ETH', 'USD'],
```
justinkaseman, line 10: These aren't forex symbols.

packages/sources/ncfx/src/endpoint/crypto.ts
```diff
@@ -4,8 +4,18 @@ import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
...
+  base: {
+    aliases: ['from', 'coin'],
+    description: 'The symbol of the currency to query',
+    options: ['BTC', 'ETH', 'USD'],
+    required: true,
+  },
+  quote: {
+    aliases: ['to', 'market'],
+    description: 'The symbol of the currency to convert to',
+    options: ['BTC', 'ETH', 'USD'],
```
justinkaseman, line 16: This endpoint supports many more symbols than this. We should either pull an accurate list from the data provider or leave this off.

CC: @austinborn

justinkaseman: Actually, this is just used for the README huh? Not actually enforced?

karen-stepanyan: > Actually, this is just used for the README huh? Not actually enforced?

no, enforced. if there is 'options', then values should match

austinborn: Yeah it's enforced. If these are only a sample of options, then it should go in the description section

justinkaseman (approved):

PR #1360, opened 2022-01-03
Fix WS issues with CF Benchmarks adapter

cds95:
## Description

 - Address failing HTTP requests to CF Benchmarks
 - Fix bug where sending requests for different feeds always returned the price of the first one.  This issue was caused due to the WS `subscriptionKey` always being the same as the `id` field in the WS subscription message was always being ignored.  We should come up with a better solution for this in the future.  I've created a ticket for this here https://app.shortcut.com/chainlinklabs/story/24721/gracefully-handle-cases-where-ws-subscription-payload-messages-contains-ignored-key

## Changes

- Update adapter HTTP endpoint
- Add a `cacheID` field to ensure different feeds use different keys

## Steps to Test

Make sure you enable websockets before testing.  

1. Send this payload and get a successful response.  Additionally verify that websockets is running from the adapter logs.

```
{
  "id": "1",
  "data": {
    "index": "ETHUSD_RTI"
  }
}
```
2. Send below payload and verify response
```
{
  "id": "1",
  "data": {
    "index": "BRTI"
  }
}
```

## Quality Assurance

### DIFF
--- a/packages/sources/cfbenchmarks/src/adapter.ts
+++ b/packages/sources/cfbenchmarks/src/adapter.ts
@@ -29,7 +29,7 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
   }
   const getSubscription = (type: 'subscribe' | 'unsubscribe', id?: string) => {
     if (!id) return
-    return { type, id, stream: 'value' }
+    return { type, id, stream: 'value', cacheID: id } // Temporarily add another key named cacheID that won't be ignored when generating the subscription key in getSubsID
   }
   return () => {
     const defaultConfig = config || makeConfig()
--- a/.changeset/chilly-cycles-confess.md
+++ b/.changeset/chilly-cycles-confess.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/cfbenchmarks-adapter': minor
+---
+
+Bugfix for broken adapter WS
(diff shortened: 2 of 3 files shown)


packages/sources/cfbenchmarks/src/adapter.ts
```diff
@@ -29,7 +29,7 @@ export const makeWSHandler = (config?: Config): MakeWSHandler => {
   }
   const getSubscription = (type: 'subscribe' | 'unsubscribe', id?: string) => {
     if (!id) return
-    return { type, id, stream: 'value' }
+    return { type, id, stream: 'value', cacheID: id } // Temporarily add another key named cacheID that won't be ignored when generating the subscription key in getSubsID
```
cds95, line 32: I've created this ticket to handle this better in the future. https://app.shortcut.com/chainlinklabs/story/24721/gracefully-handle-cases-where-ws-subscription-payload-messages-contains-ignored-key.

austinborn (dismissed):

packages/sources/cfbenchmarks/src/adapter.ts
austinborn: Have you checked if this would affect any other adapters?

austinborn: It looks like every other definition of `getSubscription` for EAs uses a single asset pair, so I'm guessing this is unique to `cfbenchmarks`.

cds95: yup it's for `cfbenchmarks` only.  It doesn't affect other adapters.

boxhock (dismissed):

alejoberardino (dismissed):

justinkaseman (approved):

PR #1392, opened 2022-01-06
limit ws message updates

cds95:
- Rate limit the number of times a WS message is handled for Tiingo


packages/sources/tiingo/src/adapter.ts
```diff
@@ -139,6 +139,7 @@ export const makeWSHandler = (config?: Config): MakeWSHandler | undefined => {
         }
         return Requester.success('1', { data: { result } }, true)
       },
+      minTimeToNextMessageUpdateInS: 1,
```
justinkaseman, line 142: Could we put this on an environment variable so that it is easy to configure?

justinkaseman (approved):

PR #1423, opened 2022-01-11
Remove resultPath from all input params

austinborn:
## Description
- Remove all `resultPath` objects from `inputParameters` since it should be excluded as a shared property [Shortcut #25085](https://app.shortcut.com/chainlinklabs/story/25085/add-references-to-shared-inputs-in-each-ea-readme)
- Add `excludableInternalAdapterRequestProperties` to Validator so that properties like `resultPath` and `overrides` are still present in `validated.data`
- Add safe accessors for `input.data` in the Validator
- Remove unit test that required `data` to be present on inputs, even if no data is required for endpoint
- Update unit test for `blockstream` to change error code

## Quality Assurance


austinborn:
This may not be the best solution. Looking at failing tests and noticing a lot of places where `resultPath` is assumed to be present in `validated.data`, which it wouldn't be with this change.

packages/sources/trueusd/src/endpoint/trueusd.ts
```diff
@@ -35,7 +27,7 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   if (validator.error) throw validator.error
 
   const jobRunID = validator.validated.id
-  const resultPath = validator.validated.data.resultPath
+  const resultPath = validator.validated.data.resultPath || 'totalTrust'
```
justinkaseman, line 30: We have something for this by adding:
```
export const endpointResultPaths = {
  trueusd: 'totalTrust'
}
```

It will fill in here https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/external-adapter/builder.ts#L82

packages/core/bootstrap/test/unit/validator.test.ts
```diff
@@ -7,15 +7,6 @@ describe('Validator', () => {
       endpoint: false,
     }
 
-    it('errors if no input data is supplied', () => {
```
justinkaseman, line 10: We still want this right? Seems like a useful test. Might just have to try catch it.

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -19,6 +19,14 @@ import presetIncludes from './overrides/presetIncludes.json'
 import { Requester } from './requester'
 import { inputParameters } from './builder'
 
+const sharedInputConfig = excludableInternalAdapterRequestProperties.reduce<Record<string, false>>(
+  (config, name) => {
+    config[name] = false
```
justinkaseman, line 24: Could also make `sharedInputConfig` using the inputParameter syntax (gives it a type and may be more readable)

alejoberardino:
@austinborn I do think this is the right approach, sounds reasonable that tests are failing

packages/core/bootstrap/src/lib/external-adapter/validator.ts
```diff
@@ -19,25 +19,42 @@ import presetIncludes from './overrides/presetIncludes.json'
...
-  constructor(input = {}, inputConfigs = {}, options = {}, shouldLogError = true) {
+  constructor(
+    input: InputType = { id: '1', data: {} },
+    inputConfigs = {},
+    options = {},
+    shouldLogError = true,
+  ) {
     this.input = { ...input }
-    this.inputConfigs = { ...inputConfigs }
+    if (!this.input.id) this.input.id = '1' //TODO Please remove these once "no any" strict typing is enabled
```
cds95, line 52: when will we enable "no any" strict typing?  Also is there a Shortcut ticket for this so that we don't forget?

packages/sources/trueusd/src/config.ts
```diff
@@ -9,5 +9,6 @@ export const DEFAULT_BASE_URL = 'https://core-api.real-time-attest.trustexplorer
 export const makeConfig = (prefix?: string): Config => {
   const config = Requester.getDefaultConfig(prefix)
   config.api.baseURL = config.api.baseURL || DEFAULT_BASE_URL
+  config.defaultEndpoint = DEFAULT_ENDPOINT
```
cds95, line 12: is this supposed to be part of this PR?

cds95 (commented):
LGTM.  Just had some questions.

packages/sources/trueusd/src/config.ts
austinborn: Yes, we did this so we could add
```
export const endpointResultPaths = {
  trueusd: 'totalTrust',
}
```
since the `resultPath` is otherwise removed.

packages/core/bootstrap/src/lib/external-adapter/validator.ts
austinborn: Good point, I'll add a note in https://app.shortcut.com/chainlinklabs/story/11106/disable-allowing-of-any-for-new-adapters

cds95 (approved):

PR #1427, opened 2022-01-12
Hotfix/metalsapi fix

austinborn:
## Description

Two issues resolved here:
1) Batched `base` symbols are caught and error is thrown
2) `from` symbol should be queried as `base` in `latest` endpoint request

## Steps to Test
`yarn && yarn setup && yarn test packages/sources/metalsapi/test/integration/adapter.test.ts`

## Quality Assurance


packages/sources/metalsapi/src/endpoint/latest.ts
```diff
@@ -61,13 +61,19 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
-  const from = validator.overrideSymbol(AdapterName)
+  const base = validator.overrideSymbol(AdapterName)
   const to = validator.validated.data.quote
   const url = `latest`
+  if (Array.isArray(base))
+    throw new AdapterError({
+      jobRunID,
+      message: `Base symbol ${base} is not batchable.`,
+      statusCode: 400,
+    })
```
boxhock, line 72: I assume we'd never run into this under normal operation since we have the following set:
```typescript
export const batchablePropertyPath = [{ name: 'quote' }]
```

Am I understanding it right? cc @justinkaseman

boxhock (approved):
Looks good to me. Just want to confirm my understanding here is correct

packages/sources/metalsapi/src/endpoint/latest.ts
justinkaseman: Yes, that is how it previously worked. I'll dig deeper into what is happening.

boxhock: Ok thanks. Btw this was not what broke it today, I think this is just adding more input validation.

boxhock: Ah no we should probably keep it. I just wanted to confirm that `batchablePropertyPath` would prevent us from sending an invalid request payload.

PR #1431, opened 2022-01-12
[Fix] Metalsapi includes symbols for batched data

justinkaseman:
## Description
Fix to `metalsapi` missing sending `symbols` to the data provider.

NOTE: doesn't need a new changeset as @austinborn 's last PR added it for this

## Changes
- Uses inputParameters type to stop `base` as an array
- Sends `symbols` with data provider request
- Amend tests

## Steps to Test
1. Run `metalsapi` adapter and redux dev tools
2. Request a pair
3. The pair should be seen in redux dev tools subscription state, along with starting a batch warmer for that pair
4. Request another pair with the same `base`. It should join the previously made batch warmer.
5. Request a pair with a different `base`. It should start a new batch warmer because base is not a batchable property.
6. Request `base` as an array. It should fail with 400.

## Quality Assurance


packages/sources/metalsapi/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -281,7 +73,7 @@ exports[`execute latest api should error with batched base symbols 1`] = `
 Object {
   "error": Object {
     "feedID": "[BTC|XAU]/USD",
-    "message": "Base symbol XAU,BTC is not batchable.",
```
justinkaseman, line 284: @austinborn I am surprised that this error message did not end up as the type validation one.

austinborn (approved):

packages/sources/metalsapi/test/integration/__snapshots__/adapter.test.ts.snap
austinborn: Until the input params are updated for an adapter, the only validation they go through is checking that the param exists if `inputParameters.[param]` is `true` or an alias array. Did you add the new `inputParameters` or did that come from some other commit?

justinkaseman: I added them here with the type property

PR #1441, opened 2022-01-13
Fix/rate limit ignore health chceks

austinborn:
## Description
Ignore health checks in rate limiter and return the usual health check response instead.

## Steps to Test

`export SERVER_RATE_LIMIT_MAX=1`, then query `localhost:8080/health` while an adapter is running at `localhost:8080`. The health check should succeed even when sending multiple requests within seconds of each other.

## Quality Assurance

### DIFF
--- a/packages/core/bootstrap/src/lib/server.ts
+++ b/packages/core/bootstrap/src/lib/server.ts
@@ -150,9 +150,13 @@ function initExpressMiddleware(app: express.Express) {
     windowMs,
     max, // limit each IP's requests per windowMs
     keyGenerator: () => '*', // use one key for all incoming requests
-    handler: ((_, res) => {
-      httpRateLimit.inc()
-      res.status(429).send('Too many requests, please try again later.')
+    handler: ((req, res) => {
+      if (req.url === '/health') {
+        res.status(200).send({ message: 'OK', version })
+      } else {
+        httpRateLimit.inc()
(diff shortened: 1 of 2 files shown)


packages/core/bootstrap/src/lib/server.ts
```diff
@@ -150,9 +150,13 @@ function initExpressMiddleware(app: express.Express) {
     windowMs,
     max, // limit each IP's requests per windowMs
     keyGenerator: () => '*', // use one key for all incoming requests
-    handler: ((_, res) => {
-      httpRateLimit.inc()
-      res.status(429).send('Too many requests, please try again later.')
+    handler: ((req, res) => {
+      if (req.url === '/health') {
+        res.status(200).send({ message: 'OK', version })
```
alejoberardino, line 155: I'd change this to `next()` just so we can have the option of using more complex healthchecks (I see in the code we already have a more complex redis check commented out)

alejoberardino (changes requested):

packages/core/bootstrap/src/lib/server.ts
justinkaseman: Seconded. It should just pass through.

austinborn: Good idea, just updated it

alejoberardino (approved):

alejoberardino (approved):

PR #1445, opened 2022-01-14
Give nicer names to GH workflows

alejoberardino:
Simple PR to change names for existing GH workflows


austinborn (approved):
Woo!

justinkaseman:
> The branch settings are going to have to be modified with this change, as they are enforcing the old names.

I don't follow this - there are no branch name changes here.

I'm merging this, if the pipeline breaks I will revert.

boxhock:
> > The branch settings are going to have to be modified with this change, as they are enforcing the old names.
> 
> I don't follow this - there are no branch name changes.
> 
> Edit: Looks like one of the checks isn't running though thinking

It's the required check names that are changing - not the branch names. In the branch settings, we need to unmark the old required check names and select the checks with the updated names.

eddiejhong:
> > > The branch settings are going to have to be modified with this change, as they are enforcing the old names.
> > 
> > 
> > I don't follow this - there are no branch name changes.
> > Edit: Looks like one of the checks isn't running though thinking
> 
> It's the required check names that are changing - not the branch names. In the branch settings, we need to unmark the old required check names and select the checks with the updated names.

@boxhock I would love to get this change merged in but I don't understand what we need to do to unblock this. Where can the above settings be updated?

PR #1458, opened 2022-01-18
Allow setting custom Ogmios URLs for ADA balance adapter

cds95:
## Description

There is a node provider that wishes to provide a Cardano node using a URL with a path other than the root.  The Cardano Ogmios library does not support this out of the box.  This PR modifies their library to allow for this use case.

## Changes

- Only allow 2 required environment variables for adapter.  These are `WS_OGMIOS_URL` and `HTTP_OGMIOS_URL`.
- Port over Cardano Ogmios `createInteractionContext` function from the Cardano Ogmios client library and repurpose to allow setting custom URLs.

## Steps to Test

1. Remove all environment variables and set `WS_OGMIOS_URL` and `HTTP_OGMIOS_URL` environment variables
2. Fire request to adapter and verify a that you get a successful response.

## Quality Assurance


packages/sources/ada-balance/src/endpoint/ogmios/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  return {
+    ...base,
+    address: {
+      http: httpOgmiosURL,
+      webSocket: wsOgmiosURL,
+    },
+  }
+}
+
+export const createInteractionContext = async (
```
cds95, line 33: This is the function that has been repurposed for our use case.

packages/sources/ada-balance/src/config.ts
```diff
@@ -4,20 +4,18 @@ import { Config as BaseConfig } from '@chainlink/types'
...
   const baseConfig = Requester.getDefaultConfig(prefix)
-  baseConfig.api.baseWsUrl = util.getRequiredEnv('WS_API_ENDPOINT', prefix)
-  const port = util.getEnv('RPC_PORT', prefix)
-  baseConfig.rpcPort = port ? parseInt(port) : DEFAULT_RPC_PORT
   baseConfig.defaultEndpoint = DEFAULT_ENDPOINT
   return {
     ...baseConfig,
-    isTLSEnabled: !!util.getEnv('IS_TLS_ENABLED', prefix),
+    httpOgmiosURL: util.getRequiredEnv('HTTP_OGMIOS_URL', prefix),
+    wsOgmiosURL: util.getRequiredEnv('WS_OGMIOS_URL', prefix),
```
boxhock, line 19: Can we make this fall back to the old env var, just in case? It should log an error about needing to update if that's the case.

boxhock: If we have good enough backwards compatibility, we could potentially get away with only doing a minor bump

boxhock (changes requested):

packages/sources/ada-balance/src/config.ts
boxhock: Yeah, error out if none of the env vars are set (we're not able to get an HTTP or WS URL for Ogimos)

boxhock (approved):

PR #1460, opened 2022-01-18
Add versions script for printing all package versions

justinkaseman:
## Description
When posting EA releases it is important to mention which EA version to migrate to.
Additionally it may be useful to have an API that returns the latest version of all EAs.

This PR adds a script that prints tables of the latest EA versions.
The `composite` `source` and `target` versions should be posted whenever there is a core change, as they will all have their versions bumped.

The next steps might be to have a similar script that will diff against the latest version on ECR to make a "from -> to" table.

## Changes
Run `yarn versions`


boxhock (approved):
Nice :+1:

PR #1466, opened 2022-01-18
Revert tradermade batch endpoint usage

justinkaseman:


austinborn (approved):

packages/sources/tradermade/src/endpoint/batch.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export const inputParameters: InputParameters = {
+  base: {
+    aliases: ['from', 'symbol', 'market'],
+    required: true,
+    description: 'The symbol of the currency to query',
+  },
+  quote: {
+    aliases: ['to', 'market', 'convert'],
+    required: true,
```
austinborn, line 28: Can't have same alias "market" on both input params

austinborn (changes requested):

packages/sources/tradermade/src/endpoint/batch.ts
justinkaseman: Woah, nice catch.

austinborn: Must've missed it the first time though...

justinkaseman: Yeah that has been in there for ~4 months :sweat_smile:

justinkaseman: https://app.shortcut.com/chainlinklabs/story/26519/unit-test-for-duplicate-input-parameter-aliases

austinborn (approved):

austinborn (approved):

PR #1484, opened 2022-01-21
Fix dxfeed result path

alejoberardino:
## Description

Fixes result path building in dxfeed adapter

## Quality Assurance


austinborn (approved):
Love it!

packages/sources/dxfeed/src/endpoint/price.ts
```diff
@@ -82,10 +88,7 @@ const handleBatchedRequest = (
           base: response.data[events][base],
         },
       },
-      Requester.validateResultNumber(
-        response.data,
-        isArray ? [events, base, 0, 'price'] : [events, base, 'price'],
-      ),
+      Requester.validateResultNumber(response.data, getResultPath(events as Events, base, isArray)),
```
justinkaseman, line 91: Would turn into `resultPath` from the Validator data

packages/sources/dxfeed/src/endpoint/price.ts
```diff
@@ -56,15 +58,19 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
 
   // NOTE: may need to force entries quoteEventSymbols to not use batching
 
-  const quotePath = ['Quote', symbol, 'bidPrice']
-  const tradePath = ['Trade', symbol, 'price']
   response.data.result = Requester.validateResultNumber(
     response.data,
-    events === 'Quote' ? quotePath : tradePath,
+    getResultPath(events, symbol),
```
justinkaseman, line 63: Would turn into `resultPath` from the Validator data

packages/sources/dxfeed/src/endpoint/price.ts
```diff
@@ -56,15 +58,19 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
   )
   return Requester.success(jobRunID, response, config.verbose, batchablePropertyPath)
 }
 
+const getResultPath = (events: Events, symbol: string, isArray = false): Array<string | number> => {
```
justinkaseman, line 68: What if we turned this into the way that other adapters use resultPath from `builder.ts` by exporting `endpointResultPaths`?
It accepts a function to handle dynamic resultPaths like this.

https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/coingecko/src/endpoint/crypto.ts#L27
shows how to export it by endpoint, so using a similar pattern of a factory function this one would look like:
```
export const endpointResultPaths: {
  [endpoint: string]: ReturnType<typeof getResultPath>
} = {
  price: getResultPath(),
  crypto: getResultPath(),
  stock: getResultPath(),
  forex: getResultPath(),
  commodities: getResultPath(),
}
```

`getResultPath` would turn into something like:
```
const getResultPath = () => (request: AdapterRequest) => {
  const validator = new Validator(request, inputParameters)
  const base = validator.validated.data.base
 const events = quoteEventSymbols[symbol] ? 'Quote' : 'Trade' (could abstract this into its own function)
  if (Array.isArray(base)) return
  return [events, symbol, events === 'Quote' ? 'bidPrice' : 'price']
}
```

justinkaseman: The benefit of this is that the core framework has access to the `resultPath`, which will be useful once we start caching full data responses.

packages/sources/dxfeed/src/endpoint/price.ts
```diff
@@ -56,15 +58,19 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
...
+  const path: Array<string | number> = [events, symbol, events === 'Quote' ? 'bidPrice' : 'price']
+  if (isArray) path.splice(2, 0, 0)
```
justinkaseman, line 70: Dunno about you guys but I find this hard to read.
I suggested a different style in the above comment. How do you feel about something like:
```
  if (Array.isArray(base)) return [events, base, 0, path]
  return [events, base, path]
```
I find this a lot more intuitive.

packages/sources/dxfeed/src/endpoint/price.ts
justinkaseman: This was already in there - but if the base is batched and included `'USO/USD:AFX'` it would never be picked up by `quoteEventSymbols` because it will key in with all of the comma delimited bases.

That's one reason to not pass through `events` into handleBatchedRequest and instead use this new `getResultPath` repeatedly.

justinkaseman:
Needs a changeset as well :socks:

packages/sources/dxfeed/src/endpoint/price.ts
alejoberardino: Agreed :+1:

packages/sources/dxfeed/src/endpoint/price.ts
alejoberardino: I've addressed the other comments, but didn't make these changes, I'm not fully understanding the desired outcome of this

packages/sources/dxfeed/src/endpoint/price.ts
austinborn: Yeah, that looks better IMO

justinkaseman (approved):
Exactly how I was thinking, thanks for making those changes.

PR #1488, opened 2022-01-24
Feature/sc 17228 websocket integration fw

alejoberardino:
## Description

Nock does not provide functionality to mock websockets. This PR adds a few helper libraries, which along the use of [mock-socket](https://www.npmjs.com/package/mock-socket) allows for intercepting a WebSocket connection and its exchange of messages.

## Changes

- Adds helpers to mock WS adapters

## Quality Assurance


CONTRIBUTING.md
```diff
@@ -100,6 +100,14 @@ The follow steps is the general pattern for writing an integration test.
...
+
+1. Run your tests, using live API endpoints with recording on (`export RECORD=true`)
+2. You will see a log statement with the message "Recorded WebSocketMessages: {JSON Object}". This JSON object contains all the WebSocket messages sent and received, but they are not printed as code as in the case of the Nock features, due to their asynchronous nature.
+3. Using the recorded messages, write a `fixtures.ts` file with "Exchanges", i.e. request and response(s) pairs that will be asserted as part of your test (see this [ncfx test fixtures example](./packages/sources/ncfx/test/integration/fixtures.ts)).
+4. Write your tests (example in [ncfx adapter test](./packages/sources/ncfx/test/integration/adapter.test.ts)) using the helper functions from the `test-helpers` package. Note the necessary setup performed in the `beforeAll` function.
+5. Finally, run your tests with recording disabled (`unset RECORD`). The WebSocket connection should be replaced and mocked.
+
 For more information on Jest, see the [Jest docs](https://jestjs.io/docs/cli).
```
austinborn, line 111: I would turn this and the previous set of steps into sub-sections:
```
### Testing HTTP Requests
1. ...
2. ...
...
### Testing Websocket Messages
1. ...
2. ...
...
```

packages/core/bootstrap/src/lib/ws/recorder.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class WsMessageRecorder {
+  private static messages: Array<WsMessage> = []
+
+  static add(message: WsMessage): void {
+    this.messages.push(message)
+  }
+
+  static print(): void {
+    console.log(`Recorded WebSocketMessages: ${JSON.stringify(this.messages)}`)
```
austinborn, line 16: Can you use the logger here?
```
import { logger } from '../external-adapter'
```

packages/core/test-helpers/src/websocket.ts
```diff
@@ -0,0 +1,65 @@
...
+ * We need this here, because the tests will connect using their instance of WebSocketClassProvider;
+ * fetching from this library to the @chainlink/ea-bootstrap package would access _another_ instance
+ * of the same constructor. Although it should be a singleton, dependencies are different so that
+ * means that the static classes themselves are also different.
+ *
+ * @param provider singleton WebSocketClassProvider
+ */
+export const mockWebSocketProvider = (provider: typeof WebSocketClassProvider): void => {
+  // Need to disable typing, the mock-socket impl does not implement the ws interface fully
+  provider.set(WebSocket as any) // eslint-disable-line @typescript-eslint/no-explicit-any
```
austinborn, line 15: Could you do something like this?
```
import { Server, WebSocket as MockWebSocket } from 'mock-socket'
import WebSocket from 'ws'
...
provider.set(MockWebSocket as WebSocket)
```
We really shouldn't be using `any` moving forward.

austinborn (commented):
Looks great for the most part, just some minor comments.

packages/core/bootstrap/src/lib/ws/recorder.ts
alejoberardino: The logger will print the inner json payload escaping strings, so I went with `console.log` to make it easier to copy and paste from the console to fixtures

packages/core/test-helpers/src/websocket.ts
alejoberardino: Unfortunately the typing issue lies in the incompatibility of the types specified by the rxjs library and the Node websocket, seen originally [here](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/ws/epics.ts#L183). Additionally, the WebSocket impl from the mock not being a full impl from the WebSocket interface, makes it so again the types have to be manually coerced. It's a bit of an unfortunate situation, because ideally all libraries would have correct types, but it's not up to us in this case.

austinborn: Yeah, makes sense. I guess it's only for the mock testing, so it's not too bad...

austinborn (approved):

PR #1492, opened 2022-01-25
chore: add granular status code

### LINKED ISSUE #1454: Granular Status Codes
Instead of just throwing an error with 400 or 500, we should have separate status codes for different known errors.

For example:
- If something times out we should know if it was the adapter didn't process it in time or if the request hung in flight to the data provider.
- If a data provider returns `null` data that we can't handle we should have a status code specifically for this scenario

bogos:
## Description

Add granular status codes

## Quality Assurance

- [X ] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)

### DIFF
--- a/packages/core/bootstrap/src/lib/external-adapter/requester.ts
+++ b/packages/core/bootstrap/src/lib/external-adapter/requester.ts
@@ -52,6 +52,19 @@ export class Requester {
         response = await axios(config)
       } catch (error) {
         // Request error
+        if (error.code === 'ECONNABORTED') {
+          // axios timeout code
+          throw new AdapterError({
+            statusCode: 200,
+            name: 'Request Timeout error',
+            providerStatusCode: error?.response?.status ?? 408,
+            message: error?.message,
+            cause: error,
+            errorResponse: error?.response?.data?.error,
+            url,
+          })
+        }
+
         if (n === 1) {
           throw new AdapterError({
             statusCode: 200,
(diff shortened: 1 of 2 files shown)


packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -52,6 +52,19 @@ export class Requester {
...
+        if (error.code === 'ECONNABORTED') {
+          // Axios timeout code
+          throw new AdapterError({
+            statusCode: 408,
+            name: 'Request Timeout error',
+            providerStatusCode: error?.response?.status ?? 408,
+            message: error?.message,
+            cause: error,
+            errorResponse: error?.response?.data?.error,
+            url,
```
boxhock, line 64: This should be a 504 instead.

From Wikipedia: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

> 408 Request Timeout
> The server timed out waiting for the request. According to HTTP specifications: "The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time."

> 504 Gateway Timeout
> The server was acting as a gateway or proxy and did not receive a timely response from the upstream server.

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -103,15 +116,23 @@ export class Requester {
...
-    if (typeof result === 'undefined') {
-      const message = 'Result could not be found in path'
+
+    if (typeof result === 'undefined' || result === null) {
+      const message = 'Result could not be found in path or is empty'
       logger.error(message, { data, path })
-      throw new AdapterError({ message })
+      throw new AdapterError({
+        message,
+        statusCode: 404,
```
boxhock, line 125: This should probably be a 502:

> 502 Bad Gateway
> The server was acting as a gateway or proxy and received an invalid response from the upstream server.

packages/core/bootstrap/src/lib/external-adapter/requester.ts
```diff
@@ -103,15 +116,23 @@ export class Requester {
...
+      })
     }
+
     if (Number(result) === 0 || isNaN(Number(result))) {
       const message = 'Invalid result received'
       logger.error(message, { data, path })
-      throw new AdapterError({ message })
+      throw new AdapterError({
+        message,
+        statusCode: 400,
```
boxhock, line 134: I also feel like this should be a 502.

boxhock (changes requested):
I'm not sure what you and @justinkaseman have discussed already, but I feel like the status code numbers currently used can be misleading, as they already have a specified use that conflicts with how we're using them here.

My requested change includes status code numbers that have a specification more fitting to how we're using them, but gives us less granularity.

@justinkaseman @bogos curious to hear both of your opinions.

eddiejhong:
@bogos Any updates on this PR?

boxhock (approved):

PR #1516, opened 2022-02-01
update rate limits for coinpaprika

cds95:


packages/core/bootstrap/src/lib/provider-limits/limits.json
```diff
@@ -378,10 +378,22 @@
   "coinpaprika": {
     "http": {
       "free": {
-        "rateLimit1s": 10,
-        "rateLimit1m": 600,
-        "rateLimit1h": 34246.5,
-        "note": "600/m and 25M/mo for free"
+        "rateLimit1s": 0.0193,
```
cds95, line 381: Not sure if this is correct but I just converted the # calls / month to per second, per minute and per hour.  There was no information given about per second, per minute and per hour limits.  The numbers do seem very low to me though.

coinpaprika

boxhock: Agreed, these limits are really strict. As long as they don't explicitly limit per second, per minute, we should not set that. We should only set per hour limits.

boxhock (approved):

PR #1519, opened 2022-02-01
Export endpoint descriptions from READMEs

austinborn:
## Description
Add description as exported attribute for endpoints with README details.

## Changes
- Add `export const description = '...'` to endpoint files
- Remove descriptions from `schema/env.json` in cases where the top-level description was taken from specific endpoint descriptions (since there wasn't an alternative at the time)

## Steps to Test
Generic tests (`yarn test:unit`, `yarn test:integration`)

## Quality Assurance


packages/sources/1forge/src/endpoint/quotes.ts
```diff
@@ -11,6 +11,9 @@ import { NAME as AdapterName } from '../config'
 export const supportedEndpoints = ['quotes', 'forex', 'price']
 export const batchablePropertyPath = [{ name: 'base' }, { name: 'quote' }]
 
+export const description =
+  'Returns a batched price comparison from a list currencies to a list of other currencies.\n\n[`/quotes`](https://1forge.com/api#quotes) - Convert from one currency to another.\n\n**NOTE: the `price` endpoint is temporarily still supported, however, is being deprecated. Please use the `quotes` endpoint instead.**'
+
```
alejoberardino, line 16: I'd recommend using backticks here and other descriptions with newlines, it'd make it cleaner imo

austinborn: How would you format this one for example? Here's my attempt at it:
```suggestion
export const description =
`Returns a batched price comparison from a list currencies to a list of other currencies.

[\`/quotes\`](https://1forge.com/api#quotes) - Convert from one currency to another.
**NOTE: the \`price\` endpoint is temporarily still supported, however, is being deprecated. Please use the \`quotes\` endpoint instead.**`
```

alejoberardino: Yeah, that's great imo 🙂

austinborn: Ok, all descriptions that used newline characters have been updated with this pattern

alejoberardino (approved):

PR #1539, opened 2022-02-08
Mock nop olympics ea

cds95:
This is a mock EA that the integration team will use to soak test.  All it does is return a mock response that changes after a fixed interval.  This interval is configurable through an environment variable.


packages/sources/mock-ea/src/endpoint/price.ts
```diff
@@ -0,0 +1,45 @@
...
+export const execute: ExecuteWithConfig<ExtendedConfig> = async (request, _, config) => {
+  const validator = new Validator(request, inputParameters)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const lastUpdated = responseInfo.lastUpdated
+  const now = Date.now()
+  if (!lastUpdated || now - lastUpdated >= config.updateIntervalInMS) {
+    responseInfo.result =
+      now % 2 === 0
```
boxhock, line 33: This won't guarantee that the result will change in the last x interval, but that it will change 50% of the time we hit the update interval.

boxhock: Could consider doing a percentage change instead?

packages/sources/mock-ea/src/endpoint/price.ts
```diff
@@ -0,0 +1,45 @@
...
+const responseInfo: ResponseInfo = {
+  result: 1000,
+}
+
+export const execute: ExecuteWithConfig<ExtendedConfig> = async (request, _, config) => {
+  const validator = new Validator(request, inputParameters)
+  if (validator.error) throw validator.error
+
+  const jobRunID = validator.validated.id
+  const lastUpdated = responseInfo.lastUpdated
```
boxhock, line 29: Will this ever be set?

packages/sources/mock-ea/src/endpoint/price.ts
cds95: hmm I'm not sure I follow.  Won't the value go up or down?

boxhock: `now % 2 === 0` will only be true half the time, right? So when we hit `updateIntervalInMS`, we have a 50% chance of the result updating or not.

packages/sources/mock-ea/src/endpoint/price.ts
boxhock: But it won't ever be set _before_ the `if` statement, right? Just wondering if this check actually does something since it's not set before we're inside the if statement.

packages/sources/mock-ea/src/endpoint/price.ts
cds95: right half the time we increment the result and the other half we decrement the result

packages/sources/mock-ea/src/endpoint/price.ts
cds95: so this check will run when a request comes in and it'll get set for the first time after the first request.  Subsequent requests then check against it and update the result accordingly.  This seems to work when as the results do vary when I tried it locally.

boxhock: Ooooh because it's a global variable -- I see now

cds95: yup exactly

boxhock (approved):

PR #1547, opened 2022-02-11
Update Linear Finance EA to pull data from APIs instead of CSV files

cds95:
## Description

This PR updates the Linear Finance EA to pull data from APIs instead of the manually uploaded CSV files.  It fetches the data from `https://pro-api.xangle.io/v1/index/xangle-bluechip` for bluechip data and from `https://pro-api.xangle.io/v1/index/xangle-largecap ` for largecap data.  Additionally NOPs should run the EA with the settings below to ensure up time if the API goes down.

```
WARMUP_UNHEALTHY_THRESHOLD=-1
```

## Changes

- Remove code to parse csv files
- Update Linear Finance EA to fetch data from endpoints above
- Update EA to require a `API_KEY` environment variable to be set

## Steps to Test

1. Set required environment variables and run EA.  Make sure to set a source url so that the Token Allocation EA can run as well as the Linear Finance EA depends on it.  Refer to the Token Allocation's README for instructions on what to set.
2. Send both requests below one after another and verify that both are successful

XBCI
```
{
    "id": "1",
    "data": {
        "index": "xbci",
        "source": "coinmarketcap"
    }
}
```

XLCI
```
{
    "id": "1",
    "data": {
        "index": "xlci",
        "source": "coinmarketcap"
    }
}
```

## Quality Assurance


packages/composites/linear-finance/src/tokenAllocationDeriver.ts
```diff
@@ -0,0 +1,46 @@
...
+    portfolio_reference_utc: string
+    portfolio: Portfolio[]
+  }
+}
+
+export const deriveAllocations = async (
+  config: Config,
+  path: string,
+): Promise<types.TokenAllocations> => {
+  const now = moment().tz('GMT').format('YYYY-MM-DD[T]hh:mm:ss')
```
cds95, line 31: Not sure if this is correct.  I need to verify this requirement.

packages/composites/linear-finance/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -81,13 +67,6 @@ Object {
           },
         },
       },
-      "CELO": Object {
```
cds95, line 84: Not sure if this is intentional but these values are not returned by the API but were previously present in the CSV files.  Are they supposed to be time dependent? @boxhock

boxhock: They were probably in a previous version of the index, I'm guessing

packages/composites/linear-finance/src/tokenAllocationDeriver.ts
austinborn: Yeah be careful with how the timezone is utilized. I've had lots of trouble with timezones in the past...

austinborn (dismissed):

packages/composites/linear-finance/src/tokenAllocationDeriver.ts
cds95: @boxhock what time should we pass to Linear Finance's API?  Currently I'm just setting whatever time it is in GMT to standardize across all NOPs.

boxhock: Let's please confirm with the SA on this

.changeset/itchy-files-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/linear-finance-adapter': minor
```
boxhock, line 2: We make a pretty significant change from reading local allocations to requiring an API key for an external call. Do you think it makes sense to make this a major bump instead?

packages/composites/linear-finance/src/tokenAllocationDeriver.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  return new Promise((resolve, reject) => {
+    const middleware = makeMiddleware(execute, undefined, endpointSelector)
+    withMiddleware(execute, context, middleware)
+      .then((executeWithMiddleware) => {
+        executeWithMiddleware(options, context)
+          .then((value) => resolve(value.data.allocations))
+          .catch(reject)
+      })
+      .catch((error) => reject(error))
```
boxhock, line 29: Can do like this or like the above, just would like to keep it consistent.

```suggestion
      .catch(reject)
```

boxhock (changes requested):
Just noticed these things, but looks good otherwise

.changeset/itchy-files-jam.md
cds95: hmm fair point.  Let's make it major.

.changeset/sixty-pumpkins-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/linear-finance-adapter': major
+---
+
+update ea to use api instead of reading from csv file
```
boxhock, line 5: Just thought about this as well. Could we explicitly call out what has changed, making it easier to do the release notes for this?

E.g.:
```suggestion
Update ea to use api instead of reading from csv file. NOTE: This version of the adapter requires an API key to be set.
```

boxhock (changes requested):
Seems there were some changes in `develop` you'll have to rebase with anyways, thought I'd add in this suggested change as well

boxhock (approved):

PR #1542, opened 2022-02-09
[Refactor] EA presets and config

karen-stepanyan:
## Changes

This PR modifies every adapter and core with following changes. No functional changes, only refactor

- For  all adapters. `config.ts` is removed. New folder is created called `config`. `index.ts` inside config folder contains the content of original `config.ts`
- `presetSymbols.json` from `core` is removed. Each adapter will now have only its own symbol overrides defined in `symbols.json` inside config folder
- `presetIncludes.json` from `core` is removed. Each adapter will now have only its own include overrides defined in `includes.json` inside config folder
- `validator.ts` is modified. `ValidatorOptions` now accepts `includes` and `overrides` as options. If an adapter has symbol overrides or includes those should be provided to validator. for example `new Validator(request, inputParameters, {}, { includes, overrides })` where `includes, overrides` were imported from config folder
- `limits.json` from `core` is removed. Each adapter will now have only its own rate limits defined in `limits.json` inside config folder
- `expose` function from core/bootstrap/src/index.ts was modified. Now the first parameter is of type AdapterContext instead of string `name`
- top level `index.ts` for each adapter was modified to provide AdapterContext to `expose` function. for example 
`expose(adapterContext, makeExecute(), undefined, endpointSelector)` where adapterContext is `{ name: NAME }` or can be `{name: NAME, rateLimit}` where rateLimit is the content of rate limits from config/limits.json

 

## Quality Assurance

- [x] Ran `yarn changeset` if adapter source code was changed
- [x] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `<ADAPTER_PACKAGE>/schemas/env.json` and `<ADAPTER_PACKAGE>/README.md`
- [x] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.
- [x] The branch naming follows git flow (`feature/x`, `chore/x`, `release/x`, `hotfix/x`, `fix/x`) or is created from Clubhouse/Shortcut
- [x] This is related to a maximum of one Clubhouse/Shortcut story or GitHub issue
- [x] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)

PR #1551, opened 2022-02-11
[Refactor] EA presets and config for kaiko

karen-stepanyan:
## Changes

This PR modifies kaiko EA and core with following changes. No functional changes, only refactor

- For  kaiko adapter. `config.ts` is removed. New folder is created called `config`. `index.ts` inside config folder contains the content of original `config.ts`
- `presetSymbols.json` from `core` is removed. Each adapter will now have only its own symbol overrides defined in `symbols.json` inside config folder
- `presetIncludes.json` from `core` is removed. Each adapter will now have only its own include overrides defined in `includes.json` inside config folder
- `validator.ts` is modified. `ValidatorOptions` now accepts `includes` and `overrides` as options. If an adapter has symbol overrides or includes those should be provided to validator. for example `new Validator(request, inputParameters, {}, { includes, overrides })` where `includes, overrides` were imported from config folder
- `limits.json` from `core` is removed. Each adapter will now have only its own rate limits defined in `limits.json` inside config folder
- `expose` function from core/bootstrap/src/index.ts was modified. Now the first parameter is of type AdapterContext instead of string `name`
- top level `index.ts` for each adapter was modified to provide AdapterContext to `expose` function. for example 
`expose(adapterContext, makeExecute(), undefined, endpointSelector)` where adapterContext is `{ name: NAME }` or can be `{name: NAME, rateLimit}` where rateLimit is the content of rate limits from config/limits.json


karen-stepanyan:
Closing because of this [PR](https://github.com/smartcontractkit/external-adapters-js/pull/1542)

PR #1554, opened 2022-02-13
Safe getFeedId and data.includes spreading

austinborn:
## Description

If `getFeedId` throws an error in a `catch` block, the process throws an unhandled error which causes node to exit. Back on node v14, it would log an error message and continue operation, but since the [upgrade to node v17](https://github.com/smartcontractkit/external-adapters-js/pull/1219), unhandled exceptions simply cause the process to crash and exit. There are likely other areas where outside of this `try-catch` which will also need to be addressed (like anywhere `createReducer` is used), but this is a start.

## Changes

- Ensure `includes` is an array before attempting to use it in `validateIncludeOverrides`
- Add `try-catch` inside `getFeedId` since this method is sometimes called within `catch` blocks

## Steps to Test

1. Run `kaiko` adapter using node v17.
2. Query the adapter with `{"jobRunId": 1, "data": { "base": "ETH", "quote": "USD", "includes": {} } }`. Expect that an error message will be logged, but the adapter should respond with an error and continue operation.

### DIFF
--- a/packages/core/bootstrap/src/lib/metrics/util.ts
+++ b/packages/core/bootstrap/src/lib/metrics/util.ts
@@ -65,51 +65,58 @@ function buildSymbolString(data: string | string[]): string {
  * @returns {string}
  */
 export const getFeedId = (input: AdapterRequest): string => {
-  const commonFeedParams = {
-    base: ['base', 'from', 'coin', 'symbol', 'asset'],
-    quote: ['quote', 'to', 'convert'],
-  }
-
-  // check if string is within array
-  function includesCheck(param: string) {
-    return Object.keys(input.data).includes(param)
-  }
-
-  /**
-   * If the request is coming from the cache warmer, use a fixed id. This is to reduce the
-   * cardinality of the feed_id label in prometheus, which can be overloaded quickly
-   */
-  if (input.debug?.warmer) {
-    return WARMER_FEED_ID
-  }
-
-  // run through validator if input.data object has keys that match potential base and quote parameters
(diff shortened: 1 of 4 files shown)


packages/core/bootstrap/src/lib/metrics/util.ts
```diff
@@ -65,51 +65,58 @@ function buildSymbolString(data: string | string[]): string {
...
-  const entries = Object.keys(input)
-    .filter((prop) => !excludableAdapterRequestProperties[prop])
-    .map((k) => [k, input[k as keyof AdapterRequest]])
 
-  const rawFeedId = JSON.stringify(Object.fromEntries(entries))
+    // run through validator if input.data object has keys that match potential base and quote parameters
+    if (commonFeedParams.base.some(includesCheck) && commonFeedParams.quote.some(includesCheck)) {
+      const validationResult = new Validator(input, commonFeedParams)
+      if (validationResult.error) {
+        logger.debug('Unable to validate feed name')
```
cds95, line 89: could be good to log out the param being checked here as well

cds95 (commented):
LGTM.  Just had one suggestion.

packages/core/bootstrap/src/lib/metrics/util.ts
austinborn: The `logger` here now looks like:
```
logger.debug('Unable to validate feed name', {
          input,
          error: validationResult.error.toString(),
        })
```

cds95 (approved):

justinkaseman (approved):

PR #1572, opened 2022-02-16
Upgrade all node versions to 17.x

austinborn:
## Description
Since we're upgrading the `node` version when running the adapters, we should upgrade `node` for other processes as well.

## Changes
- Use Node 17.x for all github actions/workflows that previously used Node 14.x
- Use `NODE_OPTIONS=--openssl-legacy-provider` in a number of areas so that renvm-address-set does not throw errors like:
```
Error: error:0308010C:digital envelope routines::unsupported
```
- Upgrade yarn to v3.1.1 so that errors like this don't occur during `yarn install --check-cache`:
```
➤ YN0001: │ Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at new NodeError (node:internal/errors:371:5)
    at PassThrough.onclose (node:internal/streams/end-of-stream:136:30)
    at PassThrough.emit (node:events:526:28)
    at emitCloseNT (node:internal/streams/destroy:145:10)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
```
- Add the new `.pnp.loader.mjs` (from upgrade to yarn 3.1.1) to `.prettierignore` to pass `yarn format:check`

## Quality Assurance


packages/core/bootstrap/test/unit/rate-limit-cache-tests/1-rate-limit-cache.test.ts
```diff
@@ -184,8 +184,8 @@ describe('Rate Limit/Cache - Integration', () => {
     const rlPerMinute = getRLTokenSpentPerMinute(state.rateLimit.heartbeats)
 
     Object.values(rlPerMinute).forEach((req) => {
-      // TODO: check that + 30 is the right capacity
-      expect(req).toBeLessThan(capacity + 30)
```
eddiejhong, line 188: Should this change be a part of this PR?

.pnp.loader.mjs
```diff
@@ -0,0 +1,249 @@
+import { URL, fileURLToPath, pathToFileURL } from 'url';
```
eddiejhong, line 1: Forgive my ignorance, but what is this code used for?

packages/core/bootstrap/test/unit/rate-limit-cache-tests/1-rate-limit-cache.test.ts
austinborn: It doesn't necessarily have to be a part of this, but this unit test failed at one point since `req` was `81` when `capacity + 30` was `80`. Probably worth investigating in its own right.

.pnp.loader.mjs
austinborn: Looks like this was a change from the latest `yarn` version upgrade. It is generated automatically when `yarn` detects that it may be needed, but it's still experimental. We can disable this by `pnpEnableEsmLoader: false` in the `.yarnrc.yml`: https://yarnpkg.com/configuration/yarnrc/#pnpEnableEsmLoader

austinborn: Here's an interesting article on [ESM Loaders](https://ar.al/2021/05/27/make-anything-a-javascript-module-using-node.js-esm-module-loaders/)

austinborn: I ended up removing it since it's unnecessary

justinkaseman (approved):
LGTM, I'd like this to soak all week to be confident in releasing it next week.

PR #1578, opened 2022-02-17
Refactored Terra RPC_URL environment variables

silaslenihan:
## Description
What is provided is actually a LCD url, we should reflect this in the name and documentation, while still supported the legacy name of RPC URL.
......

## Changes
Made slight changes to terra-view-controller adadpter and the documentation to prefer an LCD_URL environment variable over an RPC_URL one. 

## Steps to Test

1. Ensure the Adapter still works as normal using both LCD_URL and RPC_URL environment variables

## Quality Assurance

### DIFF
--- a/packages/sources/terra-view-function/src/config/index.ts
+++ b/packages/sources/terra-view-function/src/config/index.ts
@@ -11,10 +11,11 @@ export type ChainId = typeof SUPPORTED_CHAIN_IDS[number]
 export const ENV_DEFAULT_CHAIN_ID = 'DEFAULT_CHAIN_ID'
 export const DEFAULT_CHAIN_ID = 'columbus-5'
 
+export const ENV_LCD_URL = 'LCD_URL'
 export const ENV_RPC_URL = 'RPC_URL'
 
 export type Config = BaseConfig & {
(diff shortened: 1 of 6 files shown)


packages/sources/terra-view-function/src/config/index.ts
```diff
@@ -35,13 +36,19 @@ const buildRpcUrlMapping = () => {
...
+    } else if (legacyEnvVar) {
+      console.log(
+        'The RPC_URL environment variable is being deprecated, please use an LCD_URL instead. See README for more information.',
+      )
+      output[chainId] = legacyEnvVar
+      hasAtLeastOneURL = true
     }
   }
-
   if (!hasAtLeastOneURL) throw new Error('At least one RPC URL must be defined')
```
eddiejhong, line 52: We should also update this error message to "At least one LCD URL must be defined"

silaslenihan: Done, also there were some other variables I missed using rpc URL classifications that I fixed

justinkaseman (approved):
Nice work! Thanks. This will be a lot clearer for node operators.

PR #1579, opened 2022-02-17
Update Validator Input Options Behavior

### LINKED ISSUE #1531: InputParameters' "default" value used when an invalid "options" value is provided
Hi all!  I was wondering if I can get some insight for the behavior of the `default` property for the `InputParameters` type.  I have an input parameter initialized as follows:

```typescript
const inputParams = {
  item: {
    aliases: ['item'],
    required: false,
    type: 'string',
    default: 'first',
    options: ['first', 'second', 'third']
  }
}
```

If I do not provide a value for the `item` parameter, then it defaults to `first`.  If I specify any of the values in the `options` property, then that value is used instead; however, if I was to provide a value that isn't part of the `options` property, say `this should break`, it defaults to `first`.  Even though a default value has been specified, I assumed that the adapter would throw an error as a value has been provided that isn't part of the `options` property.  As an external user, if I wanted a `fourth` item, I would keep receiving back the `first` item.

With that, is the `default` value intended to be used in the event a user does not provide a value, or does it also play a dual role in being the default value if a wrong `options` value is provided?

austinborn:
## Description
In line 22 of [normalize/index.ts](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/middleware/normalize/index.ts#L22), the Validator runs without throwing errors and below that undefined values are removed. If an input is not in the options set then an error would be thrown, but due to `shouldThrowError: false` it just returns undefined and is removed. Then in a second pass to Validator  the value is set to the default.

## Changes

- Preserve input values when `shouldThrowError = false`
- Compare input options as lowercase values

## Steps to Test

1. Start an adapter with some input that has options and a default
2. If you send a request with some input not in the options list, the response should now return an error instead of using the default value
3. If you send a request with some input in the options list with different capitalization, the response should still succeed

## Quality Assurance

### DIFF
--- a/.changeset/angry-windows-chew.md
+++ b/.changeset/angry-windows-chew.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
+---
+
+Update Validator behavior with input options sets
--- a/packages/core/bootstrap/src/lib/modules/validator.ts
+++ b/packages/core/bootstrap/src/lib/modules/validator.ts
@@ -72,7 +72,7 @@ export class Validator {
             ? this.validateRequiredParam(this.input.data[key], key, options)
             : this.validateOptionalParam(this.input.data[key], key, options)
         } else {
-          this.validateObjectParam(key)
+          this.validateObjectParam(key, this.validatorOptions.shouldThrowError)
         }
       }
     } catch (e) {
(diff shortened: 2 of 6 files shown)


justinkaseman (approved):
Thanks for that! LGTM

PR #1629, opened 2022-03-02
Fix JPEG'd adapter to default to latest block

0xJPEG:
## Description

Instead of requiring a changing block number parameter, changes the JPEG'd source adapter to default to the latest value.

## Changes

- Makes block number parameter optional, defaults it to "latest"
- Updates tests

## Steps to Test

1. Run tests

## Quality Assurance

### DIFF
--- a/packages/composites/nftx/test/integration/fixtures.ts
+++ b/packages/composites/nftx/test/integration/fixtures.ts
@@ -1,42 +1,5 @@
 import nock from 'nock'
 
-export function mockAdapterResponseSuccess(): void {
-  nock('http://localhost:8080', { encodedQueryParams: true })
-    .post('/', { id: '1', data: { address: '0x269616D549D7e8Eaa82DFb17028d0B212D11232A' } })
-    .reply(
-      200,
-      {
-        jobRunID: '1',
-        statusCode: 200,
-        data: {
-          fee: '0.02',
-          price: '65.319915591679174162',
-          priceWithFee: '66.626313903512757645',
-        },
-      },
-      [
-        'X-Powered-By',
-        'Express',
-        'X-RateLimit-Limit',
-        '250',
-        'X-RateLimit-Remaining',
-        '249',
-        'Date',
-        'Tue, 22 Feb 2022 06:40:29 GMT',
-        'X-RateLimit-Reset',
-        '1645512035',
-        'Content-Type',
-        'application/json; charset=utf-8',
-        'Content-Length',
-        '126',
-        'ETag',
-        'W/"7e-lYTaizPLbvEPoDsBsih9JgqTY9c"',
-        'Connection',
-        'close',
-      ],
-    )
-}
-
 export function mockEthereumResponseSuccess(): void {
   nock('http://localhost:8545')
     .persist()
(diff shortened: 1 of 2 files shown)


packages/sources/jpegd/src/endpoint/punks.ts
```diff
@@ -11,10 +11,11 @@ export interface ResponseSchema {
 
 export const inputParameters: InputParameters = {
   block: {
-    required: true,
+    required: false,
     description: 'The block number for which information is being queried',
     aliases: ['blockNumber', 'blockNum'],
-    type: 'number',
+    type: 'string',
```
austinborn, line 17: If the input can be multiple types, you can just remove this line. We don't have a way to define multiple types, but it will not pass validation if the type doesn't match the value passed in.

packages/sources/jpegd/src/endpoint/punks.ts
```diff
@@ -26,7 +27,7 @@ export const execute: ExecuteWithConfig<Config> = async (request, _, config) =>
   const url = `/punks`
 
   const params = {
-    block: validator.validated.data.block,
+    block: validator.validated.data.block || 'latest',
```
austinborn, line 30: The validator will automatically set this to the default if present in `inputParameters`

austinborn (changes requested):
Looks good aside from two comments about the input param types

packages/sources/jpegd/src/endpoint/punks.ts
austinborn: So you can keep it as:
```suggestion
    block: validator.validated.data.block,
```

austinborn (dismissed):
Awesome, thank you!

packages/sources/jpegd/test/integration/fixtures.ts
```diff
@@ -1,27 +1,52 @@
...
+      'application/json',
+      'Transfer-Encoding',
+      'chunked',
+      'Connection',
+      'close',
+      'Vary',
+      'Accept-Encoding',
+    ])
+
+export const mockPunksValueResponseSuccess = (): nock =>
```
boxhock, line 20: This mock should be removed. Shouldn't mock what we're trying to test 😅

boxhock (changes requested):

packages/sources/jpegd/test/integration/fixtures.ts
austinborn: Good catch

austinborn: @0xJPEG  Sorry, he meant that the`localhost:8080` mock should be removed

0xJPEG: Sorry, fixed

boxhock (commented):
Changes look good now, but there is however a unit test that is failing. Should be good to merge once that's been resolved.

0xJPEG:
I pushed some changes to the tests alongside the mock changes, should be good now

0xJPEG:
Unnecessary validation error test removed as adapter now falls back to default value. Should be good now, I think.

PR #1641, opened 2022-03-04
18580 (PR 1) - URL Path Builder (method, unit tests, accuweather, amberdata, ap-election)

mjk90:
## Description
 Added buildUrlPath methods & unit tests to util. Updated the following adapters to use buildUrlPath: accuweather, amberdata, ap-election
......

## Steps to Test

1. yarn test:unit
2. yarn test:integration
3. Run an adapter with user input included in a url and check the URL it attempts to call.

## Quality Assurance


packages/core/bootstrap/src/lib/util.ts
```diff
@@ -373,3 +373,101 @@ export const getRequiredEnvWithFallback = (
...
+const charsToEncode = {
+  ':': '%3A',
+  '/': '%2F',
+  '?': '%3F',
+  '#': '%23',
+  '[': '%5B',
+  ']': '%5D',
+  '@': '%40',
+  '!': '%21',
+  $: '%24',
```
boxhock, line 388: Any reason this isn't in quotes when the others are?

```suggestion
  '$': '%24',
```

packages/core/bootstrap/test/unit/utils.test.ts
```diff
@@ -24,4 +24,102 @@ describe('utils', () => {
...
+
+    it(`builds path with non-latin characters`, () => {
+      const actual = buildUrlPath('/from/:from/to/:to', {
+        from: 'abcÂÃ',
+        to: '你好世界',
+      })
+      expect(actual).toEqual('/from/abc%C3%82%C3%83/to/%E4%BD%A0%E5%A5%BD%E4%B8%96%E7%95%8C')
+      expect(decodeURI(actual)).toEqual('/from/abcÂÃ/to/你好世界')
+    })
+  })
```
boxhock, line 84: Would be also great to see a test of something similar to this:

```typescript
buildUrlPath('/price/:from::to', {from: 'ETH', to: 'USD'}) // Expect /price/ETH:USD
```

boxhock (changes requested):
Great job! Looks good for the most part and seems like it will be a good improvement. Left a couple of comments.

packages/core/bootstrap/src/lib/util.ts
mjk90: It looks like the quotes are being removed by Prettier. I checked and it looks like the dollar sign is successfully used as a key, even without the quotes. I can investigate further if you think this could be an issue

austinborn: This would be why (apparently): https://mothereff.in/js-properties

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -373,3 +373,101 @@ export const getRequiredEnvWithFallback = (
...
+  '`': '%60',
+  '\\': '%5C',
+}
+
+/**
+ * Check whether the given string contains characters in the given whitelist.
+ * @param str The string to check.
+ * @param whitelist The string array of whitelist entries. Returns true if any of these are found in 'str', otherwise returns false.
+ * @returns {boolean}
+ */
```
austinborn, line 416: The JSDoc descriptions are a great touch 👍

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -373,3 +373,101 @@ export const getRequiredEnvWithFallback = (
...
+    const value = params[param as keyof typeof params]
+    if (!value) continue
+
+    // If string contains a whitelisted character: manually replace any non-whitelisted characters with percent encoded values. Otherwise, encode the string as usual.
+    const encodedValue = stringHasWhitelist(value, allowedChars)
+      ? percentEncodeString(value, allowedChars)
+      : encodeURIComponent(value)
+
+    if (pathTemplate.includes(`:${param}`))
+      pathTemplate = pathTemplate.replace(`:${param}`, encodedValue)
```
austinborn, line 456: I think you can move the `if` statement up to here (lmk if I'm missing something):
```suggestion
    const value = params[param as keyof typeof params]
    if (!value || !pathTemplate.includes(`:${param}`)) continue

    // If string contains a whitelisted character: manually replace any non-whitelisted characters with percent encoded values. Otherwise, encode the string as usual.
    const encodedValue = stringHasWhitelist(value, allowedChars)
      ? percentEncodeString(value, allowedChars)
      : encodeURIComponent(value)

      pathTemplate = pathTemplate.replace(`:${param}`, encodedValue)
```

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -373,3 +373,101 @@ export const getRequiredEnvWithFallback = (
...
+  const allowedChars = whitelist.split('')
+
+  for (const param in params) {
+    const value = params[param as keyof typeof params]
+    if (!value) continue
+
+    // If string contains a whitelisted character: manually replace any non-whitelisted characters with percent encoded values. Otherwise, encode the string as usual.
+    const encodedValue = stringHasWhitelist(value, allowedChars)
+      ? percentEncodeString(value, allowedChars)
+      : encodeURIComponent(value)
```
austinborn, line 453: Is there any particular benefit to using `encodeURIComponent` over `percentEncodeString`? If not, I would go straight to `const encodedValue = percentEncodeString(value, allowedChars)` to avoid having to run through the arrays multiple times.

mjk90: `percentEncodeString` is more limited than `encodeURIComponent`, in that it only replaces the characters found in `charsToEncode`, so if we use only `percentEncodeString`, then we will have the edge case of non-latin characters being skipped (this is handled by `encodeURIComponent`)

mjk90: If this is a real concern, then I could update `percentEncodeString` to be more comprehensive. Let me know what you think :-)

austinborn: Got it, that clears things up. No change needed then.

austinborn: > If this is a real concern, then I could update `percentEncodeString` to be more comprehensive. Let me know what you think :-)

No worries, I'm happy with it for now. The urls are relatively small so even if it has a sub-optimal run time, it shouldn't be noticeable.

packages/core/bootstrap/src/lib/util.ts
mjk90: You're right about the `if (pathTemplate.includes(':${param}'))`, but actually, I think it can probably be removed altogether rather than being moved up. That `if` was needed for an earlier version of the method but it shouldn't be needed at this point, since string.replace shouldn't do anything if the param isn't found

austinborn (approved):
Nice, lgtm!

boxhock (approved):
Looks good! I just merged in some changes to GH actions, so this PR will need a quick rebase 😅  Should be good to merge once rebased!

PR #1644, opened 2022-03-07
18580 (PR 2) - URL Path Builder (adapters B-C)

mjk90:
## Description
Updated the following adapters to use buildUrlPath: bitex, blockchain.com, blockchair, btc.com, coinapi, coinbase, coincodex, coinmetrics, coinpaprika, covidtracker, cryptoapis, cryptoapis-v2, cryptocompare, cryptomkt
......

## Steps to Test

1. yarn test:unit
2. yarn test:integration
3. Run an adapter with user input included in a url and check the URL it attempts to call.

## Quality Assurance


austinborn:
I think reviewers will have a hard time reviewing this until the first part is merged. Same for parts 3 and 4.

packages/sources/blockchair/src/endpoint/balance.ts
```diff
@@ -21,10 +21,13 @@ const getBalances: balance.GetBalances = async (accounts, config) => {
 
   const reqConfig: RequestConfig = {
     ...config.api,
-    url: getBalanceURI(addresses, coin as string, chain as string),
+    url: getBalanceURI(coin as string, chain as string),
```
boxhock, line 24: Are these `... as string` necessary? Do we not have types covered for them?

packages/sources/blockchair/src/endpoint/balance.ts
```diff
@@ -21,10 +21,13 @@ const getBalances: balance.GetBalances = async (accounts, config) => {
...
+    params: { addresses: addresses.join(',') },
   }
 
+  console.log({ reqConfig })
```
boxhock, line 28:
```suggestion
```

packages/sources/blockchair/src/endpoint/balance.ts
```diff
@@ -21,10 +21,13 @@ const getBalances: balance.GetBalances = async (accounts, config) => {
...
   const response = await Requester.request(reqConfig)
+  console.log({ response })
```
boxhock, line 30:
```suggestion
```

boxhock (changes requested):

packages/sources/blockchair/src/endpoint/balance.ts
mjk90: `coin` and `chain` come from the `Account` type, which defines them as optional, so if we don't cast them, they cause a typescript error. I could update this method to check if they're valid strings before doing anything else if you think this is an issue

boxhock: Hmm yeah I think we should do that. If they're not required, I guess we could just do `... || ''` instead?

austinborn: +1 for `... || ''`, or `... ?? ''` if the type is limited to `string | undefined | null`

boxhock (approved):

PR #1651, opened 2022-03-09
Catch unhandled promise rejections

alejoberardino:
## Description

When upgrading adapters to from Node version 14 to 17, behavior that would previously fail now additionally causes the process to exit. This PR adds a small global handler set up in the expose function of the core framework, to effectively simulate the old behavior without making changes to infrastructure.

## Changes

- Add global handler for unhandled promise rejections

## Steps to Test

1. Without these changes, add background process to any adapter, with a failing unhandled promise 
2. The rejection should cause the process to exit
3. Pull this PR, and apply the same changes as before
4. The error should still occur, but the process should stay alive

## Quality Assurance


packages/core/bootstrap/src/lib/util.ts
```diff
@@ -470,3 +471,22 @@ export const buildUrl = (baseUrl: string, pathTemplate = '', params = {}, whitel
   new URL(buildUrlPath(pathTemplate, params, whitelist), baseUrl).toString()
 
 //  URL Encoding
```
boxhock, line 473: Is this comment relevant to the section below? Just looks like it didn't belong here

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -470,3 +471,22 @@ export const buildUrl = (baseUrl: string, pathTemplate = '', params = {}, whitel
...
+
+let unhandledRejectionHandlerRegistered = false
+/**
+ * Adapters use to run with Node 14, which by default didn't crash when a rejected promised bubbled up to the top.
+ * This function registers a global handler to catch those rejections and just log them to console.
+ */
+export function registerUnhandledRejectionHandler(): void {
```
boxhock, line 480: Nitpick, but just for style consistency

```suggestion
let unhandledRejectionHandlerRegistered = false

/**
 * Adapters use to run with Node 14, which by default didn't crash when a rejected promised bubbled up to the top.
 * This function registers a global handler to catch those rejections and just log them to console.
 */
export const registerUnhandledRejectionHandler = (): void => {
```

packages/core/bootstrap/src/lib/util.ts
alejoberardino: Seems to be the "end" tag of a begin-end comment pair, I'd probably refactor it and extract it out into its own file rather than comment, but I'll leave that for another PR so as not to clutter this one

boxhock (approved):

PR #1671, opened 2022-03-14
Return Anchor EA price to 8dp

cds95:


.changeset/purple-adults-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/anchor-adapter': minor
+---
+
+allow the number of decimals for answer to be configurable with a default of 8dp
```
boxhock, line 5: This should normally be considered a breaking change. Are we covered in all ways to ensure this wouldn't be a problem?

It shouldn't be a problem if the previous version hasn't been published to the public ECR, but I'm also wondering if we should just make this a major version bump, just to be on the safe side?

cds95: yeah I'm fine with either option.  It isn't a breaking change as the previous version has not gone out yet.

cds95: made this major to be on the safe side

boxhock (approved):

PR #1688, opened 2022-03-17
fix beth calculations

cds95:
Fix below issues with the `bETH` calculations

1) The result from calling the Anchor Vault contract's `get_rate` function returns the `bETH/stETH` rate.  The EA was previously using this value incorrectly as it assumed that the rate was `stETH/bETH`.

2) The `get_dy` function on the `stETH-ETH` contract is currently called with the parameters `get_dy(1, 0, 10**18)` which returns the `ETH/stETH` rate.  This was used incorrectly as the calculations used it as `stETH/ETH`.  The code for this was confusing so I named the variables a bit better and added a few more comments.  The parameters that have been passed in have not been changed as we are still interested in the `ETH/stETH` rate.  Instead the calculation in the `bETH`'s `execute` function has been updated to correctly use this value.


boxhock (dismissed):
Approving the code changes, but would like someone who's more familiar with the calculations to also review before merging this.

ainsleybc (approved):
The calculations look as intended by the code & comments 👍

packages/composites/anchor/src/endpoint/price/beth.ts
```diff
@@ -43,24 +43,32 @@ const getStETHExchangeRate = async (
...
+  const result = await stEthPoolContract.get_dy(
+    baseTokenIndex,
+    quoteTokenIndex,
+    ethers.BigNumber.from(10).pow(18),
+  )
   throwErrorForInvalidResult(
     jobRunID,
     result,
-    `stETH/ETH Exchange Rate from Curve Pool address ${config.stEthPoolContractAddress}`,
+    `ETH/stETH Exchange Rate from Curve Pool address ${config.stEthPoolContractAddress}`,
```
thodges-gh, line 58: This comment was correct. Base = stETH; Quote = ETH.

thodges-gh (changes requested):

packages/composites/anchor/src/endpoint/price/beth.ts
cds95: reverted the comment

packages/composites/anchor/src/endpoint/price/beth.ts
```diff
@@ -43,24 +43,32 @@ const getStETHExchangeRate = async (
...
- * getStEthBEthExchangeRate returns a promise for the value of stETH/bETH
- * from the stETH/bETH Anchor Vault contract
+ * getBEthStETHExchangeRate returns a promise for the value of bEth/StETH
+ * from the bEth/StETH Anchor Vault contract
  * @param jobRunID string
  * @param config Config
  * @param provider ethers.providers.JsonRpcProvider
  * @returns Promise<ethers.BigNumber>
  */
-const getStEthBEthExchangeRate = async (
```
eddiejhong, line 63: old: stETH/bETH exchange rate
new: bEth/stETH exchange rate

Is there a reason for changing the name of the function if we're not changing the calculation? If we're only changing the name, then that means we're not truly making the correct adjustment and the math is incorrect.

cds95: actually looking at this again `bETH/stETH` is correct.  The previous version is wrong.

boxhock (approved):

boxhock (dismissed):

thodges-gh (dismissed):

austinborn (approved):

PR #1695, opened 2022-03-21
Feature/ea env var context

austinborn:
## Description
Add `baseEnvDefaults` to consolidate env var defaults, and extend adapter context to support default overrides with `envDefaultOverrides`. If an env var is defined then that value is used, but if not then the adapter override default is used, followed by the base env default if no override is given. Override defaults have been added for `ncfx` and `onchain-gas` for the `WS_ENABLED` env var as an example.

## Changes

- Consolidate env var defaults in `baseEnvDefaults`
- Replace all `process.env` usage with `getEnv` for consistency
- Add `envDefaultOverrides` to adapter context passed into `expose` to allow adapter-specific env var default overrides
- Coalesce the env var base and adapter override defaults in `getEnv`, with adapter override defaults taking precedent over base defaults
- Add default overrides for `WS_ENABLED` in `ncfx` and `onchain-gas` adapters to demonstrate usage
- Add default override for `CACHE_ENABLED` in `layer2-sequencer-health` adapter

## Steps to Test
1. Comment out `WS_ENABLED: 'true',` from `packages/sources/ncfx/src/config/envDefaultOverrides.ts`.
2. Export proper API_USERNAME and API_PASSWORD for the `ncfx` adapter.
3. Build and start the `ncfx` adapter.
4. Run a request like `{"jobRunId": 1, "data": {"base": "ETH", "quote": "USD"}}` and verify the result is a `500` error with the following message:
```
"The NCFX adapter does not support making HTTP requests. Make sure WS is enabled in the adapter configuration."
 ```
5. Uncomment the `WS_ENABLED: 'true',` line.
6. Re-build and re-start the `ncfx` adapter.
7. Run the same request and verify the result is a successful response.

## Quality Assurance

### CODE REVIEW

File: packages/core/types/@chainlink/index.d.ts [hunk 1]
```diff
@@ -130,6 +131,18 @@ declare module '@chainlink/types' {
...
+  export type EnvDefaults = { [name: string]: string }
+
+  /**
+   * To add a default override for an env var not present here,
+   * ensure that the adapter context from `expose` is passed
+   * to the `getEnv` call for the variable in question. See
+   * `WS_ENABLED` for example usage.
+   */
+  export type EnvDefaultOverrides = {
+    WS_ENABLED?: 'true' | 'false'
```
justinkaseman, line 143: What if we held these as their proper type (e.g. `boolean` here) and then `String(envDefaultOverride)` them later?

Might be more intuitive for developers.

justinkaseman: Actually I see why this is a problem, we `getEnv` would need to be heavily reworked with a schema of sorts. 
```
{
CACHE_MAX_ITEMS: {
default: 1000,
type: number
}
}
```
Then we would have helper functions to convert the environment variable string to a number.
So if it comes from env var use the `number` helper, otherwise use the default which is already a number.

justinkaseman:
Could you add a unit test or two for `getEnv` so that we have a codified guard on priority:
1. user provided env override
2. EA level default override
3. Bootstrap level default override

packages/core/types/@chainlink/index.d.ts
austinborn: Yeah, I like the concept but I would like to avoid that for this initial implementation. The defaults act as if they were any other env var pulled in, which is always a string if present.

austinborn:
> Could you add a unit test or two for `getEnv` so that we have a codified guard on priority:
> 
> 1. user provided env override
> 2. EA level default override
> 3. Bootstrap level default override

Sure thing

packages/core/bootstrap/src/lib/middleware/cache-warmer/config.ts
```diff
@@ -35,7 +35,7 @@ export interface Config {
 }
 
 export function get(): Config {
-  const warmupInterval = Number(process.env.WARMUP_INTERVAL)
+  const warmupInterval = Number(getEnv('WARMUP_INTERVAL'))
```
justinkaseman, line 38: So ones like this because they don't have the context passed in won't be able to have an EA coded default override.. could we write that as a comment somewhere, like in the JSdoc for `getEnv` maybe?

I"ll also add it to the project ideas board to refactor for all being default config overridden.

packages/core/types/@chainlink/index.d.ts
```diff
@@ -130,6 +131,19 @@ declare module '@chainlink/types' {
...
   }
 
+  export type EnvDefaults = { [name: string]: string }
+
+  /**
+   * To add a default override for an env var not present here,
+   * ensure that the adapter context from `expose` is passed
+   * to the `getEnv` call(s) for the variable in question. See
+   * `WS_ENABLED` for example usage.
+   */
```
austinborn, line 141: @justinkaseman Here's a note I added about how to add the default overrides. I'll add a similar note to the `getEnv` description as well.

packages/core/bootstrap/src/lib/middleware/cache-warmer/config.ts
austinborn: Yeah good idea to add it to the board. I added more detail to the `getEnv` so it should be clear that `context` needs to be passed in.

packages/core/types/@chainlink/index.d.ts
austinborn: Done.

justinkaseman (approved):
LGTM :raised_hands:

PR #1699, opened 2022-03-22
[Refactor]: added endpoints folders structure to b-c composite adapters

bogos:
## Description

Added endpoints folders structure to b-c composite adapter

## Changes

Add endpoints folders in composite adapters

## Quality Assurance

- [ ] Ran `yarn changeset` if adapter source code was changed
- [X ] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types)


packages/composites/bitcoin-json-rpc/src/config/index.ts
```diff
@@ -1,7 +1,7 @@
 import { Requester, util } from '@chainlink/ea-bootstrap'
 import { Config } from '@chainlink/types'
 
-export const DEFAULT_ENDPOINT = 'difficulty'
+export const DEFAULT_ENDPOINT = 'getblockchaininfo'
```
boxhock, line 4: Why was this changed?

packages/composites/bitcoin-json-rpc/src/endpoint/getblockchaininfo.ts
```diff
@@ -1,19 +1,41 @@
 import * as JSONRPC from '@chainlink/json-rpc-adapter'
-import { Config, ExecuteWithConfig } from '@chainlink/types'
+import { Config, ExecuteWithConfig, InputParameters } from '@chainlink/types'
 import { Validator, Requester } from '@chainlink/ea-bootstrap'
 
 export const NAME = 'getblockchaininfo'
```
boxhock, line 5: This one should have an alias, as per the original switch-case.

packages/composites/crypto-volatility-index/src/config/index.ts
```diff
@@ -1,2 +1,13 @@
+import { Requester } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/types'
+
 export const DEFAULT_NETWORK = 'ETHEREUM'
 export const NAME = 'CRYPTO_VOLATILITY_INDEX'
+export const DEFAULT_ENDPOINT = 'volatility'
```
boxhock, line 6: I think this endpoint name is misleading - suggest naming it "index"

boxhock (changes requested):

packages/composites/bitcoin-json-rpc/src/config/index.ts
bogos: `getblockchaininfo` is the method that is executed. 'difficulty' is the default resulth path, but if this is changed, other information can be obtained. It seemed better to me to call the endpoint as the name of the method to execute. Do you think it's OK?

packages/composites/crypto-volatility-index/src/config/index.ts
bogos: we have 2 files in endpoint folder. The `index.ts` that exports all the methods defined in another file where we have the logic of the endpoints. So how about I call it `indexVolatility` ? to keep that file structure of `index.ts` to export everything.

boxhock: Oooh, I see. Let's call it `volatilityIndex` then

boxhock (approved):

austinborn (approved):

PR #1704, opened 2022-03-22
Refactor out sleep function into bootstrap

alejoberardino:
## Description

In many places, primarily websocket tests, a pattern for a quick sleep function using promises and setTimeout is used. This PR takes all those repeated functions and refactors them out into bootstrap utils.


justinkaseman (approved):
:clap:

PR #1710, opened 2022-03-23
(feat): Add error backoff middleware

justinkaseman:
## Description
To prevent repeated failing requests from draining the available API requests per second/minute we can back off from known errors.

## Changes

- Adds a new middleware: `ErrorBackoff`

## Steps to Test

Success case:
1. Run an EA
2. Request any valid request
3. It should return a success
4. Repeat this 3+ times. It should continually return from the cache.

Failure case:
1. Run an EA
2. Request any invalid request
3. It should return an error
4. Repeat this 3 times. On the 3rd, attempt it should return a back off error.
5. Wait one minute and re-request, it should return an error.

## Quality Assurance


alejoberardino:
Just to check that I understand the approach, the idea is to record errors in a similar way to the rate limiting middleware, but handling in a separate one so as to block the request in a simple fashion instead of the cache ttl numbers that the former relies on.
I that's the case I think we won't be able to technically enforce the fixed rate limit "properly", but as long as we make sure that the error specific rate limiting is always less permissive (much lower errors/s than successful req/s) than the normal limits, we _probably_ shouldn't see this being an issue.

Would these assessments be correct?

alejoberardino:
Just to check that I understand the approach, the idea is to record errors in a similar way to the rate limiting middleware, but handling in a separate one so as to block the request in a simple fashion instead of the cache ttl numbers that the former relies on.
I that's the case I think we won't be able to technically enforce the fixed rate limit "properly", but as long as we make sure that the error specific rate limiting is always less permissive (much lower errors/s than successful req/s) than the normal limits, we _probably_ shouldn't see this being an issue.

Would these assessments be correct?

justinkaseman:
> Just to check that I understand the approach, the idea is to record errors in a similar way to the rate limiting middleware, but handling in a separate one so as to block the request in a simple fashion instead of the cache ttl numbers that the former relies on. I that's the case I think we won't be able to technically enforce the fixed rate limit "properly", but as long as we make sure that the error specific rate limiting is always less permissive (much lower errors/s than successful req/s) than the normal limits, we _probably_ shouldn't see this being an issue.
> 
> Would these assessments be correct?

Rate Limiting middleware doesn't record errors. Rate Limit tracks successful requests seen.
This new one is similar to Rate Limit in the way that it tracks requests (sliding time slice), but this middleware tracks only tracks errors.
If  Error Backoff blocks the request it will never be recorded by Rate Limit.

alejoberardino:
My question was that because the rate limits middleware doesn't block requests per se and uses the ttl, if that was the main reason to separate into a different middleware, to avoid using that complex mechanic or wiring the outright rejection into it

alejoberardino:
My question was that because the rate limits middleware doesn't block requests per se and uses the ttl, if that was the main reason to separate into a different middleware, to avoid using that complex mechanic or wiring the outright rejection into it

justinkaseman:
> My question was that because the rate limits middleware doesn't block requests per se and uses the ttl, if that was the main reason to separate into a different middleware, to avoid using that complex mechanic or wiring the outright rejection into it

It's that the middleware state is different (storing successes vs. fails)
And it's also a different purpose - preventing bad feeds vs. adjusting TTL. A user may want the error backoffs should happen even if Rate Limit is shut off.

alejoberardino (approved):

PR #1711, opened 2022-03-23
Add coverage for simple middlewares

alejoberardino:
## Description

Adds coverage for the simple middlewares in core bootstrap package.


justinkaseman (approved):
:chart_with_upwards_trend: coverage go uppp

PR #1722, opened 2022-03-25
EA upper limit load test tooling

boxhock:
## Changes

- Added tooling to run k6 RPS load tests

## Steps to Test

1. Follow "Upper limit testing" steps in k6 README

## Quality Assurance


alejoberardino:
I think we should change the approach to make it so we can get an approximate number with one operation/command. We can configure stress tests to ramp up a bit: https://k6.io/docs/test-types/stress-testing/
This is still not a dynamic increase and we will have to figure out what are appropriate steps, but we can start with something like a few mins to stabilize, then ramping up the load until the error rate becomes too high

boxhock:
@alejoberardino That sounds like a good suggestion. I will work on including this.

boxhock:
@alejoberardino 
> ramping up the load until the error rate becomes too high
From what I could tell, this isn't possible. However, I now made it gradually warm up the request count, which should avoid any cold start problems.

packages/k6/src/test-limits.ts
```diff
@@ -0,0 +1,141 @@
...
+export const options = {
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(90)<1000'], // 90% of requests should be below 1s
+  },
+  stages: [
+    { duration: '5m', target: uniqueRequests }, // 5m warmup from 0 to `uniqueRequests`
+    { duration: '1m', target: VU }, // 1m warmup from `uniqueRequests` to target RPS
+    { duration: testDuration, target: VU }, // `testDuration` with constant target RPS
+  ],
```
alejoberardino, line 83: @boxhock This is what I had in mind as a stress test, the way the k6 docs show to do it is by setting many stages like
```suggestion
  stages: [
    { duration: '5m', target: uniqueRequests }, // 5m warmup from 0 to `uniqueRequests`
    { duration: testDuration / 10, target: (VU / 10) * 1 },
    { duration: testDuration / 10, target: (VU / 10) * 2 },
    { duration: testDuration / 10, target: (VU / 10) * 3 },
    { duration: testDuration / 10, target: (VU / 10) * 4 },
    { duration: testDuration / 10, target: (VU / 10) * 5 },
    { duration: testDuration / 10, target: (VU / 10) * 6 },
    { duration: testDuration / 10, target: (VU / 10) * 7 },
    { duration: testDuration / 10, target: (VU / 10) * 8 },
    { duration: testDuration / 10, target: (VU / 10) * 9 },
    { duration: testDuration / 10, target: VU },
  ],
```
We'd change the RPS variable to signify "max RPS", with the expectation that the test would fail somewhere along the line. Another option could be to use one of the [ramping executors](https://k6.io/docs/using-k6/scenarios/executors/ramping-vus/)

boxhock: I don't think we'd be able to make the test fail along the line at any point? With these stages, we'd just see the RPS increase gradually for each stage, and then see if it worked in the end or not afaict

alejoberardino: With the thresholds we'd be able to see when they fail, and optionally we can also abort execution as well https://k6.io/docs/using-k6/thresholds/#aborting-a-test-when-a-threshold-is-crossed

boxhock: @alejoberardino Thanks for this, I think I got it in a good state now, so please have a look. Only difference is that I added a scaleup period between each stage, allowing us to test for a constant RPS for some time instead of constantly scaling up. Let me know what you think

alejoberardino (approved):

PR #1732, opened 2022-03-25
Add tests to increase cache coverage

alejoberardino:


packages/core/bootstrap/test/unit/cache.test.ts
```diff
@@ -136,5 +137,144 @@ describe('cache', () => {
...
+      const response = {
+        jobRunID: '1',
+        data: { jobRunID: '1', statusCode: 200, data: {}, result: 123 },
+        result: 123,
+        statusCode: 200,
+      }
+      const execute = async () => response
+      const middleware = await withCache()(execute, context)
+      const result = await middleware(request, {})
+      expect(result).toBe(response)
```
austinborn, line 158: Where does the `123` in the result come from? Are you sure the `expect(...).toBe(...)` check is also checking nested values?

alejoberardino: The execute function returns the response object, and in this case the expectation is that the middleware floats that up to the surface; the `toBe` function will compare the objects by reference so it's guaranteed they're the same instance

austinborn (approved):

PR #1740, opened 2022-03-29
bitcoin-json-rpc height endpoint

### LINKED ISSUE #1715: EA to get difficulty and height from BTC node RPC
SC-ID: 31978

We need an EA to query the current difficulty and block height from a Bitcoin node JSON-RPC.

It should be a single EA with two endpoints:
- difficulty
- height

karen-stepanyan:
## Changes

- Added new `height` alias in supportedEndpoints
- If the endpoint is `height` the adapter will return `block` property from getblockchaininfo result

## Quality Assurance


packages/composites/bitcoin-json-rpc/src/endpoint/getblockchaininfo.ts
```diff
@@ -37,10 +37,16 @@ export interface ResponseSchema {
...
+  const endpointName = validator.validated.data.endpoint
+  let resultPath
+
+  if (endpointName === DEFAULT_FIELD) {
+    resultPath = DEFAULT_FIELD
+  } else if (endpointName === 'height') {
+    resultPath = 'blocks'
+  } else {
+    resultPath = validator.validated.data.resultPath || DEFAULT_FIELD
+  }
```
boxhock, line 49: Would this maybe be more readable with a switch case?

boxhock (changes requested):
Looks good, but could use an integration test. Should be good to merge once added!

packages/composites/bitcoin-json-rpc/src/endpoint/getblockchaininfo.ts
```diff
@@ -5,7 +5,7 @@ import { Validator, Requester } from '@chainlink/ea-bootstrap'
 export const NAME = 'getblockchaininfo'
 const DEFAULT_FIELD = 'difficulty'
 
-export const supportedEndpoints = [NAME, 'difficulty']
+export const supportedEndpoints = [NAME, 'difficulty', 'height']
```
boxhock, line 8: Should we be adding a convention like this here?

https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/coingecko/src/endpoint/global.ts#L6-L9

karen-stepanyan: indeed. we won't need to use switch statement in this case

boxhock (approved):

boxhock:
Looks good, thank you!

PR #1771, opened 2022-04-06
Add confirmations param to eth-balance adapter

alejoberardino:
## Description

Adds `confirmations` parameter to the `eth-balance` adapter to specify the number of confirmed blocks that must have passed to fetch the balance.

## Changes

- Add confirmations param to balance endpoint in eth-balance adapter

## Steps to Test

1. Run eth-balance adapter
2. Fetch balance from a known address with a very recent transaction
3. Specify confirmations param to a number of blocks higher than the ones that have passed since that transaction
4. Check that the balance received does not include the recent tx, as it does not have enough confirmaitons in the chain

## Quality Assurance

### DIFF
--- a/packages/sources/eth-balance/src/endpoint/balance.ts
+++ b/packages/sources/eth-balance/src/endpoint/balance.ts
@@ -15,9 +15,9 @@ export const inputParameters: InputParameters = {
     description:
       'An array of addresses to get the balances of (as an object with string `address` as an attribute)',
   },
-  confirmations: {
+  minConfirmations: {
     required: false,
-    aliases: ['minConfirmations'],
+    aliases: ['confirmations'],
     type: 'number',
     default: 0,
     description:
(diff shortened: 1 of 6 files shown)


packages/sources/eth-balance/src/endpoint/balance.ts
```diff
@@ -15,6 +15,14 @@ export const inputParameters: InputParameters = {
     description:
       'An array of addresses to get the balances of (as an object with string `address` as an attribute)',
   },
+  confirmations: {
+    required: false,
+    aliases: ['minConfirmations'],
```
eddiejhong, line 20: It seems like `minConfirmations` should be the primary alias as it's much more descriptive than `confirmations`

packages/sources/eth-balance/test/integration/balance.test.ts
```diff
@@ -86,4 +86,27 @@ describe('execute', () => {
       expect(response.body).toMatchSnapshot()
     })
   })
+
```
eddiejhong, line 89: Might also want to add tests here for when minConfirmations is out of bounds (<0 or >64)

alejoberardino: This is already covered in the unit tests 👍

eddiejhong: Oh I see this now, ty

eddiejhong (approved):

PR #1780, opened 2022-04-08
Core Type #4 - Composite adapter types

justinkaseman:


packages/composites/bob/src/endpoint/format.ts
```diff
@@ -59,21 +61,24 @@ export const execute: ExecuteWithConfig<ExtendedConfig> = async (request, contex
...
-  const _execute: ExecuteWithConfig<Config> = JSONRPC.makeExecute()
+  const _execute = JSONRPC.makeExecute(config)
   const response = await _execute(
     {
       ...request,
-      data: { ...request.data, method: 'eth_getBlockByHash', params: [block.hash, false] },
+      data: {
+        ...(request.data as JSONRPC.types.request.TInputParameters),
+        method: 'eth_getBlockByHash',
+        params: [block.hash, 'false'],
```
boxhock, line 71: This shouldn't be a string

```suggestion
        params: [block.hash, false],
```

packages/composites/bsol-price/src/endpoint/price.ts
```diff
@@ -124,14 +140,16 @@ export const getStSolPerUSD = async (
   const allocations = [
     {
       symbol: 'SOL',
-      balance: new BigNumber(10).pow(LAMBERT_DECIMALS),
+      balance: new BigNumber(10).pow(LAMBERT_DECIMALS) as unknown as TBN,
```
boxhock, line 143: Why do we need this conversion?

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -151,14 +155,14 @@ const getValidFilecoinAddress = (id: string, address: string): string | undefine
...
+      `JobId ${id}: The address "${address}" is not a valid Dogecoin address and has been removed.`,
     )
     return
   }
-  if (address[1] !== '0' && address[1] !== '1' && address[1] !== '2' && address[1] !== '3') {
+  if (address[1] !== '1' && address[1] !== '3') {
     Logger.warn(
       { warning: 'Invalid address detected' },
-      `JobId ${id}: The address "${address}" is not a valid Filecoin address and has been removed.`,
+      `JobId ${id}: The address "${address}" is not a valid Dogecoin address and has been removed.`,
```
boxhock, line 165: I'm guessing most of the diff in this file will be resolved when rebasing?

packages/composites/token-allocation/src/config/index.ts
```diff
@@ -15,20 +15,21 @@ import * as Nomics from '@chainlink/nomics-adapter'
...
+  CoinGecko as any,
+  CoinMarketCap as any,
+  CoinMetrics as any,
+  CoinPaprika as any,
+  CryptoCompare as any,
+  Finage as any,
+  Kaiko as any,
+  NCFX as any,
+  Nomics as any,
+  Tiingo as any,
```
boxhock, line 32: Seems like we used `unknown` as an intermediary type in the past. Should we do that here as well?

boxhock (changes requested):

PR #1799, opened 2022-04-18
Add ETH to coinmetrics WS valid quote params

alejoberardino:
## Description

Needed for SUSD/ETH feed, coinmetrics ws api now supports `ReferenceRateETH` (and for BTC as well). This is a simple PR to add that.

### DIFF
--- a/.changeset/ninety-meals-confess.md
+++ b/.changeset/ninety-meals-confess.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/coinmetrics-adapter': patch
+---
+
+Add ETH to valid quotes for WS endpoint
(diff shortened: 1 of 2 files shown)


.changeset/ninety-meals-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/coinmetrics-adapter': patch
+---
+
+Add ETH to valid quotes for WS endpoint
```
eddiejhong, line 5: Looks like we're also adding BTC to the array of valid quotes, was that intended?

alejoberardino: It was, updated the changelog 👍

eddiejhong (approved):

PR #1808, opened 2022-04-20
update how how Galaxis API URL is formed

cds95:


packages/sources/galaxis/src/endpoint/nba.ts
```diff
@@ -65,7 +65,8 @@ const getDate = async (config: ExtendedConfig, paramsDate?: string): Promise<str
 
 const getURL = async (config: ExtendedConfig, date: string): Promise<string> => {
   if (isDateNotSet(date)) return config.api.baseURL
-  return `${config.api.baseURL}/nightly_achievements_${date}.json` // YYYY-MM-DD
+  const year = date.split('-')[0]
```
thodges-gh, line 68: Could we use `date.getFullYear()` instead? It's probably clearer for the reader.

thodges-gh (dismissed):

thodges-gh (approved):

PR #1817, opened 2022-04-22
update terra view function README example request

cds95:


mjk90 (approved):

austinborn:
If the response from the API has changed, you'll need to change this in the integration tests too. Otherwise, this change will be overridden the next time the README is generated.

PR #1827, opened 2022-04-25
Remove express rate limiting middlewares

alejoberardino:
## Description

The use of `express-rate-limit` and `express-slow-down` middlewares was necessary in the initial stages of the EA framework, but they have become more of a pain point that limits performance since. This PR removes both from the core framework.

## Changes

- Removed `express-rate-llimit` middleware.
- Removed `express-slow-down` middleware.


mjk90 (approved):

boxhock (approved):
👏👏

PR #1829, opened 2022-04-26
Add address validation and duplicate address filtering for Ethereum PoR

KuphJr:
## Changes

- Implemented address validation & duplicate address filtering for Ethereum addresses in the proof-of-reserves EA

## Steps to Test

`yarn test proof-of-reserves`

## Quality Assurance


packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,52 @@
+import { utils } from 'ethers'
+import { Logger } from '@chainlink/ea-bootstrap'
+
+type AddressObject = { address: string }
+type AddressArray = AddressObject[]
+
+export const validateAddresses = (indexer: string, addresses: AddressArray): AddressArray => {
+  switch (indexer) {
+    case 'eth_balance':
```
boxhock, line 9: Assuming a `network` param is passed with each address, can we switch on that primarily? And if it's not present, fall back to the indexer?

I'm afraid of us forgetting to put in the right indexer here. Another problem is that we have indexers that work for multiple blockchains (bitcoin + dogecoin).

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -38,13 +39,14 @@ const inputParameters: InputParameters = {
...
 export const execute: ExecuteWithConfig<Config> = async (input, context, config) => {
   const validator = new Validator(input, inputParameters, paramOptions)
-
   const jobRunID = validator.validated.jobRunID
   const protocol = validator.validated.data.protocol.toUpperCase()
   const indexer: Indexer = validator.validated.data.indexer.toUpperCase()
   const confirmations = validator.validated.data.confirmations
 
   const protocolOutput = await runProtocolAdapter(jobRunID, context, protocol, input.data, config)
+  protocolOutput.result = validateAddresses(indexer, protocolOutput.result)
```
boxhock, line 48: We should have a way to disable address validation, probably with an input param.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -38,13 +39,14 @@ const inputParameters: InputParameters = {
...
+  protocolOutput.data.result = protocolOutput.result
```
boxhock, line 49: Can we store the validated output as a new variable instead of overriding the existing one?

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,52 @@
...
+  const validAddresses: AddressArray = []
+  for (const { address } of addresses) {
+    try {
+      validAddresses.push({ address: utils.getAddress(address) })
+    } catch (error) {
+      Logger.warn(
+        error,
+        `The address "${address}" is invalid or has an invalid checksum and has been removed from the request.`,
+      )
+    }
```
boxhock, line 32: Good use of the ethers util function here :+1:

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
boxhock: Or... have a way to do address validation, but still allow duplicate addresses. I just want us to think ahead and not have to do code changes in the future if we suddenly want duplicate addresses for some use case.

packages/composites/proof-of-reserves/test/unit/addressValidator.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    ]
+    expect(validateAddresses('eth_balance', lowercaseAddresses)).toEqual(validChecksumAddresses)
+  })
+
+  it('Validates an address containing only upper case letters', () => {
+    const uppercaseAddresses = [
+      { address: '0x' + '8288c280F35Fb8809305906C79BD075962079Dd8'.toUpperCase() },
+      { address: '0x' + '81910675DbaF69deE0fD77570BFD07f8E436386A'.toUpperCase() },
+    ]
+    console.log(JSON.stringify(uppercaseAddresses))
```
boxhock, line 26:
```suggestion
```

boxhock (changes requested):

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
KuphJr: Added option to disable validation altogether & added param info to readme

.changeset/wild-colts-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/proof-of-reserves-adapter': minor
+---
+
+Addressed PR notes from @boxhock
```
boxhock, line 5: No need for a new changeset for this, you can delete this one.

boxhock (changes requested):

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -38,19 +39,23 @@ const inputParameters: InputParameters = {
...
   const validator = new Validator(input, inputParameters, paramOptions)
-
   const jobRunID = validator.validated.jobRunID
   const protocol = validator.validated.data.protocol.toUpperCase()
   const indexer: Indexer = validator.validated.data.indexer.toUpperCase()
   const confirmations = validator.validated.data.confirmations
 
   const protocolOutput = await runProtocolAdapter(jobRunID, context, protocol, input.data, config)
+  const validatedInput = { ...protocolOutput }
+  if (validator.validated.data.disableAddressValidation !== 'true') {
```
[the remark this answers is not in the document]
KuphJr: I used the Boolean constructor to resolve this issue.

`if (!Boolean(validator.validated.data.disableAddressValidation)) { ... }`

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -38,19 +38,24 @@ const inputParameters: InputParameters = {
...
   const protocol = validator.validated.data.protocol.toUpperCase()
   const indexer: Indexer = validator.validated.data.indexer.toUpperCase()
   const confirmations = validator.validated.data.confirmations
 
   const protocolOutput = await runProtocolAdapter(jobRunID, context, protocol, input.data, config)
+  const validatedInput = { ...protocolOutput }
+  if (Boolean(validator.validated.data.disableAddressValidation) === false)
+    validatedInput.result = validateAddresses(indexer, validatedInput.result)
+  if (Boolean(validator.validated.data.disableDuplicateAddressFiltering) === false)
+    validatedInput.result = filterDuplicates(validatedInput.result)
```
boxhock, line 51: A simple `Boolean()` is not going to work, unfortunately...

```
> Boolean("false")
true
```

boxhock (changes requested):

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,75 @@
...
+  address: string
+  network?: string
+}
+
+export const validateAddresses = (indexer: string, addresses: AddressObject[]): AddressObject[] => {
+  const validatedAddresses: AddressObject[] = []
+  for (const addressObj of addresses) {
+    const { address, network } = addressObj
+    let validatedAddress: string | undefined
+    if (network) {
```
boxhock, line 14: Instead of checking if `network` is set and doing two switch-es here, I would prefer we do something like this:

```typescript
const indexerToNetwork = {
  'eth_balance': 'ethereum'
}

const validationNetwork = network || indexerToNetwork[indexer]
```

This would allow us to only do a single switch, so we don't duplicate this logic. Does this make sense?

boxhock (changes requested):

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -38,19 +38,30 @@ const inputParameters: InputParameters = {
...
+  if (
+    validator.validated.data.disableAddressValidation &&
+    validator.validated.data.disableAddressValidation !== 'false'
+  )
+    validatedInput.result = validateAddresses(indexer, validatedInput.result)
+  if (
+    validator.validated.data.disableDuplicateAddressFiltering &&
+    validator.validated.data.disableDuplicateAddressFiltering !== 'false'
+  )
+    validatedInput.result = filterDuplicates(validatedInput.result)
```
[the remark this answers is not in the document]
KuphJr: Yeah I see the issue.  Both methods technically work if the value is only ever set to "true" or "false", but in the event of a mistype, it would default to true. (ex: "fase" would be set to true and not false).  I committed your suggestion to fix this.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
KuphJr: Yeah this is a much better way to do this in order to avoid duplicated logic.  I have implemented it as you suggested.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -34,23 +34,48 @@ const inputParameters: InputParameters = {
...
+    type: 'boolean',
+    description: 'Gives the option to disable address validation before the balances are fetched.',
+  },
+  disableDuplicateAddressFiltering: {
+    required: false,
+    type: 'boolean',
+    description:
+      'Gives the option to disabled the filtering of duplicate addresses in a request. ' +
+      'If this is set to `true` and a duplicate address is contained in the request, the balance of that address will be counted twice.',
+  },
```
boxhock, line 48: We probably want to add a default value here. Also need to add `disableDuplicateAddressFiltering` to README

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -34,23 +34,48 @@ const inputParameters: InputParameters = {
...
+  ) {
+    validatedInput.result = validateAddresses(indexer, validatedInput.result)
+  }
+  if (
+    !validator.validated.data.disableDuplicateAddressFiltering ||
+    validator.validated.data.disableDuplicateAddressFiltering !== 'true'
+  ) {
+    validatedInput.result = filterDuplicates(validatedInput.result)
+  }
+  validatedInput.data.result = validatedInput.result
```
boxhock, line 72: I'm concerned with how big this logic is growing here. Let's extract it just like how we do it with the Reduce step: https://github.com/smartcontractkit/external-adapters-js/blob/85f6de8abad564af166a11c340b54fce1578469f/packages/composites/proof-of-reserves/src/utils/reduce.ts#L24-L51

boxhock (changes requested):

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -34,24 +35,61 @@ const inputParameters: InputParameters = {
...
+      'Gives the option to disabled the filtering of duplicate addresses in a request. ' +
+      'If this is set to `true` and a duplicate address is contained in the request, the balance of that address will be counted twice.',
+    default: 'false',
+  },
 }
 
 export const execute: ExecuteWithConfig<Config> = async (input, context, config) => {
   const validator = new Validator(input, inputParameters, paramOptions)
-
   const jobRunID = validator.validated.jobRunID
```
boxhock, line 56: Let's keep the space here to keep it consistent with other adapters.

```suggestion

  const jobRunID = validator.validated.jobRunID
```

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -34,24 +35,61 @@ const inputParameters: InputParameters = {
...
   const protocol = validator.validated.data.protocol.toUpperCase()
   const indexer: Indexer = validator.validated.data.indexer.toUpperCase()
   const confirmations = validator.validated.data.confirmations
-
   const protocolOutput = await runProtocolAdapter(jobRunID, context, protocol, input.data, config)
```
boxhock, line 60: Same here

```suggestion

  const protocolOutput = await runProtocolAdapter(jobRunID, context, protocol, input.data, config)
```

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -34,24 +35,61 @@ const inputParameters: InputParameters = {
...
+  }
+  if (
+    !validator.validated.data.disableDuplicateAddressFiltering ||
+    validator.validated.data.disableDuplicateAddressFiltering !== 'true'
+  ) {
+    validatedInput.result = filterDuplicates(validatedInput.result)
+  }
+  validatedInput.data.result = validatedInput.result
+  return validatedInput
+}
```
boxhock, line 95: Let's move this to `packages/composites/proof-of-reserves/src/utils/addressValidator.ts`.

Also, to stay consistent with the naming here, suggest calling it something like `runAddressValidationAdapter`

boxhock (changes requested):
Just a few nitpicks, is good to merge once resolved.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const getValidAddresses = (
+  protocolOutput: AdapterResponse,
+  validator: Validator,
+): AdapterResponse => {
+  const validatedInput = { ...protocolOutput }
+  if (
+    !validator.validated.data.disableAddressValidation ||
+    validator.validated.data.disableAddressValidation !== 'true'
+  ) {
```
boxhock, line 23:
```suggestion
  if (!parseBool(validator.validated.data.disableAddressValidation)) {
```

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,92 @@
...
+    validatedInput.result = validateAddresses(
+      validator.validated.data.indexer,
+      validatedInput.result,
+    )
+  }
+  if (
+    !validator.validated.data.disableDuplicateAddressFiltering ||
+    validator.validated.data.disableDuplicateAddressFiltering !== 'true'
+  ) {
```
boxhock, line 32:
```suggestion
  if (!parseBool(validator.validated.data.disableDuplicateAddressFiltering)) {
```

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
boxhock: Need to import this from utils as well

boxhock (changes requested):

KuphJr:
Another PR included these changes

PR #1841, opened 2022-05-04
(feat): Add ENS adapter

justinkaseman:
## Description
Adds a new ENS External Adapter to read from Ethereum mainnet ENS.

## Changes

- Add ENS adapter

## Quality Assurance


packages/sources/ens/src/endpoint/lookup.ts
```diff
@@ -0,0 +1,80 @@
...
+  const splitName = name.split('.')
+  if (splitName.length < 2)
+    throw new AdapterError({
+      jobRunID,
+      message: `Invalid ENS name. Format must be [domain].[top level domain], (e.g. chainlink.eth)`,
+      statusCode: 400,
+    })
+
+  const networkProvider = new ethers.providers.JsonRpcProvider(config.rpcUrl)
+  const contracts = await initializeENS(networkProvider)
```
alejoberardino, line 37: Looking at the impl it seems to make a lot of async requests, would moving this initialization outside of the endpoint logic affect behavior?

alejoberardino (changes requested):

packages/sources/ens/src/endpoint/lookup.ts
justinkaseman: Yeah, I was thinking about that too. I'll initialize it on startup instead and hold the contracts in the config.

packages/sources/ens/src/endpoint/lookup.ts
```diff
@@ -0,0 +1,80 @@
...
+    status: 200,
+    statusText: '',
+    headers: {},
+    config: {},
+  }
+
+  try {
+    const isEthTLD = splitName[splitName.length - 1] === 'eth'
+    const isSubdomain = splitName.length > 2
+    const labelHash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes(splitName[0]))
```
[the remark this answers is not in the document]
justinkaseman: Yeah, I believe it is because the Registrar itself is what the top level domain is, so for example the `.eth` registrar knows would only need `test` to get to `test.eth`.

https://docs.ens.domains/dapp-developer-guide/ens-as-nft#deriving-tokenid-from-ens-name

PR #1847, opened 2022-05-06
add telemetry data for EA responses

### LINKED ISSUE #1842: Include telemetry data in EA responses
See separate PDF for spec.

We want to include telemetry data from the EA that can be ingested by the Chainlink node. This feature should be behind a feature flag.

karen-stepanyan:
## Description

Added support to provide telemetry data in adapter's response with `TELEMETRY_DATA_ENABLED` env var

## Quality Assurance


packages/core/bootstrap/src/lib/modules/requester.ts
```diff
@@ -219,6 +225,32 @@ export class Requester {
       adapterResponse.providerStatusCode = response.status
     }
 
+    if (parseBool(getEnv('TELEMETRY_DATA_ENABLED'))) {
+      adapterResponse.telemetry = {
+        rateLimitEnabled: parseBool(getEnv('RATE_LIMIT_ENABLED')),
+        wsEnabled: parseBool(getEnv('WS_ENABLED')),
```
austinborn, line 231: One big problem with the current approach is that it overlooks the `envDefaultOverrides` for each adapter, since `getEnv` needs the adapter context to get the overrides. An example is `WS_ENABLED`, which is overridden to `true` by default for some adapters.

One option is to pass in the `context` to every `Requester.success` call, although that would get messy. Another option is to set this up as a separate middleware, which I believe should be able to attach the telemetry data to the result object. See [here](https://github.com/smartcontractkit/external-adapters-js/blob/54514ec52436a9a16fb45d05738e76a011a3fd46/packages/core/bootstrap/src/index.ts#L81-L94) for other middleware examples.

austinborn (changes requested):
Due to the env var overrides used in `getEnv`, the current changes do not reflect a complete picture of the env var config. See the comment for details.

packages/core/bootstrap/src/lib/modules/requester.ts
karen-stepanyan: Hey Austin, thanks for review and comment. Indeed i missed this part. What about another option, to merge envDefaultOverrides with process.env when we are starting the server ?  IDK if it's intentional that we keep them separate. 

```
 const envDefaultOverrides = adapterContext.envDefaultOverrides
 for (const key in envDefaultOverrides) {
   if (!process.env[key]) {
     process.env[key] = envDefaultOverrides[key];
   }
}
```

austinborn: > IDK if it's intentional that we keep them separate.

Not intentional, just a design-by-omission. I am open to the idea of setting the env in a single place when the server is started, but it's unconventional IMO. Let's see what @boxhock thinks.

alejoberardino: @karen-stepanyan I like the idea, I think that's how it should work instead of checking for base defaults or overrides in each get. Ideally we wouldn't depend on global variables (harder to mock) but it might make sense to make the change easier.

boxhock: Good catch @austinborn. Global variables should generally be avoided IMO, but I do think that this use case warrants this solution. As @alejoberardino says, we should have had something like this for input params as well, but that's way out of scope for this PR (and likely won't be changed until v3 anyways).

TL;DR: I'm good with setting envDefaultOverrides in process.env on startup. Let's make sure that we properly test this though, as this is a change that's relevant regardless of whether the feature flag for telemetry is enabled or not.

packages/core/bootstrap/src/lib/modules/requester.ts
```diff
@@ -48,7 +48,13 @@ export class Requester {
       let response: AxiosResponse<T>
       const url = join(config.baseURL || '', config.url || '')
       try {
+        const startTime = process.hrtime()
+
         response = await axios(config)
+
+        const endTime = process.hrtime(startTime)
+        const milliseconds = Math.round(endTime[0] * 1000 + endTime[1] / 1000000)
```
boxhock, line 56: We're doing a network request to an external resource here, so exact ns timings are not going to be very relevant. What's the impact of doing a simple `Date.now()` minus `Date.now()` at the start of the request? Seems simpler to manage, and more readable, right?

karen-stepanyan: no particular reason, just thought to be more precise, but i guess `Date.now()` will be more manageable

alejoberardino: Actually I prefer hrtime, it's meant for these measurements (no idea if it's more performant)

boxhock: Had a closer look in the docs:
> These times are relative to an arbitrary time in the past, and not related to the time of day and therefore not subject to clock drift.

While I do think this one is overkill in terms of complexity, it would also be nice to not be affected by things like clock drift.

Let's keep `process.hrtime`, but let's use `process.hrtime.bigint()`, [as `process.hrtime([time])` is deprecated](https://nodejs.org/docs/latest-v16.x/api/process.html#processhrtimetime).

packages/core/bootstrap/src/lib/modules/requester.ts
```diff
@@ -48,7 +48,13 @@ export class Requester {
...
+        response.headers['request-duration'] = milliseconds.toString()
```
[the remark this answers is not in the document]
boxhock: Hmm, it might be too much work, yeah. I'm just not a very big fan of dumping data into headers just for the purpose of us passing variables.

Let's keep using the headers, but would prefer if we make it more explicit that this is an EA header (e.g. `ea-request-duration`).

packages/core/types/@chainlink/index.d.ts
```diff
@@ -107,6 +108,20 @@ declare module '@chainlink/types' {
...
+    rateLimitEnabled: boolean
+    wsEnabled: boolean
+    cacheEnabled: boolean
+    cacheType?: string
+    cacheWarmingEnabled: boolean
+    cacheMaxAge?: string
+    metricEnabled: boolean
+    rateLimitApiTier?: string
+    requestCoalescingEnabled: boolean
+    protocol?: 'ws' | 'http'
```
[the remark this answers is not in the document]
boxhock: It seems like this metric might be too difficult to keep accurate.

Let's please remove this from the spec for now. I think if we were to add it now, it wouldn't serve its original purpose. This would be easier to implement in EAv3.

boxhock (changes requested):
Just one small change, otherwise LGTM

packages/core/types/@chainlink/index.d.ts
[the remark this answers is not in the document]
karen-stepanyan: oh sorry, didn't notice that

alejoberardino (dismissed):

boxhock (approved):

austinborn (approved):

PR #1856, opened 2022-05-10
PoR Address List EA

cds95:
## Description

This PR implements the new PoR Address list EA that fetches a list of custodial addresses that support a PoR feed for a wrapped token

## Changes

- Implement `por-address-list` EA

## Steps to Test

1. Set required `RPC_URL` env var.  This EA was tested using contracts deployed on Rinkeby
2. Send request below
3. Verify successful response
4. View the contract on **Rinkeby** Etherscan and verify addresses match what the EA responds with

Request
```
{
    "id": "1",
    "data": {
        "contractAddress": "0x203E97cF02dB2aE52c598b2e5e6c6A778EB1987B"
    }
}
```

## Quality Assurance


packages/sources/por-address-list/src/endpoint/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+    const endIdx = nextEndIdx.gt(numAddresses) ? numAddresses.sub(1) : nextEndIdx
+    // element at endIdx is included in result
+    const latestAddresses = await addressManager.getPoRAddressList(startIdx, endIdx, {
+      blockTag,
+    })
+    addresses = addresses.concat(latestAddresses)
+    startIdx = endIdx.add(1)
+  }
+  // Filter out duplicate addresses
+  return [...new Set(addresses)]
```
boxhock, line 25: We are trying to do the same in #1829, just more configurable. I think we shouldn't have to worry about duplicate addresses here.

packages/sources/por-address-list/src/endpoint/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { ethers } from 'ethers'
+
+export const fetchAddressList = async (
+  addressManager: ethers.Contract,
+  confirmations = 0,
+  batchSize = 10,
+): Promise<string[]> => {
+  const blockTag = confirmations ? -confirmations : undefined
```
boxhock, line 8: How does this work? Don't you have to fetch the current block height and subtract confirmations? Can you pass in a negative number as the block tag?

Not doubting that it works, just unfamiliar with this

boxhock: Might be safer to do a check of the block number at the start, so the block number between all of these requests doesn't change

packages/sources/por-address-list/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@chainlink/por-address-list-adapter",
+  "version": "0.0.3",
```
justinkaseman, line 3:
```suggestion
  "version": "0.0.0",
```

packages/sources/por-address-list/src/endpoint/utils.ts
cds95: yeah I think we can pass in a negative number.  I got this info from the docs here https://docs.ethers.io/v5/api/providers/types/#providers-BlockTag

boxhock: Thanks, TIL!

Still think that we should fetch the current block number and subtract, that way we can ensure we are using the same block number for all requests - otherwise we could be calling over two blocks, where the list could have been changed.

cds95: gotcha ok.  Just updated the PR

packages/sources/por-address-list/CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
+# @chainlink/example-source-adapter
```
justinkaseman, line 1:
```suggestion
# @chainlink/por-address-list-adapter
```

packages/sources/por-address-list/CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
...
+
+## 0.0.3
```
justinkaseman, line 3: Clear these

packages/sources/por-address-list/README.md
```diff
@@ -0,0 +1,58 @@
+# PoR Address List Adapter
+
+![0.0.3](https://img.shields.io/github/package-json/v/smartcontractkit/external-adapters-js?filename=packages/sources/por-address-list/package.json)
```
justinkaseman, line 3:
```suggestion
![0.0.1](https://img.shields.io/github/package-json/v/smartcontractkit/external-adapters-js?filename=packages/sources/por-address-list/package.json)
```

justinkaseman:
Also need to `yarn` locally to regen the lockfile

justinkaseman (approved):

PR #1875, opened 2022-05-17
Fix batch cache key generation by ignoring overrides & other excl props

alejoberardino:
## Description

The changes to a more performant cache key generation made it so a few additional params that are sent to the adapter (like overrides) affected the algorithm that calculated them and resulted in different keys for batch and non batch items. This PR updates the relevant function to exclude those properties to simplify and make keys predictable again.


austinborn (approved):
LGTM

PR #1896, opened 2022-05-24
Fix/por address validation and filtering

KuphJr:
## Changes

- Added address validation and duplicate filtering for Ethereum, Bitcoin, Cardano, Dogecoin and Filecoin in the proof of reserves adapter

## Steps to Test

`yarn test proof-of-reserves`

## Quality Assurance

- [ ] Ran `yarn changeset` if adapter source code was changed


packages/composites/proof-of-reserves/README.md
```diff
@@ -42,12 +42,14 @@ See the [Composite Adapter README](../README.md) for more information on how to
...
+|           |          `confirmations`           | The number of confirmations required for a transaction to be counted when getting an address balance |                                                                                                                                                                         |      6      |
+|           |            `addresses`             |           An array of addresses to get the balance from, when "protocol" is set to `list`            |                                                                                                                                                                         |             |
+|           |     `disableAddressValidation`     |               if this parameter is set to `true` address validation will be turned off               |                                                                             `true`, `false`                                                                             |   `false`   |
```
justinkaseman, line 51: When would one want to use this feature?

KuphJr: Great question.  Please ask Jonas because I 100% agree.  I see no reason to ever disable these options.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -4,39 +4,55 @@ import { makeOptions } from '../config'
...
-      'The number of confirmations required for a transaction to be counted when getting an address balance',
-    default: 6,
-  },
-  addresses: {
-    required: false,
-    type: 'array',
-    description: 'An array of addresses to get the balance from, when `protocol` is set to `list`',
-  },
-}
+  const inputParameters: InputParameters = {
```
justinkaseman, line 13: Why is this moving within `execute`'s function scope? It means that it can no longer get exported and utilized by the EA framework for things such as cache key generation.

KuphJr: This resolves that issue I had with not being able to set env vars for testing.  I can elaborate further in a huddle, but this is necessary.

KuphJr: The problem is, `makeOptions`, which gets the env vars, runs BEFORE the first line of `adapter.test.ts` if it is left outside of the `execute` function.  So by the time I try to set env vars, the options have already been created so I can't change them.  This is because `makeOptions` is executed when Jest imports all the required deps to run the tests, which is BEFORE any of the tests actually execute.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,209 @@
...
+
+/**
+ * Returns either a valid Ethereum-style address with a valid checksum
+ * or logs a warning and returns undefined
+ */
+const getValidEvmAddress = (address: string): string | undefined => {
+  try {
+    return utils.getAddress(address)
+  } catch (error) {
+    Logger.warn(
```
justinkaseman, line 79: We may want to consider two things:
- Turning this to log level error
- Erroring out completely

We could have the behavior connected to an environment variable.
It depends on how serious it is for an invalid address to be included. A warning log will be extremely easy for a node operator to miss.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
KuphJr: also, the options object is not exported, so it shouldn't have any external impact where the function is located.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
KuphJr: I asked Jonas about this and he preferred a log message.  If there is any debate, I would prefer to leave it as-is unless you think it is critical.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+export const filterDuplicates = (addresses: AddressObject[]): AddressObject[] => {
+  const uniqueMap: Record<string, boolean> = {}
+  const uniqueAddresses: AddressObject[] = []
+  for (const addressObject of addresses) {
+    if (uniqueMap[addressObject.address]) {
+      Logger.warn(
+        { warning: 'Duplicate address detected' },
+        `The address "${addressObject.address}" is duplicated in the request and the duplicate has been removed.`,
```
justinkaseman, line 150: May want to also include the job run ID to help link this warning to the job that is throwing it.

KuphJr: Good call on that. I will add it

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -0,0 +1,209 @@
...
+      )
+    } else {
+      uniqueMap[addressObject.address] = true
+      uniqueAddresses.push(addressObject)
+    }
+  }
+  return uniqueAddresses
+}
+
+const parseBoolean = (value: unknown) => {
```
justinkaseman, line 160: Seems like this may be useful to have in bootstrap's utils if it isn't already there.

KuphJr: Agreed.  I can add it if its not there.

packages/composites/proof-of-reserves/src/utils/protocol.ts
```diff
@@ -23,15 +23,23 @@ export const adapters: AdapterImplementation[] = [
...
 export const runProtocolAdapter = async (
   jobRunID: string,
   context: AdapterContext,
   protocol: Protocol,
-  data: { token: string; chainId: string; network: string } | { addresses: string[] },
+  data: AddressData,
   config: Config,
 ): Promise<AdapterResponse> => {
-  if (protocol === LIST_ADAPTER) return listAdapter(jobRunID, data as { addresses: string[] })
+  if (protocol === LIST_ADAPTER) return listAdapter(jobRunID, data as AddressList)
```
justinkaseman, line 42: Isn't `data` already typed as `AddressList`?

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
alejoberardino: I agree with Justin's points, we do need the input params to be exported before and kept consistent with other EAs. For the particular issue of testing it, you can use Jest to mock and shadow objects in the global scope.

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
alejoberardino: To offer a different perspective, some NOPs have alerting even for warn logs afaik, so we do need to be careful that these are sporadic and if we add this it's for NOPs to be warned

packages/composites/proof-of-reserves/src/utils/protocol.ts
justinkaseman, line 42: `data` is being coerced to `AddressList` here, but we don't yet know that it is actually that type. That validation happens within the function. We should leave it as so and allow the narrowing within the function to get us to the correct type.

packages/composites/proof-of-reserves/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import { server as startServer } from '../../src'
+import { mockPoRindexerSuccess, mockEthBalanceSuccess } from './fixtures'
+import * as nock from 'nock'
+import { AddressInfo } from 'net'
+
+beforeAll(() => {
+  if (process.env.RECORD) {
+    nock.recorder.rec()
+  }
+  process.env.ETH_BALANCE_ADAPTER_URL = 'https://eth-balance-adapter.com'
```
justinkaseman, line 13: When modifying environment variables we should ensure that there are no side effects by saving the current env and restoring it at the end.

See [other integration tests](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/composites/anchor/test/integration/prices/beth.test.ts#L88) as an example.

packages/composites/proof-of-reserves/README.md
justinkaseman: @boxhock

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
justinkaseman: +1 to that this should be resolved within the test and the endpoint maintained

packages/composites/proof-of-reserves/src/utils/addressValidator.ts
justinkaseman: I would think that an invalid address being fed in is a fairly serious thing that should be immediately caught by nops.
I would advocate bumping this to an `error` level log.

@boxhock what do you think?
Though if we have enough nops that have warn log alerting it should be okay this way.

justinkaseman (dismissed):

justinkaseman (approved):

PR #1903, opened 2022-05-27
Enhance Single-command docker app instructions

aelmanaa:
## Description

3 minor changes to help future users :

- instructions in README#single-command-docker-ap must specify that all dependencies mentioned under grafana should be installed. In my case, I ran into issues because jsonnet was not installed
- generate-prom-config.sh script should create generated folder
- I've also removed leading "$" otherwise we cannot easily copy/paste commands:
```
$ yarn setup
zsh: command not found: $
```

## Steps to Test

1. Follow README#single-command-docker-ap

## Quality Assurance

I tested again README#single-command-docker-ap and I was able to boostrap grafana, prometheus & adapters without any issues


justinkaseman (approved):
Thank you for the contribution! :raised_hands: 

Always glad to have things be more usable.

PR #1913, opened 2022-05-31
Metis support for the Layer2 Sequencer Health EA

cds95:
## Description

Adds support to fetch the health of the Metis Sequencer using the Layer2 Sequencer Health EA

## Changes

- Add `metis` as an option for `network` when calling the Metis Sequencer EA

## Steps to Test

1. Run EA
2. Call EA and verify that it returns the latest Sequencer health

## Quality Assurance


packages/sources/layer2-sequencer-health/README.md
```diff
@@ -39,13 +41,17 @@ For the adapter to be useful on the desired network, at least one endpoint (RPC
...
+  "result": 0,
+  "statusCode": 200,
   "data": {
-    "result": true,
-    "isHealthy": true
-  },
-  "statusCode": 200
+    "isHealthy": 0,
+    "result": 0
+  }
```
boxhock, line 55: This doesn't sit right with me:
`"isHealthy": 1 == sequencer is unhealthy`

Can we improve this? Change to `isUnhealthy`?

boxhock (changes requested):

packages/sources/layer2-sequencer-health/README.md
cds95: What about if we just return the Boolean value for isHealthy? I don’t think we can switch the 0/1 value for the result field as we already have feeds that are dependent on 0 representing a healthy sequencer and 1 representing an unhealthy sequencer.

boxhock: Yeah that could work as well. I was not suggesting that we change the 0/1 value, but rather rename the return param `isHealthy` to `isUnhealthy`. I'm guessing we're only using the `result` param, so renaming this additional param wouldn't have any effect?

cds95: Gotcha. Yeah think that we’re only using the result param so this should not have any effect.

boxhock (approved):

PR #1926, opened 2022-06-02
Rename Durin EA to CCIP Read

cds95:


boxhock (approved):
Having a changeset for an EA rename is a bit weird, and will effectively just create a new EA that starts on v2.0.0. Having gone back and forth in my head, I think that it ultimately does make sense - but keep it in mind if you can think of better ways to handle it.

PR #1936, opened 2022-06-07
Update etherchain gasprice endpoint to use /api/gasnow api

mjk90:
Updated etherchain gasprice endpoint to use /api/gasnow api. Updated tests to match new endpoint.

## Steps to Test

1. `yarn test:unit`
2. `yarn test:integration`
3. Manually test etherchain gasprice endpoint

## Quality Assurance


packages/sources/etherchain/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -3,11 +3,11 @@
 exports[`execute gasprice api should return success 1`] = `
 Object {
   "data": Object {
-    "result": 1500000000,
+    "result": 38200000000000000000,
   },
   "jobRunID": "1",
   "providerStatusCode": 200,
-  "result": 1500000000,
+  "result": 38200000000000000000,
```
boxhock, line 10: This result is way higher than the previous result. Does the new API endpoint already return the gas price as wei? If so, we need to remove the `* 1e9` before setting the result.

boxhock (changes requested):

packages/sources/etherchain/test/integration/__snapshots__/adapter.test.ts.snap
mjk90: Ahh good point, it's returning the prices in wei. I'll update it

boxhock (approved):

PR #1943, opened 2022-06-08
Fix/42679 - Reduce metrics cardinality

mjk90:
## Description

- Removed request_origin label from metrics to reduce metrics cardinality. The IP & hostname are being logged for each request so it's not worth it to keep it as a metrics label if it increases the size of metrics data too much.
- Added 'result' to excluded properties in cache key & feed ID generation

## Changes

- Removed request_origin label from metrics & tests
- Added 'result' to excluded properties in cache key & feed ID generation

## Steps to Test

1. `yarn test:unit`
2. `yarn test:integration`
3. Run an EA locally with metric enabled & check that request_origin label is no longer present in metrics

## Quality Assurance

### DIFF
--- a/.changeset/twelve-queens-sell.md
+++ b/.changeset/twelve-queens-sell.md
@@ -2,4 +2,4 @@
 '@chainlink/ea-bootstrap': minor
 ---
 
-Removed request_origin label from metrics
+Removed request_origin label from metrics & tests. Added 'result' to excluded properties in cache key & feed ID generation.
--- a/.changeset/twelve-queens-sell.md
+++ b/.changeset/twelve-queens-sell.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': minor
+---
+
+Removed request_origin label from metrics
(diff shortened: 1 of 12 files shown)


justinkaseman:
Do we know 100% that this was the part causing the problem? Like do we have a before vs. after benchmark?

justinkaseman:
It seems to me like request origin should be relatively stable.

boxhock (dismissed):

mjk90:
> Instead of outright removing it, have you considered just removing the outbound port?
> 
> If that's not a viable option or if the request origin is not needed in metrics, then this PR LGTM

I did consider that but I though the overhead of finding and replacing a string on each request would probably not be worth it

.changeset/twelve-queens-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': minor
+---
+
+Removed request_origin label from metrics
```
boxhock, line 5: Should add a comment about the `result` key being removed as welll

boxhock (changes requested):
One small comment for the changeset, otherwise LGTM

justinkaseman:
Not for this PR, but something I am thinking about is how we can better catch issues like this in the future. Maybe setting up an alert on the cardinality of our metrics with something like https://github.com/thought-machine/prometheus-cardinality-exporter

PR #1945, opened 2022-06-08
feature/Created new codified dashboard EA Release

silaslenihan:
## Description
This PR contains a new codified dashboard called EA Release that can be easily deployed using the grafana/scripts/deploy.sh script.
......

## Changes
- Created eaRelease.jsonnet and altered the relevant scripts to enable deployment. 

1. Launch a few adapters (one websocket, one regular) with Grafana and Prometheus
2. Send requests to each adapter, some valid, some purposefully to error.
3. Launch deployment script, make sure it runs and see that dashboard panels are filled in with relevant info.

## Quality Assurance


grafana/src/eaRelease.jsonnet
```diff
@@ -0,0 +1,276 @@
...
+    content='# Inbound Requests (Node → EA)',
+);
+
+local ErrorRate5xxPanel = graphPanel.new(
+  title='Top 10 - 5xx Error Rate',
+  datasource=cortexDataSource,
+  format='percent',
+).addTarget(
+  prometheus.target(
+    'topk(10, sum by(app_name) (rate(http_requests_total{' + instanceFilter + ',status_code=~"^5..",app_name=~"$adapter.*"}' + interval + ')) / sum by(app_name) (rate(http_requests_total{app_name!="",' + instanceFilter + ',app_name=~"$adapter.*"}' + interval + ')) * 100)'
```
alejoberardino, line 30: Check out this query in the dashboard, because it uses the nginx metric instead of the adapter one; the former is much more reliable

grafana/src/eaRelease.jsonnet
```diff
@@ -0,0 +1,276 @@
...
+local graphPanel = grafana.graphPanel;
+local timeseries = grafana.timeseries;
+local textPanel = grafana.text;
+local prometheus = grafana.prometheus;
+
+local shared = import './shared.libsonnet';
+local addSideLegend = shared.helpers.addSideLegend;
+
+local cortexDataSource = shared.constants.cortexDataSource;
+local instanceFilter = shared.constants.instanceFilter;
```
alejoberardino, line 12: This instance filter also adds a filter for service, which unfortunately won't work for all queries in the dashboard (nginx vs ea metrics)

grafana/src/eaRelease.jsonnet
```diff
@@ -0,0 +1,276 @@
...
+  )
+);
+
+local Bot10CacheHitTotalReqCount = graphPanel.new(
+  title='Bot 10 - (Cache Hit / Total Req Count) Ratio',
+  datasource=cortexDataSource,
+  format='percent',
+).addTarget(
+  prometheus.target(
+    'bottomk(10, (sum(rate(http_requests_total{' + instanceFilter + ' ,type=\"cacheHit\",app_name=~"$adapter.*"}' + interval + ')) by (app_name) / sum(rate(http_requests_total{' + instanceFilter + ' ,app_name=~"$adapter.*"}' + interval + ')) by (app_name))) * 100'
```
alejoberardino, line 50: This should filter by service instead of app_name, to differentiate between sub-adapters (i.e. adapter-x, adapter-x-solana)

alejoberardino (changes requested):

grafana/src/eaRelease.jsonnet
silaslenihan: I think we changed it to be http_requests_total instead of nginx because this version is supposed to be for the node operators to make, and they don't have nginx set up

grafana/src/eaRelease.jsonnet
silaslenihan: I don't see a filter field in http_requests_total, should I use the instance field instead?

grafana/src/eaRelease.jsonnet
silaslenihan: I created a new filter namespaceFilter that only filters namespace and leaves the service unfiltered and replaced all instances of instanceFilter with namespaceFilter in the code, Does that resolve this issue?

alejoberardino: The adapter label in instanceFilter will clash with the service filter it already has, better to leave that one alone. Idk also 
 if that service label is coming from k8s, I think so

silaslenihan: What is this service filter doing? I don't see how it effects anything?

alejoberardino (approved):

PR #1949, opened 2022-06-09
transform ether.js rpc errors

### LINKED ISSUE #1942: Improve clarity of RPC errors
### Description
Communication to various JSON-RPC available blockchains is handled using [ethers.js](https://docs.ethers.io/v5/single-page/).
When a revert or error occurs within this framework the EA will return a non-descriptive 500 error.
It will be helpful for debugging to have the exact cause more clearly stated in logs & EA responses.

### Suggested solutions
Common error description strings could be transformed into a more human friendly description.
Handling RPC level errors in a re-usable and extensible way would be preferred. For example keeping a map of error strings to transformed error strings in the core framework and then importing it in EAs. That way more could be added in the future.

It may be possible to inject error handling logic into ethers. The documentation should be checked for this possibility.

If not, errors could be caught in the EA framework's error handling logic.

### Reproduction
1. Start the `curve` adapter with a garbage `RPC_URL='no'` set.
2. Send the running adapter a request:
```
{
    "data": {
        "from": "LINK",
        "to": "USD"
    }
}
```
3. Notice that the respone's error message could be more clear that the problem is that the provided RPC_URL could not be connected to:
```
{
    "jobRunID": "1",
    "status": "errored",
    "statusCode": 500,
    "error": {
        "name": "AdapterError",
        "message": "could not detect network (event=\"noNetwork\", code=NETWORK_ERROR, version=providers/5.4.5)",
        "feedID": "LINK/USD"
    }
}
```

### Completion
Completion of this ticket looks like the ability to have error handling coverage for all of the different adapters that use Ethers.
At a minimum this would be turning the message "could not detect network (event=\"noNetwork\", code=NETWORK_ERROR, version=providers/5.4.5)" into "The provided RPC_URL environment variable of ${RPC_URL} could not be connected to.".

These can be found by searching `ethers.Contract(` within the repository.

Some adapters use others adapters to make RPC calls for them (e.g. JSON-RPC). It should be ensured that in these cases the original adapter still displays the correct error.

karen-stepanyan:
## Description

This PR adds a list of error types from ether.js mapped to descriptive messages. Currently there are two error types that were mapped because in general other error types have clear messages. [This](https://github.com/ethers-io/ethers.js/blob/a71f51825571d1ea0fa997c1352d5b4d85643416/packages/logger/src.ts/index.ts#L53) is the list of all possible error codes from ether.js. In the future the list can be modified and extended. 
......

## Changes

- Added check for ether.js errors to map and transform  error message
- Renamed `providerResponseStatusCode` to `providerStatusCode`. I didn't find any usage of `providerResponseStatusCode` in the project

## Quality Assurance


packages/core/bootstrap/src/lib/errors.ts
```diff
@@ -1,5 +1,14 @@
...
+
 export const HTTP_ERROR_NOT_IMPLEMENTED = 501
 
 export const HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE = 415
 export const HTTP_ERROR_UNSUPPORTED_MEDIA_TYPE_MESSAGE =
   'Only application/json media type is supported.'
+
+export const RPCErrorMap = {
+  NETWORK_ERROR: `The provided RPC_URL environment variable ${
+    process.env.RPC_URL || process.env.ETHEREUM_RPC_URL
```
justinkaseman, line 11: We should use `util.getEnv` for this

packages/core/bootstrap/src/lib/errors.ts
```diff
@@ -1,5 +1,14 @@
...
+  } could not be connected to.`,
+  TIMEOUT: 'Timeout exceeded',
```
justinkaseman, line 13: Let's make this more descriptive
```suggestion
  TIMEOUT: 'Request to the RPC has timed out',
```

packages/core/bootstrap/src/lib/errors.ts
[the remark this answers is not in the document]
karen-stepanyan: @boxhock No, only message, type and param which in this particular case contains `event: "noNetwork"` .

packages/core/bootstrap/src/lib/util.ts
```diff
@@ -537,3 +537,21 @@ export const registerUnhandledRejectionHandler = (): void => {
...
+
+export const mapRPCErrorMessage = (errorCode: string, errorMessage: string): string => {
+  // Try to transform error message if error is thrown from ether.js
+  if (
+    errorCode &&
+    errorMessage &&
+    RPCErrorMap[errorCode as keyof typeof RPCErrorMap] &&
+    errorMessage.includes('version')
+  ) {
+    return RPCErrorMap[errorCode as keyof typeof RPCErrorMap]
```
KuphJr, line 554: why do we need to use "as keyof typeof" here?  Shouldn't this work without explicitly defining the type?

KuphJr (approved):

KuphJr (dismissed):

KuphJr (approved):

PR #1958, opened 2022-06-12
Added validation for Filecoin miner addresses ("f0" format)

KuphJr:


packages/composites/proof-of-reserves/src/utils/addressValidator.ts
```diff
@@ -162,6 +162,7 @@ const getValidFilecoinAddress = (id: string, address: string): string | undefine
     )
     return
   }
+  if (address[1] === '0' && isBase10(address.slice(2))) return address
```
austinborn, line 165: What was the reason to omit `address[0] === 'f'`?

KuphJr: Line 151 checks to ensure the address prefix is either f or t (f for mainnet, t for testnet).  Then, I strip away the prefix and validate that the address is either base10 or base32.  Only addresses starting with `f0` or `t0` can be base10.  All the other address types must be base32.

austinborn (approved):
Thanks for the response, this looks good!

PR #1988, opened 2022-06-28
Add better logging to catch cache warmer error

cds95:


boxhock (commented):
Instead of making this a temp change, can we enable this extra logging when `DEBUG=true` is set, or something like that?

boxhock (approved):

PR #2019, opened 2022-07-12
Publish shasum artifacts for the release workflow to consume by downs…

itmustbejj:
…tream actions

## Changes

- Publishes shasum digests as github artifacts to be aggregated by the consume-release job downstream.


justinkaseman (approved):
I'll test with the hotfix release that I am about to put out.

PR #2039, opened 2022-07-22
Add support for Starkware to layer2 sequencer health EA

cds95:
## Description

Add support to `layer2-sequencer-health` EA to track Starkware Sequencer health

## Changes

- Update Layer2 Sequencer Health EA to check for the Starkware Sequencer's health.
- This is a two step process
   1. Poll Sequencer directly for the pending block to see if either a new block is produced or if there are new pending transactions
   2. If step 1 fails, then try send a dummy transaction to the network and see if we get back an expected error.  Receiving an expected error means that the Sequencer is still healthy.

## Steps to Test

 - Run unit and integration tests
 - Run EA
 - Send request payload below and verify that the EA returns a successful response.  Repeat this with other networks.
 ```
{
    "data": {
        "network": "starkware"
    }
}
```

## Quality Assurance

### CODE REVIEW

File: packages/sources/layer2-sequencer-health/src/config/index.ts [hunk 1]
```diff
@@ -46,14 +51,30 @@ export const HEALTH_ENDPOINTS = {
...
+    gatewayUrl: string
+    argentAccountAddr: string
+  }
 }
 
+const DEFAULT_STARKWARE_SEQUENCER_ENDPOINT = 'https://alpha-mainnet.starknet.io'
+const DEFAULT_STARKWARE_FEEDER_GATEWAY_URL = 'feeder_gateway'
+const DEFAULT_STARKWARE_GATEWAY_URL = 'gateway'
+const DEFAULT_STARKWARE_ARGENT_ACCOUNT_ADDR =
+  '0xd175dcf2fcf9d858d8d686826d56db7aeb29c3490110b4cfe0d8442944b828'
```
krebernisak, line 76: Is this an account deployed to testnet or mainnet? 
How do you envision handling this for any network?

cds95: this is an account I deployed to mainnet.  We can handle this for other networks by deploying to that network and setting an environment variable to overwrite the default.  The reason I've added a default is because EngOps often asks for this so that NOPs don't have to worry about it when running the EA.

packages/sources/layer2-sequencer-health/src/config/index.ts
```diff
@@ -46,14 +51,30 @@ export const HEALTH_ENDPOINTS = {
...
+    responsePath: [],
+  },
 }
 
 export interface ExtendedConfig extends Config {
   delta: number
   deltaBlocks: number
   timeoutLimit: number
+  starkwareConfig: {
+    sequencerUrl: string
```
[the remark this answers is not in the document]
cds95: so we don't actually have to do this but I'm trying to follow a similar pattern to other EAs where we pass the env vars through the config object returned in `makeConfig()`.

packages/sources/layer2-sequencer-health/src/endpoint/health.ts
```diff
@@ -63,18 +65,18 @@ export const makeNetworkStatusCheck = (
   }
 }
 
-const networks: Record<Networks, (delta: number, deltaBlocks: number) => Promise<boolean>> = {
+const networks: Record<Networks, (config: ExtendedConfig) => Promise<boolean>> = {
   [Networks.Arbitrum]: makeNetworkStatusCheck(Networks.Arbitrum),
   [Networks.Optimism]: makeNetworkStatusCheck(Networks.Optimism),
   [Networks.Metis]: makeNetworkStatusCheck(Networks.Metis),
+  [Networks.Starkware]: checkStarkwareSequencerPendingTransactions(),
```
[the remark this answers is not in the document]
cds95: So `makeNetworkStatusCheck` isn't doing the same thing as `checkStarkwareSequencerPendingTransactions`.  `makeNetworkStatusCheck` is checking that the block height is growing after `delta` minutes whilst `checkStarkwareSequencerPendingTransactions` is specifically pinging the Sequencer and checking the list of pending transactions.  Maybe we should rename to `makeNetworkStatusCheck` to `checkBlockHeightIsGrowing` to make it clearer?

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,140 @@
...
+    ) {
+      Logger.debug(
+        `Skipping check for Starkware Sequencer health as it has been less than ${delta} seconds since last check`,
+      )
+      return lastSeenPendingBlock.lastStatus
+    }
+    let pendingBlockParams: StarkwareBlockResponse
+    try {
+      const { data } = await Requester.request<StarkwareBlockResponse>({
+        url: `${config.starkwareConfig.sequencerUrl}/${config.starkwareConfig.feederGatewayUrl}/get_block?blockNumber=pending`,
```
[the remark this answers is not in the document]
cds95: so I'm initializing a provider instance somewhere else like this

```
const provider = new Provider({
    baseUrl: config.starkwareConfig.sequencerUrl,
    feederGatewayUrl: config.starkwareConfig.feederGatewayUrl,
    gatewayUrl: config.starkwareConfig.gatewayUrl,
  })
```

is this ok with you?  The `RPCProvider` class does not seem to be included in the latest `3.18.2` release

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,140 @@
+import { Logger, Requester } from '@chainlink/ea-bootstrap'
+import { DEFAULT_PRIVATE_KEY, ExtendedConfig, Networks } from './config'
+import { ec, Account, AddTransactionResponse, Provider, ProviderInterface } from 'starknet'
+import { race } from './network'
+
+export interface StarkwareBlockResponse {
```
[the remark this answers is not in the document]
cds95: removed redeclared interface

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
krebernisak: Yeah, I've asked ShardLabs in the shared channel about the next release.

I'd prefer we migrate to the RPCProvider.

In the meantime, we can try pulling from git directly: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#git-urls-as-dependencies

packages/sources/layer2-sequencer-health/src/config/index.ts
```diff
@@ -64,5 +84,15 @@ export const makeConfig = (prefix?: string): ExtendedConfig => {
   const deltaBlocks = Number(util.getEnv('DELTA_BLOCKS', prefix)) || DEFAULT_DELTA_BLOCKS
   const timeoutLimit = Number(util.getEnv('NETWORK_TIMEOUT_LIMIT', prefix)) || DEFAULT_TIMEOUT_LIMIT
   config.defaultEndpoint = DEFAULT_ENDPOINT
-  return { ...config, delta, deltaBlocks, timeoutLimit }
+  const starkwareConfig: ExtendedConfig['starkwareConfig'] = {
```
krebernisak, line 87: We should only append this config if we're running a starknet instance of the EA.

packages/sources/layer2-sequencer-health/src/endpoint/health.ts
```diff
@@ -7,19 +7,20 @@ import {
...
 import {
   requestBlockHeight,
   getSequencerHealth,
   NetworkHealthCheck,
   getStatusByTransaction,
 } from '../network'
+import { checkStarkwareSequencerPendingTransactions } from '../starkware'
 
-export const supportedEndpoints = ['health']
+export const supportedEndpoints = ['health', 'health-evm']
```
krebernisak, line 19: Why did we add this in? It doesn't seem used.

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
krebernisak: Version [4.1.0](https://www.npmjs.com/package/starknet/v/4.1.0) is available on NPM.

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,118 @@
...
+      Logger.debug(
+        `Skipping check for Starkware Sequencer health as it has been less than ${delta} seconds since last check`,
+      )
+      return lastSeenPendingBlock.lastStatus
+    }
+    let pendingBlockParams: GetBlockResponse
+    try {
+      pendingBlockParams = await config.starkwareConfig.provider.getBlock('pending')
+    } catch (e: any) {
+      if (e.providerStatusCode === 504) {
```
krebernisak, line 77: We should retry 2 times with some backoff, before marking it unhealthy. It could just be flaky.

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,118 @@
...
+      currentTime - lastSeenPendingBlock.lastUpdated < delta
+    ) {
+      Logger.debug(
+        `Skipping check for Starkware Sequencer health as it has been less than ${delta} seconds since last check`,
+      )
+      return lastSeenPendingBlock.lastStatus
+    }
+    let pendingBlockParams: GetBlockResponse
+    try {
+      pendingBlockParams = await config.starkwareConfig.provider.getBlock('pending')
```
krebernisak, line 75: Consider extracting this part in a separate function as you did with the `checkBatcherHealthy`. You explain two workflows in the main fn, so let's have them encoded as two internal functions.

krebernisak: It's gonna make the code more readable!

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,118 @@
...
+      lastUpdated: currentTime,
+      lastStatus: isBatcherHealthy,
+    }
+    return isBatcherHealthy
+  }
+}
+
+const checkBatcherHealthy = (
+  lastSeenPendingBlock: PendingBlockState,
+  pendingBlockParams: GetBlockResponse,
```
krebernisak, line 99: Two different types used to compare two of the same concepts (last seen & current pending block) make it hard to read/reason.

```suggestion
  pendingBlockPrevious: GetBlockResponse,
  pendingBlockCurrent: GetBlockResponse,
```

packages/sources/layer2-sequencer-health/src/config/index.ts
krebernisak: If that is the case, please add in-line comments documenting this.

Please also take this comment into account before you do that: https://github.com/smartcontractkit/external-adapters-js/pull/2039#discussion_r937584988

packages/sources/layer2-sequencer-health/src/endpoint/health.ts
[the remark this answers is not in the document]
krebernisak: No, I suggest the opposite. To define a single fn call API as `makeNetworkStatusCheck(network)` (like you already have), and you check your status in an appropriate way internally, depending on the network.

So Optimistic rollups can check " that the block height is growing after delta minutes" while for StarkNet we check that the number of pending transactions is growing as expected.

Additionally, did we consider writing tests for this:

- How to test this on StarkNet Devnet (simulator) where 1 tx == 1 block?
-

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
cds95: hmm I need to try this but if I'm not wrong we need to ping the Sequencer directly to get the list of pending transactions.  This is because they haven't implemented a gossip protocol to obtain the info from full nodes.  Can we use `SequencerProvider` instead for now?

packages/sources/layer2-sequencer-health/src/endpoint/health.ts
cds95: ah this was leftover.  Just removed this.

packages/sources/layer2-sequencer-health/src/config/index.ts
cds95: So every instance of this EA typically supports fetching the health of all networks.  The only way we'd be able to differentiate between EAs that run Starknet vs those that don't is to set another env var.  I think we shouldn't do this though as it adds an unnecessary env var.

packages/sources/layer2-sequencer-health/src/starkware.ts
cds95: These are different.  `PendingBlockState` stores the last status and last updated at timestamp, which used internally by the EA.  I'll rename to make it clearer.

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,118 @@
...
+      { maxFee: '0' },
+    ),
+    error: `Transaction receipt not received in ${config.timeoutLimit} milliseconds`,
+  })
+  Logger.info(
+    `Transaction receipt received with hash ${receipt.transaction_hash} for EVM network: ${Networks.Starkware}`,
+  )
+  await provider.waitForTransaction(receipt.transaction_hash)
+  const tx = await provider.getTransaction(receipt.transaction_hash)
+  return tx.status !== 'REJECTED'
```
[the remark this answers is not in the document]
cds95: It'll actually throw an error and we won't actually reach this point.  Actually looking at this again we'd never actually reach this point.  I think I was trying to follow up something like this https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/layer2-sequencer-health/src/network.ts#L120 where we check something.  I may be wrong but shouldn't we also never reach the check on line 120?

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,118 @@
...
+  const starkKeyPub = ec.getStarkKey(starkKeyPair)
+  const provider = config.starkwareConfig.provider
+  const account = new Account(provider, config.starkwareConfig.argentAccountAddr, starkKeyPair)
+
+  const receipt = await race<AddTransactionResponse>({
+    timeout: config.timeoutLimit,
+    promise: account.execute(
+      {
+        contractAddress: config.starkwareConfig.argentAccountAddr,
+        entrypoint: 'initialize',
```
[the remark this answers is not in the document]
cds95: 1.  It should work as expected as we are setting `max_fee` to be 0 so we should always be getting the `max_fee must be bigger than 0.\n0 >= 0` error.
2. I would actually prefer to do it with a fake txn but I couldn't find a way to do this in their docs.  Did you come across anything?

packages/sources/layer2-sequencer-health/src/config/index.ts
krebernisak: You're right!

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
krebernisak: > I would actually prefer to do it with a fake txn but I couldn't find a way to do this in their docs. Did you come across anything?

My question is will it fail with some other error if the specific contract was not pre-deployed? If yes, then this is basically a fake tx that never goes over the `max_fee` check. 

If it fails with a different error maybe we could do two checks:
1. underfunded and fake tx to a nonexistent contract
2. underfunded tx to an existing contract

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
krebernisak: Then let's revisit this and remove code that isn't necessary, like that REJECTED check as it will never happen.

packages/sources/layer2-sequencer-health/src/starkware.ts
krebernisak: I know they are currently different, but you can avoid having that custom type altogether and just use  `PendingBlockState` + in-line calculations.

Slightly more computation for more readability.

packages/sources/layer2-sequencer-health/src/config/index.ts
cds95: todo:  Check to see if we still get 0 gas error with 0 address.

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
cds95: so I tried this with a dummy account `0x00000000000000000000000000000000000000000000000000000000000001` and got a different error message saying `StarknetErrorCode.UNINITIALIZED_CONTRACT`. 

I've updated the code to use this dummy address and check for either of the following errors:

 - `StarknetErrorCode.UNINITIALIZED_CONTRACT`
 - `'StarknetErrorCode.OUT_OF_RANGE_FEE`

The first error will get thrown when the dummy contract is still uninitialized.  The second error is a safeguard incase someone goes ahead and initializes the dummy contract.  In this case, the dummy transaction will fail with the `OUT_OF_RANGE_FEE` error due to the EA setting `max_fee` to 0.

packages/sources/layer2-sequencer-health/src/evm.ts
```diff
@@ -0,0 +1,120 @@
+import {
```
cds95, line 1: This is just code moved from `health.ts`

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -34,120 +47,87 @@ export const getSequencerHealth: NetworkHealthCheck = async (
   return isHealthy
 }
 
-export const requestBlockHeight = async (network: Networks): Promise<number> => {
```
cds95, line 37: moved to `evm.ts`

packages/sources/layer2-sequencer-health/src/config/index.ts
```diff
@@ -1,8 +1,11 @@
 import { AdapterConfigError, Requester, util } from '@chainlink/ea-bootstrap'
 import { AdapterContext, Config } from '@chainlink/ea-bootstrap'
 import { envDefaultOverrides } from './envDefaultOverrides'
+import { SequencerProvider } from 'starknet'
 
 export const NAME = 'L2_SEQUENCER_HEALTH'
+export const DEFAULT_PRIVATE_KEY =
```
cds95, line 7: This was just moved from a different file.  It was previously used to instantiate a dummy wallet to make dummy transactions to EVM chains.

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -34,120 +50,99 @@ export const getSequencerHealth: NetworkHealthCheck = async (
...
-    [Networks.Metis]: ['cannot accept 0 gas price transaction'],
+  let isSequencerHealthy = true
+  try {
+    Logger.info(`Submitting empty transaction for network: ${network}`)
+    await sendEmptyTransaction(network, config)
+  } catch (e) {
+    isSequencerHealthy = isExpectedErrorMessage(network, e as Error)
   }
+  return isSequencerHealthy
+}
```
krebernisak, line 65:
```suggestion
  try {
    Logger.info(`Submitting empty transaction for network: ${network}`)
    await sendEmptyTransaction(network, config)
  } catch (e) {
    return isExpectedErrorMessage(network, e as Error)
  }
  return true
}
```

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
krebernisak: This is a StarkNet network error, so it's not that you are calling an account contract that is not initialized (has an init fn), but you are calling a contract that doesn't exist (not deployed, not initialized in the network) so the network returns an error.

At first I thought you are going to do 2x transactions and check specific error, but looks like you just do 1x tx and leave room for that contract to be deployed or not, right?

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,135 @@
...
+ *  - The batcher is considered to be down if we a new block has not been produced AND no new transactions have been added to the
+ *  pending block within a given time interval.
+ * status when fetching
+ * @param networks
+ * @returns true if Sequencer is healthy
+ */
+export const checkStarkwareSequencerPendingTransactions = (): ((
+  config: ExtendedConfig,
+) => Promise<boolean>) => {
+  let recordedStarkwareState: StarwareState = {
```
krebernisak, line 48: Verbose to read, suggestion:

```suggestion
  let state: StarwareState = {
```

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,135 @@
...
+      promise: async () => config.starkwareConfig.provider.getBlock('pending'),
+      retryConfig: config.retryConfig,
+    })
+  } catch (e: any) {
+    if (e.providerStatusCode === 504) {
+      Logger.warn(
+        `Request to fetch pending block timed out.  Status Code: ${e.providerStatusCode}.  Sequencer: UNHEALTHY`,
+      )
+    } else {
+      throw e
```
krebernisak, line 103: Where is this error actually caught?

Even if it is, it's very hard to follow the code. I couldn't find who handles this. I suggest we should handled it here, and not throw.

packages/sources/layer2-sequencer-health/src/starkware.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+const checkBatcherHealthy = (
+  recordedStarkwareState: GetBlockResponse | null,
+  pendingBlockParams: GetBlockResponse,
+): boolean => {
+  if (!recordedStarkwareState) {
+    return pendingBlockParams.transactions.length > 0
+  }
```
krebernisak, line 117:
```suggestion
const checkBatcherHealthy = (
  blockPrevious: GetBlockResponse | null,
  blockCurrent: GetBlockResponse,
): boolean => {
  if (!blockPrevious) return blockCurrent.transactions.length > 0
```

packages/sources/layer2-sequencer-health/src/config/index.ts
```diff
@@ -63,6 +85,39 @@ export const makeConfig = (prefix?: string): ExtendedConfig => {
...
+  }
+
+  return { ...config, delta, deltaBlocks, timeoutLimit, retryConfig, starkwareConfig }
+}
+
+const DEFAULT_STARKWARE_SEQUENCER_ENDPOINT = 'https://alpha-mainnet.starknet.io'
+const DEFAULT_STARKWARE_FEEDER_GATEWAY_URL = 'feeder_gateway'
+const DEFAULT_STARKWARE_GATEWAY_URL = 'gateway'
+const DEFAULT_STARKWARE_DUMMY_ACCOUNT_ADDRESS =
+  '0x00000000000000000000000000000000000000000000000000000000000001'
```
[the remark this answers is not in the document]
cds95: `env.json` is currently only used for generating the ReadMe file.  We don't typically read env vars from there.

packages/sources/layer2-sequencer-health/src/endpoint/health.ts
```diff
@@ -133,15 +72,15 @@ export const execute: ExecuteWithConfig<ExtendedConfig> = async (request, _, con
...
+    const isHealthy = await method(network, config)
     if (!isHealthy) {
       Logger.info(`Checking unhealthy network ${network} with transaction submission`)
       let isHealthyByTransaction
       try {
-        isHealthyByTransaction = await getStatusByTransaction(network, config.timeoutLimit)
+        isHealthyByTransaction = await getStatusByTransaction(network, config)
       } catch (e: any) {
         throw new AdapterDataProviderError({
           network,
```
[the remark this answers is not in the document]
cds95: Right now we just respond with `ok` and we don't do the check again.  We can definitely improve the workflow but I think that this is a larger change as it would impact the flow for all other networks.  Can we create a separate ticket and discuss this for the future?

packages/sources/layer2-sequencer-health/src/starkware.ts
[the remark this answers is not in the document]
cds95: Yes that's right.  We're only sending one transaction here and checking to see if we get an expected error.  The second error is a safety net incase someone actually deploys and initializes the contract.

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -34,120 +50,99 @@ export const getSequencerHealth: NetworkHealthCheck = async (
...
       [Networks.Metis]: ['error', 'message'],
+      [Networks.Starkware]: ['errorCode'],
     }
     return (Requester.getResult(e, paths[network]) as string) || ''
   }
+  const error = e as Error
+  if (sequencerOnlineErrors[network].includes(_getErrorMessage(error))) {
+    Logger.debug(`Transaction submission failed with an expected error ${_getErrorMessage(error)}.`)
+    return true
+  }
```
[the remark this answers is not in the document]
cds95: See comment above

packages/sources/layer2-sequencer-health/src/starkware.ts
cds95: The error is caught [here](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/layer2-sequencer-health/src/endpoint/health.ts#L123).  Similar to this [comment](https://github.com/smartcontractkit/external-adapters-js/pull/2039#discussion_r942593110), this has been done this way as it was the general flow the EA took to check for the Sequencer health.  I do agree that this EA needs a general refactor but I'm worried that we're introducing too many changes in this PR.

krebernisak (approved):

krebernisak (approved):
Nice work, @cds95 🙌  LGTM!

krebernisak (approved):

PR #2041, opened 2022-07-22
[SC-35528] Extend Tiingo EA to handle inverse prices

amit-momin:
## Description

Added functionality to the Tiingo EA to handle inverse currency pairs. The market convention is for only queen's currencies (AUD, NZD, EUR, and GBP) to be quoted with USD as the quote. In all other cases, USD should be used as the base. To follow this convention, the existing non-queen's currency pairs used by feeds (CAD, CHF, and JPY) were configured to pull their rates by their inverses.

Updated other EA's to avoid future tech debt.

Note: Future feeds with new non-queen's currency pairs against USD will require an update to the includes.json configuration 

## Changes

- Utilized getPairOptions() in the core util to retrieve and use inverse configurations
- Used the inverse option in the Requester validateResultNumber() method to invert the results if needed
- Added includes.json to Tiingo to store the inverse currency pair configurations
- Removed Tiingo from the readme generator blacklist
- Updated getPairOptions() logic to use 2 distinct methods for batching and non-batching to avoid type coercion at the EA level
- Updated Amberdata, Kaiko, and Openexchangerates to use the new methods and to remove the type coercion
- Updated Kaiko and Openexchangerates to use _ instead of lint ignore tags

## Steps to Test

1. yarn test tiingo/test/unit
2. yarn test tiingo/test/integration
3. yarn test packages/core/bootstrap/test/unit/utils.test.ts

## Quality Assurance


packages/sources/tiingo/README.md
```diff
@@ -230,6 +230,8 @@ Aliases: `fx`, `commodities`
 
 https://api.tiingo.com/documentation/forex
 
+This endpoint has the ability to leverage inverses in the scenario a specific pair exists but not its inverse on the Tiingo forex API.
```
justinkaseman, line 233: This will get lost because the READMEs are auto-generated from code.

Add it to `packages/sources/tiingo/src/endpoint/forex.ts`'s `description`.

packages/sources/tiingo/src/endpoint/forex.ts
```diff
@@ -37,15 +38,38 @@ interface ResponseSchema {
...
+  inverse?: boolean
+}
+
+const getUrl = (from: string, to: string) => ({
+  url: util.buildUrlPath('/tiingo/fx/:ticker/top', { ticker: `${from}${to}`.toLowerCase() }),
+})
+
+const getIncludesOptions = (
+  //@ts-expect-error no-unused-vars
+  validator: Validator<TInputParameters>,
```
justinkaseman, line 52:
```suggestion
  _: Validator<TInputParameters>,
```
Declaring the unused parameter as an underscore will tell the TS transpiler that it is unused.

packages/sources/tiingo/src/endpoint/forex.ts
```diff
@@ -37,15 +38,38 @@ interface ResponseSchema {
...
-  const ticker = `${base}${quote}`.toLowerCase()
   const resultPath = (validator.validated.data.resultPath || '').toString()
-  const url = util.buildUrlPath('/tiingo/fx/:ticker/top', { ticker })
+
+  const { url, inverse } = util.getPairOptions<TOptions, TInputParameters>(
+    AdapterName,
+    validator,
+    getIncludesOptions,
+    getUrl,
+  ) as TOptions // If base and quote cannot be batched, getPairOptions will return TOptions
```
justinkaseman, line 72: I wonder if we can improve this. I don't like needing to coerce the type here.
It was introduced [here](https://github.com/smartcontractkit/external-adapters-js/pull/2028/files#diff-db405e686a22de740fe2caac9fa7e96ac17b56ab1ff5fa2ff40d49453a025fa2R601). I've reached out to Austin for context.

austinborn: This may be better as 2 distinct methods, like `getPairOptions` and a wrapper `getBatchedPairOptions` (or distinct methods with common helpers). That would prevent the coercion here.

packages/sources/tiingo/src/endpoint/forex.ts
austinborn: Could you do the same in `kaiko` since I did the same thing there originally?

packages/sources/tiingo/README.md
austinborn: In src/endpoint/forex.ts:
```
export const description = "..."
```

packages/sources/tiingo/src/endpoint/forex.ts
amit-momin: I see the same in openexchangerates too. Should I take care of both of those under this PR?

packages/sources/tiingo/README.md
amit-momin: I've updated this in the latest commit. But I noticed afterwards Tiingo is in the readmeBlacklist.json.

packages/sources/tiingo/src/endpoint/forex.ts
justinkaseman: Yes please!

packages/sources/tiingo/src/endpoint/forex.ts
justinkaseman: We could probably also do this with a [conditional type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) based on the input. Amit would you like to flex some Typescript muscles and take a stab at this?

[Here](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/core/bootstrap/src/lib/modules/requester.ts#L388) is an example where I have done this before in the codebase.

amit-momin: Sure I'll take a shot at it!

packages/sources/tiingo/README.md
austinborn: Hmm I wonder why it's still on the blacklist. Can you try removing it and running `yarn generate:readme tiingo` and see if it overwrites the README properly?

amit-momin: The generated README looks good to me

amit-momin: The latest commit should take care of this

packages/sources/tiingo/src/endpoint/forex.ts
amit-momin: The latest commit includes the change for Kaiko and Openexchangerates

packages/sources/tiingo/src/endpoint/forex.ts
amit-momin: After some discussions, we decided to go with the 2 disctinct methods approach. I updated Amberdata, Kaiko, Openexchangerates, and Tiingo accordingly

justinkaseman (approved):

PR #2083, opened 2022-08-05
Updated NCFX Subscription handling

mjk90:
- Updated NCFX Subscription handling
- Updated NCFX README

### DIFF
--- a/.changeset/dull-hounds-remain.md
+++ b/.changeset/dull-hounds-remain.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ncfx-adapter': patch
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/dull-hounds-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ncfx-adapter': patch
```
justinkaseman, line 2: We don't need a changeset. There are no source code changes, which means it does not need to re-build a new image.

justinkaseman:
Can we clean up this git history? Looks like a rebase will help.

austinborn (approved):

PR #2098, opened 2022-08-10
Update 'consume-release' workflow for image-dispatcher

justinkaseman:
### Description
Trigger for image-dispatcher after release workflow runs.


itmustbejj (dismissed):

itmustbejj (approved):

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
   workflow_run:
     workflows: [Release]
     types: [completed]
+    branches: [develop]
 
 jobs:
-  on-success:
+  get-images:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
HenryNguyen5, line 13:
```suggestion
    if:  github.event.workflow_run.conclusion == 'success' 
```

You can omit the expression evaluation syntax here, GHA treats `if` statements as expressions by default :) https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+    runs-on: ubuntu-latest
     environment: InfraK8s
+    steps:
+      - id: get-artifacts
+        name: 'Get artifacts from release workflow'
+        uses: actions/github-script@v6
```
HenryNguyen5, line 19: I would recommend pinning this to a specific sha instead of a version, we try to do this as a best practice to avoid pulling in a malicious version bump

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+        env:
+          AWS_PRIVATE_ECR_PATH: ${{ secrets.AWS_PRIVATE_ECR_PATH }}
+        with:
+          result-encoding: string
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
```
HenryNguyen5, line 25:
```suggestion
            const allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
```

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+            const adapterImages = allArtifacts.data.artifacts.map((artifact) => {
+              const basePath = process.env.AWS_PRIVATE_ECR_PATH + '/adapters/'
+              return basePath + artifact.name.replace('-shasum', '')
+            });
```
HenryNguyen5, line 33:
```suggestion
	    // Create whitespace separated URLs for downloading ECR adapter images based on uploaded workflow artifacts 
	    const pathPrefix = process.env.AWS_PRIVATE_ECR_PATH
            return allArtifacts.data.artifacts
              .map(({name}) => `${pathPrefix}/adapters/${name.replace('-shasam', '')}`)
              .join(' ')
```

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+        name: 'Get artifacts from release workflow'
+        uses: actions/github-script@v6
+        env:
+          AWS_PRIVATE_ECR_PATH: ${{ secrets.AWS_PRIVATE_ECR_PATH }}
+        with:
+          result-encoding: string
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
```
HenryNguyen5, line 27:
```suggestion
               ...context.repo,
```

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+               run_id: context.payload.workflow_run.id,
+            });
+            const adapterImages = allArtifacts.data.artifacts.map((artifact) => {
+              const basePath = process.env.AWS_PRIVATE_ECR_PATH + '/adapters/'
+              return basePath + artifact.name.replace('-shasum', '')
+            });
+            return adapterImages.join(' ');
```
HenryNguyen5, line 34:
```suggestion
```

HenryNguyen5 (changes requested):
Looks great, left a few suggestions around code style for terse-ness + clarity

.github/workflows/consume-release.yml
```diff
@@ -3,16 +3,43 @@
 # a corresponding event handler to generate releases based on this signal
 
 on:
-  workflow_run:
-    workflows: [Release]
-    types: [completed]
+  workflow_dispatch:
+    inputs:
+      workflowRunId:
```
HenryNguyen5, line 8: Shouldn't we keep around `workflow_run`? Then we can automatically extract the `workflowRunId` from the event payload. It's also good to keep around the `workflow_dispatch` just in-case we need to do manual runs.

I think having the `workflow_run` trigger around is better in terms of semantics, since `workflow_dispatch` can be triggered from anywhere.

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,44 @@ on:
...
+  get-images:
+    runs-on: ubuntu-latest
     environment: InfraK8s
+    steps:
+      - id: get-artifacts
+        name: 'Get artifacts from release workflow'
+        uses: actions/github-script@d50f485531ba88479582bc2da03ff424389af5c1 # v6.1.1
+        env:
+          AWS_PRIVATE_ECR_PATH: ${{ secrets.AWS_PRIVATE_ECR_PATH }}
+          WORKFLOW_RUN_ID: ${{ inputs.workflowRunId }}
```
HenryNguyen5, line 26: This needs to be conditional based on the trigger. I would test the following block, i havent tried it yet.

```
${{ inputs.workflowRunId  || github.event.workflow_run.id  }}

```

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,44 @@ on:
...
+            // Create whitespace separated URLs for downloading ECR adapter images based on uploaded workflow artifacts 
+            const pathPrefix = process.env.AWS_PRIVATE_ECR_PATH
+            return allArtifacts.data.artifacts
+              .map(({name}) => `${pathPrefix}/adapters/${name.replace('-shasum', '')}`)
+              .join(' ')
+    outputs:
+      published-adapters: ${{ steps.get-artifacts.outputs.result }}
+  image-dispatch:
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
HenryNguyen5, line 14: `if: ${{ inputs.workflowRunId || github.event.workflow_run.conclusion == 'success' }}`

We dont want to dispatch if there was a failure on the workflow run

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,44 @@ on:
...
+    branches: [develop]
+  workflow_dispatch:
+    inputs:
+      workflowRunId:
+        description: The workflow_run.id of the release workflow that has published artifacts
+        required: true
 
 jobs:
-  on-success:
+  get-images:
```
HenryNguyen5, line 17: We should only get artifacts if we were successful in publishing them, it might be easier to merge this job into the next one and use the same job execution conditional

.github/workflows/consume-release.yml
```diff
@@ -6,13 +6,44 @@ on:
...
+      published-adapters: ${{ steps.get-artifacts.outputs.result }}
+  image-dispatch:
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    environment: InfraK8s
+    needs: get-images
     steps:
-      - run: gh workflow run --repo smartcontractkit/infra-k8s updater.yaml -F image-updater-head-branch=image-updater/adapters -F pr-delay=60
+      - name: 'Trigger Image Dispatcher'
+        run: gh workflow run --repo smartcontractkit/infra-k8s image-dispatcher.yaml -F imageRepos="${{needs.get-images.outputs.published-adapters}}" -F gitRepo=external-adapter-js
```
HenryNguyen5, line 47:
```suggestion
        run: gh workflow run --repo smartcontractkit/infra-k8s image-dispatcher.yaml -F imageRepos="${{needs.get-images.outputs.published-adapters}}" -F gitRepo=${{ github.event.repository.name }}
```

HenryNguyen5 (changes requested):

HenryNguyen5 (approved):
Looks great! I would keep an eye out when this runs to make sure that the conditionals are being properly evaluated

PR #2114, opened 2022-08-17
Refactor/remove any types from EA

bogos:
## Description

Remove any types from EA

## Steps to Test

yarn lint 

## Quality Assurance


packages/core/test-helpers/src/helpers.ts
```diff
@@ -3,7 +3,9 @@ import { AdapterRequest, Execute } from '@chainlink/ea-bootstrap'
 
 const mockContext = {}
 
-export function assertError(statusCode: any, data: any, expectedJobId: any) {
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function assertError(statusCode: any, data: any, expectedJobId: any): void {
```
boxhock, line 8: Are these really necessary? Seems like we should be able to get types for a few of these, right?

packages/sources/1forge/src/adapter.ts
```diff
@@ -36,6 +36,7 @@ export const makeExecute: ExecuteFactory<DefaultConfig, endpoints.TInputParamete
   return async (request, context) => execute(request, context, config || makeConfig())
 }
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
boxhock, line 39: Is this one also needed?

packages/sources/amberdata/src/adapter.ts
```diff
@@ -48,6 +48,7 @@ interface Message {
   }
 }
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
 export const makeWSHandler = (defaultConfig?: DefaultConfig): MakeWSHandler<Message | any> =>
```
boxhock, line 52: Would it be better to do Message | UnknownWSMessage instead, so we avoid disabling the any check?

packages/sources/cfbenchmarks/src/adapter.ts
```diff
@@ -38,6 +38,7 @@ interface Message {
   time: number
 }
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
 export const makeWSHandler = (config?: Config): MakeWSHandler<Message | any> =>
```
boxhock, line 42: Same as above

boxhock (changes requested):
Only had a quick look for now. Mostly had questions around accepting `any` in WS EAs. Would it not be better to use the unknown WS message type instead of explicitly disabling the `any` check there?

alejoberardino:
Closing this, as the new framework supersedes this work

PR #2125, opened 2022-08-23
Elwood Adapter

austinborn:
## Description
Build new Elwood external adapter for WS-only connections

## Changes
- Add Elwood EA for WS price updates only
- Add `await` for WS subscription (since Elwood subscriptions are through REST)
- Remove `console.log` from WS test framework

## Steps to Test
`yarn && yarn setup && yarn test packages/sources/elwood/test/`

## Quality Assurance


justinkaseman:
Try running `yarn` locally to rebuild the lock file

packages/sources/elwood/src/adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+  stream: 'index',
+  symbol: pairSymbol,
+  index_freq: 1_000, //ms
+})
+
+export const makeWSHandler =
+  (
+    config?: Config,
+  ): MakeWSHandler<
+    ResponseMessage | any // TODO: full WS message types
```
boxhock, line 58: Can we get this TODO resolved before merging?

packages/sources/elwood/src/adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+            url: wsConfig.api?.baseURL,
+            headers: {
+              ...wsConfig.api?.headers,
+              apiKey: wsConfig.apiKey ?? '',
+            },
+            method: 'post',
+            data: subscriptionMsg,
+          }
+
+          Requester.request(data)
```
boxhock, line 79: Can you add some clarification as to what's happening here? It seems this is a WS handler, but we're also making a POST request without catching the response data?

packages/sources/elwood/src/adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+
+          return subscriptionMsg
+        } catch (e) {
+          const error = e as any
```
boxhock, line 83: Can we avoid `any`?

packages/sources/elwood/src/adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+
+    return {
+      connection: { url: `${wsConfig?.ws?.baseWsURL}?apiKey=${wsConfig.apiKey}` },
+      noHttp: true,
+      subscribe: handleSubscription('subscribe'),
+      unsubscribe: handleSubscription('unsubscribe'),
+      subsFromMessage: (message: ResponseMessage) => {
+        Logger.debug({ msg: `subsFromMessage`, message })
```
boxhock, line 95: Is this logging needed?

packages/sources/elwood/src/adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (!('type' in message && message.type === 'Index')) return
+        const subInMessage = getSubscribeRequest(message.data?.symbol, 'subscribe')
+        if (!message.data?.symbol || !subInMessage) return
+        return subInMessage
+      },
+      toResponse: (message: PriceResponse, input: AdapterRequest) => {
+        Logger.debug({ msg: `toResponse`, message })
```
boxhock, line 102: Is this logging needed?

boxhock (changes requested):

packages/sources/elwood/src/adapter.ts
austinborn: At this point I'll remove it

packages/sources/elwood/src/adapter.ts
austinborn: At this point I'll remove it

packages/sources/elwood/src/adapter.ts
austinborn: I'll change this to `Error`

packages/sources/elwood/src/adapter.ts
austinborn: The Elwood API only allows using POST requests to subscribe to WS feeds, so this was my attempt to send the subscription request and return `subscriptionMsg` without needing to mess with the framework. The `Requester.request` call is async, so waiting for the return value would require `awaiting` in every instance of `wsHandler.subscribe` and `wsHandler.unsubscribe`. I'm aware it's not ideal, so we can investigate a longer-term solution if desired.

boxhock: Could you add this explanation to the code, so we don't lose this knowledge? :sweat_smile:

packages/sources/elwood/src/adapter.ts
austinborn: Removed

packages/sources/elwood/src/adapter.ts
austinborn: Removed

packages/sources/elwood/src/adapter.ts
austinborn: Updated to `const error = e as Error`

packages/sources/elwood/src/adapter.ts
austinborn: Yeah for sure

austinborn: Done

boxhock (dismissed):
Looks good to me, but would like someone to give a second opinion on including `any`here.

austinborn:
> Looks good to me, but would like someone to give a second opinion on including `any`here.

@justinkaseman Any thoughts?

packages/sources/elwood/src/adapter.ts
alejoberardino: If the response types that are used in the adapter code are all included here, I don't think we need to go and get every possible ws message for the API, as long as the code itself is typesafe I'm good

KuphJr:
> Looks good to me, but would like someone to give a second opinion on including `any`here.

If we could use the `unknown` type and cast it if you try to access a specific property, I think that is the ideal solution.  I didn't spend too much time looking at all the code though, so if that is also going to be too difficult, I think using `any` is fine.  Just be sure to check that all the properties you access are not `undefined` so you don't cause a crash.

austinborn:
> > Looks good to me, but would like someone to give a second opinion on including `any`here.
> 
> If we could use the `unknown` type and cast it if you try to access a specific property, I think that is the ideal solution. I didn't spend too much time looking at all the code though, so if that is also going to be too difficult, I think using `any` is fine. Just be sure to check that all the properties you access are not `undefined` so you don't cause a crash.

I get the following when I try to set it to `unknown`:
```
packages/sources/elwood/src/index.ts:11:58 - error TS2345: Argument of type 'MakeWSHandler<unknown>' is not assignable to parameter of type 'MakeWSHandler<UnknownWSMessage>'.
  Type 'WSHandler<unknown> | Promise<WSHandler<unknown>>' is not assignable to type 'WSHandler<UnknownWSMessage> | Promise<WSHandler<UnknownWSMessage>>'.
    Type 'WSHandler<unknown>' is not assignable to type 'WSHandler<UnknownWSMessage> | Promise<WSHandler<UnknownWSMessage>>'.
      Type 'WSHandler<unknown>' is not assignable to type 'WSHandler<UnknownWSMessage>'.
        The types returned by 'subscribe(...)' are incompatible between these types.
          Type 'unknown' is not assignable to type 'UnknownWSMessage | undefined'.

11 const { server } = expose(adapterContext, makeExecute(), makeWSHandler(), endpointSelector)
```

packages/sources/elwood/src/adapter.ts
austinborn: I can add `util.isObject` checks to ensure we at least know the input is an object before trying to operate on it

boxhock:
@austinborn what about `UnknownWSMessage`?

boxhock (approved):

PR #2143, opened 2022-08-29
Lower Default API_TIMEOUT to 10s

austinborn:
## Description
Lower the default API timeout from `30s` to `10s`, since it's rare that an API would successfully respond if it hasn't within the first 10 seconds. Open to decreasing this further, but we need to validate that it doesn't increase API timeout failures (the lower we go, the more likely timeouts will be triggered).

## Changes
- Decrease default `API_TIMEOUT` to `10000` (10s)
- Set the default override for Kaiko's `API_TIMEOUT` back to `30000` (30s)

## Quality Assurance


justinkaseman:
Brendon has requested this in the past, so I'm sure he will be thrilled :grin:.

@austinborn We can test this on our monitoring before making the actual code change. Let's do that first so that we don't have to roll this back if any complications arise.

nzleet:
Hi guys. I see no issue with this for all EAs other than Kaiko. We should really try and figure out what is causing Kaiko to require such a long timeout though - is it purely down to their API taking so long to respond?

boxhock:
@austinborn To proceed with this PR, can we change the Kaiko default to still be 30s, while setting the framework default to 10s?

That will alleviate concerns around Kaiko for now, and gives us time to investigate this further.

austinborn:
> @austinborn To proceed with this PR, can we change the Kaiko default to still be 30s, while setting the framework default to 10s?
> 
> That will alleviate concerns around Kaiko for now, and gives us time to investigate this further.

I've set Kaiko's default `API_TIMEOUT` back to `30000` through the `envDefaultOverrides`. Does it look alright?

.changeset/modern-apples-travel.md
```diff
@@ -0,0 +1,6 @@
+---
+'@chainlink/ea-bootstrap': patch
+'@chainlink/kaiko-adapter': patch
+---
+
+Add API_TIMEOUT to env default overrides, with override for Kaiko
```
boxhock, line 6: Feel free to re-phrase, just want to include the actual value in the changeset.

```suggestion
Add API_TIMEOUT to env default overrides, with an override of 30s for Kaiko
```

boxhock (changes requested):
Just a tiny changeset request, otherwise LGTM!

.changeset/modern-apples-travel.md
austinborn: Done!

boxhock (approved):

PR #2155, opened 2022-09-07
[SC-51579] Expanded coverage of redacting secrets in logs

amit-momin:
## Description

Some secrets were getting written to logs so the coverage of redacting secrets was expanded to include custom config properties an adapter may use.

## Changes

- Leveraged pino redact configurations to include config property paths that are secret
- Removed previous implementation of cloning config without secrets prior to logging
- Added new URL param match pattern `/client/i` to cover Tradingeconomics ws URLs

## Steps to Test

1. Find adapter with a custom secret config such as privateKey (Harmony, Ethwrite, etc.)
2. Set the appropriate env vars
3. Start the adapter
4. Submit a request
5. Check logs for the config and ensure the field is [REDACTED]
6. Repeat steps for the other redacted paths

## Quality Assurance


packages/core/bootstrap/src/lib/modules/logger.ts
```diff
@@ -2,8 +2,9 @@ import { uuid } from '../util'
 import pino from 'pino'
 import { wsRedactPaths } from '../middleware/ws/config'
 import { cloneDeep } from 'lodash'
+import { configRedactPaths } from '../config'
 
-export const paths = [...wsRedactPaths]
+const paths = [...wsRedactPaths, ...(configRedactPaths || [])]
```
amit-momin, line 7: Had issues with config.test.ts throwing an error that configRedactPaths is not iterable. Unsure why it has problems with this static string[] and not with wsRedactPaths. Applied this check to avoid the issue but not a fan of having to do this. If anyone has a suggestion on what could be happening, that'd be super helpful.

boxhock: I do not fully understand what's causing this issue, but I found a better way to work around it.
It's failing because it's reading `configRedactPaths` from `index.ts`. If you create a new file, e.g. `config/logging.ts`, and place the array there, it will work.

I guess that's better than `(configRedactPaths || [])`? :man_shrugging:

amit-momin: Interesting... I agree that's definitely better. Just updated it in the latest commit.

boxhock (approved):
Looks good!

PR #2160, opened 2022-09-09
Release/v1.37.0

alejoberardino:
### Breaking Changes

- Makes WS_SOCKET_KEY required for Finage
- Added a check to verify the chain ID for external adapters that use RPC nodes.  If an RPC node returns a chain ID that does not match the expected ID, it will cause an error.  Ensure the chain ID configuration variable is set correctly.

### Features

- Removes PASSWORD from config and adds rate limit tier for Bank Frick
- Update Coinbase NFT floor price
- Set inverse pair for IDR/USD for Tradermade

### Bug Fixes

- Removes UNI mapping override for Kaiko
- Changes EA framework to prevent WS reducer from updating properties of connection/subscription keys that cannot be found in the state
- Set default PAGE_SIZE for bank-frick

### Notable Adapter Updates

| Adapter | Version |         Description          |
| :-----: | :-----: | :--------------------------: |
| finage  | v1.6.0  | Makes WS_SOCKET_KEY required |
|  kaiko  | v1.5.9  |     Removes UNI override     |
| bank-frick  | v0.0.6  | Removes PASSWORD from config, adds rate limit tier, sets default PAGE_SIZE |
| tradermade | v1.8.2 | Added inverse pair |


austinborn (approved):

boxhock (changes requested):
Blocking this: There are commits included here that probably shouldn't be included (see unconsumed changesets)

PR #2165, opened 2022-09-09
feat: [sc-52088] Work on coingecko optimizations

ad0ll:
Story details: https://app.shortcut.com/chainlinklabs/story/52088

This PR ports the coingecko adapter from the framework repo to the monorepo. The changes consist of adding the coingecko framework-based adapter as "coingecko-test" to the monorepo, and then making several changes to github actions, core, and scripts to make the monorepo capable of accommodating framework adapters.

Specific changes are:
1. The addition of the coingecko-test adapter with a handful of small tweaks to make it compliant with the monorepo
2. Metrics register will always be cleared when importing an adapter to avoid collisions when a package imports both ea-bootstrap and the framework (legos, generate scripts)
3. `generate-readme` will evaluate if the EA is a framework adapter, and will source and transform the generation config from the adapter itself if it is (as opposed to using env.json). 
4. Same as above for `generate-master-list`
5. Same as above for `legos`
6. Same as above for `schema-flatten`, with one additional note. This change was for `docker-build`. When running docker-build, environment variables for compose.yml will be {framework.BaseSettings + adapter.customSettings}. Bootstrap envvars are not (and should not) be considered for framework EAs
7. Custom GHA added for setting up yarn, including supplying the NPM_AUTH_TOKEN. All instances of `npm i yarn > yarn > yarn setup` replaced with this custom `actions/setup` action.


packages/scripts/src/generate-master-list/generator.ts
```diff
@@ -64,18 +65,36 @@ const getEndpoints = async (adapterPath: string, verbose = false) => {
   let batchableEndpoints = 'Unknown'
   try {
     const indexPath = adapterPath + '/src/endpoint/index.ts'
+    const schemaPath = adapterPath + '/src/schemas/env.json'
 
     const endpointDetails: EndpointDetails = await require(path.join(process.cwd(), indexPath))
-
+    const schema = getJsonFile(schemaPath) as Schema
```
alejoberardino, line 71: Let's make it so for adapters using the new version, we don't add a schema for those, it was an old initiative but we've gone in a different direction. We can use its lack of existence as a way to determine the major version

packages/scripts/src/generate-readme/generator.ts
```diff
@@ -80,7 +81,7 @@ export class ReadmeGenerator {
     this.name = schema.title ?? packageJson.name ?? ''
     this.envVars = schema.properties ?? {}
     this.requiredEnvVars = schema.required ?? []
-
+    this.frameworkVersion = schema.frameworkVersion ?? 'v2'
```
alejoberardino, line 84: See comment above

alejoberardino (changes requested):

packages/core/legos/package.json
```diff
@@ -135,7 +136,6 @@
     "@chainlink/poa-adapter": "workspace:*",
     "@chainlink/polygon-adapter": "workspace:*",
     "@chainlink/por-address-list-adapter": "workspace:*",
-    "@chainlink/reduce-adapter": "workspace:*",
```
alejoberardino, line 138: Why is this one removed?

packages/scripts/package.json
```diff
@@ -6,6 +6,7 @@
     "@apidevtools/json-schema-ref-parser": "9.0.9",
     "@chainlink/ea-bootstrap": "workspace:*",
     "@chainlink/ea-factories": "workspace:*",
+    "@chainlink/external-adapter-framework": "0.0.42",
```
alejoberardino, line 9: Let's make sure to bump before merging

packages/sources/coingecko-test/src/config/index.ts
```diff
@@ -0,0 +1,14 @@
+import { SettingsMap } from '@chainlink/external-adapter-framework/config'
+
+export const NAME = 'COINGECKO'
```
alejoberardino, line 3: Also let's make sure to copy the latest code from this EA over here

alejoberardino (approved):

alejoberardino:
Looks great overall, let's make sure we cover those few comments I left but we should be good to merge tomorrow

alejoberardino (approved):

PR #2171, opened 2022-09-13
fix unit tests making request to DP

karen-stepanyan:
## Changes

- Removed unit tests from EAs without input params
- Modified unit tests of EAs that were making http request to DP

## Steps to Test

yarn test:ci:unit

## Quality Assurance


packages/sources/gemini/test/unit/adapter.test.ts
```diff
@@ -9,12 +9,16 @@ describe('execute', () => {
   const execute = makeExecute()
 
   describe('validation error', () => {
-    const requests = [{ name: 'empty body', testData: {} }]
```
boxhock, line 12: According to the README, this EA should still be able to have a unit test on empty body, as it requires `token` to be set.

boxhock (changes requested):

packages/sources/gemini/test/unit/adapter.test.ts
karen-stepanyan: `empty body` check will make a DP call because it has default value, it will use it.

boxhock: I see. Could you update the schema to make it not-required since it has a default?

karen-stepanyan: I don't really understand why it has both `required` and `default` specified, doesn't make sense to me

austinborn (approved):

boxhock (approved):

PR #2179, opened 2022-09-19
Fix/update galaxy auth

mjk90:
## Description
- Updated Galaxy WS EA to fetch access token if existing token is older than 30 seconds.
- Added fetchingToken check it avoid making many duplicated token requests
- Added default env overrides

## Steps to Test

1. `yarn test:unit`
2. `yarn test:integration`

## Quality Assurance


packages/sources/galaxy/src/adapter.ts
```diff
@@ -69,7 +72,32 @@ export const getAccessToken = async (config: Config): Promise<string> => {
...
+        return
+      }
+
+      // Refresh token if it is older than 30 seconds
+      if (new Date().getTime() - token.created > 30000) {
+        fetchingToken = true
+        token = await getAccessToken(config)
+        fetchingToken = false
+        return
+      }
```
austinborn, line 96: Would this work?

```suggestion
      if (!token || new Date().getTime() - token.created > 30000) {
        fetchingToken = true
        token = await getAccessToken(config)
        fetchingToken = false
        return
      }
```

mjk90: Yes it would. I'll add this

KuphJr (approved):

PR #2191, opened 2022-09-26
add por-address-list to PoR

karen-stepanyan:
Changed response schema of por-address-list EA so that now it returns array of objects each containing `address` property. 

Added por-address-list protocol option for Proof-of-reserves EA


justinkaseman (approved):
LGTM :+1:

PR #2196, opened 2022-09-27
[SC-53080] Added new logging censorship implementation

amit-momin:
- Compiles a list of RegExps to censor based on a static list of sensitive env vars
- Uses the compiled list to replace values in the logger with a redacted message i.e. [API_KEY REDACTED]
- Maintains existing pino redact config for paths than are not covered by the new implementation. Like when sensitive values are encoded

### Before
Screen Shot 2022-09-27 at 2 56 12 PM

### After
Screen Shot 2022-09-27 at 2 57 13 PM


boxhock (changes requested):
I love this change. Only thing I'd request is that we document this in the bootstrap package README and in the CONTRIBUTING files. Just want to make sure we document how EA developers can add a secret that should be redacted.

packages/core/bootstrap/test/unit/logger.test.ts
```diff
@@ -17,5 +23,27 @@ describe('Logger', () => {
...
+    it('properly redacts API_KEY (multiple nested values)', () => {
+      const redacted = censorLog(
+        { apiKey: 'mock-api-key', config: { headers: { auth: 'mock-api-key' } } },
+        CensorList.getAll(),
+      )
+      expect(redacted).toEqual({
+        apiKey: '[API_KEY REDACTED]',
+        config: { headers: { auth: '[API_KEY REDACTED]' } },
+      })
+    })
```
boxhock, line 47: Does it also censor if the value has additional text in addition to the secret? If so, can we add a test for it?

E.g. `Bearer mock-api-key` -> `Bearer [API_KEY REDACTED]`

amit-momin: It does so good call! Adding that in.

boxhock (approved):

PR #2207, opened 2022-09-29
feat: [sc-48609] "customError" can return a string

ad0ll:
Story details: https://app.shortcut.com/chainlinklabs/story/48609

Lets customError optionally return a string, and updates requestor's main function (`request`) to print the customError return value if its a string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chainlink/external-adapters-js",
-  "version": "1.38.0",
+  "version": "1.39.0",
   "license": "MIT",
   "private": true,
   "workspaces": [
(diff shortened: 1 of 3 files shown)


.changeset/bright-doors-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': patch
```
boxhock, line 2: Seems to me like this should be `minor`

boxhock (changes requested):
LGTM, only want to change the changeset a bit

boxhock (approved):

PR #2221, opened 2022-10-06
Patch ethereum cryptography

cds95:


.changeset/flat-flies-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/layer2-sequencer-health-adapter': minor
+---
+
+Add patch for the ethereum-cryptopgraph package so that the EA can successfully build using the StarknetJS dependency
```
boxhock, line 5: Should probably also mention support for Starkware?

cds95: Added

boxhock (approved):
This PR builds on previously approved #2039, which was later reverted. This PR fixes the issue that caused the original PR to be reverted.

Can confirm building and running this Docker image works.

PR #2222, opened 2022-10-06
Update EA framework and use new alias types

boxhock:


alejoberardino:
Changes have been slowly rolled in in other PRs, final one is #2242

PR #2223, opened 2022-10-07
Bug/sc 55035

ad0ll:


amit-momin (dismissed):
LGTM! Ran the same tests used to recreate the problem and metrics are back up here

amit-momin (approved):

PR #2224, opened 2022-10-09
Bump coingecko test

alejoberardino:


packages/sources/coingecko-test/src/crypto-utils.ts
```diff
@@ -30,7 +27,7 @@ export const cryptoInputParams: InputParameters = {
     description: 'The symbol of the currency to convert to',
     required: true,
   },
-}
+} as const
```
boxhock, line 30: What does this do? Isn't it already `const`?

packages/sources/coingecko-test/src/crypto-utils.ts
```diff
@@ -1,18 +1,15 @@
-import { AxiosRequestConfig, AxiosResponse } from 'axios'
```
boxhock, line 1: I already did the work for this in #2222. Wanna get that in first, or is that redundant with this PR?

alejoberardino: I think we can get this one in first which is newer, and then I can rebase the other one and check if there are pending changes

packages/sources/coingecko-test/src/crypto-utils.ts
boxhock: Got it, thanks!

boxhock (approved):

PR #2225, opened 2022-10-10
Remove UUID env var

karen-stepanyan:


packages/core/bootstrap/src/lib/middleware/cache/index.ts
```diff
@@ -18,18 +18,26 @@ import * as redis from './redis'
...
   shouldUseLocal?: boolean,
   adapterContext?: AdapterContext,
 ): CacheOptions => {
+  let cacheGroupKey = ''
+  if (adapterContext?.name) {
+    cacheGroupKey =
+      adapterContext?.name + (getEnv('CACHE_KEY_GROUP') ? `-${getEnv('CACHE_KEY_GROUP')}` : '')
+  } else {
+    cacheGroupKey = UUID + (getEnv('CACHE_KEY_GROUP') ? `-${getEnv('CACHE_KEY_GROUP')}` : '')
+  }
```
boxhock, line 33: Could we simplify this?

```suggestion
  const cacheGroupKey = `${adapterContext?.name || UUID}-${ getEnv('CACHE_KEY_GROUP') || '' }`
```

I don't think it would be a problem that there's an extra `-` at the end in most cases.

packages/core/bootstrap/README.md
```diff
@@ -129,7 +128,7 @@ To configure caching these environment variables are available:
...
-|           |     `CACHE_KEY_GROUP`     |                                      Set to specific group ID to group the cached data, for this adapter, with other instances in the same group. Applicable only in remote cache scenarios, where multiple adapter instances share the cache.                                      |                    |                          UUID of the adapter                          |
+|           |     `CACHE_KEY_GROUP`     |                                      Set to specific group ID to group the cached data, for this adapter, with other instances in the same group. Applicable only in remote cache scenarios, where multiple adapter instances share the cache.                                      |                    |                          Name of the adapter                          |
```
boxhock, line 131: Default should be `undefined`, no?

.changeset/yellow-hairs-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ea-bootstrap': minor
+---
+
+Removed UUID environment variable
```
boxhock, line 5: Could you add an explanation for how this impact users, and what users that relied on `UUID` should do now? (Basically, no action needed for EAs with the same name. Need to set `CACHE_KEY_GROUP` to split EAs with the same name).

Also wondering if this justifies a major version bump - since it can be a breaking change... @alejoberardino opinions?

boxhock (changes requested):

.changeset/yellow-hairs-lay.md
karen-stepanyan: @boxhock Does this look better ? 
`Removed UUID environment variable. UUID was used for generating cache keys. If you were not using UUID environment variable and/or running single instance of EA there is no action needed. If you were using UUID to run multiple instances of the same EA with shared cache, you should use CACHE_KEY_GROUP environment variable instead. Applicable only in remote cache scenarios like when using redis for cache.`

As for major version bump i also thought about it. I don't know how common is the case where this change can break something.

boxhock: > If you were using UUID to run multiple instances of the same EA with shared cache, you should use CACHE_KEY_GROUP environment variable instead.

Technically they shouldn't have to set anything now if running multiple EAs.`CACHE_KEY_GROUP` should be used if they want to separate caches between multiple identical EAs using the same Redis cache.

packages/core/bootstrap/src/lib/middleware/cache/index.ts
```diff
@@ -18,7 +18,7 @@ import * as redis from './redis'
 import * as metrics from './metrics'
 import { CacheOptions, CacheImplOptions, CacheEntry, Cache } from './types'
 
-const DEFAULT_CACHE_KEY_GROUP = uuid()
+const UUID = uuid()
```
alejoberardino, line 21: If this isn't used anywhere can we remove it?

alejoberardino (changes requested):

packages/core/bootstrap/src/lib/middleware/cache/index.ts
karen-stepanyan: Is is very unlikely, but still possible that there will be no name for adapter and UUID is the fallback

alejoberardino: I thought EAs were force to have a name, but I also don't trust it enough haha. Even in that case we can take care of this in a separate PR

alejoberardino (approved):

PR #2238, opened 2022-10-17
Add Cryptocompare EAv3

boxhock:
## Steps to Test

1. yarn test packages/sources/cryptocompare-test/test/integration/

## Quality Assurance


packages/sources/cryptocompare-test/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.3.6",
```
alejoberardino, line 31: This was merged after work on this was done, but would you mind upgrading to `0.5.1`? That version has the noticeable refactor of the type generics, and would be good to do it before merging imo

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -0,0 +1,84 @@
...
+  MESSAGE: string
+  PARAMETER: string
+  INFO: string
+}
+
+const logger = makeLogger('CryptoCompareCryptoEndpoint')
+
+export const transport = new WebSocketTransport({
+  url: (context: AdapterContext) =>
+    `${DEFAULT_WS_API_ENDPOINT}?api_key=${context.adapterConfig.API_KEY}`,
```
alejoberardino, line 40: This does not allow the ws api endpoint to be modified with the env var from the config, is this intended?

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+    message(message: Message): ProviderResult<PriceEndpointParams>[] | undefined {
+      logger.trace(message, 'Got response from websocket')
+      if (message.TYPE === '5' && 'PRICE' in message) {
+        return [
+          { params: { base: message.FROMSYMBOL, quote: message.TOSYMBOL }, value: message.PRICE },
+        ]
+      }
+      if (message.MESSAGE === 'INVALID_SUB') {
+        logger.error(message, 'asset not supported by data provider')
```
alejoberardino, line 64: We should add a TODO here to return an error response once I add support for it

packages/sources/cryptocompare-test/src/endpoint/index.ts
```diff
@@ -0,0 +1,2 @@
+export { endpoint as cryptoWsEndpoint } from './crypto-ws'
+export { endpoint as cryptoEndpoint } from './crypto'
```
alejoberardino, line 2: Imma be annoying here too, to see if we can kill two birds with one stone: let's add a RoutingTransport that takes both of these, and turns it into one endpoint that chooses where to send the request based on the endpoint name ('crypto' would go to the batch one, 'crypto-ws' to the websocket one). This way we can start testing this new functionality with NOPs 😄

alejoberardino (changes requested):
This looks great, left a few requests to use things from the new framework release. We can potentially do them in a separate PR, but I think it'd be easier to just change this one. Lmk

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
boxhock: This is how it was done in the v3 repo, so I didn't intend anything further than that. Is there a more ideal way to do this?

boxhock: This has been updated now. Please confirm if resolved

alejoberardino (approved):

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
alejoberardino: It is, ty

PR #2245, opened 2022-10-17
Feature/universal ea io format changes

KuphJr:


packages/sources/universal/src/endpoint/sandbox.ts
```diff
@@ -4,23 +4,31 @@ import {
...
   'This adapter endpoint sends a custom request to a FaaS sandbox to be executed.'
 
 export type TInputParameters = {
-  code: string
-  httpQueries?: HttpQuery[]
+  source: string
+  queries?: HttpQuery[]
   args?: string[]
   secrets?: string
   codeLocation?: 'onchain'
```
bolekk, line 34: This field as well as secretsLocation and language are encoded as integers (= 0). You might want to create enums equivalent to the ones in the Contract: https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/dev/ocr2dr/OCR2DR.sol#L15-L28

Same for "HttpVerb" below.

justinkaseman (commented):
LGTM

PR #2255, opened 2022-10-21
Rocket Pool Composite EA

austinborn:
## Description
- Add new Rocket Pool Composite EA with `reth` endpoint to calculate `rETH/ETH` and `rETH/USD` prices across different chains (using `rETH` exchange rate from Ethereum L1)
- Endpoint allows passing `network` to designate specific network RPC to query, along with `quote` as either `ETH` or `USD`, with `ethUsdProxyAddress` to specify where the Chainlink ETH/USD price feed proxy is located on the given network (if not Ethereum mainnet).

## Steps to Test
`yarn && yarn setup && yarn test packages/composites/rocket-pool/test/`

## Quality Assurance


packages/composites/rocket-pool/schemas/env.json
```diff
@@ -0,0 +1,31 @@
...
+    },
+    "PRICE_CHAIN_ID": {
+      "type": "string",
+      "description": "Chain ID for EVM-compatible network from PRICE_RPC_URL. Defaults to `1` for Ethereum mainnet."
+    },
+    "ETH_USD_PROXY_ADDRESS": {
+      "type": "string",
+      "description": "Proxy address for ETH/USD Chainlink price feed on PRICE_RPC_URL network.",
+      "default": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419"
+    }
```
austinborn, line 24:
```suggestion
    }
```

packages/composites/rocket-pool/src/config/index.ts
```diff
@@ -0,0 +1,28 @@
+import { Requester, util } from '@chainlink/ea-bootstrap'
+import { Config } from '@chainlink/ea-bootstrap'
+import { ethers } from 'ethers'
+import rethAbi from '../abis/rethAbi.json'
+
+export const NAME = 'ROCKET_POOL'
+
+export const DEFAULT_ENDPOINT = 'reth'
+export const DEFAULT_ETH_USD_PROXY_ADDRESS = '0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419'
+export const RETH_TOKEN_CONTRACT = '0xae78736Cd615f374D3085123A210448E74Fc6393'
```
justinkaseman, line 10: Should we allow overriding this with an environment variable or input parameter? Similar to `ETH_USD_PROXY_ADDRESS`

packages/composites/rocket-pool/src/config/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export type RocketPoolConfig = Config & {
+  rethContract: ethers.Contract
+}
+
+export const makeConfig = (prefix?: string): RocketPoolConfig => {
+  const ethereumRpcUrl = util.getRequiredEnv('ETHEREUM_RPC_URL')
```
justinkaseman, line 17: The EA endpoint looks like it is set up to be cross-chain. Should this support multiple networks as the `network` input param implies?

```
description: 'Network to query for price feed (EA must have `<network>_RPC_URL` configured).',
```
We have previously used a combo of a dynamic prefix from the `network` + a fallback.
```
// warning: pseudocode
// build a dynamic env variable look up here
  const rpcURL = util.getRequiredEnvWithFallback(
    ENV_RPC_URL,
    [ENV_FALLBACK_RPC_URL],
    prefix,
  )
```

This is probably something we will want to standardize into a utility function.

packages/composites/rocket-pool/src/config/index.ts
justinkaseman: Another, suggestion if the testnet addresses are known is to hold a lookup for `{ [network: string]: contractAddress: string }`
(and still allow overrides)

justinkaseman (dismissed):
Left some nitpicky comments. If this is high priority it LGTM.

packages/composites/rocket-pool/src/config/index.ts
austinborn: Note to self: this would also require extending the source to be able to come from a different chain.

qs (approved):

packages/composites/rocket-pool/src/config/index.ts
austinborn: I like the override idea, I will add that in. For the lookup map, I don't know any of the other addresses so I'd rather not look into that now. Could be useful in the future though so I'll leave a comment.

packages/composites/rocket-pool/src/config/index.ts
austinborn: Yeah the idea here is that the EA would support multiple chains for the `ETH/USD` price feed, but the `rETH/ETH` exchange rate comes solely from mainnet (for current implementation).

austinborn: I didn't know how to best capture dynamic env vars in the `schema/env.json`, so I simply referenced the `reference-data-reader` and tried to explain in the README.

justinkaseman (approved):

qs (approved):

PR #2259, opened 2022-10-21
Bump coingecko test

alejoberardino:


packages/sources/coingecko-test/src/endpoint/global-marketcap.ts
```diff
@@ -1,25 +1,20 @@
...
+const batchEndpointTransport = new BatchWarmingTransport<GlobalEndpointTypes>({
+  prepareRequest: (_, config): HttpRequestConfig<never> => {
     return buildGlobalRequestBody(config.API_KEY)
   },
-  parseResponse: (
-    params: AdapterRequestParams[],
-    res: HttpResponse<ProviderResponseBody>,
-  ): ProviderResult<AdapterRequestParams>[] => {
-    const entries = [] as ProviderResult<AdapterRequestParams>[]
+  parseResponse: (params, res): ProviderResult<GlobalEndpointTypes>[] => {
```
amit-momin, line 16: I thought after the types changes we wouldn't have to define the return type like the crypto endpoints?

alejoberardino: We don't have to necessarily, it's just convenient sometimes haha

amit-momin (approved):

PR #2276, opened 2022-10-25
Add script to bump framework versions, bump coingecko

alejoberardino:
Adds a script to bump the new framework version for adapters that use it, as well as the scripts package. Additionally, bumps the coingecko version.


amit-momin (approved):
Nice! This'll be super useful

PR #2289, opened 2022-11-01
experimental/test-node18-workflows

ad0ll:
**This PR can be completely ignored. I'm doing a quick experiment to see what, if anything breaks, when upgrading to nodejs 18**

GitHub should be supporting this as the default for runners soon: https://github.com/actions/runner-images/discussions/5429

Things that broke/how to fix:
`runs-on: [self-hosted, sdlc-ghr-prod]`, this runner image results in a library error when running setup-node. I fixed it **temporarily** by changing runners to ubuntu-latest (other steps in the workflow were using this already)

Ongoing:
Unit and integration tests fail because they run out of heap. Wonder if this is an issue with ubuntu-latest?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,14 +16,16 @@ env: # NPM_AUTH_TOKEN is needed for any yarn command. Having it at the top is no
 jobs:
   packages:
     name: Verify dependency package archives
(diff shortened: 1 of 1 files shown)


alejoberardino:
@ad0ll our self hosted runners are probably higher spec'd than the default gh ones

PR #2292, opened 2022-11-03
Chore/Spammy layer2 sequencer health log

amit-momin:
## Description

`layer2-sequencer-health` logs when a health endpoint is missing for the network. For Arbitrum, this is expected behavior causing NOPs to get spammed in the logs for every call for this network. The log statement was moved to on-startup to alleviate the spam.

## Changes

- Moved unavailable endpoint log to startup

## Steps to Test

1. Start the `layer2-sequencer-health` adapter
2. Notice ```Health endpoint not available for network:``` log for `arbitrum` and `starkware` 
3. Send a request for the `arbitrum` or `starkware` network
4. Check to make sure the log is not printing on request

## Quality Assurance

### DIFF
--- a/packages/sources/layer2-sequencer-health/src/network.ts
+++ b/packages/sources/layer2-sequencer-health/src/network.ts
@@ -41,7 +41,6 @@ export const checkSequencerHealth: NetworkHealthCheck = async (
   network: Networks,
 ): Promise<undefined | boolean> => {
   if (!HEALTH_ENDPOINTS[network]?.endpoint) {
-    Logger.debug(`Health endpoint not available for network: ${network}`)
     return
   }
   const response = await Requester.request({
--- a/packages/sources/layer2-sequencer-health/src/network.ts
+++ b/packages/sources/layer2-sequencer-health/src/network.ts
@@ -41,7 +41,7 @@ export const checkSequencerHealth: NetworkHealthCheck = async (
   network: Networks,
 ): Promise<undefined | boolean> => {
   if (!HEALTH_ENDPOINTS[network]?.endpoint) {
-    Logger.info(`Health endpoint not available for network: ${network}`)
+    Logger.debug(`Health endpoint not available for network: ${network}`)
     return
   }
   const response = await Requester.request({
--- a/.changeset/many-seahorses-smile.md
+++ b/.changeset/many-seahorses-smile.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/layer2-sequencer-health-adapter': patch
+---
+
+Changed spammy missing endpoint log to debug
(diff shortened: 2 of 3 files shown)


packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -41,7 +41,7 @@ export const checkSequencerHealth: NetworkHealthCheck = async (
   network: Networks,
 ): Promise<undefined | boolean> => {
   if (!HEALTH_ENDPOINTS[network]?.endpoint) {
-    Logger.info(`Health endpoint not available for network: ${network}`)
+    Logger.debug(`Health endpoint not available for network: ${network}`)
```
alejoberardino, line 44: Wouldn't it be better to do it for Arbitrum only?

amit-momin: I had the same idea but felt wrong hardcoding it for a network. I would have done that if the original log was maybe a warn since it's not problematic to not have the endpoint with fallback checks. Since it didn't seem like something actionable for us or NOPs, I thought I could hide it for all networks behind the debug flag? But I don't mind changing it if you still think that's better.

alejoberardino: I think it should be actionable, no? If the endpoint is not responding the EA probably won't function as expected, let's discuss with the team just in case

amit-momin: So the check is more so that an endpoint is not configured and we have defaults for the other networks except starkware which might be in the same boat as Arbitrum. But you're probably right. I'll drop this in the team channel to check.

qs: I think, the log message should be send only in the case of config has been initiated or changed. I may be wrong, but the config isn't changed during the execution. Probably, the is no need to dive into `checkSequencerHealth` if there is nothing to check actually, the method could be filtered out from the [checks list](https://github.com/smartcontractkit/external-adapters-js/blob/develop/packages/sources/layer2-sequencer-health/src/endpoint/health.ts#L77) using the config.

amit-momin: It makes sense to skip `checkSequencerHealth` when an endpoint isn't configured if we're assuming that when a network doesn't have one it's intentional. I was thinking it'd be better to keep the log for tracing purposes if ever needed but mark it as debug since it's not necessary info. What do you think @qs?

austinborn: If we know that `network` is a closed list, then I agree with @qs. The log ideally should only be triggered at startup or on change (though I don't think EAs can change config dynamically). Whether it's `info` or `debug` wouldn't be as much of an issue then since it wouldn't be triggering constantly. That said, if setting it to `debug` won't cause any unintended consequences, then that's probably the easiest course of action.

austinborn (approved):
Nice, I like this

PR #2295, opened 2022-11-04
(feat): Add cfbenchmarks v3

justinkaseman:
## Description
Migrates the CFBenchmarks adapter to EAv3.

## Changes
- The original adapter had capabilities for providing input param overrides for `index`. This is not currently possible with v3.

## Steps to Test
`yarn test packages/sources/cfbenchmarks-test`

## Quality Assurance


packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
```diff
@@ -0,0 +1,127 @@
...
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+    required: false,
+  },
+}
+
+export const inputParameters = {
+  ...priceEndpointInputParameters,
+  ...extendedInputParameters,
+}
```
alejoberardino, line 41: Right now the obj merging is not doing much, imo we can simplify like this:

```suggestion
const inputParameters: InputParameters = {
  index: {
    description: 'The ID of the index. Takes priority over base/quote when provided.',
    type: 'string',
    required: false,
  },
  base: {
    ...priceEndpointInputParameters.base,
    required: false
  },
  quote: {
    ...priceEndpointInputParameters.quote,
    required: false
  },
}
```

packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
```diff
@@ -0,0 +1,127 @@
...
+  const override = overrideId(base, quote)
+  if (override) return override
+
+  if (type === 'secondary') return getSecondaryId(base, quote)
+  return getPrimaryId(base, quote)
+}
+
+const overridenBaseQuoteFromId: { [id: string]: { base: string; quote: string } } = {
+  BRTI: { base: 'BTC', quote: 'USD' },
+}
```
alejoberardino, line 79: Imo we should only hardcode one of these objects in the code and build the inverse at runtime (only once, e.g. when the EA starts or in the global scope) so that we don't update one and not the other by mistake

packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
```diff
@@ -0,0 +1,127 @@
...
+  base: string,
+  quote: string,
+  type: 'primary' | 'secondary',
+): string => {
+  const override = overrideId(base, quote)
+  if (override) return override
+
+  if (type === 'secondary') return getSecondaryId(base, quote)
+  return getPrimaryId(base, quote)
+}
```
alejoberardino, line 75: We could maybe move all these helper fns to a separate file

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// NOTE: This is using the BatchWarming transport, but the actual API endpoint is not batchable
+export const transport = new BatchWarmingTransport<RestEndpointTypes>({
+  prepareRequest: ([{ base, quote }], { DEFAULT_API_ENDPOINT }) => {
+    return {
+      baseURL: DEFAULT_API_ENDPOINT,
+      url: '/v1/values',
+      method: 'GET',
+      params: {
+        id: getIdFromBaseQuote(base, quote, 'primary'),
```
alejoberardino, line 29: This is failing in a bunch of CI runs 🤔 
`Argument of type 'string | undefined' is not assignable to parameter of type 'string'.`

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+    }
+  },
+  parseResponse: ([{ base, quote }], res) => {
+    const values = res.data.payload.sort((a, b) => {
+      if (a.time < b.time) return 1
+      if (a.time > b.time) return -1
+      return 0
+    })
```
alejoberardino, line 38:
```suggestion
    const values = res.data.payload.sort((a, b) => b.time - a.time) // Descending
```

alejoberardino: Kinda annoying they don't already come sorted, this is potentially a lot of processing

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type RestEndpointTypes = EndpointTypes & {
+  Provider: {
+    RequestBody: EmptyObject
+    ResponseBody: ProviderResponse
+  }
+}
+
+// NOTE: This is using the BatchWarming transport, but the actual API endpoint is not batchable
+export const transport = new BatchWarmingTransport<RestEndpointTypes>({
```
alejoberardino, line 22: Since this one is just like the one below but with a different endpoint, we should abstract it into a builder function, subclass, etc

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export type WsEndpointTypes = EndpointTypes & {
+  Provider: {
+    WsMessage: Message
+  }
+}
+
+const logger = makeLogger('CfbenchmarksCryptoWebsocketEndpoint')
+
+export const transport = new WebSocketTransport<WsEndpointTypes>({
```
alejoberardino, line 30: Same comment as in the batch ones, these two ws transports are identical save for 1 string

packages/sources/cfbenchmarks-test/src/index.ts
```diff
@@ -0,0 +1,15 @@
+import { expose, ServerInstance } from '@chainlink/external-adapter-framework'
+import { PriceAdapter } from '@chainlink/external-adapter-framework/adapter'
+import { SettingsMap } from '@chainlink/external-adapter-framework/config'
+import { customSettings } from './config'
+import { crypto } from './endpoint'
+
+export const makeAdapter = (): PriceAdapter<SettingsMap> =>
```
alejoberardino, line 7: The correct type for the generic param would be `typeof customSettings`, but it should be able to infer it, does 
```suggestion
export const makeAdapter = (): PriceAdapter =>
```
work?

alejoberardino (changes requested):

packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
justinkaseman: Oop, good catch. Left over from before I realized CFBenchmarks has optional base/quote. Removed.

packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
justinkaseman: Sure. I believe they are specific to the use of this crypto endpoint. But may be cleaner.

packages/sources/cfbenchmarks-test/src/index.ts
justinkaseman: It does not.

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
```diff
@@ -0,0 +1,150 @@
...
+    },
+
+    message(message) {
+      logger.trace(message, 'Got response from websocket')
+      if (message.type === 'value') {
+        const { base, quote } = getBaseQuoteFromId(message.id)
+        const value = Number(message.value)
+        return [
+          {
+            params: { base, quote },
```
justinkaseman, line 65: Just realized ... @alejoberardino does this cache based on the `params`? If so this may be a problem.
The WS message we get back is the id - we don't know whether it came from a request base/quote or id.
I guess we can cache on both and/or on every combo.

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
justinkaseman: Damn, from the required input parameter groups. Coercing them and adding a note.

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
justinkaseman: Agreed. Done.

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
justinkaseman: 3 lines achstually :nerd_face: 

Agreed. Done.

packages/sources/cfbenchmarks-test/src/endpoint/common/crypto.ts
justinkaseman: Done

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
alejoberardino: Ah, good point. For these the way to make sure this is not an issue is to use the `Adapter.requestTransforms`: these will be applied to the parameters after validation but before cache keys are generated. What we should make sure is that if you would receive from the DP the same message for different params, we should be consolidating the params before in these transformations so that we can be sure it's a 1 to 1 relationship

packages/sources/cfbenchmarks-test/src/index.ts
alejoberardino: Ah, mb. Have a PR up in the repo that should fix it

justinkaseman: Still not working as expected with 0.10.1

image

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
justinkaseman: Not sure if I follow that second sentence.

Do you mean: when writing to the cache we need to match back up to the original params, but we consolidated them down, so we will need to mark `transport` and `type: primary | secondary` in the .requestTransforms and result params?

packages/sources/cfbenchmarks-test/src/index.ts
alejoberardino: I'll pull the branch and check

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
justinkaseman: Regardless, because of the primary vs secondary thing `requestTransform`s does not work cleanly for this adapter without changing the framework to also expose the `AdapterConfig`. I could read it directly from the environment, but definitely not  a good practice.

alejoberardino: I think exposing the config to the routing function makes sense, we should make the change to the framework on that one

justinkaseman: Made the change in the latest commit, check it out.

packages/sources/cfbenchmarks-test/src/endpoint/rest/crypto.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export type RestEndpointTypes = EndpointTypes & {
+  Provider: {
+    RequestBody: EmptyObject
+    ResponseBody: ProviderResponse
+  }
+}
+
+// NOTE: This is using the BatchWarming transport, but the actual API endpoint is not batchable
```
alejoberardino, line 21: I apologize, I missed this earlier. I looked at the cfbenchmarks API and I thought we were using the `latest_values` endpoint that does support batching (although it has a limit on the number of ids we can send). For the priorities relevant to this migration, I think it's fine if we leave this in an incomplete state and throw an error when not enabling websockets, and come back to this EA later on to rework the rest api

alejoberardino (changes requested):

packages/sources/cfbenchmarks-test/src/endpoint/websocket/crypto.ts
alejoberardino: Great, that should take care of the problem, now whether you request the EA with base+quote or id, after the request is validated and transformations applied, all it will use is the id, therefore making it 1-1 with the responses from the provider and always match requests to their responses in the cache

packages/sources/cfbenchmarks-test/src/index.ts
alejoberardino: `0.10.2` should fix this, second time's the charm

alejoberardino (approved):

PR #2296, opened 2022-11-04
Feature/56346 coinmetrics v3 (REST only)

mjk90:
## Description
Coinmetrics V3 adapter, with `price`, `burned` and `total-burned` REST endpoints. No WS endpoint yet, will add that in another PR as it looks like it may be blocked for now.

## Steps to Test

1. `yarn test packages/sources/coinmetrics-test/test/integration`

## Quality Assurance


packages/sources/coinmetrics-test/src/endpoint/burned.ts
```diff
@@ -0,0 +1,66 @@
...
+      assets: (asset as string).toLowerCase(),
+      metrics: METRICS,
+      frequency,
+      page_size: 1,
+      api_key: config.API_KEY as string,
+      isBurnedEndpointMode: true,
+    }
+
+    return {
+      url: `${config.API_ENDPOINT}/timeseries/asset-metrics`,
```
boxhock, line 42:
```suggestion
      url: `${config.API_ENDPOINT || DEFAULT_API_ENDPOINT}/timeseries/asset-metrics`,
```

packages/sources/coinmetrics-test/src/endpoint/burned.ts
```diff
@@ -0,0 +1,66 @@
...
+      providerStatusCode: res.status,
+      statusCode: 200,
+      result: ethers.utils.formatEther(res.data.totalBurnedTKN.toString()),
+    }
+  },
+  options: {
+    requestCoalescing: {
+      enabled: true,
+    },
+  },
```
boxhock, line 59: Should we remove this here to not forcefully enable it?

packages/sources/coinmetrics-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,79 @@
...
+      providerStatusCode: res.status,
+      statusCode: 200,
+      result,
+    }
+  },
+  options: {
+    requestCoalescing: {
+      enabled: true,
+    },
+  },
```
boxhock, line 72: Same comment as endpoint above

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
```diff
@@ -0,0 +1,93 @@
...
+      metrics: METRICS,
+      frequency,
+      page_size: pageSize,
+      api_key: config.API_KEY as string,
+      start_time: startTime,
+      end_time: endTime,
+    }
+
+    return {
+      url: `${config.API_ENDPOINT}/timeseries/asset-metrics`,
```
boxhock, line 69:
```suggestion
      url: `${config.API_ENDPOINT || DEFAULT_API_ENDPOINT}/timeseries/asset-metrics`,
```

packages/sources/coinmetrics-test/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  rateLimiting: {
+    tiers: {
+      community: {
+        rateLimit1m: 100,
+      },
+      paid: {
+        rateLimit1s: 300,
+      },
+    },
+  },
```
boxhock, line 20: Just want to confirm - no overrides?

packages/sources/coinmetrics-test/src/transports/totalBurnedTransport.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+    burnedTKN = burnedTKN.add(feeTotNTV.sub(revNtv.sub(issTotNtv)))
+  })
+  return burnedTKN
+}
+
+export class TotalBurnedTransport extends RestTransport<TotalBurnedEndpointTypes> {
+  async foregroundExecute(
+    req: AdapterRequest<TotalBurnedEndpointTypes['Request']>,
+    config: AdapterConfig<TotalBurnedEndpointTypes['CustomSettings']>,
```
boxhock, line 97: Are the explicit types necessary?

boxhock (changes requested):

packages/sources/coinmetrics-test/src/endpoint/burned.ts
mjk90: It looks like `options.requestCoalescing` is required so we need to give it `true` or `false`. I checked the framework and that value is overwritten in the `initialize` method by an env value:
```
this.config.options.requestCoalescing.enabled = config.REQUEST_COALESCING_ENABLED
```
so I think it should be ok

packages/sources/coinmetrics-test/src/index.ts
mjk90: Yeah I think so. The original `coinmetrics` adapter has no adapter level overrides

packages/sources/coinmetrics-test/src/transports/totalBurnedTransport.ts
mjk90: Due to the custom request loop in this transport, we need to pull a few values from the DP response which are specific to `TotalBurnedTransport`, so if we don't make the types explicit at the top, then we have to cast the DP response to an explicit value below. Do you think it's better to do the casting inside the `foregroundExecute` method?

alejoberardino:
This work was completed in a separate PR

PR #2303, opened 2022-11-08
add coinpaprika-test v3 ea

karen-stepanyan:
coinpaprika-test v3 ea port


packages/sources/coinpaprika-test/src/endpoint/crypto-single.ts
```diff
@@ -0,0 +1,48 @@
+import { PriceEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { RestTransport } from '@chainlink/external-adapter-framework/transports'
+import { DEFAULT_API_ENDPOINT, PRO_API_ENDPOINT } from '../config'
+import { buildUrlPath, EndpointTypesSingle, inputParameters } from '../crypto-utils'
+
+const restEndpointTransport = new RestTransport<EndpointTypesSingle>({
```
alejoberardino, line 6: v2 had support for this endpoint, but since coinpaprika has a batched one we should drop support for this one (that is, just delete it from the v3 EA)

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,219 @@
...
+  params: CryptoRequestParams[],
+  config: AdapterConfig,
+): HttpRequestConfig<CryptoRequestBody> => {
+  const headers: { Authorization?: string } = {}
+  if (config.API_KEY) {
+    headers['Authorization'] = config.API_KEY
+  }
+  return {
+    baseURL: config.API_KEY ? PRO_API_ENDPOINT : DEFAULT_API_ENDPOINT,
+    url: 'v1/tickers',
```
alejoberardino, line 186: The coinpaprika tickers response returns information for all active tickers on their end (many thousands). We should reevaluate this as it's been a problem before, even if batching is theoretically possible

alejoberardino (changes requested):

packages/sources/coinpaprika-test/src/config/index.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_API_ENDPOINT = 'https://api.coinpaprika.com'
+export const PRO_API_ENDPOINT = 'https://api-pro.coinpaprika.com'
+
+export const customSettings = {
+  API_KEY: {
+    description: 'An API key Coinpaprika',
+    type: 'string',
+    required: false,
+  },
```
amit-momin, line 9: We should mark this config as sensitive so it gets redacted in logs
```suggestion
  API_KEY: {
    description: 'An API key Coinpaprika',
    type: 'string',
    required: false,
    sensitive: true
  },
```

packages/sources/coinpaprika-test/src/config/index.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_API_ENDPOINT = 'https://api.coinpaprika.com'
+export const PRO_API_ENDPOINT = 'https://api-pro.coinpaprika.com'
+
+export const customSettings = {
```
amit-momin, line 4: We should also add the `API_ENDPOINT` config to custom settings so that it's configurable through env var. I think you could follow `coingecko` as an example

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+      ath_date: string
+      percent_from_price_ath: number
+    }
+  }
+  cost?: number
+}
+
+export interface CryptoRequestParams {
+  coinid?: string
+  base?: string
```
amit-momin, line 66: Since base is a required input param, shouldn't it be here too?

packages/sources/coinpaprika-test/src/config/index.ts
amit-momin, line 4: I also noticed `coinpaprika` has some unique functionality based on an env var called `IS_TEST_MODE`. We may need to include that here too unless there was a discussion on not needing this anymore

packages/sources/coinpaprika-test/src/endpoint/vwap.ts
```diff
@@ -0,0 +1,114 @@
...
+interface Response {
+  timestamp: string
+  price: number
+  volume_24h: number
+  market_cap: number
+}
+
+export interface RequestParams {
+  coinid?: string
+  base?: string
```
amit-momin, line 33: Same question here as above, since base is a required input param for this endpoint too.

packages/sources/coinpaprika-test/src/global-utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  const marketMap: { [key: string]: string } = {
+    BTC: 'bitcoin',
+  }
+
+  const propertyPath =
+    resultPath === '_dominance_percentage'
+      ? `${marketMap[requestPayload.market.toUpperCase()]}${resultPath}`
+      : `market_cap_${requestPayload.market.toLowerCase()}`
```
amit-momin, line 76: This is me nitpicking but just to keep both sides of the conditional consistent
```suggestion
  const propertyPath =
    resultPath === '_dominance_percentage'
      ? `${marketMap[requestPayload.market.toUpperCase()]}_dominance_percentage`
      : `market_cap_${requestPayload.market.toLowerCase()}`
```
Or you could replace `market_cap_` with `${resultPath}`. Whichever you prefer

packages/sources/coinpaprika-test/src/config/index.ts
amit-momin: This would also require changes where you prepare requests

amit-momin: This would also require changes where you prepare requests since you'd have to set a header based on this.

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,142 @@
...
+        baseURL: getApiEndpoint(config),
+        url: 'v1/tickers',
+        method: 'GET',
+        headers: getApiHeaders(config),
+        params: {
+          quotes: [...new Set(chunkedParams.map((p) => p.quote.toUpperCase()))].join(','),
+        },
+      },
+    }
+  })
```
boxhock, line 105: Nitpick, but imo this seems cleaner - what do you think?

```suggestion
  return chunkedMatrix.map((chunkedParams) => ({
    params: chunkedParams,
    request: {
      baseURL: getApiEndpoint(config),
      url: 'v1/tickers',
      method: 'GET',
      headers: getApiHeaders(config),
      params: {
        quotes: [...new Set(chunkedParams.map((p) => p.quote.toUpperCase()))].join(','),
      },
    },
  }))
```

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export const constructEntry = (
+  res: CryptoResponseSchema[],
+  requestPayload: CryptoRequestParams,
+  resultPath: 'price' | 'market_cap' | 'volume_24h',
+) => {
```
boxhock, line 112: Can we please enforce return type here?

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+const chunkArray = (params: CryptoRequestParams[], size = 3): CryptoRequestParams[][] =>
+  params.length > size ? [params.slice(0, size), ...chunkArray(params.slice(size), size)] : [params]
+
+export const buildBatchedRequestBody = (
+  params: CryptoRequestParams[],
+  config: AdapterConfig<typeof customSettings>,
+) => {
```
boxhock, line 88: Enforce return type

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,142 @@
...
+        },
+      },
+    }
+  })
+}
+
+export const constructEntry = (
+  res: CryptoResponseSchema[],
+  requestPayload: CryptoRequestParams,
+  resultPath: 'price' | 'market_cap' | 'volume_24h',
```
boxhock, line 111: If we want to do this properly, we should split out the type starting on L41-42 into it's own named type, then here do `keyof TYPE_NAME_HERE`

boxhock (changes requested):

packages/sources/coinpaprika-test/src/crypto-utils.ts
karen-stepanyan: i think for this one we need to export `ProviderRequestConfig` type from the framework

boxhock: Hmm I see. For now, can we do `HttpRequestConfig<never>[]` like we did in the coingecko EA?

karen-stepanyan: I think`HttpRequestConfig` is not available for newer version of the framework that coinpaprika is using.

boxhock: Yeah you're right. Do you think it makes sense to export `ProviderRequestConfig` from the framework? If so, could you get a PR up for that?

I'd just prefer enforcing types here to avoid future issues

karen-stepanyan: I think having explicit type would be great, tho in this particular case it doesn't change anything. Since every return property of `buildBatchedRequestBody` or `constructEntry` is stated clearly with their own types, even without having return type for the whole function if i change return structure to something wrong i will get an error for `prepareRequests` or `parseResponse` at the end .

boxhock: Ok makes sense. Let's skip this one and others that would clearly fail

packages/sources/coinpaprika-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,141 @@
...
+      baseURL: getApiEndpoint(config),
+      url: 'v1/tickers',
+      method: 'GET',
+      headers: getApiHeaders(config),
+      params: {
+        quotes: [...new Set(chunkedParams.map((p) => p.quote.toUpperCase()))].join(','),
+      },
+    },
+  }))
+}
```
boxhock, line 105: This part needs to be re-worked a bit.

Currently, this will split up the params set into chunks of 3 _price pair_ requests. However, we should split it into chunks of 3 _quotes_.

Currently, the following input leads to the following output:
```
[ ETH/USD, BTC/USD, LINK/USD, BUSD/USD ] -> [ [ ETH/USD, BTC/USD, LINK/USD ], [ BUSD/USD ] ]
```

Instead, these should all be sent in the same request (`/USD`) - and it should be able to support 2 more unique quotes until it splits requests up into a new chunk.

boxhock (changes requested):

packages/sources/coinpaprika-test/src/crypto-utils.ts
karen-stepanyan: Yes, you are absolutely right. Somehow i lost the context and thought about limit of 3 requests. I will change this.

boxhock (approved):
Good to merge after the next EA release goes out later today

packages/sources/coinpaprika-test/src/crypto-utils.ts
amit-momin: Since we can only batch 3 unique quotes in a single request, I noticed testing locally that the background execute takes longer to complete than the default cache max age (1.5min) when we have more than 6 unique quotes in the subscription set. In that scenario, we're calling the `v1/tickers` endpoint more than twice. Not sure what the alternative is though. The only other endpoint we can use is for a single base. @boxhock if you have any thoughts/suggestions.

boxhock: > I noticed testing locally that the background execute takes longer to complete than the default cache max age (1.5min) when we have more than 6 unique quotes in the subscription set

This is if you're on the lowest API rate limit tier, right? If so, I'd argue that it's simply a limitation of how many different requests you can send to Coingecko on the free tier.

Is this also happening on the paid tier?

amit-momin: I was testing with a paid tier but realized I didn't set the tier in the env vars. That's probably what caused the slow background execute. I'll test again with that

amit-momin: Retested with the rate limit tier set and it took 13 unique quotes without any issue. And that's not any sort of limit, it's just where I stopped testing. I'm ok with resolving this conversation if you guys are.

amit-momin: I'm going to mark this as resolved since the v3 version seems to be capable of handling multiple calls to this endpoint per background execute cycle.

packages/sources/coinpaprika-test/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  crypto,
+  crypto_market_cap,
+  crypto_volume,
+  dominance,
+  globalmarketcap,
+  coins,
+  vwap,
+} from './endpoint'
+
+export const adapter = new Adapter({
```
amit-momin, line 14:
```suggestion
export const adapter = new PriceAdapter({
```

amit-momin (dismissed):

amit-momin (approved):

PR #2309, opened 2022-11-09
Feature/56347 - Galaxy v3

mjk90:
## Description
Added galaxy-test adapter using v3 framework.

......

## Steps to Test

1. `yarn test packages/sources/galaxy-test/test`

## Quality Assurance

- [x If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `<ADAPTER_PACKAGE>/schemas/env.json` and `<ADAPTER_PACKAGE>/README.md`


packages/sources/galaxy-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,50 @@
...
+// Split by the characters "_" and "/" as pairs are in this format: "markPrice_BTC/USD"
+const SPLIT_PAIR_REGEX = /[_/]+/
+
+const getPair = (params: AdapterRequestParams) => {
+  const { base = '', quote = '' } = params
+  return base && quote && `markPrice_${base.toUpperCase()}/${quote.toUpperCase()}`
+}
+
+export const priceTransport = new WebSocketTransport<PriceEndpointTypes>({
+  url: () => DEFAULT_WS_API_ENDPOINT,
```
alejoberardino, line 19: We can instead add WS_API_ENDPOINT to the custom settings with a default to what's now in this constant

packages/sources/galaxy-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,50 @@
...
+  priceEndpointInputParameters,
+} from '@chainlink/external-adapter-framework/adapter'
+import { getAccessToken } from '../util'
+import { AdapterRequestParams, PriceEndpointTypes } from '../types'
+
+// Split by the characters "_" and "/" as pairs are in this format: "markPrice_BTC/USD"
+const SPLIT_PAIR_REGEX = /[_/]+/
+
+const getPair = (params: AdapterRequestParams) => {
+  const { base = '', quote = '' } = params
```
alejoberardino, line 14: Are the empty strings a valid case for the params? Base and Quote should be required, so we don't need the defaults here

packages/sources/galaxy-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,50 @@
...
+  return base && quote && `markPrice_${base.toUpperCase()}/${quote.toUpperCase()}`
```
alejoberardino, line 15: Since they are required we also wouldn't need to check for their validity here

packages/sources/galaxy-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,50 @@
...
+          params: { base, quote },
+          value: message.value,
+        },
+      ]
+    },
+  },
+  builders: {
+    subscribeMessage: (params) => JSON.stringify({ type: 'subscribe', signals: [getPair(params)] }),
+    unsubscribeMessage: (params) =>
+      JSON.stringify({ type: 'unsubscribe', signals: [getPair(params)] }),
```
alejoberardino, line 42: The websocket transport stringifies by default, so it shouldn't be necessary to do it here

alejoberardino (changes requested):

packages/sources/galaxy-test/src/endpoint/price.ts
KuphJr: Also, my understanding is that `base` is always going to be defined as it is initialized to the empty string by default.
Therefore, based on the function of logical `and` with strings, won't this function always return whatever the value for `base` is?

If I am incorrect, please clarify as I would like to know how this works.

mjk90: you're right, this was from the v2 adapter. Will update it

packages/sources/galaxy-test/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "^0.10.0",
+    "mock-socket": "^9.1.5",
```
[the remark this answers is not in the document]
alejoberardino: Woops I don't know what the github comment showed, I meant mock-socket should be dev, the framework needs to be a normal one

alejoberardino (changes requested):

alejoberardino (approved):
LGTM 🚀

packages/sources/galaxy-test/src/index.ts
```diff
@@ -0,0 +1,16 @@
+import { expose, ServerInstance } from '@chainlink/external-adapter-framework'
+import { PriceAdapter } from '@chainlink/external-adapter-framework/adapter'
+import { customSettings } from './config'
+import { price } from './endpoint'
+import { Adapter } from '@chainlink/external-adapter-framework/adapter'
+import { SettingsMap } from '@chainlink/external-adapter-framework/config'
+
+export const adapter = new PriceAdapter({
+  name: 'galaxy',
```
alejoberardino, line 9: Should be uppercase

alejoberardino (dismissed):

boxhock (approved):

PR #2320, opened 2022-11-11
feature/56346-coinmetrics-v3-redo

ad0ll:
This is the MVP for the coinmetrics framework EA. This PR is only concerned with price-ws, although some of the work from feature/56346-coinmetrics-v3 appears here as well (but it's on the cutting room floor, temporarily, due to a tight deadline on price-ws specifically)

It should be functionally identical to the original, but the section relating to the websocket has been all but rewritten:
https://github.com/smartcontractkit/external-adapters-js/pull/2312/files#diff-0f6abfc65e95022e708bd46edd285bf2da216190b3627f6cdcbfc5717452a1bc
(see: [here for original implementation](https://github.com/smartcontractkit/external-adapters-js/blob/ae96beaf3f886130dd6fb88b767e28d37b55cdd8/packages/sources/coinmetrics/src/adapter.ts#L1)


packages/sources/coinmetrics-test/src/config/index.ts
```diff
@@ -0,0 +1,63 @@
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+
+export const DEFAULT_API_ENDPOINT = 'https://api.coinmetrics.io/v4'
+export const DEFAULT_WS_API_ENDPOINT = 'wss://api.coinmetrics.io/v4'
```
alejoberardino, line 4: We shouldn't export these so we don't use them by mistake, better to just put the hardcoded string directly in the custom settings

packages/sources/coinmetrics-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,124 @@
...
+  generated.searchParams.append('api_key', API_KEY)
+  logger.debug(`Generated URL: ${generated.toString()}`)
+  return generated.toString()
+}
+
+export const handleAssetMetricsMessage = (
+  message: WsAssetMetricsMessage,
+): ProviderResult<WsAssetMetricsEndpointTypes>[] | undefined => {
+  logger.trace(message, 'Got response from websocket')
+  //Cast message to all of its known types so we can check which variant we received
```
alejoberardino, line 80: Can remove, we're no longer doing it that way

packages/sources/coinmetrics-test/src/config/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  WS_API_ENDPOINT: {
+    description: 'The websocket url for coinmetrics',
+    type: 'string',
+    required: true,
+    default: DEFAULT_WS_API_ENDPOINT,
+  },
+  API_ENDPOINT: {
+    description: 'The API url for coinmetrics',
+    type: 'string',
+    required: true,
```
alejoberardino, line 31: These shouldn't be required since they have defaults

packages/sources/coinmetrics-test/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.5.63",
+    "ethers": "^5.7.2",
```
alejoberardino, line 32: this dependency is unnecessary I think

alejoberardino (changes requested):

alejoberardino (approved):

PR #2322, opened 2022-11-14
 add nomics-test v3 ea

karen-stepanyan:


packages/sources/nomics-test/src/endpoint/filtered.ts
```diff
@@ -0,0 +1,94 @@
...
+  CustomSettings: typeof customSettings
+  Provider: {
+    RequestBody: ProviderRequestBody
+    ResponseBody: ResponseSchema
+  }
+}
+
+const restEndpointTransport = new RestTransport<FilteredEndpointTypes>({
+  prepareRequest: (p, config) => {
+    const baseURL = config.API_ENDPOINT
```
alejoberardino, line 55: Seems like nomics supports batches on the base side, but not the quote. Same thing as the kaiko one, we should block this one.

alejoberardino (dismissed):

packages/sources/nomics-test/src/config/index.ts
```diff
@@ -0,0 +1,13 @@
+export const customSettings = {
+  API_KEY: {
+    description: 'An API key that can be obtained from https://p.nomics.com/pricing#free-plan',
+    type: 'string',
+    required: true,
+  },
```
boxhock, line 6: Make this sensitive?

packages/sources/nomics-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,153 @@
...
+  high_timestamp: string
+  '1d': {
+    volume: string
+    price_change: string
+    price_change_pct: string
+    volume_change: string
+    volume_change_pct: string
+    market_cap_change: string
+    market_cap_change_pct: string
+  }
```
boxhock, line 55: Can we extract this into its own type and re-use that? Assuming it's the same payload for each of these time ranges

packages/sources/nomics-test/src/endpoint/filtered.ts
boxhock: Can this be unblocked now? Can we make this EA send batches on the base, like the crypto endpoint?

boxhock (changes requested):

packages/sources/nomics-test/src/endpoint/filtered.ts
karen-stepanyan: I don't really think `/prices/restricted` API endpoint is partially batchable. As a matter of fact i didn't even find any documentation for that. There is a `prices` API endpoint where just one ticker is allowed.

packages/sources/nomics-test/src/config/index.ts
bogos: Done

packages/sources/nomics-test/src/endpoint/filtered.ts
boxhock: Should we just mark this as resolved then?

boxhock (approved):
LGTM

amit-momin:
I think this is good to go but let's do one more framework bump before we merge this. The recent patches had some important fixes

amit-momin (approved):

PR #2331, opened 2022-11-15
 add kaiko-test v3 ea

karen-stepanyan:


packages/sources/kaiko-test/src/config/index.ts
```diff
@@ -0,0 +1,13 @@
+export const DEFAULT_INTERVAL = '1m'
+export const DEFAULT_SORT = 'desc'
+export const DEFAULT_MILLISECONDS = 1800000
+export const DEFAULT_ENDPOINT = 'trades'
+export const DEFAULT_API_ENDPOINT = 'https://us.market-api.kaiko.io/v2/data/trades.v1'
```
alejoberardino, line 5: Let's not use a separate var for this, it shouldn't be used outside of the adapter index so better to hardcode the value in the adapter obj

packages/sources/kaiko-test/src/endpoint/trades.ts
```diff
@@ -0,0 +1,151 @@
...
+const restEndpointTransport = new RestTransport<EndpointTypes>({
+  prepareRequest: (req, config) => {
+    const data = req.requestContext.data
+    const base = data.base.toLowerCase()
+    const quote = data.quote.toLowerCase()
+    const url = `/spot_exchange_rate/${base}/${quote}`
+
+    const interval = data.interval || DEFAULT_INTERVAL
+    const start_time = calculateStartTime(data.millisecondsAgo || DEFAULT_MILLISECONDS)
+    const sort = data.sort || DEFAULT_SORT
```
alejoberardino, line 107: the input parameters have a default value, they should never be undefined so there's no reason to use the defaults here too

packages/sources/kaiko-test/src/config/index.ts
alejoberardino: Same thing goes for the others actually, they shouldn't be used other places

packages/sources/kaiko-test/src/endpoint/trades.ts
```diff
@@ -0,0 +1,151 @@
...
+    default: 'desc',
+  },
+} as const
+
+export interface RequestParams {
+  base: string
+  quote: string
+  interval?: string
+  millisecondsAgo?: number
+  sort?: string
```
alejoberardino, line 44: These will always have a value because of the defaults

packages/sources/kaiko-test/src/endpoint/trades.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+const calculateStartTime = (millisecondsAgo: number) => {
+  const date = new Date()
+  date.setTime(date.getTime() - millisecondsAgo)
+  return date
+}
+
+const restEndpointTransport = new RestTransport<EndpointTypes>({
```
alejoberardino, line 98: Rest transport is going away so we'll block this one until we can either move to a batch endpoint (there seem to be a few options for kaiko) or have support for the more inefficient single requests

packages/sources/kaiko-test/src/endpoint/trades.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (data.length == 0) {
+      throw new AdapterError({
+        message:
+          'Kaiko is not returning any price data for this price pair, likely due to too low trading volume for the requested interval. This is not an issue with the external adapter.',
+      })
+    }
+    let price = Number(res.data.data[0].price)
+    if (inverse && price != 0) {
+      price = 1 / price
+    }
```
alejoberardino, line 131: inverses are already calculated by the price constructs, this should not not needed

alejoberardino (changes requested):
There's a few requested changes we can tackle, but since we don't support non-batch endpoints right now we won't push this through until we resolve that

packages/sources/kaiko-test/src/config/index.ts
```diff
@@ -0,0 +1,12 @@
+export const customSettings = {
+  API_KEY: {
+    description: 'API KEY for  KAIKO',
+    type: 'string',
+    required: true,
+  },
```
amit-momin, line 6:
```suggestion
  API_KEY: {
    description: 'API KEY for  KAIKO',
    type: 'string',
    required: true,
    sensitive: true,
  },
```

packages/sources/kaiko-test/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  rateLimiting: {
+    tiers: {
+      default: {
+        rateLimit1h: -1,
+        note: 'claims no rate limits',
+      },
+    },
+  },
+  envDefaultOverrides: {
+    API_TIMEOUT: 10000,
```
amit-momin, line 22: Based on Kaiko v2 I think we should set this to 30000
```suggestion
    API_TIMEOUT: 30000,
```

packages/sources/kaiko-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.15.1",
```
boxhock, line 31: This EA looks good to me, but let's wait until we have the metrics fix in the EA framework

boxhock (approved):

amit-momin (approved):

PR #2335, opened 2022-11-16
Release/v1.44.1 (DROPPED)

ad0ll:
### Breaking Changes
- stader-address-list: Update contract ABI

### Notable Adapter Updates
|    Adapter    | Version | Description |
| :-----------: | :-----: | :---------: |
| stader-address-list  | v1.2.0  | Update contract ABI |


amit-momin (dismissed):

amit-momin (approved):

justinkaseman:
It looks like this release is a full release, and not just the stader fix - which is okay to do, but is that intentional?
If so we will want to document all changes in the changelog.

PR #2380, opened 2022-12-02
fix/PDI-169: Update bank-frick-test to latest framework version

ad0ll:
<b style="color:red">This PR shouldn't be merged until the circular structure issue is resolved (PDI-182) and the framework dependency in bank-frick-test has been updated to use it</b>

This updates bank-frick-test to the latest framework version. It was using a very old version, so some of the changes are substantial.


packages/sources/bank-frick-test/src/config/index.ts
```diff
@@ -9,6 +9,17 @@ export const signingAlgorithms: SigningAlgorithm[] = ['rsa-sha256', 'rsa-sha384'
 const MAX_PAGE_SIZE = 500
 
 export const customSettings = {
+  API_ENDPOINT: {
+    description: 'The endpoint to use for making requests to the Bank Frick API',
+    type: 'string',
+    required: false,
+    default: 'https://olbsandbox.bankfrick.li',
```
amit-momin, line 16: Noticed a discrepancy in this default from the old env var override `https://olbsandbox.bankfrick.li/webapi/v2`. Is this intentional? Don't think I see `/webapi/v2` appended later.

amit-momin (dismissed):

amit-momin (approved):

amit-momin (approved):

PR #2385, opened 2022-12-05
Upgrade GSR adapter from v2 to v3

boxhock:


.changeset/weak-plums-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/gsr-adapter': minor
```
alejoberardino, line 2: Should we release this as a major version instead?

alejoberardino (dismissed):

.changeset/weak-plums-ring.md
boxhock: Yeah, I wouldn't mind, since there are so many framework changes. Will update

alejoberardino (approved):

PR #2404, opened 2022-12-09
ready for draft review

KuphJr:


packages/sources/universal-v3/CHANGELOG.md
```diff
@@ -0,0 +1,303 @@
+# @chainlink/example-source-adapter
```
justinkaseman, line 1: Rename title and clear contents

packages/sources/universal-v3/src/endpoint/lambda.ts
```diff
@@ -0,0 +1,203 @@
...
+    await this.lambdaController.initalize()
+    return
+  }
+
+  // This function handles each request sent to the EA
+  foregroundExecute = async (
+    req: AdapterRequest<Request>,
+    config: AdapterConfig<typeof customSettings>,
+  ): Promise<AdapterResponse<Response>> => {
+    const requestId = req.requestContext.data.requestId
```
justinkaseman, line 129: More readable to destructure? ¯\_(ツ)_/¯
```suggestion
    const { requestId, source, subscriptionId, secrets, secretsOwner, args, numAllowedQueries, maxResponseBytes} = req.requestContext.data
```

packages/sources/universal-v3/src/endpoint/lambda.ts
```diff
@@ -0,0 +1,203 @@
...
+    required: true,
+  },
+  requestId: {
+    description: 'Request Id unique to each request',
+    required: true,
+  },
+  maxResponseBytes: {
+    description: 'Maximum size of response data, in bytes',
+  },
+  numAllowedQueries: {
```
justinkaseman, line 39: Where does `numAllowedQueries` and `maxResponseBytes` come from? They aren't in the request CBOR - so they are set staticly in the OCR plugin?

packages/sources/universal-v3/src/endpoint/lambda.ts
```diff
@@ -0,0 +1,203 @@
...
+    const subscriptionId = req.requestContext.data.subscriptionId
+    const secrets = req.requestContext.data.secrets
+    const secretsOwner = req.requestContext.data.secretsOwner
+    const args = req.requestContext.data.args
+    const numAllowedQueries = req.requestContext.data.numAllowedQueries
+    const maxResponseBytes = req.requestContext.data.maxResponseBytes
+
+    const responseBuilder = new ResponseBuilder(config)
+
+    let decryptedSecrets: Record<string, unknown> = {}
```
justinkaseman, line 140: Secrets have to be strings right?
```suggestion
    let decryptedSecrets: Record<string, string> = {}
```

packages/sources/universal-v3/src/endpoint/lambda.ts
KuphJr: They are optional params that we can add in the future if we want to charge different amounts based on the size of the response or the number of queries.  They are fully functional right now, but not required params.  They have default values within the Lambda sandbox (256 byte max response & 5 max queries, both of which are configurable via env var)

packages/sources/universal-v3/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+    description:
+      'Number of milliseconds to wait before trying to invoke a Lambda function after it is created',
+    default: 500,
+  },
+  LAMBDA_MEMORY_SIZE_MB: {
+    type: 'number',
+    description: 'The maximum amount of RAM that can be used by a Lambda function',
+    default: 512,
+  },
+  MAX_RESPONSE_BYTES: {
```
justinkaseman, line 102: I wonder about whether the NOP should have control over all of these via environment variable.

This one, for example, should probably be tied to the input `maxResponseBytes`?

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+    if (this.deployedLambdaFunctions.count < this.config.LAMBDA_MAX_DEPLOYED_FUNCTIONS) {
+      return this.config.LAMBDA_PRUNER_LOOP_WAIT_MS
+    }
+
+    const leastRecentlyUsedFunction = await this.getLeastRecentlyUsedFunction()
+
+    try {
+      await this.deleteLambdaFunction(leastRecentlyUsedFunction)
+    } catch (error) {
+      this.logger.error(`Error deleteing lambda function ${leastRecentlyUsedFunction}: ${error}`)
```
justinkaseman, line 70:
```suggestion
      this.logger.error(`Error deleting lambda function ${leastRecentlyUsedFunction}: ${error}`)
```
typo

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+      return 1
+    }
+
+    delete this.deployedLambdaFunctions.timeLastUsed[leastRecentlyUsedFunction]
```
justinkaseman, line 74: Does this only clean it up in the EA state?
What about cleaning it up from AWS Lambda?

packages/sources/universal-v3/src/index.ts
KuphJr: I am not sure, hence I left it as an env var for now.  We should probably have a discussion about what NOPs should and should not be able to configure.

packages/sources/universal-v3/src/utils/LambdaController.ts
KuphJr: On line 68, we try to delete the Lambda function.  If the delete succeeds (ie: no error is caught), we update the EA state by deleting the entry in `deployedLambdaFunctions.timeLastUsed`

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+        let deployedLambdaFunctions: DeployedLambdaFunctions = { count: 0, timeLastUsed: {} }
+
+        // Each call to getDeployedLambdaFunctions fetches up to 50 Lambda functions at a time.
+        // If more than 50 Lambda functions exist, the next page must be fetched using the NextMarker.
+        if (data.NextMarker) {
+          deployedLambdaFunctions = await this.getDeployedLambdaFunctions(data.NextMarker)
+        }
+
+        functions.forEach(({ FunctionName }) => {
+          if (FunctionName) deployedLambdaFunctions.timeLastUsed[FunctionName] = 0
```
justinkaseman, line 301: Would it be better to use something from the `listFunctions` response instead of 0? Such as:
```
         "LastModified": "string",
         "LastUpdateStatus": "string",
```

packages/sources/universal-v3/src/utils/LambdaController.ts
justinkaseman: Should this be in the try block then?

Like if we can't delete it from AWS, then it still exists. Maybe re-try the deletion?

KuphJr: Oh, there is a return statement in the catch block, but yeah I could improve this for readability.

packages/sources/universal-v3/src/utils/LambdaController.ts
KuphJr: hmmmm that's a good call.

On first initialization, we should probably keep the OLDEST lambda function as that is the one that is most likely to be most frequently used.  Then, after the initialization, we would keep the existing logic which deletes the least-recently-used function.

I'll put some more thought into this.

KuphJr: `LastUpdateStatus` will always be null for our use case since functions are never updated after being created (I have tested this).  They are only either created or deleted.  Additionally, `LastModified` will always equal the time of creation.

As for a solution to the initial ordering:

Imagine we have deployed Lambda functions `0` through `n`, in sorted order, with `0` being the oldest function and `n` being the newest.

We assign the `timeLastUsed` in reverse order, such that the oldest function is assigned `n` and the newest is assigned `0`.  This means that, as soon as the max # of functions is exceeded, the newest deployed function is first to be deleted.

However, since `n` will always be less than the current timestamp, if a new function is created _after_ initialization, we maintain the functionality where the least-recently-used function is deleted first.

Does that make sense?

packages/sources/universal-v3/src/utils/buildResponse.ts
```diff
@@ -0,0 +1,183 @@
...
+        (!sandboxResponse.error.message || typeof sandboxResponse.error.message === 'string') &&
+        (!sandboxResponse.error.details || typeof sandboxResponse.error.details === 'string') &&
+        (!sandboxResponse.error.userHttpQueries ||
+          Array.isArray(sandboxResponse.error.userHttpQueries))
+      ) {
+        return true
+      }
+    }
+
+    return false
```
alejoberardino, line 168:
```suggestion
       if (typeof unknownResponse !== 'object') {
      return false
    }

    const sandboxResponse = unknownResponse as SandboxError
    return (
      typeof sandboxResponse.error === 'object' &&
      (typeof sandboxResponse.error.message === 'string' ||
        typeof sandboxResponse.error.details === 'string' ||
        Array.isArray(sandboxResponse.error.userHttpQueries))
    )
```

packages/sources/universal-v3/src/utils/secretsDecrypter.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // check signature (NOTE: the ability off signature checking temporary & will be removed)
+  if (signer !== secretsOwner) {
+    throw Error('Encrypted secrets not signed by subscription owner')
+  }
+
+  try {
+    return JSON.parse(decryptedPayload.message)
+  } catch {} // eslint-disable-line no-empty
+  throw Error('Decrypted secrets are not a valid JSON string')
```
alejoberardino, line 45:
```suggestion
  try {
    return JSON.parse(decryptedPayload.message)
  } catch {
    throw Error('Decrypted secrets are not a valid JSON string')
   }
```

packages/sources/universal-v3/src/utils/secretsDecrypter.ts
```diff
@@ -0,0 +1,46 @@
...
+export const decrypt = async (
+  privateKey: string,
+  encryptedBytesString: Base64ByteString,
+  secretsOwner: Address,
+): Promise<Record<string, unknown>> => {
+  let encryptedHexString: HexString
+  let encryptedObject: Encrypted
+  let decrypted: JSONstring
+  let decryptedPayload: { signature: string; message: string }
+  let signer: Address
```
alejoberardino, line 19: The only things used outside of the try catch are the signer and decrypted payload, everything else should be a const

alejoberardino (changes requested):

packages/sources/universal-v3/src/endpoint/lambda.ts
```diff
@@ -0,0 +1,205 @@
...
+      })
+    }
+
+    return responseBuilder.buildResponse(sandboxOutput)
+  }
+
+  // This function in the background in a loop and waits for returned number of milliseconds between each iteration
+  backgroundExecute = async (): Promise<number> => {
+    return await this.lambdaController.lambdaPruner()
+  }
```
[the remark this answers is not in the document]
KuphJr: Gotcha. I can update this such that the lambda container is recursive.

However, how do I implement a graceful shutdown within my `lambdaPruner`? Is there some sort of shutdown flag?

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+      this.logger.error(`Lambda retry count exceeded`)
+      reject(err)
+      return
+    }
+
+    // Retry after LAMBDA_RETRY_TIME_MS
+    setTimeout(
+      () => resolve(this.executeRequestInLambda(FunctionName, sandboxRequestData, retryCount--)),
+      this.config.LAMBDA_RETRY_TIME_MS,
+    )
```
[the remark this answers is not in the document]
KuphJr: Correct.  We decrement retries.  This final condition should never be met, however I want to catch and report any unexpected error messages from the AWS SDK.

packages/sources/universal-v3/src/utils/buildResponse.ts
```diff
@@ -0,0 +1,183 @@
...
+        error: this.stringToValidHexString(errorString),
+        errorString,
+        details: sandboxOutput.error.details,
+        userHttpQueries: sandboxOutput.error.userHttpQueries,
+      },
+      statusCode: 200,
+      timestamps: {
+        providerDataReceived: 0,
+        providerDataRequested: 0,
+        providerIndicatedTime: 0,
```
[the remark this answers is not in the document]
KuphJr: Apparently `undefined` is not an accepted type, but I can use `NaN`.

Since this is out of scope and I need to deliver by EoD, I am going to ignore this for now to make the deadline.  However, I will make a ticket and add it to the sprint.

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    const invokeRequest: Lambda.InvocationRequest = {
+      FunctionName,
+      Payload: JSON.stringify(sandboxRequestData),
+    }
+
+    return new Promise((resolve, reject) => {
+      this.lambda.invoke(invokeRequest, (err, data) => {
+        this.logger.trace(`Invoking Lambda function name ${FunctionName}`)
```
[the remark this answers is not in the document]
KuphJr: It does not unfortunately AFAIK

packages/sources/universal-v3/src/utils/LambdaController.ts
```diff
@@ -0,0 +1,348 @@
...
+        })
+
+        deployedLambdaFunctions.count = Object.keys(deployedLambdaFunctions.timeLastUsed).length
+
+        resolve(deployedLambdaFunctions)
+      })
+    })
+  }
+
+  private getLeastRecentlyUsedFunction = async (): Promise<string> => {
```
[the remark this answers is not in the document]
KuphJr: This is true, however if you implement your suggestion, you are incurring more compute cost each time the function is called as the array needs to be re-ordered.

I instead chose the simple approach and decided to incur the cost on getLeastRecentlyUsedFunction as this will be called less frequently than invokeLambdaFunction.

packages/sources/universal-v3/src/utils/LambdaController.ts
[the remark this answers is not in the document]
alejoberardino: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Lambda.html#createFunction-property

In v2 of the aws-sdk you can do it by not providing a callback and chaining .promise() on the returned request, i.e.
```typescript
this.lambda.invoke(invokeRequest).promise()
```
It would be better to migrate to v3 of the sdk since it's the current one and v2 will be deprecated (or is now, I'm not sure), but I think v3 is built for Node 18 so there might be some additional work needed, we don't want to run EAs on different node versions

packages/sources/universal-v3/src/utils/LambdaController.ts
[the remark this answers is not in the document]
alejoberardino: You don't need to reorder an array, implementing a simple double linked list makes shifting a known node to the head or tail a constant time operation (there are more efficient methods still, that's just an easy one). It's true that it adds a few OPs more on each invocation, that's why I was asking about the number of items that will be relevant here, but yeah if it's low 4 digits and this is called infrequently I agree it's not worth the extra complexity atm

packages/sources/universal-v3/src/utils/buildResponse.ts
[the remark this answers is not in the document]
alejoberardino: Undefined is only valid for the `providerIndicatedTime` (since not all providers specify one), but for the others we should be able to get values for all of them. Maybe it's better to keep them as 0 instead of NaN if the work will be done later, but it should be carried out at some point to avoid us having outliers in our metrics

packages/sources/universal-v3/src/utils/buildResponse.ts
```diff
@@ -0,0 +1,169 @@
...
+      return (
+        typeof sandboxResponse.error === 'object' &&
+        (!sandboxResponse.error.message || typeof sandboxResponse.error.message === 'string') &&
+        (!sandboxResponse.error.details || typeof sandboxResponse.error.details === 'string') &&
+        (!sandboxResponse.error.userHttpQueries ||
+          Array.isArray(sandboxResponse.error.userHttpQueries))
+      )
+    }
+
+    return false
```
alejoberardino, line 154:
```suggestion
     if (typeof unknownResponse !== 'object') {
      return false
    }
    
    const sandboxResponse = unknownResponse as SandboxError
    return (
      typeof sandboxResponse.error === 'object' &&
      (typeof sandboxResponse.error.message === 'string' ||
        typeof sandboxResponse.error.details === 'string' ||
        Array.isArray(sandboxResponse.error.userHttpQueries))
    )
```

alejoberardino (changes requested):

PR #2412, opened 2022-12-12
feature/PDI-169: Promote bank-frick v3/Deprecate bank-frick v2

ad0ll:
This deprecates the v2 bank-frick adapter for v3.

Note, for some reason, bank-frick never had a v1.0.0 release (last version before the port was v0.1.1), most likely because we weren't certain if it was going to get blocked for issues surrounding BF's daily maintenance. Although most EAs are seeing their major version ticked from v1.x.x to v2.0.0, **bank-frick is getting ticked from v0.x.x to v1.0.0**

### DIFF
--- a/packages/core/legos/tsconfig.json
+++ b/packages/core/legos/tsconfig.json
@@ -44,9 +44,6 @@
       "path": "../../sources/bank-frick"
     },
     {
-      "path": "../../sources/bank-frick"
-    },
-    {
       "path": "../../sources/bea"
     },
     {
--- a/packages/tsconfig.json
+++ b/packages/tsconfig.json
@@ -156,9 +156,6 @@
       "path": "./sources/bank-frick"
     },
     {
-      "path": "./sources/bank-frick"
-    },
-    {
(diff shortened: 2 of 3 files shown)


packages/core/legos/tsconfig.json
```diff
@@ -44,7 +44,7 @@
       "path": "../../sources/bank-frick"
```
amit-momin, line 44: I think just changing the name left 2 `bank-frick`'s in the tsconfig

packages/tsconfig.json
```diff
@@ -156,7 +156,7 @@
       "path": "./sources/bank-frick"
```
amit-momin, line 156: Same with this one

packages/tsconfig.test.json
```diff
@@ -156,7 +156,7 @@
       "path": "./sources/bank-frick/tsconfig.test.json"
```
amit-momin, line 156: And this one

amit-momin (approved):

PR #2414, opened 2022-12-14
added dxfeed-test EA

karen-stepanyan:
dxfeed v3 port
[PDI-191](https://smartcontract-it.atlassian.net/browse/PDI-191)

[PDI-191]: https://smartcontract-it.atlassian.net/browse/PDI-191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/dxfeed-test/src/config/index.ts
```diff
@@ -0,0 +1,36 @@
+export const customSettings = {
+  API_USERNAME: {
+    description: 'username for dxfeed API',
+    type: 'string',
+    default: '',
```
boxhock, line 5: Should this have a default? Or remove the default and make it `required: true`?

packages/sources/dxfeed-test/src/config/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  API_PASSWORD: {
+    description: 'password for dxfeed API',
+    type: 'string',
+    sensitive: true,
+    default: '',
```
boxhock, line 11: Same here

packages/sources/dxfeed-test/src/config/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  WS_API_ENDPOINT: {
+    description: 'The websocket url for dxfeed',
+    type: 'string',
+    default: '',
```
boxhock, line 16: Remove default?

packages/sources/dxfeed-test/src/endpoint/price-router.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    WS: wsTransport,
+    REST: batchTransport,
+  },
+  (_, adapterConfig) => {
+    if ((adapterConfig && adapterConfig.WS_ENABLED) || process.env['WS_ENABLED'] === 'true') {
```
boxhock, line 82: We should only rely on `adapterConfig`

```suggestion
    if (adapterConfig.WS_ENABLED) {
```

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,146 @@
...
+  {
+    id: 3,
+    channel: META_CONNECT,
+    clientId: '',
+    connectionType: 'websocket',
+  },
+]
+
+class DxFeedWebsocketTransport extends WebSocketTransport<EndpointTypes> {
+  connectionClientId = ''
```
boxhock, line 73: Instead of doing the messages as consts above, can we move them inside of this class and have them just get the `connectionClientId` from within the class? That way we avoid having to remember to set it in the messages before sending

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,146 @@
...
+      //If dxfeed errors there is no information about failed feeds/params in the message, returning empty array
+      if (message[0].successful === false) {
+        logger.warn('Dxfeed returned unsuccessful message')
+        return []
+      }
+
+      if (Array.isArray(message) && message[0].channel === SERVICE_DATA) {
+        const base = message[0].data[1][0]
+        const price = message[0].data[1][6]
+        console.log('returned to be saved in cache', base, price)
```
boxhock, line 109:
```suggestion
```

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,146 @@
...
+    message(message) {
+      //If dxfeed errors there is no information about failed feeds/params in the message, returning empty array
+      if (message[0].successful === false) {
+        logger.warn('Dxfeed returned unsuccessful message')
+        return []
+      }
+
+      if (Array.isArray(message) && message[0].channel === SERVICE_DATA) {
+        const base = message[0].data[1][0]
+        const price = message[0].data[1][6]
```
boxhock, line 108: Can we store these indexes (`0` and `6`) as consts to make it a bit easier to track?

packages/sources/dxfeed-test/src/endpoint/price-router.ts
```diff
@@ -0,0 +1,94 @@
...
+      return 'WS'
+    }
+    return 'REST'
+  },
+)
+
+export const endpoint = new AdapterEndpoint<EndpointTypes>({
+  name: 'price-ws',
```
boxhock, line 90:
```suggestion
  name: 'price',
```

boxhock (changes requested):

packages/sources/dxfeed-test/src/config/index.ts
karen-stepanyan: I'll remove the default but not make it required since it is optional for DP

packages/sources/dxfeed-test/src/endpoint/price-router.ts
```diff
@@ -0,0 +1,93 @@
...
+import { batchTransport } from './price'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { wsTransport } from './price-ws'
+import { customSettings } from '../config'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin', 'market'],
+    type: 'string',
+    description: 'The symbol of symbols of the currency to query',
```
boxhock, line 12: I know this is a typo on the v3 framework types, but we're not extending it here, so we can fix it

```suggestion
    description: 'The symbol of the currency to query',
```

packages/sources/dxfeed-test/src/endpoint/price-router.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    WS: wsTransport,
+    REST: batchTransport,
+  },
+  (_, adapterConfig) => {
+    if (adapterConfig && adapterConfig.WS_ENABLED) {
```
boxhock, line 81: I want to acknowledge that `adapterConfig` is possibly undefined here, but I don't understand why...

Made a PR for it: https://github.com/smartcontractkit/ea-framework-js/pull/20

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,157 @@
...
+  id = 1
+
+  get connectionClientId() {
+    return this._connectionClientId
+  }
+
+  set connectionClientId(id) {
+    this._connectionClientId = id
+  }
+  get handshakeMsg() {
```
boxhock, line 51: Would be great with some spacing here and between the other functions

```suggestion
  }
  
  get handshakeMsg() {
```

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,157 @@
...
+              result: price,
+              data: {
+                result: price,
+              },
+            },
+          },
+        ]
+      } else {
+        return []
+      }
```
boxhock, line 134: Can we flip this and make the left path the happy path?

```suggestion
      if (!Array.isArray(message) || message[0].channel !== SERVICE_DATA) {
        return []
      ...
```

packages/sources/dxfeed-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,65 @@
+import { BatchWarmingTransport } from '@chainlink/external-adapter-framework/transports/batch-warming'
+import { makeLogger } from '@chainlink/external-adapter-framework/util'
+import { EndpointTypes } from './price-router'
+
+const logger = makeLogger('DxFeed Price Batched')
+
+const quoteEventSymbols: { [key: string]: boolean } = {
+  'USO/USD:AFX': true,
+}
```
boxhock, line 9: While we're at it, should we split this out into a JSON file along with overrides?

boxhock (changes requested):

packages/sources/dxfeed-test/src/endpoint/price-ws.ts
```diff
@@ -0,0 +1,157 @@
...
+            connection.send(JSON.stringify(wsTransport.heartbeatMsg))
+            resolve()
+          }
+        })
+        connection.send(JSON.stringify(wsTransport.handshakeMsg))
+      })
+    },
+    message(message) {
+      //If dxfeed errors there is no information about failed feeds/params in the message, returning empty array
+      if (message[0].successful === false) {
```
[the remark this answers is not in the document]
karen-stepanyan: i wouldn't do this. Dxfeed sometimes sends '/service/state' info messages that don't contain `successful` property.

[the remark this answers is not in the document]
boxhock: Hmm I see. I wonder if we need to change the type then to be explicitly `boolean | undefined`? I am not sure about this, but WebStorm treats that differently from how it is now (`successful?: boolean`)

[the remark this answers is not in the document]
boxhock: Maybe it is fine as it is, but WebStorm complains about it. Might be useful to add a comment explaining why we need to explicitly do a strict check against `false`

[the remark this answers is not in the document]
karen-stepanyan: strange, i also use Webstorm but have no issues with this. but you're right, i'll add a comment why we need ===

packages/sources/dxfeed-test/src/endpoint/price.ts
karen-stepanyan: do you mean to have it in separate JSON file and import it ?

boxhock: Correct

boxhock (dismissed):

packages/sources/dxfeed-test/src/endpoint/price-router.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    WS: wsTransport,
+    REST: batchTransport,
+  },
+  (_, adapterConfig) => (adapterConfig.WS_ENABLED ? 'WS' : 'REST'),
+)
+
+export const endpoint = new AdapterEndpoint<EndpointTypes>({
```
amit-momin, line 83: I'm solely basing this off the name of the endpoint but shouldn't this be a PriceEndpoint?
```suggestion
export const endpoint = new AdapterEndpoint<EndpointTypes>({
```
```suggestion
export const endpoint = new PriceEndpoint<EndpointTypes>({
```

karen-stepanyan: Actually this price endpoint accepts only `base` , so technically this is not a `PriceEndpoint`. Same way this can't be `PriceAdapter`

amit-momin: Ah I see. Never mind then

amit-momin (approved):

amit-momin (approved):

PR #2429, opened 2022-12-29
add coinapi-test v3 ea

karen-stepanyan:
[PDI-79](https://smartcontract-it.atlassian.net/browse/PDI-79)

[PDI-79]: https://smartcontract-it.atlassian.net/browse/PDI-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/coinapi-test/src/endpoint/crypto-ws.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+
+  builders: {
+    subscribeMessage: (params) => {
+      return {
+        type: 'hello',
+        apikey: apikey,
+        heartbeat: false,
+        subscribe_data_type: ['exrate'],
+        subscribe_filter_asset_id: [`${params.base}/${params.quote}`],
```
amit-momin, line 57: Looking at CoinAPI v2, I think it passes base and quote separately in the `subscribe_filter_asset_id` array. Was it changed to send a specific pair like this for a reason? Based on CoinAPI docs, it seems like this is valid too but unsure if there's any behavior difference

karen-stepanyan: Yes there is a behavior difference. In v2 when we pass two separate arguments coinapi treats them as two 'bases' and returns every possible quote information. For example for ['BTC' , 'ETH'] coinapi will return every price info for BTC/USD, BTC/DAO, BTC/SOL and so on, same for ETH. 
Whereas if we send ['BTC/ETH'] coinapi will return only prices for this particular pair. 
Not sure which one is better for us.

amit-momin: Ok we can keep it how you have it now. Probably better we don't have process extra pairs that would be unused. Don't think we have too many pairs configured for the same base where this would help. And it would only help in response time for the initial request anyways.

amit-momin (approved):

PR #2430, opened 2022-12-29
add tiingo-test v3 ea

karen-stepanyan:
[PDI-74](https://smartcontract-it.atlassian.net/browse/PDI-74)

[PDI-74]: https://smartcontract-it.atlassian.net/browse/PDI-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/tiingo-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,111 @@
...
+      params: chunkedParams,
+      request: {
+        baseURL: config.API_ENDPOINT,
+        url,
+        params: {
+          token: config.API_KEY,
+          tickers: [...new Set(chunkedParams.map((p) => `${p.base}${p.quote}`.toLowerCase()))].join(
+            ',',
+          ),
+        },
```
amit-momin, line 77: Based on Tiingo v2, I think we're missing the `resampleFreq: '24hour'` param for the crypto/price endpoint

packages/sources/tiingo-test/src/endpoint/common/forex-router.ts
```diff
@@ -0,0 +1,44 @@
...
+    HTTP: httpTransport,
+  },
+  (_, adapterConfig) => (adapterConfig?.WS_ENABLED ? 'WS' : 'HTTP'),
+)
+
+export const endpoint = new PriceEndpoint<ForexEndpointTypes>({
+  name: 'forex',
+  aliases: ['fx', 'commodities'],
+  transport: routingTransport,
+  inputParameters: inputParameters,
```
amit-momin, line 43: I think the inputParameters for forex have different aliases from crypto based on v2. `base` would also need the `asset` alias

karen-stepanyan: good catch

packages/sources/tiingo-test/src/endpoint/ws/crypto.ts
```diff
@@ -0,0 +1,69 @@
...
+    WsMessage: Message
+  }
+}
+
+let apiKey = ''
+export const wsTransport = new WebSocketTransport<EndpointTypes>({
+  url: (context) => {
+    apiKey = context.adapterConfig.API_KEY
+    return `${context.adapterConfig.WS_API_ENDPOINT}/crypto-synth`
+  },
```
boxhock, line 31: Instead of setting global variables, should we instead create a class that extends WebSocketTransport which can hold the API key? @alejoberardino opinions?

alejoberardino: I think a subclass would be cleaner, yeah. Another option is to expose the config to the builders, but maybe that's too much of a hassle

karen-stepanyan: Done

packages/sources/tiingo-test/src/ws-utils.ts
```diff
@@ -0,0 +1,25 @@
+import { TransportGenerics } from '@chainlink/external-adapter-framework/transports'
+import {
+  WebsocketReverseMappingTransport,
+  WebSocketTransport,
+} from '@chainlink/external-adapter-framework/transports/websocket'
+
+// Once WebsocketTransportGenerics type is exported from EA framework this can be removed
+type WebsocketTransportGenerics = TransportGenerics & {
```
amit-momin, line 8: I think we can get a quick PR up for this so we don't have to come up to tweak just this later. We can maybe bundle it with https://github.com/smartcontractkit/ea-framework-js/pull/52

boxhock: I've updated my PR

amit-momin (approved):

PR #2431, opened 2022-12-30
 add tradermade-test v3 ea

karen-stepanyan:
[PDI-194](https://smartcontract-it.atlassian.net/browse/PDI-194)

[PDI-194]: https://smartcontract-it.atlassian.net/browse/PDI-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/tradermade-test/src/config/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  API_ENDPOINT: {
+    description: 'API endpoint for tradermade',
+    default: 'https://marketdata.tradermade.com/api/v1/live',
+    type: 'string',
+  },
+  API_KEY: {
+    description:
+      'An API key that can be obtained from [here](https://marketdata.tradermade.com/docs/restful-api)',
+    type: 'string',
+  },
```
amit-momin, line 11: Based on v2, I think this should be marked as `required: true`. We should also mark it as sensitive.

packages/sources/tradermade-test/src/config/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  WS_ENABLED: {
+    description: 'Whether data should be returned from websocket or not',
+    type: 'boolean',
+    default: false,
+  },
+  WS_API_KEY: {
+    description:
+      'An API key that can be obtained from [here](https://marketdata.tradermade.com/docs/restful-api)',
+    type: 'string',
+  },
```
amit-momin, line 21: We should mark this as sensitive too

packages/sources/tradermade-test/src/endpoint/live.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const inputParameters = {
+  base: {
+    aliases: ['from', 'coin', 'symbol', 'market'],
+    required: true,
+    type: 'string',
+    description: 'The symbol of symbols of the currency to query',
+  },
+  quote: {
+    aliases: ['to', 'market'],
```
amit-momin, line 13: Based on v2 I think we'll need `convert` here too

packages/sources/tradermade-test/src/endpoint/forex-ws.ts
```diff
@@ -0,0 +1,61 @@
...
+    handlers: {
+      message(message) {
+        const pair = wsTransport.getReverseMapping(message.symbol.toLowerCase())
+
+        if (!pair) {
+          return []
+        }
+        return [
+          {
+            params: { base: pair.base, quote: pair.quote },
```
amit-momin, line 40: This is me just nitpicking but I think we can just set params to pair directly like 
```suggestion
            params: pair,
```

packages/sources/tradermade-test/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      enterprise: {
+        rateLimit1h: 833.33,
+      },
+      'enterprise-xl': {
+        rateLimit1h: 1736.11,
+      },
+    },
+  },
+  bootstrap: async (adapter) => {
+    if (!adapter.config.WS_ENABLED && !adapter.config.API_KEY) {
```
amit-momin, line 38: Since `API_KEY` is always required in v2, we can remove this case when we add the required flag in the custom configs

packages/sources/tradermade-test/src/price-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+  return res.quotes.map((entry) => {
+    if (entry.error) {
+      const pair = params.find(
+        (param) => `${param.base}${param.quote ?? ''}`.toUpperCase() === entry.instrument,
+      ) as BatchRequestParams
+      return {
+        params: pair,
+        response: {
+          errorMessage: `No data for base - ${pair.base}, quote - ${pair.quote} `,
+          statusCode: 400,
```
amit-momin, line 86: Unsure if it's technically the correct code but `coingecko-test` sets 502 when data is missing in the response. We should maybe do the same for consistency

packages/sources/tradermade-test/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["src/**/*", "src/**/*.json"],
+  "exclude": ["dist", "**/*.spec.ts", "**/*.test.ts"],
+  "references": [{ "path": "../../core/test-helpers" }, { "path": "../../core/bootstrap" }]
```
amit-momin, line 9: I think we can remove references to core here

packages/sources/tradermade-test/src/endpoint/live.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const inputParameters = {
+  base: {
+    aliases: ['from', 'coin', 'symbol'],
+    required: true,
+    type: 'string',
+    description: 'The symbol of symbols of the currency to query',
+  },
+  quote: {
+    aliases: ['to', 'market', 'convert'],
```
amit-momin, line 13: I think `market` should be an alias for `base` instead of `quote` based on the v2 adapter.

karen-stepanyan: Right, i also missed that point

packages/sources/tradermade-test/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.23.2",
```
amit-momin, line 31: Sorry to keep tacking things on but we got a new version of the fw out if you could bump this when you get the chance

karen-stepanyan: Done, Also tested this EA manually

packages/sources/tradermade-test/src/endpoint/forex-router.ts
```diff
@@ -0,0 +1,21 @@
+import { RoutingTransport } from '@chainlink/external-adapter-framework/transports/meta'
+import { PriceEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { BatchEndpointTypes } from '../price-utils'
+import { httpTransport, inputParameters } from './forex'
+import { wsTransport } from './forex-ws'
+
+const transports = {
+  websocket: wsTransport,
+  batch: httpTransport,
+}
```
amit-momin, line 10: Could you update the names to `ws` and `rest`, respectively? I believe that's what EngOps will be configuring the feeds to.

packages/sources/tradermade-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,27 @@
...
+    aliases: ['to', 'market', 'convert'],
+    required: true,
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+  },
+  transport: {
+    description: 'which transport to route to',
+    required: false,
+    type: 'string',
+  },
```
amit-momin, line 21: We can also add `ws` and `rest` as options for this input param to enforce the correct transport names

karen-stepanyan: Currently typescript isn't allowing to use options (throws type errors). I think we need to change [this](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/validation/input-params.ts#L11) to be readonly as other array values

amit-momin: I think there's also some specific typing on the transport input param for routing transports. We can skip on this then

packages/sources/tradermade-test/src/endpoint/forex-router.ts
```diff
@@ -0,0 +1,21 @@
...
+export const routingTransport = new RoutingTransport<BatchEndpointTypes>(transports, {
+  defaultTransport: 'rest',
+})
+
+export const endpoint = new PriceEndpoint<BatchEndpointTypes>({
+  name: 'forex',
+  aliases: ['batch'],
+  transport: routingTransport,
+  inputParameters: inputParameters,
+})
```
amit-momin, line 21: Shoot sorry I keep realizing things when I'm doing my "final" pass. Since the bootstrap validation is removed, I think we need to add a new one to make sure `ws` requests are rejected if `WS_API_KEY` isn't set. Since the configuration is at the request level now, we can use a `customInputValidation` for this endpoint instead. I think `ncfx` should have an example on its forex endpoint

amit-momin (approved):

PR #2434, opened 2023-01-03
Add Blocksize Capital v3 EA

bogos:


packages/sources/blocksize-capital-test/src/config/envDefaultOverrides.ts
```diff
@@ -0,0 +1,3 @@
+export const envDefaultOverrides = {
```
karen-stepanyan, line 1: this file should be removed

packages/sources/blocksize-capital-test/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { customSettings } from './config'
+//import { envDefaultOverrides } from './config/envDefaultOverrides'
+import { price } from './endpoint'
+
+export const adapter = new PriceAdapter({
+  name: 'BLOCKSIZECAPITAL',
+  endpoints: [price],
+  defaultEndpoint: price.name,
+  customSettings,
+  //envDefaultOverrides,
```
karen-stepanyan, line 12: should be removed

packages/sources/blocksize-capital-test/src/config/index.ts
```diff
@@ -0,0 +1,35 @@
+export const DEFAULT_BASE_WS_URL = 'wss://data.blocksize.capital/marketdata/v1/ws'
+export const defaultEndpoint = 'price'
+
+export const customSettings = {
+  API_ENDPOINT: {
```
karen-stepanyan, line 5: Blocksize capital doesn't have REST endpoint

packages/sources/blocksize-capital-test/src/config/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    description: 'The default REST API base url',
+    type: 'string',
+    required: false,
+  },
+  API_KEY: {
+    description: 'The Blocksize Capital API key',
+    type: 'string',
+    required: false,
```
karen-stepanyan, line 13: should be required

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,119 @@
...
+    Params: { base: string; quote: string }
+  }
+  Response: SingleNumberResultResponse
+  CustomSettings: typeof customSettings
+  Provider: {
+    WsMessage: Message[]
+  }
+}
+
+let api_key: string | undefined
```
karen-stepanyan, line 54: since we are making API_KEY required it will not be `undefined` anymore

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,119 @@
...
+    },
+    message: (message) => {
+      if (Object.keys(message).length === 0) {
+        logger.debug('WS message is empty, skipping')
+        return []
+      }
+      const [_, msg] = message
+      if (!(msg.method === 'vwap' || 'method' in msg)) return []
+      const [updates] = msg.params.updates
+      const base = updates.ticker.substring(0, 3)
```
karen-stepanyan, line 78: this will fail for tickers that have more symbols (DOGE, MATIC). This part needs to be refactored once ea framework version is updated.

karen-stepanyan (dismissed):
Some comments but still this is blocked until ea framework is updated

packages/sources/blocksize-capital-test/src/config/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    sensitive: true,
+  },
+  WS_API_ENDPOINT: {
+    description: 'The default WebSocket API base url',
+    type: 'string',
+    required: false,
+    default: DEFAULT_BASE_WS_URL,
+  },
+  WS_ENABLED: {
```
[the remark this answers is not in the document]
amit-momin: I agree. We would only need this env var if we needed a routing transport. But since this adapter only has WS, we can remove the `WS_ENABLED` config.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+    updates: {
+      ticker: string
+      price?: number
+      size?: number
+      volume?: number
+    }[]
+  }
+}
+
+const inputParameters: InputParameters & PriceEndpointInputParameters = {
```
amit-momin, line 29: I think you can use just `PriceEndpointInputParameters` here. Unless it starts giving you error then it's alright.
```suggestion
const inputParameters: PriceEndpointInputParameters = {
```

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+    aliases: ['to', 'market'],
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+    required: false,
+  },
+}
+
+export type EndpointTypes = {
+  Request: {
+    Params: { base: string; quote: string }
```
amit-momin, line 46: We can use the params we export from the framework for price endpoints here
```suggestion
    Params: PriceEndpointParams
```

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+  Response: SingleNumberResultResponse
+  CustomSettings: typeof customSettings
+  Provider: {
+    WsMessage: Message[]
+  }
+}
+
+type WebsocketTransportGenerics = TransportGenerics & {
```
amit-momin, line 55: If you bump the framework version to `0.22.0`, you can use the `WebsocketTransportGenerics` exported from the framework so you can delete this type.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+          params: { api_key: websocketTransport.api_key },
+        })
+      },
+      message: (message) => {
+        if (Object.keys(message).length === 0) {
+          logger.debug('WS message is empty, skipping')
+          return []
+        }
+        const [_, msg] = message
+        if (!(msg.method === 'vwap' || 'method' in msg)) return []
```
amit-momin, line 88: I think this validation only covers if `msg` doesn't have `method` but it'd still pass even if method wasn't vwap. If we want to make sure method is `vwap` we would need something like 
```suggestion
        if (!('method' in msg) || msg.method !== 'vwap') return []
```

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+        const [updates] = msg.params.updates
+        const params = websocketTransport.getReverseMapping(updates.ticker)
+        if (!params) {
+          return undefined
```
amit-momin, line 92: I think we should return an empty array here like we do on the other validation failures above. Unless there was a reason to return `undefined`
```suggestion
          return []
```

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+        }
+        const base = params.base
+        const quote = params.quote
+        return [
+          {
+            params: { base, quote },
+            response: {
+              result: updates.price as number,
```
amit-momin, line 100: Instead coercing price here I think we should have one more validation before we return results to check if price is `undefined` in the message. If it is we should return an error response saying something like the data provider didn't provide results.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+                providerIndicatedTimeUnixMs: new Date(Date.now()).getTime(),
+              },
+            },
+          },
+        ]
+      },
+    },
+    builders: {
+      subscribeMessage: (params) => {
+        const pair = `${params.base}${params.quote}`
```
amit-momin, line 114: We should uppercase base and quote here and in unsubscribe to stay in line with v2.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+          jsonrpc: '2.0',
+          method: 'vwap_unsubscribe',
+          params: { tickers: [pair] },
+        }
+      },
+    },
+  })
+
+export const endpoint = new PriceEndpoint<EndpointTypes>({
+  name: 'price',
```
amit-momin, line 135: Adding this comment so we don't forget but we should update this to the `CryptoPriceEndpoint` once that's released with the framework. Otherwise we'd need to add an alias for `crypto` here.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+        const [updates] = msg.params.updates
+        const params = websocketTransport.getReverseMapping(updates.ticker)
+        if (!params) {
+          return undefined
+        }
+        const base = params.base
+        const quote = params.quote
+        return [
+          {
+            params: { base, quote },
```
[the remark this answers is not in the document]
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,138 @@
...
+    type: 'string',
+    description: 'The symbol of symbols of the currency to query',
+    required: false,
+  },
+  quote: {
+    aliases: ['to', 'market'],
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+    required: false,
+  },
```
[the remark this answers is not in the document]
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/config/index.ts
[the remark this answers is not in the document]
amit-momin: I think removing `WS_ENABLED` was missed. Doesn't seem like we're using it anywhere so it shouldn't be needed.

packages/sources/blocksize-capital-test/src/endpoint/price.ts
```diff
@@ -0,0 +1,145 @@
...
+      message: (message) => {
+        if (!('method' in message) || message.method !== 'vwap') return []
+        const [updates] = message.params.updates
+        const params = websocketTransport.getReverseMapping(updates.ticker)
+        if (!params) {
+          return []
+        }
+        if (!updates.price) {
+          logger.error(`The data provider didn't return any value`)
+          return []
```
amit-momin, line 101: We can return an error here  so subsequent calls would just get the error from the cache. I was thinking something like what we do in `coingecko-test` [here](https://github.com/smartcontractkit/external-adapters-js/blob/59c4a2ee41116db7317915dda0dfe6063051ce70/packages/sources/coingecko-test/src/crypto-utils.ts#L88-L101)

packages/sources/blocksize-capital-test/src/config/index.ts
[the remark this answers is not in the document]
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/src/endpoint/price.ts
bogos: Done

packages/sources/blocksize-capital-test/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["src/**/*", "src/**/*.json"],
+  "exclude": ["dist", "**/*.spec.ts", "**/*.test.ts"],
+  "references": [{ "path": "../../core/test-helpers" }, { "path": "../../core/bootstrap" }]
```
amit-momin, line 9: Sorry one last thing. We don't need references to core anymore.

amit-momin (approved):

PR #2444, opened 2023-01-11
add finage-test v3 ea

karen-stepanyan:
[PDI-192](https://smartcontract-it.atlassian.net/browse/PDI-192)

[PDI-192]: https://smartcontract-it.atlassian.net/browse/PDI-192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/finage-test/schemas/env.json
```diff
@@ -0,0 +1,18 @@
+{
```
amit-momin, line 1: I think we can delete this file and the schemas folder

packages/sources/finage-test/src/config/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  },
+  API_KEY: {
+    type: 'string',
+    required: true,
+    sensitive: true,
+    description: "An API key that can be obtained from the data provider's dashboard",
+  },
+  WS_SOCKET_KEY: {
+    type: 'string',
+    required: false,
```
amit-momin, line 15: Based on v2 I think `WS_SOCKET_KEY` should be required.
```suggestion
    required: true,
```

packages/sources/finage-test/src/endpoint/http/eod.ts
```diff
@@ -0,0 +1,95 @@
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { customSettings } from '../../config'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin'],
```
amit-momin, line 8: I think in v2 the aliases for eod should be
```suggestion
    aliases: ['from', 'symbol'],
```

packages/sources/finage-test/src/endpoint/http/eod.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export const httpTransport = new HttpTransport<EndpointTypes>({
+  prepareRequests: (params, config) => {
+    return params.map((param) => {
+      return {
+        params: [param],
+        request: {
+          baseURL: config.API_ENDPOINT,
+          url: `/agg/stock/prev-close/${param.base}`,
```
amit-momin, line 54: I think we can batch on base for this endpoint. Looks like we just send a comma-separated list of bases in v2.

amit-momin: Batching may change the ResponseSchema

packages/sources/finage-test/src/endpoint/http/commodities.ts
```diff
@@ -0,0 +1,75 @@
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { PriceEndpoint, PriceEndpointParams } from '@chainlink/external-adapter-framework/adapter'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { customSettings } from '../../config'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin'],
```
amit-momin, line 8: The aliases in v2 are just `from` and `symbol`. But if coin needs to be in there for the `PriceEndpointParams` types is it possible to just add `symbol` or would that give you an error?

packages/sources/finage-test/src/endpoint/forex-router.ts
```diff
@@ -0,0 +1,54 @@
+import { customSettings } from '../config'
+import { httpTransport } from './http/forex'
+import { PriceEndpoint, PriceEndpointParams } from '@chainlink/external-adapter-framework/adapter'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { RoutingTransport } from '@chainlink/external-adapter-framework/transports/meta'
+import { wsTransport } from './ws/forex-ws'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin'],
```
amit-momin, line 10: Same here for the `symbol` alias

packages/sources/finage-test/src/endpoint/crypto-router.ts
```diff
@@ -0,0 +1,54 @@
+import { customSettings } from '../config'
+import { httpTransport } from './http/crypto'
+import { PriceEndpoint, PriceEndpointParams } from '@chainlink/external-adapter-framework/adapter'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { RoutingTransport } from '@chainlink/external-adapter-framework/transports/meta'
+import { wsTransport } from './ws/crypto-ws'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin'],
```
amit-momin, line 10: Also here for `symbol` alias

packages/sources/finage-test/src/endpoint/http/stock.ts
```diff
@@ -0,0 +1,37 @@
...
+  parseResponse: (_, res) => {
+    // Filter null values in response
+    const response = res.data.filter((e) => e)
+
+    return response.map((entry) => {
+      return {
+        params: { base: entry.symbol },
+        response: {
+          data: {
+            result: entry.bid,
```
amit-momin, line 27: Not entirely sure why v2 behaves this way but I think when using the batch endpoint the result is supposed to be (ask + bid) / 2. There is a TODO on the batch handler to "check this" so maybe this is something I'll have to check with Data Feeds on.

packages/sources/finage-test/src/endpoint/stock-router.ts
```diff
@@ -0,0 +1,54 @@
+import { customSettings } from '../config'
+import { httpTransport } from './http/stock'
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { RoutingTransport } from '@chainlink/external-adapter-framework/transports/meta'
+import { wsTransport } from './ws/stock-ws'
+
+export const inputParameters = {
+  base: {
+    aliases: ['from', 'coin'],
```
amit-momin, line 10: `symbol` alias here too

packages/sources/finage-test/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  defaultEndpoint: stock.name,
+  name: 'FINAGE',
+  customSettings,
+  overrides: overrides.finage,
+  endpoints: [crypto, stock, eod, commodities, forex],
+  bootstrap: async (adapter: Adapter<typeof customSettings>) => {
+    if (adapter.config.WS_ENABLED && !adapter.config.WS_SOCKET_KEY) {
+      throw 'WS_SOCKET_KEY is required when WS_ENABLED is set to true'
+    }
+  },
```
amit-momin, line 17: Since `WS_SOCKET_KEY` is required in v2, we probably don't need this check anymore. But this was probably the better way to handle this so NOPs don't need a WS key to run just a REST version.

packages/sources/finage-test/src/endpoint/http/eod.ts
karen-stepanyan: Actually I don't think that batching for this endpoint is working in v2. I tried to send a request with `base: ["AAPL", "MSFT"]` but got error response. Also checking the api it seems they don't have such option, maybe it was removed. 
However finage has new api endpoint for getting all stock data for a given date. We probably can use this to get stock data for previous day.

packages/sources/finage-test/src/endpoint/http/stock.ts
boxhock: +1 on this. I have sent a message to Data Feeds for confirmation

boxhock: Confirmed we should do (ask + bid) / 2

packages/sources/finage-test/src/endpoint/http/eod.ts
amit-momin: Good catch! I took a look at the api that returns all stock data for the previous day and I think it's too heavy to use here. Looks like it returns 15k+ results. I think we're good to stick to what you have.

amit-momin (approved):

PR #2453, opened 2023-01-16
add coinmarketcap-test v3 ea

karen-stepanyan:
[PDI-76](https://smartcontract-it.atlassian.net/browse/PDI-76)

[PDI-76]: https://smartcontract-it.atlassian.net/browse/PDI-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+  Response: SingleNumberResultResponse
+  CustomSettings: typeof customSettings
+  Provider: {
+    RequestBody: ProviderRequestBody
+    ResponseBody: ProviderResponseBody
+  }
+}
+
+const groupByBaseOptions = (params: CryptoRequestParams[]) => {
```
amit-momin, line 98: I think theres some missing logic from v2 that we can add here. We maintain a list of cid that we use to convert symbols if it's in the map. https://github.com/smartcontractkit/external-adapters-js/blob/d0adc1082a233b695b58bf23d752506d85ca1fd4/packages/sources/coinmarketcap/src/endpoint/crypto.ts#L66-L102

amit-momin: Also when you add this map to v3, you may want to update OHM to `OHM: 16209`. I think there's a PR open to update this in v2

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,188 @@
...
+    const groupedParams = group[1]
+
+    return {
+      params: groupedParams,
+      request: {
+        baseURL: endpoint,
+        url: '/cryptocurrency/quotes/latest',
+        params: {
+          [queryName]: [...new Set(groupedParams.map((p) => p.cid || p.slug || p.base))].join(','),
+          convert: [...new Set(groupedParams.map((p) => p.quote))].join(','),
```
amit-momin, line 131: Based on the documentation, I think CMC limits convert to 120 symbols at a time if I'm reading the docs right. We had a limit for this in v2 also. https://github.com/smartcontractkit/external-adapters-js/blob/6e9f97e33eeaa1678d99ad0de1f5215ff45ee6fe/packages/sources/coinmarketcap/src/endpoint/crypto.ts#L14

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  return groupedParams.map((group) => {
+    const queryName = group[0]
+    const groupedParams = group[1]
+
+    return {
+      params: groupedParams,
+      request: {
+        baseURL: endpoint,
+        url: '/cryptocurrency/quotes/latest',
```
amit-momin, line 128: I think we're missing the header here right?
```suggestion
        url: '/cryptocurrency/quotes/latest',
        headers: {
          'X-CMC_PRO_API_KEY': config.API_KEY,
        },
```

packages/sources/coinmarketcap-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.19.0",
```
boxhock, line 31: I know it's frequently changing these days... but just a reminder to use the latest version of the framework :sweat_smile:

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,261 @@
...
+  RCN: 2096,
+  REN: 2539,
+  SNX: 2586,
+  STETH: 8085,
+  SUSHI: 6758,
+  UMA: 5617,
+  UNI: 7083,
+  WOM: 5328,
+  YFI: 5864,
+}
```
boxhock, line 129: Can we move this list to a JSON file instead? Relevant for future improvements.

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,261 @@
...
+const chunkArray = (params: string[], size = 120): string[][] =>
+  params.length > size ? [params.slice(0, size), ...chunkArray(params.slice(size), size)] : [params]
+
+const groupByBaseOptions = (params: CryptoRequestParams[]) => {
+  const id: Payload[] = []
+  const slug: Payload[] = []
+  const symbol: Payload[] = []
+  const group: { id?: Payload[]; slug?: Payload[]; symbol?: Payload[] } = {}
+  params.forEach((param) => {
+    let optionContainer = []
```
boxhock, line 142: I might be blind here, but are we re-initializing this array on each loop? What are we doing with it?

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    if (input.id || input.payload.cid) {
+      optionContainer = id
+    } else if (input.payload.slug) {
+      optionContainer = slug
+    } else if (input.payload.base) {
+      optionContainer = symbol
+    }
+    optionContainer.push(input)
```
boxhock, line 157: Could we just `.push()` into the respective array directly here? This seems a bit over-complicated

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,261 @@
...
+    } catch (_) {
+      return {
+        params: param,
+        response: {
+          statusCode: 400,
+          errorMessage:
+            'Could not retrieve valid data from Data Provider. This is likely an issue with the Data Provider or the input params/overrides',
+        },
+      }
+    }
```
boxhock, line 250: I would prefer if we checked for each of the keys, that way we can return more descriptive error messages about what's wrong

boxhock (changes requested):
Super excited about this one and the fixed batch requests! :tada:

packages/sources/coinmarketcap-test/src/crypto-utils.ts
karen-stepanyan: I don't think it is needed. I removed it already !

packages/sources/coinmarketcap-test/src/crypto-utils.ts
```diff
@@ -0,0 +1,258 @@
...
+
+    if (!response.data[key]) {
+      return errorResponse(param, `Data for "${param.cid || param.slug || param.base}" not found`)
+    }
+
+    const dataForQuote = response.data[key].quote[quote]
+    if (!dataForQuote) {
+      return errorResponse(
+        param,
+        `"${dataForQuote}" quote for "${param.cid || param.slug || param.base}" not found`,
```
amit-momin, line 234: Wouldn't `dataForQuote` be undefined here? Should this be just `quote`?

packages/sources/coinmarketcap-test/package.json
amit-momin: Looks like we had another version since. Could you bump it once more? Hopefully that's the last fw deployment before we get this merged.

packages/sources/coinmarketcap-test/src/crypto-utils.ts
karen-stepanyan: Yep, correct

amit-momin (approved):

PR #2460, opened 2023-01-18
refactor cryptocompare-test

karen-stepanyan:
[PDI-1987](https://smartcontract-it.atlassian.net/browse/PDI-1987)

[PDI-1987]: https://smartcontract-it.atlassian.net/browse/PDI-1987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/cryptocompare-test/src/endpoint/crypto.ts
```diff
@@ -1,52 +1,57 @@
...
+
+type BatchEndpointTypes = {
+  Request: {
+    Params: CryptoEndpointParams
+  }
+  Response: SingleNumberResultResponse
+  CustomSettings: typeof customSettings
   Provider: {
-    RequestBody: never
+    RequestBody: ProviderRequestBody
```
boxhock, line 30: Is there actually a POST request body, or is it just query params?

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -61,17 +61,34 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
...
+        const base = parameters[2]
+        const quote = parameters[3]
         logger.error(message, 'asset not supported by data provider')
-        return
+        return [
+          {
+            params: { base, quote, endpoint: 'crypto' },
+            response: {
+              errorMessage:
+                'Could not retrieve valid data from Data Provider. This is likely an issue with the Data Provider or the input params/overrides',
```
boxhock, line 85: Can we accurately have a more descriptive error message, such as the error we're logging?

packages/sources/cryptocompare-test/src/endpoint/crypto.ts
```diff
@@ -58,20 +63,31 @@ const endpointResultPaths: { [endpoint: string]: KeyOfType<ProviderCryptoQuoteDa
   volume: 'VOLUME24HOURTO',
 }
 
+const errorResponse = (payload: CryptoEndpointParams) => {
+  return {
+    params: payload,
+    response: {
+      statusCode: 400,
+      errorMessage:
+        'Could not retrieve valid data from Data Provider. This is likely an issue with the Data Provider or the input params/overrides',
```
boxhock, line 72: Can we get more descriptive error messages here, like we do in the logs?

boxhock (changes requested):

packages/sources/cryptocompare-test/src/endpoint/crypto.ts
karen-stepanyan: I think you are right. After checking this should be for POST request body, tho til now we don't send POST requests to DP. Not sure if we need this for query params, but I put a note to change this for my active PRs

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -61,17 +61,33 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
       if (message.TYPE === '5' && 'PRICE' in message) {
         return [
           {
-            params: { base: message.FROMSYMBOL, quote: message.TOSYMBOL, endpoint: 'crypto-ws' },
-            value: message.PRICE as number,
+            params: { base: message.FROMSYMBOL, quote: message.TOSYMBOL, endpoint: 'crypto' },
```
boxhock, line 64: Assuming we're (for now) only routing based on the env var, we should be able to remove "endpoint" here and from the input params (`cryptoEndpointInputParams`).

karen-stepanyan: Actually i think it makes sense to have the endpoint explicitly in the params.  That way we don't need to have separate endpoint/transport for `marketcap` and `volume` since we are using `endpointResultPaths`.  What do you think ?

boxhock: Hmm yeah good point. I think we can just proceed with how it is now then!

packages/sources/cryptocompare-test/package.json
```diff
@@ -28,7 +28,7 @@
     "start": "yarn server:dist"
   },
   "dependencies": {
-    "@chainlink/external-adapter-framework": "0.12.0",
+    "@chainlink/external-adapter-framework": "0.19.0",
```
boxhock, line 31: I'm being a bit annoying, but could we please update to the latest framework version? Wanna make sure we get the updated timestamp labels. Otherwise this PR LGTM

boxhock (changes requested):

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -61,17 +61,33 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
...
+            response: {
+              result: message.PRICE as number,
```
amit-momin, line 66: It doesn't let me add the comment above but we should maybe add a case for `PRICE` not found in the message so we can return an error.

packages/sources/cryptocompare-test/src/endpoint/crypto-ws.ts
```diff
@@ -61,17 +61,33 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
...
+              statusCode: 400,
+            },
+          },
+        ]
       }
-      return
+      return []
     },
   },
   builders: {
```
amit-momin, line 93: I couldn't add the comment lower and it might not be an issue since it was already like this. But does the DP expect upper case base and quote? We may want to ensure they are in the builder below

packages/sources/cryptocompare-test/src/endpoint/crypto.ts
```diff
@@ -105,20 +169,15 @@ const batchEndpointTransport = new BatchWarmingTransport<BatchEndpointTypes>({
...
-    if (req.requestContext.data.endpoint === 'crypto-ws') {
-      return 'WS'
-    }
-    return 'HTTP'
-  },
+  (_, adapterConfig) => (adapterConfig.WS_ENABLED ? 'WS' : 'HTTP'),
 )
 
-export const endpoint = new PriceEndpoint<CryptoEndpointTypes>({
+export const endpoint = new PriceEndpoint<BatchEndpointTypes>({
```
amit-momin, line 180: Just noticed we can update this to `CryptoPriceEndpoint`

packages/sources/cryptocompare-test/package.json
```diff
@@ -28,7 +28,7 @@
     "start": "yarn server:dist"
   },
   "dependencies": {
-    "@chainlink/external-adapter-framework": "0.12.0",
+    "@chainlink/external-adapter-framework": "0.23.2",
```
amit-momin, line 31: A new version is available to bump this

karen-stepanyan: Done. Also i tested this EA manually

packages/sources/cryptocompare-test/src/endpoint/index.ts
```diff
@@ -1,2 +1,4 @@
-export { endpoint as cryptoEndpoint } from './crypto'
+export { endpoint as cryptoRouter } from './crypto-router'
 export { endpoint as vwap } from './vwap'
+export { endpoint as marketacp } from './marketcap'
```
amit-momin, line 3: Small typo here

packages/sources/cryptocompare-test/src/endpoint/crypto-router.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const routingTransport = new RoutingTransport<BatchEndpointTypes>(transports, {
+  defaultTransport: 'rest',
+})
+
+export const endpoint = new CryptoPriceEndpoint<BatchEndpointTypes>({
+  name: defaultEndpoint,
+  transport: routingTransport,
+  inputParameters: cryptoInputParams,
+})
```
amit-momin, line 21: Since `WS_API_KEY` isn't required, we should add the customInputValidation to block requests to `ws` if the key isn't set here too like I mentioned in the `tradermade` PR

amit-momin: Actually never mind. Just noticed the logic in the ws transport uses `API_KEY` as a fallback. You can disregard this.

amit-momin (approved):

PR #2463, opened 2023-01-23
Add new Two Sigma EA

martin-cll:
## Description

This adds a new external adapter for Two Sigma using EAv3.

This is a websocket API for US stocks. Upon connection, the client sends a single message containing the api key and a list of symbols (e.g. `["AAPL/USD", "AMZN/USD"]`) to the server. The server then begins streaming price messages to the client. If the client want to subscribe to a different set of symbols, it needs to do so on a new websocket connection.

The main complication of this EA is the requirement for new connections. I've chosen to use the `ReconnectingWebSocket` 3p class which is a drop-in replacement for the usual WebSocket class. It provides the ability to reconnect the underlying websocket connection every time we have an update in the set of subscribed symbols.

This will mean about a 1-1.5s delay in receiving price updates when reconnecting, but it's still a big improvement on an up to 6.5s delay (up to 5s for next loop + 1.5s connection time) if we didn't explicitly reconnect and instead waited for the framework to reconnect for us on its next iteration loop.

TODO: Soak tests && infra-k8s secrets

## Steps to Test

1. Set `WS_API_ENDPOINT` and `WS_API_KEY` env vars.
2. Run `yarn build && yarn start` to start EA locally.
3. Run `echo '{"data":{"base":"AAPL","quote":"USD"}}' | http -v POST localhost:8080` to send sample requests.


alejoberardino (changes requested):
Instead of the current approach, an alternative more in line with the framework would be to create a subclass of the `WebsocketTransport` that overrides the `streamHandler` method:
```typescript
class ReconnectingWebsocketTransport extends WebSocketTransport<WebSocketTypes>{
  override async streamHandler(
    context: EndpointContext<WebSocketTypes>,
    subscriptions: SubscriptionDeltas<WebSocketTypes['Request']['Params']>,
  ): Promise<void> {
    // Since the DP does not accept subscriptions after the initial ones have been sent,
    // we are forced to close the connection and connect again
    if (
      this.wsConnection && (subscriptions.new.length || subscriptions.stale.length)
    ) {
      await this.wsConnection.close() // The parent method will take care of reconnecting
      subscriptions.new = subscriptions.desired
      subscriptions.stale = []
    }

    super.streamHandler(context, subscriptions)
  }
}
```

martin-cll:
@alejoberardino Thanks for the suggestion! It looks much cleaner now!

packages/sources/twosigma/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.14.5",
```
alejoberardino, line 31: This version is a few iterations behind, we should bump it to the latest one

packages/sources/twosigma/src/endpoint/price.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (this.wsConnection && (subscriptions.new.length || subscriptions.stale.length)) {
+      logger.debug(
+        `closing WS connection for new subscriptions: ${JSON.stringify(subscriptions.desired)}`,
+      )
+
+      // 'close' is an async method that doesn't return a promise. Rather than
+      // polling and waiting for the connection to close, we can simply set the
+      // variable to null and rely on the parent method to create a new connection.
+      this.wsConnection.close()
+      this.wsConnection = null
```
alejoberardino, line 83: This is potentially dangerous if the connections hang indefinitely and we end up continually creating new ones, but we have metrics that should surface this situation if it were to occur

packages/sources/twosigma/src/endpoint/price.ts
```diff
@@ -0,0 +1,182 @@
...
+      return
+    }
+
+    const payload: WebSocketRequest = {
+      api_key: process.env.WS_API_KEY || 'twosigma api key not set',
+      symbols: subscribes.map(buildSymbol).sort(),
+    }
+    this.wsConnection.send(this.serializeMessage(payload))
+  }
+}
```
alejoberardino, line 117: We already set this in the previous method, once the framework version is updated this should not be necessary

alejoberardino (changes requested):

packages/sources/twosigma/src/endpoint/price.ts
martin-cll: I can also add an async check to see if the conn hasn't closed in 1s and emit a log if so.

packages/sources/twosigma/src/endpoint/price.ts
martin-cll: I'm not sure I follow why this won't be necessary. The overridden `streamHandler` does set subscriptions.new to the entire set of tickers to subscribe to, but the issue is that `sendMessages` on its own will send 1 msg per ticker which won't work. We need to only send 1 msg for all the tickers.

packages/sources/twosigma/package.json
martin-cll: Hi @alejoberardino, I attempted to bump to the latest version but I'm running into a framework issue that I don't think I can work around. Let me know if I'm missing something: https://github.com/smartcontractkit/ea-framework-js/issues/51.

packages/sources/twosigma/src/endpoint/price.ts
alejoberardino: Ah, you're right! I was thinking of a previous version (or maybe not even, I've been ooo for a while 😅), this is correct 👍

.yarn/sdks/eslint/bin/eslint.js
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env node
 
 const {existsSync} = require(`fs`);
-const {createRequire, createRequireFromPath} = require(`module`);
+const {createRequire} = require(`module`);
```
martin-cll, line 4: These `.yarn/*` file changes are generated from running `yarn dlx @yarnpkg/sdks`. They update the yarn plugins for the versions of prettier and typescript that this repo uses.

martin-cll:
This is running into the same soak test failures that other PRs see:
```
The configuration for this run is:
 {
  "action": "start",
  "adapter": "twosigma",
  "release": "2463",
  "imageTag": "pr2463",
  "imageRepository": "***.dkr.ecr.***.amazonaws.com/adapters/",
  "helmChartDir": "chainlink/cl-adapter",
  "helmSecrets": "-f ./adapter-secrets/secure-sdlc/secrets.twosigma.yaml",
  "helmValuesOverride": "",
  "name": "qa-ea-twosigma-2463"
}
start twosigma adapter as qa-ea-twosigma-2463
[helm-secrets] Decrypt: ./adapter-secrets/secure-sdlc/secrets.twosigma.yaml
Release "qa-ea-twosigma-2463" does not exist. Installing it now.
Error: timed out waiting for the condition

[helm-secrets] Removed: ./adapter-secrets/secure-sdlc/secrets.twosigma.yaml.dec
Error: plugin "secrets" exited with error
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Failed to deploy the external adapter: Release "qa-ea-twosigma-2463" does not exist. Installing it now.
".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

martin-cll:
Ready for review again @austinborn @alejoberardino !

packages/sources/twosigma/src/endpoint/price.ts
```diff
@@ -0,0 +1,194 @@
...
+      // variable to null and rely on the parent method to create a new connection.
+      const conn = this.wsConnection
+      conn.close()
+      setTimeout(() => {
+        if (conn.readyState !== WS_CLOSED_STATE) {
+          logger.warn(
+            `WS connection still in state ${conn.readyState} after ${WS_CONN_CLOSE_CHECK_MS}ms`,
+          )
+        }
+      }, WS_CONN_CLOSE_CHECK_MS)
```
alejoberardino, line 93: An alternative to this would be to add a new listener within the context of a promise, and then await that promise, something like
```suggestion
     const closed = new Promise(resolve => this.wsConnection.on('close', resolve))
     this.wsConnection.close()
     await closed
```
I wouldn't blindly trust my code but I think this would work

martin-cll: Just tested this and confirmed that it works (with a minor tweak to use `onclose = resolve` for compatibility with the mock websocket class used in integration tests). Thanks!

alejoberardino (dismissed):

martin-cll:
@alejoberardino Mind taking another quick look? I added an alias for the endpoint: https://chainlink-core.slack.com/archives/C02FM5ZP2TC/p1677019172367189?thread_ts=1676674767.521649&cid=C02FM5ZP2TC. Thanks!

packages/sources/twosigma/test/integration/README.md
```diff
@@ -0,0 +1,3 @@
+Tests that run the adapter and send an end to end request using mocked API endpoints.
+
+See `packages/source/curve/test/integration` and `packages/composite/synth-index/test/integration` for reference. The integration tests are used for the example input/outputs during README generation.
```
austinborn, line 3: You can probably delete this

packages/sources/twosigma/src/endpoint/price.ts
```diff
@@ -0,0 +1,181 @@
...
+  // We don't want the 'subscribeMessage' and 'unsubscribeMessage' hooks because we override
+  // the 'sendMessages' on the transport to send a single message for the entire set of
+  // symbols. However, this empty object is necessary for the transport to actually call our
+  // overridden 'sendMessages'.
+  builders: {},
+}
+
+export const endpoint = new PriceEndpoint({
+  name: 'price',
+  aliases: ['stock'],
```
austinborn, line 178: Can you confirm this alias with data products?

austinborn (approved):
Looks great, and thank you for adding comments to explain the code better.

packages/sources/twosigma/src/endpoint/price.ts
martin-cll: Yup, this was requested by them: https://chainlink-core.slack.com/archives/C02FM5ZP2TC/p1676674767521649

PR #2473, opened 2023-01-24
remove rate limit config from kaiko-test

karen-stepanyan:
removed rate limit config from EA. Negative rate limits values are not supported by the EA framework

### DIFF
--- a/packages/sources/kaiko-test/src/endpoint/trades.ts
+++ b/packages/sources/kaiko-test/src/endpoint/trades.ts
@@ -43,12 +43,6 @@ export interface RequestParams {
   sort: string
 }
 
-export interface RequestBody {
-  interval: string
-  sort: string
-  start_time: string
-}
-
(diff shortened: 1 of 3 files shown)


packages/sources/kaiko-test/src/endpoint/trades.ts
```diff
@@ -84,7 +84,7 @@ type EndpointTypes = {
   Response: SingleNumberResultResponse
   CustomSettings: typeof customSettings
   Provider: {
-    RequestBody: RequestBody
+    RequestBody: never
```
amit-momin, line 87: Since `RequestBody` isn't needed, should we also remove the interface above?

karen-stepanyan: oh right, i forgot about that

amit-momin (approved):

PR #2477, opened 2023-01-26
Update coingecko-test EA framework version

boxhock:
I would like to have https://github.com/smartcontractkit/ea-framework-js/pull/52 merged first


packages/sources/coingecko-test/src/endpoint/crypto.ts
```diff
@@ -1,23 +1,23 @@
...
-      constructEntry(res, requestPayload, requestPayload.quote.toLowerCase()),
+      constructEntry(res.data, requestPayload, requestPayload.quote.toLowerCase()),
     ),
 })
 
 export const endpoint = new PriceEndpoint<CryptoEndpointTypes>({
   name: 'crypto',
-  aliases: ['crypto-batched', 'batched', 'batch'],
-  transport: batchEndpointTransport,
+  aliases: ['crypto-batched', 'batched', 'batch', 'price'],
```
amit-momin, line 20: I think we can use the new `CryptoPriceEndpoint` Alejo added here. Should be available in the latest version.

boxhock: Updated!

amit-momin (approved):

PR #2479, opened 2023-01-30
Feature/pdi 123

karen-stepanyan:
[PDI-123](https://smartcontract-it.atlassian.net/browse/PDI-123)


packages/sources/coinmetrics-test/src/endpoint/price-http.ts
```diff
@@ -0,0 +1,86 @@
...
+export const httpTransport = new HttpTransport<AssetMetricsHttpTypes>({
+  prepareRequests: (params, config) => {
+    return {
+      params,
+      request: {
+        baseURL: config.API_ENDPOINT,
+        url: `timeseries/asset-metrics`,
+        params: {
+          assets: [...new Set(params.map((p) => p.base.toUpperCase()))].join(','),
+          metrics: [...new Set(params.map((p) => `ReferenceRate${p.quote.toUpperCase()}`))].join(
```
amit-momin, line 39: Wouldn't this build the metrics param with repeat quotes potentially? Like if params are ETH/USD and BTC/USD. Wouldn't metrics be `ReferenceRateUSD,ReferenceRateUSD` in that case?

packages/sources/coinmetrics-test/src/endpoint/price-http.ts
```diff
@@ -0,0 +1,86 @@
...
+            ',',
+          ),
+          frequency: '1s',
+          api_key: config.API_KEY,
+          limit_per_asset: 1,
```
amit-momin, line 44: Now that we're moving the http endpoint to batching do we need to worry about page size? According to coinmetrics docs, it defaults to 100. Unsure if we have that many unique bases coming into this adapter but to be safe should we set a high page size here? Max is 10,000

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
```diff
@@ -0,0 +1,248 @@
...
+    type: 'string',
+    required: false,
+  },
+  endTime: {
+    description:
+      'The end time for the queried period. See [Supported DateTime Formats](#supported-datetime-formats)',
+    type: 'string',
+    required: false,
+  },
+} as const
```
amit-momin, line 108: Same with this
```suggestion
} satisfies InputParameters
```

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
```diff
@@ -0,0 +1,248 @@
...
+    settings: typeof config.settings,
+  ): Promise<AdapterResponse<EndpointTypes['Response']>> {
+    let totalBurnedTKN = BigNumber.from('0')
+
+    let lastPage = false
+    const input = req.requestContext.data
+    const isBurnedEndpoint = req.requestContext.endpointName === 'burned'
+    if (isBurnedEndpoint) {
+      input.pageSize = 1
+    }
```
amit-momin, line 160: Would it be cleaner if we add pageSize as an input parameter for the `burned` endpoint and set the default and only option to 1? So we don't need this piece here?

amit-momin: Never mind, realized you need this for the break out case in the while loop too

packages/sources/coinmetrics-test/src/config/index.ts
```diff
@@ -29,22 +29,21 @@ export const config = new AdapterConfig({
...
+    aliases: ['from', 'coin'],
+  },
+  quote: {
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+    required: true,
+    aliases: ['to', 'market'],
+    options: Object.values(VALID_QUOTES),
+  },
+} as const
```
amit-momin, line 46: Same here

```suggestion
} satisfies InputParameters
```

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
```diff
@@ -0,0 +1,248 @@
...
+    const requestConfig = this.prepareRequest(req.requestContext.data, settings)
+
+    const providerDataRequestedUnixMs = Date.now()
+    while (!lastPage) {
+      const result = await this.requester.request<ResponseSchema>(
+        calculateHttpRequestKey({
+          context: {
+            adapterSettings: settings,
+            inputParameters: inputParams,
+            endpointName: ENPDOINT_NAME,
```
amit-momin, line 170: Shouldn't this be set to the endpoint name in the request?
```suggestion
            endpointName: req.requestContext.endpointName,
```

packages/sources/coinmetrics-test/src/endpoint/price-http.ts
karen-stepanyan: It shouldn't since `Set` will remove duplicates

amit-momin: Oh shoot I skimmed this too quickly. Missed the Set inline

packages/sources/coinmetrics-test/src/endpoint/price-http.ts
alejoberardino: I'll add the max allowed, and a warning if we get more

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
```diff
@@ -0,0 +1,248 @@
...
+          },
+          data: requestConfig.params,
+          transportName: this.name,
+        }),
+        requestConfig,
+      )
+
+      const { data: assetMetricsList } = result.response.data
+
+      totalBurnedTKN = totalBurnedTKN.add(calculateBurnedTKN(assetMetricsList))
```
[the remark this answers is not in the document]
alejoberardino: Well, I can find no jobs for this currently, and in this version the type will be explicitly tested in the validator. If we need to change it in the future it'll be a new feed and we can do so with notice

packages/sources/coinmetrics-test/src/endpoint/price.ts
```diff
@@ -29,5 +48,13 @@ export const endpoint = new CryptoPriceEndpoint<AssetMetricsEndpointTypes>({
   aliases: ['price-ws'],
   transportRoutes,
   defaultTransport: 'http',
-  inputParameters: priceInputParameters,
+  inputParameters,
```
karen-stepanyan, line 51: Since we are doing validation in `customInputValidation` i think we can use `priceEndpointInputParameters` from framework and get rid of input params above. Can do the commit

packages/sources/coinmetrics-test/src/endpoint/total-burned.ts
amit-momin: I think this one got missed

karen-stepanyan: updated

packages/sources/coinmetrics-test/src/endpoint/price.ts
karen-stepanyan: I changed it

amit-momin: It may be cleaner to keep the input parameters with `options: Object.values(VALID_QUOTES),` and remove the `customInputValidation`. I'll leave it up to @alejoberardino to be the tie breaker but I'm good to approve after this.

alejoberardino: There were conflicting types with the type: string and options, options should be there for enum and the price endpoint expects string

amit-momin (approved):

PR #2480, opened 2023-01-31
feat: Include bid & ask prices for NCFX

martin-cll:
## Description

[MERC-84]

Include the bid & ask prices returned by NCFX in the adapter output. Tested against live endpoint and verified that the bid & ask fields are present in the response's data object.

[MERC-84]: https://smartcontract-it.atlassian.net/browse/MERC-84?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


austinborn (approved):
Nice

packages/sources/ncfx/src/endpoint/crypto.ts
```diff
@@ -44,10 +73,13 @@ export const cryptoTransport = new WebSocketTransport<EndpointTypes>({
             logger.debug('Got logged in response, connection is ready')
             resolve()
           } else {
-            reject(new Error('Unexpected message after WS connection open'))
+            reject(
+              new Error(`Unexpected message after WS connection open: ${event.data.toString()}`),
```
austinborn, line 77: I recommend `JSON.stringify(event.data)` here instead, unless you know `data` is a primitive variable.

packages/sources/ncfx/src/endpoint/crypto.ts
```diff
@@ -93,6 +132,10 @@ export const cryptoTransport = new WebSocketTransport<EndpointTypes>({
   },
 })
 
+const isInfoMessage = (message: WsMessage): message is WsInfoMessage => {
+  return (message as WsInfoMessage).Type !== undefined
```
austinborn, line 136: Any way to make this a more specific check, like `['Type1', 'Type2'].includes(Type)`?

packages/sources/ncfx/src/endpoint/crypto.ts
martin-cll: So if I'm reading the typescript types for the websocket lib right, event.data has type `type Data = string | Buffer | ArrayBuffer | Buffer[];`. So I don't think it's safe to pass that directly to `JSON.stringify`. However, the `toString` method is guaranteed to exist on all 4 of those union types so we can use it. I'm also not 100% sure what you mean by primitive variable, but the data field should be exist (i.e. not null or undefined) according to the TS types. Also, we are already calling `event.data.toString()` just a few lines above so if there is any issue with it then it would blow up before it even gets to this line.

packages/sources/ncfx/src/endpoint/crypto.ts
martin-cll: So far the only `Type` value I've seen is `Info`, but I don't know if there could be others (e.g. other log levels or other values entirely) and there's no API docs to reference :(. The info type messages look fairly different from the price type messages so I feel like this is sufficient?

alejoberardino:
@austinborn Just saw these new fields by happenstance from a random slack thread, and went in to take a look. The response from the adapter should not have any extra properties at the root level. These should live in the `data` property, or support should be added in the framework. It's working right now because the types are not sufficiently strict but they will be, and this will break (as well as from potential runtime validations we may add in the future).
Easier to fix now than in the future if more of these cases happen, let me know and we can discuss more if needed

PR #2484, opened 2023-02-06
Add Alphavantage v3 EA

bogos:
Close: [PDI-1996](https://smartcontract-it.atlassian.net/browse/PDI-1996)


.changeset/tender-rings-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/alphavantage-test-adapter': minor
```
amit-momin, line 2: Since this is a new EA could you bump this to `major`

packages/sources/alphavantage-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.22.0",
```
amit-momin, line 31: Just dropping a reminder here to bump this but we should maybe wait until Alejo's fw changes are in so we can go directly to that version.

packages/sources/alphavantage-test/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
+export const DEFAULT_API_ENDPOINT = 'forex'
```
amit-momin, line 1: We probably don't need this line anymore from v2

packages/sources/alphavantage-test/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export const customSettings = {
+  API_ENDPOINT: {
+    description: 'The HTTP URL to retrieve data from',
+    type: 'string',
+    default: 'https://www.alphavantage.co/query',
+    required: true,
```
amit-momin, line 7: Since we have a default, we don't need to mark this as required

packages/sources/alphavantage-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    })
+  },
+  parseResponse: (params, res) => {
+    if (!res.data) {
+      logger.error(`There was a problem getting the data from the source`)
+      return []
+    }
+    return params.map((param) => {
+      const result = Number(res.data['Realtime Currency Exchange Rate']['5. Exchange Rate'])
```
amit-momin, line 39: We should add some validation that checks whether the response has a value for `Realtime Currency Exchange Rate` and `5. Exchange Rate` separately so we could return an error response if one doesn't.

amit-momin: We could also add a check if `Error Message` returned anything

packages/sources/alphavantage-test/src/endpoint/router.ts
```diff
@@ -0,0 +1,67 @@
...
+    ResponseBody: ProviderResponseBody
+  }
+}
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    HTTP: httpTransport,
+  },
+  () => 'HTTP',
+)
```
amit-momin, line 61: Since this adapter only has HTTP, we can skip on using the routing transport and just use the http transport directly in the endpoint

packages/sources/alphavantage-test/package.json
amit-momin: Version `0.25.0` is ready to be used

packages/sources/alphavantage-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "tslib": "2.4.0"
+  },
+  "devDependencies": {
+    "@types/jest": "27.5.2",
+    "@types/node": "16.11.51",
+    "@types/supertest": "2.0.12",
+    "nock": "13.2.9",
+    "supertest": "6.2.4",
+    "typescript": "4.7.4"
```
amit-momin, line 40: Dropping a note to upgrade typescript here.
```suggestion
    "typescript": "4.9.5"
```

packages/sources/alphavantage-test/src/endpoint/router.ts
```diff
@@ -0,0 +1,67 @@
...
+    required: true,
+  },
+  quote: {
+    aliases: ['to', 'market'],
+    type: 'string',
+    description:
+      'The symbol of the currency to convert to. The full list of options can be found here [Physical Currency list](https://www.alphavantage.co/physical_currency_list/) or [Cryptocurrency list](https://www.alphavantage.co/digital_currency_list/)',
+    required: true,
+  },
+} as const
```
amit-momin, line 22: After the typescript upgrade and the fw bump past 0.25, we want to move to using the satisfies operator here like 
```suggestion
} satisfies InputParameters
```

.changeset/tender-rings-drive.md
bogos: Done

packages/sources/alphavantage-test/package.json
bogos: Done

packages/sources/alphavantage-test/src/config/index.ts
bogos: Done

packages/sources/alphavantage-test/src/config/index.ts
bogos: Done

packages/sources/alphavantage-test/src/endpoint/router.ts
bogos: Dona

packages/sources/alphavantage-test/src/endpoint/forex.ts
bogos: Done

packages/sources/alphavantage-test/src/endpoint/router.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    HTTP: httpTransport,
+  },
+  () => 'HTTP',
+)
+
+export const endpoint = new AdapterEndpoint<EndpointTypes>({
+  name: 'forex',
```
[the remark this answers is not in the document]
bogos: Done

packages/sources/alphavantage-test/package.json
bogos: Done

packages/sources/alphavantage-test/src/endpoint/router.ts
bogos: Done

packages/sources/alphavantage-test/src/endpoint/router.ts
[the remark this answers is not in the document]
amit-momin: Could you also consolidate these in the forex.ts file? The router.ts name may be misleading now

packages/sources/alphavantage-test/src/endpoint/router.ts
amit-momin: Could you also consolidate these in the forex.ts file? The router.ts name may be misleading now

bogos: Done

packages/sources/alphavantage-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,82 @@
...
+        `There was a problem getting the 'Realtime Currency Exchange Rate' data from the source`,
+      )
+      return []
+    }
+
+    const exchangeRate = res.data['Realtime Currency Exchange Rate']['5. Exchange Rate']
+    if (!exchangeRate) {
+      logger.error(`There was a problem getting the 'Exchange Rate' data from the source`)
+      return []
+    }
```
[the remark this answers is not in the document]
bogos: Dona

packages/sources/alphavantage-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.27.4",
```
amit-momin, line 31: Wouldn't be a complete PR review without asking for one last version bump

packages/sources/alphavantage-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,132 @@
...
+          to_currency: to,
+          from_symbol: from,
+          to_symbol: to,
+          symbol: from,
+          market: to,
+          apikey: settings.API_KEY,
+        },
+      }
+      return {
+        params,
```
amit-momin, line 71: Also just realized we're passing all params here. I think we only want to pass the ones associated to this request. Otherwise in parseResponse we'll get all params for every response causing us to set the wrong results for the wrong params. Since each param has a separate request, I think this should be something like 
```suggestion
        params: [{ ...param }]
```

amit-momin: I think we're still missing this change. Since we build requests for each individual param, we should only pass that param for this request
```suggestion
        params: [param],
```

packages/sources/alphavantage-test/package.json
amit-momin: Since it's been a while there's a new version we can bump to

amit-momin (approved):

PR #2488, opened 2023-02-09
Oanda External Adapter

austinborn:
## Description
Onboard the new DP Oanda with the first EA implementation of SSE. Follows [this spec](https://docs.google.com/document/d/1ZMGNv6uRhjMos-wO89mO1xFd7SuS12nA2dQR_ioDUos/edit#heading=h.4l1fd53wkrnr) for development.

## Changes
- Add Oanda EA
- Add custom "ModifiedSSETransport" since existing SSE transport had some failings (`EventSource` library doesn't play nice outside of browsers, and the subscription model for Oanda is different from the method shown in the framework SSE transport)
- Prevent attempts to subscribe with unsupported pairs since this makes the entire process fail
- Support IDR/USD and INR/USD over the `REST` transport, everything else via `SSE`
- Support a number of inverse pairs for higher precision
- Follow the bid/ask midpoint methodology to calculate prices

## Steps to Test
1. Clone branch:
```
git clone https://github.com/smartcontractkit/external-adapters-js.git
git checkout feat/oanda-ea
```
2. Save credentials:
```
export API_KEY=
export SSE_API_KEY=
export API_ACCOUNT_ID=
```
3. Run EA (command starting from root dir):
```
yarn && yarn setup
cd packages/sources/oanda
yarn start
```
4. Send requests:
```
curl -X POST "http://localhost:8080" -H "Content-Type: application/json" -d '{"jobRunID": "1","data": {"base": "INR","quote": "USD"}}'
```

## Quality Assurance


austinborn:
Soak tests have been failing for other PRs with a similar error, but we're currently unsure what the cause is. I'm going to focus on getting equivalent functional integration tests first.

austinborn:
```The configuration for this run is:
 {
  "action": "start",
  "adapter": "oanda",
  "release": "2488",
  "imageTag": "pr2488",
  "imageRepository": "***.dkr.ecr.***.amazonaws.com/adapters/",
  "helmChartDir": "chainlink/cl-adapter",
  "helmSecrets": "-f ./adapter-secrets/secure-sdlc/secrets.oanda.yaml",
  "helmValuesOverride": "",
  "name": "qa-ea-oanda-2488"
}
start oanda adapter as qa-ea-oanda-2488
[helm-secrets] Decrypt: ./adapter-secrets/secure-sdlc/secrets.oanda.yaml
Release "qa-ea-oanda-2488" does not exist. Installing it now.
Error: timed out waiting for the condition

[helm-secrets] Removed: ./adapter-secrets/secure-sdlc/secrets.oanda.yaml.dec
Error: plugin "secrets" exited with error
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Failed to deploy the external adapter: Release "qa-ea-oanda-2488" does not exist. Installing it now.
".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

**Comment by austinborn** [comment]
Here's the soak test command that deploys the EA, if anyone is curious:
https://github.com/smartcontractkit/external-adapters-js/blob/ae1cf06fb59db412335d3f16a306fd3d6a909308/packages/scripts/src/ephemeral-adapters/lib.ts#L170-L186

File: packages/sources/oanda/schemas/env.json [hunk 1]
```diff
@@ -0,0 +1,41 @@
+{
```
amit-momin, line 1: We actually don't need the env.json file anymore in v3 since all of the settings are defined in custom configs

packages/sources/oanda/src/endpoint/price.ts
```diff
@@ -0,0 +1,153 @@
...
+    {
+      headers: {
+        contentType: 'application/json',
+        Authorization: `Bearer ${context.adapterConfig.SSE_API_KEY}`,
+      },
+    },
+  )
+
+  if (!data || status !== 200) {
+    throw new AdapterError({ message: 'Could not fetch asset list', providerStatusCode: status })
```
amit-momin, line 44: We could use `AdapterDataProviderError` here since the DP is returning the error. I believe it'll have you provide timestamps but for `providerIndicatedTimeUnixMs` you can mark as undefined since it wouldn't make sense here.

packages/sources/oanda/src/transport/modifiedSSE.ts
```diff
@@ -0,0 +1,178 @@
...
+            restartStream()
+          })
+
+          stream.on('error', (error: any) => {
+            logger.error({ msg: 'Stream error occurred', error })
+            restartStream()
+          })
+
+          stream.on('data', (data: any) => {
+            logger.error({ data })
```
amit-momin, line 138: Is this expected to be a logger error?

packages/sources/oanda/src/transport/modifiedSSE.ts
```diff
@@ -0,0 +1,178 @@
...
+      logger.info({ subsList })
+
+      if (subsList.some((s) => !s)) {
+        logger.error({ msg: `Cannot update SSE subscriptions, subscriptions contains invalid sub` })
+      } else {
+        const sseConfig = this.config.prepareSSEConnectionConfig(subsList as string[], context)
+
+        const providerDataStreamEstablishedUnixMs = Date.now()
+
+        const restartStream = async () => {
```
amit-momin, line 86: Is the logic to restart the stream on every call to the streamHandler the expected subscription model for Oanda or is this something you just had to do since the default SSE transport wasn't playing nice?

amit-momin: As opposed to subscribe/unsubscribe and keep alive calls after establishing the connection once

packages/sources/oanda/src/endpoint/price.ts
```diff
@@ -0,0 +1,153 @@
...
+})
+
+// See https://developer.oanda.com/exchange-rates-api/#get-/v2/rates/spot.-ext-
+const restTransport = new HttpTransport<HttpGenerics>({
+  prepareRequests: (params: PriceEndpointParams[], config) =>
+    params.map((p) => {
+      const { base, quote } = p
+
+      return {
+        params: [p],
```
amit-momin, line 112: Shouldn't this just be `p` since the request is for a single set of params?
```suggestion
        params: p,
```

packages/sources/oanda/src/transport/modifiedSSE.ts
austinborn: Good catch, I'll change this to `logger.debug`

packages/sources/oanda/src/endpoint/price.ts
austinborn: The `ProviderRequestConfig` type requires this to be an array: https://github.com/smartcontractkit/ea-framework-js/blob/a3e9ba5a12d96a1b954c2d1bb990c9e19cd73173/src/transports/http.ts#L43

packages/sources/oanda/src/transport/modifiedSSE.ts
austinborn: If there are new or stale subscriptions then yes, we'd have to restart the stream. The stream is established with the list of pairs to subscribe to as input params that can't be updated while the stream is active (see `GET /v3/accounts/{accountID}/pricing/stream` [here](https://developer.oanda.com/rest-live-v20/pricing-ep/)).

packages/sources/oanda/schemas/env.json
austinborn: Removed

packages/sources/oanda/src/transport/modifiedSSE.ts
amit-momin: I see. All the more reason you needed a custom transport. Good to know

packages/sources/oanda/src/endpoint/price.ts
amit-momin: Oh shoot. I was mixing up `prepareRequest` and `parseResponse`. I see you have it correct below.

nikhsha:
Passes DP checks

ainsleybc (approved):

PR #2507, opened 2023-02-21
Add openexchangerates-test ea

karen-stepanyan:
[PDI-1998](https://smartcontract-it.atlassian.net/browse/PDI-1998)

[PDI-1998]: https://smartcontract-it.atlassian.net/browse/PDI-1998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/openexchangerates-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "0.24.2",
```
amit-momin, line 31: Just dropping a note here to bump this once the new fw version is out

packages/sources/openexchangerates-test/src/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  includes,
+  rateLimiting: {
+    tiers: {
+      developer: {
+        rateLimit1h: 13.69,
+        note: 'only mentions monthly limits',
+      },
+      enterprise: {
+        rateLimit1h: 136.9,
+      },
```
amit-momin, line 21: Even though the fw doesn't support -1 like v2, we should still add the unlimited tier. We would want to set a reasonable limit though so something like
```
rateLimit1s: 100,
rateLimit1m: 6000,
note: 'Considered unlimited tier, but setting reasonable limits',
```

packages/sources/openexchangerates-test/package.json
amit-momin: Version `0.25.0` is ready to be used

packages/sources/openexchangerates-test/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@chainlink/external-adapter-framework": "0.25.0",
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@types/jest": "27.5.2",
+    "@types/node": "16.11.51",
+    "@types/supertest": "2.0.12",
+    "nock": "13.2.9",
+    "supertest": "6.2.4",
+    "typescript": "4.7.4"
```
amit-momin, line 40: We have a change coming soon that's upgrading typescript in the whole monorepo to the latest version 4.9.5. Could you also bump this up?
```suggestion
    "typescript": "4.9.5"
```

amit-momin:
I noticed tests are failing because of the latest version of axios in the framework. There's a fix coming for that in the PR that upgrades typescript so we can wait to get this in after that

packages/sources/openexchangerates-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,114 @@
...
+    type: 'string',
+    description: 'The symbol of symbols of the currency to query',
+  },
+  quote: {
+    aliases: ['to', 'market'],
+    required: true,
+    type: 'string',
+    description: 'The symbol of the currency to convert to',
+  },
+} as const
```
amit-momin, line 41: After the typescript upgrade and the fw bump past 0.25, we want to move to using the satisfies operator here like
```suggestion
} satisfies InputParameters & PriceEndpointInputParameters
```

packages/sources/openexchangerates-test/src/endpoint/forex.ts
```diff
@@ -0,0 +1,103 @@
...
+            app_id: config.API_KEY,
+            base,
+          },
+        },
+      }
+    })
+  },
+  parseResponse: (params, res) => {
+    return params.map((param) => {
+      const result = res.data.rates[param.quote.toUpperCase()]
```
amit-momin, line 75: Just for safety could we check if `rates` exists first and return a generic error if it doesn't for all params?

karen-stepanyan: done

amit-momin (approved):

PR #2534, opened 2023-03-09
Feature/df-18622 coinmetrics LWBA

mjk90:
## Description
Added crypto-lwba endpoint to coinmetrics adapter
......

## Steps to Test

`yarn test packages/sources/coinmetrics-test/test`

## Quality Assurance


austinborn (dismissed):

packages/sources/coinmetrics-test/src/endpoint/lwba-router.ts
```diff
@@ -0,0 +1,58 @@
...
+  wsTransport,
+)
+
+export const endpoint = new AdapterEndpoint<CryptoLwbaEndpointTypes>({
+  name: 'crypto-lwba',
+  aliases: ['crypto_lwba'],
+  transportRoutes,
+  defaultTransport: 'ws',
+  inputParameters: priceEndpointInputParameters,
+})
```
amit-momin, line 58: Sorry to jump in after this got approved but I don't think we need to use a router here. Looks like the endpoint only uses one transport. This probably got copied over from the `price` endpoint but that's going to be updated soon to have both an http and ws transport. Think you can just specify the transport directly here 
```suggestion
export const endpoint = new AdapterEndpoint<CryptoLwbaEndpointTypes>({
  name: 'crypto-lwba',
  aliases: ['crypto_lwba'],
  transport: wsTransport,
  inputParameters: priceEndpointInputParameters,
})
```

packages/sources/coinmetrics-test/src/index.ts
```diff
@@ -1,12 +1,12 @@
 import { expose, ServerInstance } from '@chainlink/external-adapter-framework'
-import { PriceAdapter } from '@chainlink/external-adapter-framework/adapter'
+import { Adapter } from '@chainlink/external-adapter-framework/adapter'
+import { priceRouter, lwbaRouter } from './endpoint'
 import { config } from './config'
-import { priceRouter } from './endpoint'
 
-export const adapter = new PriceAdapter({
+export const adapter = new Adapter({
```
amit-momin, line 6: I think we should keep this as `PriceAdapter`. Only one endpoint needs to be a price endpoint which is the case.

austinborn (approved):

PR #2570, opened 2023-03-23
Refactor CMC v3 to reuse DP responses

alejoberardino:


packages/sources/coinmarketcap-test/src/endpoint/crypto.ts
```diff
@@ -1,25 +1,270 @@
...
+        },
+      }
+    })
   },
 })
 
 export const endpoint = new CryptoPriceEndpoint<CryptoEndpointTypes>({
-  name: 'crypto',
-  aliases: ['price'],
+  name: 'price',
```
amit-momin, line 249: I know it's covered by the CryptoPriceEndpoint but looks like the name and alias got switched around here

packages/sources/coinmarketcap-test/src/endpoint/crypto.ts
```diff
@@ -1,25 +1,270 @@
...
+    required: false,
+    type: 'string',
+  },
+  slug: {
+    description: 'The CMC coin ID (optional to use in place of base)',
+    required: false,
+    type: 'string',
+  },
+  resultPath: {
+    description: 'The path to the result within the provider response',
```
amit-momin, line 46: Nitpick: This isn't really the result within the provider response but don't want to call it the endpoint to avoid confusion. Maybe we can leave it

packages/sources/coinmarketcap-test/src/endpoint/crypto.ts
```diff
@@ -1,25 +1,270 @@
...
+        return {
+          params: p,
+          response: {
+            statusCode: 502,
+            errorMessage: `An ID was not found in the idsMap for parameter ${JSON.stringify(p)}`,
+          },
+        }
+      }
+
+      const data = res.data.data[id]
```
amit-momin, line 194: I think the results in data are keyed by the id, slug, or symbol used in the query param. Wouldn't this be an issue if we're only using id to search? We might not need an idsMap and can do something like 
```suggestion
      const data = res.data.data[p.cid || p.slug || p.base]
```

packages/sources/coinmarketcap-test/src/endpoint/crypto.ts
```diff
@@ -1,25 +1,270 @@
...
+      if (!data) {
+        return {
+          params: p,
+          response: {
+            statusCode: 502,
+            errorMessage: `ID type "${idType}" was not found in response for request: ${JSON.stringify(
```
amit-momin, line 200: If I'm right in the last comment, we might need to update this too
```suggestion
            errorMessage: `Data was not found in response for request: ${JSON.stringify(
```

amit-momin (approved):

PR #2594, opened 2023-03-30
Update READMEs manually

boxhock:
Manually re-regenerated READMEs for all EAs, as some were outdated.


packages/sources/snowflake/README.md
```diff
@@ -46,68 +46,7 @@ Queries US confirmed Covid cases per County, using the John Hopkins University t
 
 ### Example
 
-Request:
```
amit-momin, line 49: Looks like examples are getting removed here

amit-momin:
Consumed changeset that contained an update for all v3 adapters which fixed all of the README's

PR #2601, opened 2023-03-31
Upgrade coinmetrics to v3

alejoberardino:


packages/sources/coinmetrics/CHANGELOG.md
```diff
@@ -1,784 +1,102 @@
-# @chainlink/coinmetrics-adapter
```
amit-momin, line 1: I think the CHANGELOG shouldn't have any changes

packages/sources/coinmetrics/tsconfig.json
```diff
@@ -6,9 +6,5 @@
   },
   "include": ["src/**/*", "src/**/*.json"],
   "exclude": ["dist", "**/*.spec.ts", "**/*.test.ts"],
-  "references": [
-    { "path": "../../core/test-helpers" },
-    { "path": "../../core/bootstrap" },
-    { "path": "../../core/test-helpers" }
-  ]
+  "references": [{ "path": "../../core/test-helpers" }, { "path": "../../core/bootstrap" }]
```
amit-momin, line 9: Think this slipped through the cracks with the `-test` adapter but we should remove this ref to core

packages/non-deployable/token-allocation/src/config/index.ts
```diff
@@ -38,10 +38,8 @@ export const adaptersV2: v2AdapterImplementation[] = [
 ]
```
amit-momin, line 38: It doesn't let me put the comment higher but there's a ref to `coinmetrics` in the v2 list still

amit-momin (approved):

PR #2629, opened 2023-04-11
New Stader PoR Changes

amit-momin:
## Description

Stader PoR changes required more complex calculations than just simply retrieving balances from the beacon chain. To accommodate these new requirements, the existing `stader-address-list` adapter was update to be compatible with the new contracts. A new balance adapter, `stader-balance`, was created for the custom calculations. The `proof-of-reserves` composite adapter was update to include the new adapter.

## Changes

- `stader-address-list` was updated to be compatible with new contracts and logic
- `stader-address-list` was upgraded to the v3 framework
- `stader-balance` adapter was created for custom calculation logic with the v3 framework
- `proof-of-reserves` composite adapter was update to include the new balance adapter

## Steps to Test

1. yarn test packages/sources/stader-address-list/test
2. yarn test packages/sources/stader-balance/test

## Quality Assurance

[PDI-2033]: https://smartcontract-it.atlassian.net/browse/PDI-2033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+            this.getStaderStakeManagerBalance(req, blockTag, balances),
+            // Get balance of all execution layer reward addresses
+            this.getElRewardBalances(req, blockTag, balances, context.adapterSettings),
+            // Get balance of the Permissioned Pool address
+            this.getPermissionedPoolBalance(req, blockTag, balances),
+          ])
+
+          await Promise.all([
+            // Perform validator level calculations
+            await this.performValidatorCalculations(
```
alejoberardino, line 92: pretty sure this await shouldn't  be here

packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+      // Separate the address set into the specified batch size
+      // Add the batches as comma-separated lists to a new list used to make the requests
+      for (let i = 0; i < addresses.length / batchSize; i++) {
+        batchedAddresses.push(
+          addresses
+            .slice(i * batchSize, i * batchSize + batchSize)
+            .map(({ address }) => address)
+            .join(','),
+        )
+      }
```
alejoberardino, line 164: same thing as above, all these batching steps should be abstracted into a common pattern to make it easier to read

packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+        await Promise.all(
+          group.map(async (address) => {
+            const options: AxiosRequestConfig = {
+              baseURL: settings.BEACON_RPC_URL,
+              url,
+              params: { id: address, status: statusList },
+            }
+            return axios.request<ProviderResponse>(options).then((response) => {
+              responses.push(response)
+            })
```
alejoberardino, line 179:
```suggestion
            const response = await axios.request({
              baseURL: settings.BEACON_RPC_URL,
              url,
              params: { id: address, status: statusList },
            })
            responses.push(response)
            return
```

packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+        )
+      }
+
+      // Flatten the results into single array for validators and balances
+      const validatorBatches = responses.map(({ data }) => data)
+      validatorBatches.forEach(({ data }) => {
+        data.forEach((validator) => {
+          result.push(validator)
+        })
+      })
```
alejoberardino, line 190: foreach -> push doesn't look right

packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+            logger.debug(
+              `Validator (${validatorAddress}): Collateral ETH: ${collateralEth}. User Deposit: ${userDeposit}`,
+            )
+            // Convert gwei balance from Beacon to wei to align with values from execution layer in wei
+            const validatorBalance = BigNumber(
+              ethers.utils.parseUnits(state.balance, 'gwei').toString(),
+            )
+            // Add validator to deposited address list if Stader status is DEPOSITED and balance is 1 ETH on beacon
+            // Deposited address list will be used to look for 31 ETH deposit event in logs later
+            if (
```
alejoberardino, line 238: This looks too deeply nested

packages/sources/stader-balance/src/endpoint/balance.ts
```diff
@@ -0,0 +1,741 @@
...
+    addresses: ValidatorAddress[],
+    validatorStateList: ValidatorState[],
+    balances: BalanceResponse[],
+    req: RequestParams,
+    validatorDeposit: BigNumber,
+    ethDepositContract: string,
+    commissionMap: Record<number, number>,
+    collateralEthMap: Record<number, BigNumber>,
+    blockTag: number,
+  ): Promise<void> {
```
alejoberardino, line 208: If we have more than ~3 parameters, these should be passed as an obj for readability

alejoberardino: (readonly, to avoid problems)

alejoberardino (changes requested):

packages/sources/stader-address-list/src/endpoint/address.ts
```diff
@@ -1,41 +1,73 @@
...
+    type: 'string',
+  },
+  permissionedPoolAddress: {
+    description: 'The address of the Stader Permissioned Pool.',
+    type: 'string',
+  },
+  staderConfigAddress: {
+    description: 'The address of the Stader Config contract.',
     type: 'string',
   },
```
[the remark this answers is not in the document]
amit-momin: I took the design pattern from other PoR adapters which include these as input parameter so we can override the default contract address if we need to. But that does bring up a good point that these should be configured as env vars. Following other PoR adapters, the defaults are stored in a map in the code.

packages/sources/stader-address-list/src/endpoint/address.ts
```diff
@@ -51,50 +83,307 @@ export const inputParameters: InputParameters<TInputParameters> = {
...
+  async backgroundHandler(
+    context: EndpointContext<EndpointTypes>,
+    entries: RequestParams[],
+  ): Promise<void> {
+    if (!entries.length) {
+      await sleep(context.adapterSettings.BACKGROUND_EXECUTE_MS)
+      return
+    }
+    await Promise.all(
+      entries.map(async (req) => {
```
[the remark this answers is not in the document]
amit-momin: Probably not many but we could have more than one if the addresses change or if there's any reason to have requests for different confirmations

packages/sources/stader-address-list/src/utils.ts
```diff
@@ -1,49 +1,51 @@
...
+  providerDataRequestedUnixMs: number,
+): TimestampedProviderErrorResponse => {
+  return {
+    statusCode: 502,
+    errorMessage,
+    timestamps: {
+      providerDataRequestedUnixMs,
+      providerDataReceivedUnixMs: 0,
+      providerIndicatedTimeUnixMs: undefined,
+    },
```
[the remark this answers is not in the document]
amit-momin: I believe we do? I think we're usually adding them behind the scenes even for errors before it's written to response cache. Unless I've set up my types incorrectly. Although following the trend in the fw I should also set `providerDataRequestedUnixMs` to 0 on error

packages/sources/stader-address-list/src/endpoint/address.ts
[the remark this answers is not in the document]
amit-momin: I should also say I dislike this pattern of sending params to the address list adapter just for the sake of being able to pass it to the balance adapter through the response. Something I want to change in the `proof-of-reserves` composite potentially but that would maybe require some investigation

packages/sources/stader-balance/src/endpoint/balance.ts
amit-momin: I probably can't make it read-only because some of the parameters like `commissionMap` and `collateralEthMap` are updated on the fly and used in a couple different functions. But will move this to an object for sure

packages/sources/stader-address-list/src/endpoint/address.ts
```diff
@@ -51,50 +88,314 @@ export const inputParameters: InputParameters<TInputParameters> = {
...
+  }): Promise<PoolAddress[]> {
+    try {
+      logger.debug('Fetching socializing pool address list')
+      const poolCount = await poolFactoryManager.poolCount({ blockTag })
+      let socialPoolAddresses: PoolAddress[] = []
+      for (let i = 1; i <= poolCount; i++) {
+        const address = await poolFactoryManager.getSocializingPoolAddress(i, { blockTag })
+        socialPoolAddresses.push({ address, poolId: i })
+      }
+      socialPoolAddresses = filterDuplicates<PoolAddress>(socialPoolAddresses)
```
alejoberardino, line 344: I think this is something we could abstract as well
```typescript
const runAllSequentially = <T>(count, fn: (i: number) => Promise<T>): Promise<T[]> => {
	const results: T[] = []
	for (let i = 1; i <= count; i++) {
		results.push(await fn(i))
	}
	return results
}
```
And then
```suggestion
      const rawSocialPoolAddresses = await runAllSequentially(poolCount, i => poolFactoryManager.getSocializingPoolAddress)
      const socialPoolAddresses = filterDuplicates(socialPoolAddresses) 
      // Note the lack of explicit generic above, should be inferred by the param
```

alejoberardino (approved):

PR #2672, opened 2023-04-27
Stader-balance balance calculations update

amit-momin:
## Description

Previously calculations would only consider the balance found on beacon and report ETH in limbo found in deposit events separately. The logic has been updated to include the ETH found in deposit events in balance calculations.

## Steps to Test

1. yarn test packages/sources/stader-balance/test

## Quality Assurance

### DIFF
--- a/.changeset/hip-cougars-trade.md
+++ b/.changeset/hip-cougars-trade.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/stader-balance-adapter': patch
+---
+
+Updated balance calculations to include deposited ETH value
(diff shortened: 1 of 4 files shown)


alejoberardino (commented):
We should not pass arrays to be modified. Now that the validator will account for the deposited ETH, it should be created with it, and the limbo and deposited addresses should be mapped to get constant lookups

amit-momin:
> We should not pass arrays to be modified. Now that the validator will account for the deposited ETH, it should be created with it, and the limbo and deposited addresses should be mapped to get constant lookups

Map is def the better move to track the deposited eth but don't know if I can take the route to include the value on validator creation. I'd have to include the limbo logic in the beacon retrieval but I think we'd want to keep it separate for rate limiting.

I could pass the deposited validator objects to the limbo logic and update the totals directly but then I'd be updating an array that was passed

alejoberardino (approved):

PR #2761, opened 2023-05-24
Exclude test and config files from Sonar coverage

chudilka1:


mxiao-cll:
Please re-open if still relevant

PR #2769, opened 2023-05-28
Pipelines refactor

alejoberardino:
This PR refactors the current GH workflows for this repository as previously designed and discussed internally.

## Differences with RFC design
- Soak testing is left unmodified; it proved very difficult to integrate cleanly in its current state and as those are separate workflows and not required ones, it was easier to keep them as is. They are still the slowest part of the PR CI, but refactoring them needs deeper discussions.
- Since soak tests are kept separate, a difference from the design is that images are not built at the PR stage to speed up the process.

## How does it work?
- Only one branch will be used after this change: `main`.
- PR checks have been sped up significantly, where every possible step will try to only execute for the relevant adapters.
- Upon push to main (PR merge, once no option but squash is left enabled, used this way to better restrict deployment environments):
	- The relevant adapter images will be built, published, and deployed to staging.
	- A PR for the release will be automatically created / updated, containing the new versions + all the automatic readme changes.
- Once the Release PR is merged:
	- Images will be pulled from private ECR (no rebuilding, much faster this way) and published to the public one.
	- Once that's successful, a GH Release will be automatically published using the description from the PR.
	
## Steps after merging this PR
- Remove all merge options except squash to guarantee a single commit per PR (which will avoid messing up the history as well as making workflows easier).
- Change the required checks to merge a PR to the new workflow names.
- Push the current `develop` to `main`.
- Switch the default branch to `main`.


amit-momin (dismissed):

.github/actions/publish-image/action.yml
```diff
@@ -0,0 +1,68 @@
...
+        --policy-text "$(jq '.Statement[0].Principal.AWS |= [${{ inputs.aws-ecr-account-ids }}]' .github/actions/publish-image/secondary-account-access-policy.json)" &> /dev/null;
+    - name: Push to ECR
+      shell: bash
+      run: |
+        docker push ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:${{ inputs.image-version }}
+    - name: Push to ECR with latest tag
+      shell: bash
+      run: |
+        docker tag ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:${{ inputs.image-version }} ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:latest
+        docker push ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:latest
```
amit-momin, line 68: Should this have a condition? If not, is the `latest` input being used?
```suggestion
    - name: Push to ECR with latest tag
      shell: bash
      if: inputs.latest == 'true'
      run: |
        docker tag ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:${{ inputs.image-version }} ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:latest
        docker push ${{ inputs.aws-ecr-url }}/${{ inputs.aws-ecr-repo }}:latest
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,112 @@
...
+    inputs:
+      # For this workflow, BUILD_ALL will cause all adapters to have their image built and deployed
+      build-all:
+        description: whether to run steps for all adapters, regardless of whether they were changed in this event
+        required: false
+        default: 'false'
+      trigger-infra-workflow:
+        description: whether to trigger the infra deployment workflow after publishing images
+        required: false
+        default: 'true'
```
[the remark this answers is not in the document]
alejoberardino: To potentially deploy every adapter to ECR without triggering a deployment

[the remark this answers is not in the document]
amit-momin: Wouldn't we always want the digest updates in our infra if we're deploying new adapters to ECR? I guess this workflow could always just be triggered again with the flag set to `true`

amit-momin (approved):

.github/scripts/changed-adapters.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Build a list to use with matrix strategies
+CHANGED_ADAPTERS=$(
+  echo $CHANGED_PACKAGES \
+  | jq -cr '{
+      adapter: [
+        .[]
+        | select(
+          .location
+          | startswith("packages/sources") or startswith("packages/composites") or startswith("packages/targets"))
```
austinborn, line 34: Maybe we can look into decommissioning the `targets` EAs since they aren't really being developed at this point. Not for this PR, but for future ref.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up and install dependencies
+        uses: ./.github/actions/setup
+      - name: Test Master List Generation
+        run: yarn generate:master-list -v
+      - name: Test README Generation
+        run: yarn generate:readme -v
```
austinborn, line 126: You may be able to use `CHANGED_PACKAGES` here as well

austinborn (approved):
Yeah this is awesome. Huge boost to iteration speed.

.github/scripts/changed-adapters.sh
alejoberardino: Agreed, there's an ongoing plan to clean a lot of these EAs but it might take a while to make sure it's proper to do so

.github/workflows/checks.yml
alejoberardino: The scripts already use the upstream branch to make the calculations, it's a bit inefficient since we could use the packages like you say but changing the code was too much for this PR imo

PR #2821, opened 2023-06-09
Refactor v3 integration tests pt2

karen-stepanyan:
## Description

This is the second PR out of 3 where integration tests of v3 adapters are changed. In this PR integration tests of following adapters are refactored 
alongside, alphavantage-test, blocksize-capital, coinpaprika, dar, elwood, finage-test, finnhub, galaxy, gsr.

## Changes

- `supertest` and `@types/supertest`  packages are removed from v3 packages. They still remain for those where e2e tests are written. `supertest` was used for v2 framework  as ea-bootstrap was using `express` and in order to test something we had to spawn a server and make requests. Since v3 framework uses fastify, the `inject` api makes it much easier to test APIs without creating a new server every time. To achieve this new integration tests use `TestAdapter` from v3 framework which internally uses `fastify.inject` to make requests and assert results. 
- `mock-socket` package is removed from websocket v3 EAs. The two important classes that adapters use from this package are now exported from framework (`MockWebsocketServer` and `MockWebsocket`) so they can be directly imported from framework.
- `mockWebSocketProvider` function is removed from all integration tests. This function was present in every WS integration test with the same code so now it is part of framework and can be imported from there.
- setup.ts file is removed from all integration tests.
- `@sinonjs/fake-timers` and `@types/sinonjs__fake-timers`  packages are added to websocket v3 EAs. Fake timers are needed to configure the `TestAdapter` for websocket tests so that we can get rid of random sleep codes (`await sleep(5000)`)
- Rest and WS integration tests are now split into their own files. There was a mixed pattern where in some cases all tests were in one file, whereas in other cases WS tests were in separate file.  

## Steps to Test
 
yarn test:ci:integration


alejoberardino (dismissed):

amit-momin:
This was actually an issue with pt3 but since it's merged thought we could maybe just fix it under this PR. Looks like the `cyan-apricots-smile` changeset has a reference to `kaiko-test`, but now that it's been migrated could you update it to `kaiko`?

amit-momin (approved):

PR #2834, opened 2023-06-14
Northwestnodes

northwestnodes-eric:
## Description

......

## Changes

- Created new Northwest Nodes EA

## Steps to Test

1. Regular supertest

## Quality Assurance

### DIFF
--- a/packages/sources/northwestnodes/package.json
+++ b/packages/sources/northwestnodes/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "@chainlink/northwestnodes-adapter",
+  "version": "1.0",
+  "description": "Chainlink Northwest Nodes API adapter.",
+  "keywords": [
+    "Chainlink",
+    "LINK",
+    "blockchain",
+    "oracle",
+    "northwestnodes"
+  ],
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "url": "https://github.com/smartcontractkit/external-adapters-js",
+    "type": "git"
+  },
+  "license": "MIT",
+  "scripts": {
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/ea-bootstrap": "workspace:*",
+    "@chainlink/ea-test-helpers": "workspace:*",
+    "@chainlink/external-adapter-framework": "0.29.8",
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@types/jest": "27.5.2",
+    "@types/node": "16.11.51",
+    "@types/supertest": "2.0.12",
+    "nock": "13.2.9",
+    "supertest": "6.2.4",
+    "typescript": "5.0.4"
+  }
+}
--- a/packages/sources/northwestnodes/package.json
+++ b/packages/sources/northwestnodes/package.json
@@ -28,8 +28,6 @@
     "start": "yarn server:dist"
   },
   "dependencies": {
-    "@chainlink/ea-bootstrap": "workspace:*",
-    "@chainlink/ea-test-helpers": "workspace:*",
     "@chainlink/external-adapter-framework": "0.29.8",
     "tslib": "^2.3.1"
   },
--- a/packages/sources/northwestnodes/src/config/index.ts
+++ b/packages/sources/northwestnodes/src/config/index.ts
@@ -13,6 +13,6 @@ export const config = new AdapterConfig({
     description: 'The default HTTP API base url',
     type: 'string',
     required: false,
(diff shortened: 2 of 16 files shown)


packages/sources/northwestnodes/src/config/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    type: 'string',
+    required: false,
+    default: 'https://api.northwestnodes.com/v2',
+  },
+  WARMUP_SUBSCRIPTION_TTL: {
+    description: 'Reduced warmup subscription TTL to 10 seconds, for this EA',
+    type: 'number',
+    required: false,
+    default: 10000,
+  },
```
alejoberardino, line 23: This will actually fail as startup, because adding it here means you add it as a custom setting and you're not allowed to override a base framework setting. To override a default you can pass a second parameter to the `AdapterConfig` with an `envDefaultOverrides` prop

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/list.ts
```diff
@@ -0,0 +1,96 @@
...
+      const statusCode = res.status
+      const statusText = res.statusText
+
+      logger.debug('parseResponse: statusCode=' + statusCode + '; statusText=' + statusText + ';')
+
+      if (statusCode == 200) {
+        return {
+          params: p,
+          response: {
+            data: JSON.stringify(res.data),
```
alejoberardino, line 73: If this will be used in a classic feed, we need to return a non-null result as that will be the value used in OCR

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/single.ts
```diff
@@ -0,0 +1,95 @@
...
+  Provider: {
+    ResponseBody: IEpochResponse
+    RequestBody: null
+  }
+}
+
+const transport = new HttpTransport<EndpointType>({
+  prepareRequests: (params: (typeof inputParameters.validated)[], settings) => {
+    const baseURL = settings.API_ENDPOINT
+    const id = params[0].id.toLowerCase()
```
alejoberardino, line 35: Same thing as the endpoint above, you could have different requests to this adapter that want different epochs so the source requests would be different

packages/sources/northwestnodes/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "dependencies": {
+    "@chainlink/ea-bootstrap": "workspace:*",
+    "@chainlink/ea-test-helpers": "workspace:*",
```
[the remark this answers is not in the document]
northwestnodes-eric: Fixed.

packages/sources/northwestnodes/src/config/index.ts
northwestnodes-eric: Removed entirely based on conversations.

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/list.ts
```diff
@@ -0,0 +1,96 @@
...
+            result: null,
+          },
+        }
+      } else {
+        return {
+          params: p,
+          response: {
+            data: res.statusText,
+            status: res.status,
+            result: null,
```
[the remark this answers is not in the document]
northwestnodes-eric: Should be fixed

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/list.ts
northwestnodes-eric: Fixed

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/single.ts
northwestnodes-eric: Fixed

packages/sources/northwestnodes/src/endpoint/staking/ethereum/epoch/list.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    return {
+      params,
+      request: {
+        baseURL,
+        url,
+        method: 'GET',
+        params: query,
+      },
+    }
```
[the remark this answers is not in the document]
northwestnodes-eric: Fixed

alejoberardino (commented):
LGTM!
Only remaining note is that we should add a default rate limit tier to the Adapter, otherwise the framework will send requests as fast as it possibly can

northwestnodes-eric:
> LGTM! Only remaining note is that we should add a default rate limit tier to the Adapter, otherwise the framework will send requests as fast as it possibly can

done

alejoberardino (approved):

alejoberardino:
Sorry @northwestnodes-eric, one thing I just noticed is that this PR targets `develop` but the current tracking branch we use has changed to `main`, could you change the target?

Additionally, to double check the rate limits, they are all effectively translated to per second with the default strategy, so it will mean that EAs will run with 1 req / min

PR #2839, opened 2023-06-16
Standardize v3 adapter structure pt 1

karen-stepanyan:
## Description

This is the first PR out of 3 where v3 adapter structure is refactored. The goal is to have a standard that all v3 adapters will follow in terms of file structure type/code sharing. Adapters refactored in this PR 
cfbenchmarks, coinmarketcap, cryptocomapre, coingecko, coinmetrics, dxfeed, moonbeam-address-list, ncfx, tiingo, tradermade

No functional changes were made. 

In summary this is how the new structure looks like 

```
src
	-config
		*index.ts
                *other json files (includes, overrides, abi files)

	-endpoint
		*index.ts
		*crypto-router.ts (AdapterEndpoint itself, inputParams, common EndpointTypes and all other types and functions for endpoint)
		*forex-router.ts
                *utils.ts (optional. if endpoints share similar logic/types)

	-transport
		*crypto-http.ts (only the transport logic (can be custom), extended HttpTransportTypes from EndpointTypes and transport specific types like ProviderResponseBody)
		*forex-http.ts

		*crypto-ws.ts (only the transport logic, extended WsTransportTypes from EndpointTypes and transport specific types like WsMessage)
		*forex-ws.ts

		*utils.ts (optional. utility functions or if multiple http or ws transports share similar logic. in case of ws we can have custom WebSocketTransports or WebsocketReverseMappingTransports)
```

## Changes

- With the  new approach endpoint logic is separated from transport logic, therefore  new `transport` folder inside `src` is created.  `endpoint` folder now contains only endpoint files. Those files now include only endpoint related information, for example **endpoint declaration** (_new AdapterEndpoint_), **input parameters declaration**, other helper functions for endpoint declaration (**customInputValidation, requestTransforms, cacheKeyGenerator**) and base type for the endpoint (**BaseEndpointTypes**).  Other entities that are not directly needed to create an endpoint should not be there as it is most likely transport related and should be in transport files.
- For each endpoint file there is now a same named transport file in `transport` folder. The only exception is router endpoints (endpoints that have many transports) and in that case transport folder will have more files depending how many transports that specific endpoint supports. 
- Transport files  import and extend **BaseEndpointTypes** from endpoint to create transport specific types, like  `WsTransportTypes` or `HttpTransportTypes`. In terms of functionality they only export transport logic.
- In case some endpoints share similar logic/types (input params, same BaseEndpointTypes) those shared entities are in `utils.ts` file inside `endpoint` folder. 
- In case some transports share similar logic/types (same helper functions, same provider response interface) those shared entities are  in `utils.ts` file inside `transport` folder. 

## Steps to Test

yarn test:ci:integration


.changeset/ninety-rings-brake.md
```diff
@@ -0,0 +1,13 @@
+---
+'@chainlink/cfbenchmarks-adapter': patch
+'@chainlink/coingecko-adapter': patch
+'@chainlink/coinmarketcap-adapter': patch
+'@chainlink/cryptocompare-adapter': patch
+'@chainlink/dxfeed-adapter': patch
+'@chainlink/moonbeam-address-list-adapter': patch
+'@chainlink/ncfx-adapter': patch
+'@chainlink/tiingo-adapter': patch
+'@chainlink/tradermade-adapter': patch
```
amit-momin, line 10: Think `coinmetrics` and `dxfeed-secondary` are missing from the changeset

karen-stepanyan: Good catch. I'll add them

packages/sources/cfbenchmarks/src/endpoint/crypto-router.ts
```diff
@@ -6,10 +6,10 @@ import {
 } from '@chainlink/external-adapter-framework/util'
```
amit-momin, line 6: This might be me just nitpicking but I don't think we need to add the `-router` to the end of the endpoint anymore. Think it made sense to when the endpoint and transports were in the same folder to help differentiate what's what. Since they're separated now we could just use the endpoint's name like `crypto`? What do you think?

packages/sources/coinmetrics/src/endpoint/lwba-ws.ts
```diff
@@ -1,27 +1,15 @@
 import {
```
amit-momin, line 1: Similar to the `-router` comment if you agree, we might not need `-ws` here on the endpoint anymore

karen-stepanyan: I think in this case this was my mistake since lwba is `ws` only  and suffix is not needed (Regardless if i agree about `router` or not). Same in the transport

packages/sources/cfbenchmarks/src/endpoint/crypto-router.ts
karen-stepanyan: Initially i wanted to sort of highlight that this is a router endpoint, that's why there are multiple transports but now, especially after working with EA generator, i think it makes more sense to get rid of that suffix. I will update the PRs

amit-momin (approved):

PR #2846, opened 2023-06-19
Standardize v3 adapter structure pt 2

karen-stepanyan:
## Description

This is the second PR out of 3 where v3 adapter structure is refactored. The goal is to have a standard that all v3 adapters will follow in terms of file structure type/code sharing. Adapters refactored in this PR 
alongside, alphavantage, blocksize-capital, coinpaprika, dar, elwood, finage-test, finnhub, galaxy, gsr

No functional changes were made. 

In summary this is how the new structure looks like 

```
src
	-config
		*index.ts
                *other json files (includes, overrides, abi files)

	-endpoint
		*index.ts
		*crypto-router.ts (AdapterEndpoint itself, inputParams, common EndpointTypes and all other types and functions for endpoint)
		*forex-router.ts
                *utils.ts (optional. if endpoints share similar logic/types)

	-transport
		*crypto-http.ts (only the transport logic (can be custom), extended HttpTransportTypes from EndpointTypes and transport specific types like ProviderResponseBody)
		*forex-http.ts

		*crypto-ws.ts (only the transport logic, extended WsTransportTypes from EndpointTypes and transport specific types like WsMessage)
		*forex-ws.ts

		*utils.ts (optional. utility functions or if multiple http or ws transports share similar logic. in case of ws we can have custom WebSocketTransports or WebsocketReverseMappingTransports)
```

## Changes

- With the  new approach endpoint logic is separated from transport logic, therefore  new `transport` folder inside `src` is created.  `endpoint` folder now contains only endpoint files. Those files now include only endpoint related information, for example **endpoint declaration** (_new AdapterEndpoint_), **input parameters declaration**, other helper functions for endpoint declaration (**customInputValidation, requestTransforms, cacheKeyGenerator**) and base type for the endpoint (**BaseEndpointTypes**).  Other entities that are not directly needed to create an endpoint should not be there as it is most likely transport related and should be in transport files.
- For each endpoint file there is now a same named transport file in `transport` folder. The only exception is router endpoints (endpoints that have many transports) and in that case transport folder will have more files depending how many transports that specific endpoint supports. 
- Transport files  import and extend **BaseEndpointTypes** from endpoint to create transport specific types, like  `WsTransportTypes` or `HttpTransportTypes`. In terms of functionality they only export transport logic.
- In case some endpoints share similar logic/types (input params, same BaseEndpointTypes) those shared entities are in `utils.ts` file inside `endpoint` folder. 
- In case some transports share similar logic/types (same helper functions, same provider response interface) those shared entities are  in `utils.ts` file inside `transport` folder. 

## Steps to Test

yarn test:ci:integration


packages/sources/finage-test/src/endpoint/crypto-router.ts
```diff
@@ -1,13 +1,13 @@
 import { CryptoPriceEndpoint } from '@chainlink/external-adapter-framework/adapter'
```
amit-momin, line 1: Think we just need to remove `-router` from `finage-test` endpoints then this PR is good to go

karen-stepanyan: done

amit-momin (approved):

PR #2853, opened 2023-06-21
Allow skipping batching eth-beacon and stader-balance

amit-momin:
## Description

NOPs running local consensus clients (confirmed for Lighthouse as of now) do not need to batch validators. Added a way to configure the adapter to skip batching by setting `BATCH_SIZE=0` in env vars. This will help reduce the number of requests to the client and help mitigate OOM errors.

## Changes

- Updated logic to skip batching when `BATCH_SIZE=0`

## Steps to Test

1. yarn test packages/sources/eth-beacon/test/integration
2. yarn test packages/sources/stader-balance/test/integration

## Quality Assurance

[PDI-2218]: https://smartcontract-it.atlassian.net/browse/PDI-2218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/eth-beacon/src/endpoint/balance.ts
```diff
@@ -114,23 +113,26 @@ const queryInBatches = async (
 ) => {
   const url = `/eth/v1/beacon/states/${params.stateId}/validators`
   const statusList = params.validatorStatus?.join(',')
-  const batchSize = Number(config.adapterSpecificParams?.batchSize) || DEFAULT_BATCH_SIZE
+  const batchSize = Number(config.adapterSpecificParams?.batchSize)
+  const groupSize = Number(config.adapterSpecificParams?.groupSize)
```
alecgard, line 117: This looked odd to me as `Number(undefined)` = `NaN`, however it looks like these will actually always be defined, it's just our `Config` type isn't specific enough to be accurate.

This could be improved by adding an `EthBeaconConfig` type to `config.ts`, which more accurately represents the config, which can then be referenced throughout the eth-beacon EA. E.g.
```
export type EthBeaconConfig = Config & {
  adapterSpecificParams: {
    beaconRpcUrl: string,
    executionRpcUrl: string,
    batchSize: number,
    groupSize: number,
    chainId: number
  }
}
```

Feel free to take it or leave it, not required but would better enforce our types and avoid needing to have unnecessary undefined handling!

alecgard (approved):

packages/sources/eth-beacon/src/endpoint/balance.ts
amit-momin: I'll take it! The plan is to eventually move these adapters to v3 though which provider stricter typing for this but think it's worth making this change in the meantime

alejoberardino (approved):

alecgard (approved):

PR #2871, opened 2023-07-12
Fix Finnhub EA's WebSocket transports

alecgard:
## Description

* Fixes an issue where Finnhub EA's would fail for certain requests when using WebSockets.
* Previously Finnhub WebSocket requests would succeed for full symbols (e.g. {"base": "OANDA:EUR_USD"}), but fail for requests with separate base and quote (e.g. {"base": "EUR", "quote: "USD"}). This is because the WebSocket message returns a single symbol which is cached as the base, and future requests included a quote so did not match the cached key.
* This commit fixes the above by introducing a requestTransform for the endpoint. The requestTransform checks if the symbol is a full symbol (e.g. OANDA:EUR_USD), and if so discards any quote provided as only the base will be used anyway. This keeps the cache key consistent when writing and reading for both REST and WS. The full symbol is expected to be provided, either via the request or derived from adapter/provided overrides. 

## Changes

* Added requestHandler to the Finnhub Quote endpoint, to discard the quote if the full symbol is provided.
* Added new tests to catch the previous issue

## Steps to Test

Tested against both `finnhub` and `finnhub-secondary`, for `WS_ENABLED` = `true` and `false`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "GBP", "to": "USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "GBP" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "OANDA:GBP_USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "FHFX:GBP-USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "AAPL" } }'`

## Quality Assurance

### DIFF
--- a/.changeset/clever-mails-tap.md
+++ b/.changeset/clever-mails-tap.md
@@ -1,0 +1,12 @@
+---
+'@chainlink/finnhub-adapter': minor
+'@chainlink/finnhub-secondary-adapter': minor
+---
+
+Fix Finnhub EA's WebSocket transports
+
+Fixes an issue where Finnhub EA's would fail for certain requests when using WebSockets.
+Previously Finnhub WebSocket requests would succeed for full symbols (e.g. {"base": "OANDA:EUR_USD"}), but fail for requests with separate base and quote (e.g. {"base": "EUR", "quote: "USD"}). This is because the WebSocket message returns a single symbol which is cached as the
+base, and future requests included a quote so did not match the cached key.
+This commit fixes the above by introducing a requestTransform for the endpoint. The requestTransform checks if the symbol is a full symbol (e.g. OANDA:EUR_USD), and if so discards any quote provided as only the base will be used anyway. This keeps the cache key consistent when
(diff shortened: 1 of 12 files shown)


alejoberardino:
A change like this can make the logic that handles inverse flags fail, we should check that before moving forward

alecgard:
> A change like this can make the logic that handles inverse flags fail, we should check that before moving forward

Good call - I'll look into this

PR #2872, opened 2023-07-13
Finnhub WS: Maintain reverse symbol mapping.

alecgard:
## Description
- Fixes an issue where Finnhub EA's would fail for certain requests when using WebSockets.
- Previously Finnhub WebSocket requests would succeed for full symbols (e.g. {"base": "OANDA:EUR_USD"}), but fail for requests with separate base and quote (e.g. {"base": "EUR", "quote: "USD"}). This is because the WebSocket message returns a single symbol which is cached as the base, and future requests included a quote so did not match the cached key.
- This commit fixes the above issue by using WebsocketReverseMappingTransport to be able to retrieve the original params once the WebSocket response is returned. This allows for a consistent cache key between requests and responses.

## Changes

- Updated `finnhub` WebSocket transport to use `WebsocketReverseMappingTransport`. On WS subscription we store the symbol and request params, and retrieve them when we receive a message. 
- Added testing to cover the original failing case. 

## Steps to Test

Tested against both `finnhub` and `finnhub-secondary`, for `WS_ENABLED` = `true` and `false`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "GBP", "to": "USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "GBP" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "OANDA:GBP_USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "FHFX:GBP-USD" } }'`
* `curl --request POST \ --url http://localhost:8080/ \ --header 'Content-Type: application/json' \ --data '{ "id": "1", "data": { "from": "AAPL" } }'`

## Quality Assurance

### DIFF
--- a/packages/sources/finnhub/src/transport/quote-ws.ts
+++ b/packages/sources/finnhub/src/transport/quote-ws.ts
@@ -1,6 +1,6 @@
 import { BaseEndpointTypes } from '../endpoint/quote'
-import { WebSocketTransport } from '@chainlink/external-adapter-framework/transports'
-import { makeLogger } from '@chainlink/external-adapter-framework/util'
+import { WebsocketReverseMappingTransport } from '@chainlink/external-adapter-framework/transports'
+import { ProviderResult, makeLogger } from '@chainlink/external-adapter-framework/util'
 
 const logger = makeLogger('Finnhub quote endpoint WS')
 
@@ -28,7 +28,7 @@ type WsEndpointTypes = BaseEndpointTypes & {
   }
 }
 
-export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
+export const wsTransport = new WebsocketReverseMappingTransport<WsEndpointTypes, string>({
   url: ({ adapterSettings }) =>
     `${adapterSettings.WS_API_ENDPOINT}?token=${adapterSettings.API_KEY}`,
   handlers: {
@@ -40,18 +40,30 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
 
       if (message.type === 'trade') {
         const trades = message.data
-        return trades.map(({ s, p, t }) => ({
-          params: { base: s },
-          response: {
-            result: p,
-            data: {
+        const results: ProviderResult<WsEndpointTypes>[] = []
+
+        trades.forEach(({ s, p, t }) => {
+          const params = wsTransport.getReverseMapping(s)
+
+          if (!params) {
+            return
+          }
+
+          results.push({
+            params,
+            response: {
               result: p,
+              data: {
+                result: p,
+              },
+              timestamps: {
+                providerIndicatedTimeUnixMs: t,
+              },
             },
-            timestamps: {
-              providerIndicatedTimeUnixMs: t,
-            },
-          },
-        }))
+          })
+        })
+
+        return results
       }
 
       return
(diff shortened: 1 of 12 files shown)


packages/sources/finnhub/src/transport/quote-ws.ts
```diff
@@ -40,29 +40,44 @@ export const wsTransport = new WebSocketTransport<WsEndpointTypes>({
...
       }
 
       return
     },
   },
   builders: {
     subscribeMessage: (params) => {
-      return { type: 'subscribe', symbol: `${params.base}`.toUpperCase() }
+      const symbol = params.base.toUpperCase()
+      wsTransport.setReverseMapping(symbol, params)
```
alejoberardino, line 75: This is only using the base, which is fine for cases where the entire symbol is sent there but will cause collisions for requests that have base and quote separate I think.
As a separate comment, the reverse mapping transport should be avoided unless absolutely necessary, for cases where symbols are glued together and it's not possible to deterministically deconstruct them

alecgard: This EA is currently a bit strange and only respects the `base` - we always ignore the passed in `quote`. This looks to be because this EA heavily relies on overrides, which only support base. 

What I'd like to do is to split out the base, quote and exchange to separate input parameters and construct the symbol at time of request. We should still be able to support passing in full symbols with a bit of pattern matching. This probably will require splitting out Forex from the current single Quote endpoint which supports Forex, Equities and Crypto, so we can accurately pattern match. 

Seeing as this EA only uses `base` today, do you have any concerns with this fix as an intermediate step?

> the reverse mapping transport should be avoided unless absolutely necessary

Is this mainly due to risk of collisions? I have a PR up for the framework (https://github.com/smartcontractkit/ea-framework-js/pull/204) to pass the adapter context into the subscribe/unsubscribe handlers, which would allow us to support the symbol creation and deconstruction there rather than need to persist the mapping.

alejoberardino: Gotcha, this would've worked then if we only ever use the base. That being said, it makes the use of the mapping unnecessary then; the mapping is only necessary for those ambiguous cases but it looks like the problem here is not that we can't match the incoming data with the inbound requests, but rather that the EA requests themselves are not standardized. A better fix for those cases imo is to add a request transform to normalize the input to the adapter.

That is the main reason why I am against the reverse mapping ws transport existing, it was created as a workaround but every case I've seen so far usually has a more "proper" solution to it.

The FW PR is approved nonetheless, I think it's a good change regardless.

alecgard: Agree that standardising the requests would be a better fix. I can add a requestTransform that transforms the data into the separate params (base, quote and exchange), and we can derive the same from the WS message response based on the format. 

To be able to derive the default exchange for Finnhub, we need to be able to read the adapter settings in the `requestTransform`. I've put up a PR to add this to the framework: https://github.com/smartcontractkit/ea-framework-js/pull/206

alecgard: Thanks for your suggestion - I've put up a separate PR to standardise the inbound and DP response requests, by having a request transform that normalises the input into the separate params: https://github.com/smartcontractkit/external-adapters-js/pull/2873

Closing this PR in favour of that one.

alecgard:
Closing this PR in favour of https://github.com/smartcontractkit/external-adapters-js/pull/2873

PR #2888, opened 2023-07-23
CVI cache smoothing

vladi-coti:
## Description

Changing the way CVI is smoothed

## Changes

- Changed smoothing from using on chain last value to using cached value
- Changed some default
- Updated README


mxiao-cll:
Please re-open if still relevant

PR #2907, opened 2023-08-08
Upgrade wbtc-address-set to v3 framework

karen-stepanyan:
## Description
Upgraded `wbtc-address-set` to the v3 framework

## Steps to Test

yarn test packages/sources/wbtc-address-set


alejoberardino (dismissed):

amit-momin (approved):

packages/sources/wbtc-address-set/src/index.ts
```diff
@@ -1,10 +1,13 @@
...
+import { config } from './config'
+import { addresses, members } from './endpoint'
 
-const adapterContext = { name: NAME }
+export const adapter = new PoRAdapter({
+  defaultEndpoint: addresses.name,
+  name: 'WBTC',
+  config,
+  endpoints: [addresses, members],
+})
```
amit-momin, line 11: Since this adapter uses HttpTransports, we should set a reasonable limit. @alejoberardino What do you think about 10 req/s? PoR feeds don't update often so we can even go 1 req/s
```suggestion
export const adapter = new PoRAdapter({
  defaultEndpoint: addresses.name,
  name: 'WBTC',
  config,
  endpoints: [addresses, members],
  rateLimiting: {
    tiers: {
      default: {
        rateLimit1s: 10,
        note: 'Considered unlimited tier, but setting reasonable limits',
      },
    },
  },
})
```

alejoberardino: I think we can go even lower haha, like once every 10s

amit-momin (approved):

PR #2908, opened 2023-08-08
Upgrade eth-beacon to v3 framework

karen-stepanyan:
## Description
Upgraded `eth-beacon` to the v3 framework

## Steps to Test

yarn test packages/sources/eth-beacon


packages/sources/eth-beacon/src/config/index.ts
```diff
@@ -1,54 +1,32 @@
...
+  ETH_CONSENSUS_RPC_URL: {
+    description: 'RPC URL of an Ethereum consensus client (beacon node)',
+    type: 'string',
+    required: true,
+  },
+  ETH_EXECUTION_RPC_URL: {
+    description:
+      'RPC URL of an Ethereum execution client (archive node). Required for requests that need a limbo validator search',
+    type: 'string',
+    default: '',
```
amit-momin, line 13: I think we can remove this default. It's ok if the value for the env var is undefined. We only had the `|| ''` in the v2 adapter to force the type for `adapterSpecificParams`. Luckily in v3 we don't have to anymore

amit-momin (changes requested):

packages/sources/eth-beacon/src/config/index.ts
```diff
@@ -1,54 +1,32 @@
...
+      'Number of requests to execute asynchronously before the adapter waits to execute the next group of requests. Setting this lower than the default may result in lower performance from the adapter. Unused if BATCH_SIZE is set to 0.',
+    type: 'number',
+    default: 15,
+  },
+  CHAIN_ID: {
+    description: 'The chain id to connect to',
+    type: 'number',
+    default: 1,
+  },
+})
```
amit-momin, line 32: We'll want to include an envDefaultOverride for `API_TIMEOUT: 60000` here too

packages/sources/eth-beacon/src/config/index.ts
karen-stepanyan: Removing the default causes type issues in proof of reserves EA. balance.ts file imports eth-beacon EA

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -0,0 +1,360 @@
...
+    if (batchSize === 0) {
+      const addresses = params.addresses.map(({ address }) => address).join(',')
+      responses.push(
+        await this.queryBeaconChain(
+          adapterSettings,
+          req.requestContext.endpointName,
+          url,
+          addresses,
+          statusList,
+        ),
```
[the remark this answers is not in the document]
karen-stepanyan: I think same applies to adapter config as well. It is also used in different places

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -53,43 +54,70 @@ export type Address = {
...
       const addresses = params.addresses.map(({ address }) => address).join(',')
       responses.push(
-        await this.queryBeaconChain(
-          adapterSettings,
-          req.requestContext.endpointName,
-          url,
-          addresses,
-          statusList,
-        ),
+        await this.queryBeaconChain(this.config, this.endpointName, url, addresses, statusList),
```
amit-momin, line 120: Could we just use `this.config` and `this.endpointName` directly in `queryBeaconChain`? It'll clear up the parameter clutter here

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -149,11 +177,7 @@ export class BalanceTransport implements Transport<BalanceTransportTypes> {
     // Otherwise, record 0 for the balance of missing validators
     if (params.searchLimboValidators) {
       balances.push(
-        ...(await this.searchLimboValidators(
-          adapterSettings,
-          unfoundValidators,
-          req.requestContext.endpointName,
-        )),
+        ...(await this.searchLimboValidators(this.config, unfoundValidators, this.endpointName)),
```
amit-momin, line 180: Same here

packages/sources/eth-beacon/src/transport/balance.ts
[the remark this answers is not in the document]
amit-momin: Resolving this comment since it's not necessarily needed for the adapter. With the latest changes it'd only help reduce the clutter in the initialize method

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -0,0 +1,376 @@
...
+    })
+
+    // Returns map of validators found in limbo with balances in wei
+    const limboBalances = await this.fetchLimboEthBalances(
+      limboAddressMap,
+      String(config.ETH_CONSENSUS_RPC_URL),
+      this.provider,
+      config,
+      endpointName,
+    )
```
amit-momin, line 261: Also here

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -0,0 +1,343 @@
...
+  private async searchLimboValidators(unfoundValidators: Address[]): Promise<BalanceResponse[]> {
+    const balances: BalanceResponse[] = []
+    const limboAddressMap: Record<string, Address> = {}
+    // Parse unfound validators into a map for easier access in limbo search
+    unfoundValidators.forEach((validator) => {
+      limboAddressMap[validator.address.toLowerCase()] = validator
+    })
+
+    // Returns map of validators found in limbo with balances in wei
+    const limboBalances = await this.fetchLimboEthBalances(limboAddressMap, this.provider)
```
amit-momin, line 239: I think we can also remove `this.provider` from here and use it directly in `fetchLimboEthBalances`

amit-momin (approved):

PR #2916, opened 2023-08-09
Upgrade avalanche-platform to v3 framework

karen-stepanyan:
## Description
Upgraded `avalanche-platform` to the v3 framework

## Steps to Test

yarn test packages/sources/avalanche-platform


alejoberardino (approved):

packages/sources/avalanche-platform/src/index.ts
```diff
@@ -1,10 +1,13 @@
...
+import { config } from './config'
+import { balance } from './endpoint'
 
-const adapterContext = { name: NAME, envDefaultOverrides }
+export const adapter = new PoRAdapter({
+  defaultEndpoint: balance.name,
+  name: 'AVALANCHE_PLATFORM',
+  config,
+  endpoints: [balance],
+})
```
amit-momin, line 11: Since this adapter uses an HttpTransport, we'll want to add a reasonable limit here. I believe this RPC call is a heavier call on the node so we can set it conservatively at 10 req/s.
```suggestion
export const adapter = new PoRAdapter({
  defaultEndpoint: balance.name,
  name: 'AVALANCHE_PLATFORM',
  config,
  endpoints: [balance],
  rateLimiting: {
    tiers: {
      default: {
        rateLimit1s: 10
      },
    },
  },
})
```

amit-momin (changes requested):

amit-momin (approved):

PR #2918, opened 2023-08-09
Set eth-beacon BACKGROUND_EXECUTE_TIMEOUT default to max

amit-momin:
Set `eth-beacon` v3 env default override for `BACKGROUND_EXECUTE_TIMEOUT` to the max 180s since PoR requests take longer to process

### DIFF
--- a/.changeset/curvy-pets-cheat.md
+++ b/.changeset/curvy-pets-cheat.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/eth-beacon-adapter': patch
+---
+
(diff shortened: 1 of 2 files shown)


karen-stepanyan:
Hey @amit-momin . I think all PoR adapters (regardless if they use background execute or not) [use 180s](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/adapter/por.ts#L137) as a default value.

amit-momin:
You're right! Forgot the env default override is included in the PorAdapter subclass we're using

PR #2921, opened 2023-08-10
Upgrade lotus to v3 framework

karen-stepanyan:
## Description
Upgraded `lotus` to the v3 framework

## Steps to Test

yarn test packages/sources/lotus


packages/sources/lotus/src/config/index.ts
```diff
@@ -1,12 +1,22 @@
...
+    description:
+      'Your Lotus node [API key/token](https://docs.filecoin.io/build/lotus/api-tokens/#obtaining-tokens)',
+    type: 'string',
+    required: true,
+    sensitive: true,
+  },
+  FILECOIN_RPC_URL: {
+    description: 'RPC URL of Filecoin node',
+    type: 'string',
+    default: 'http://localhost:8545',
```
amit-momin, line 14: I think I see the part of the `json-rpc` adapter this was taken from but I don't think it carries over in the config `lotus` is using. Think we should just mark this as required and remove the default

packages/sources/lotus/src/transport/balance.ts
```diff
@@ -0,0 +1,141 @@
...
+    dependencies: TransportDependencies<BalanceTransportTypes>,
+    _adapterSettings: BalanceTransportTypes['Settings'],
+    _endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    await super.initialize(dependencies, _adapterSettings, _endpointName, transportName)
+    this.requester = dependencies.requester
+    this.settings = _adapterSettings
+    this.endpointName = _endpointName
+  }
```
amit-momin, line 39: Think we can skip on setting the requester here and also remove the _ from the property names since we're using them
```suggestion
export class TotalBalanceTransport extends SubscriptionTransport<BalanceTransportTypes> {
  settings!: BalanceTransportTypes['Settings']
  endpointName!: string

  async initialize(
    dependencies: TransportDependencies<BalanceTransportTypes>,
    adapterSettings: BalanceTransportTypes['Settings'],
    endpointName: string,
    transportName: string,
  ): Promise<void> {
    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
    this.settings = adapterSettings
    this.endpointName = endpointName
  }
```

packages/sources/lotus/src/transport/balance.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  async backgroundHandler(
+    context: EndpointContext<BalanceTransportTypes>,
+    entries: RequestParams[],
+  ) {
+    if (!entries.length) {
+      await sleep(context.adapterSettings.BACKGROUND_EXECUTE_MS)
+      return
+    }
+    await Promise.all(entries.map(async (param) => this.handleRequest(param)))
```
amit-momin, line 49: Like the other adapters, I think we'll want to sleep after all bg cycles
```suggestion
    await Promise.all(entries.map(async (param) => this.handleRequest(param)))
    await sleep(context.adapterSettings.BACKGROUND_EXECUTE_MS)
```

packages/sources/lotus/src/index.ts
```diff
@@ -1,10 +1,13 @@
...
+import { balance } from './endpoint'
+import { PoRAdapter } from '@chainlink/external-adapter-framework/adapter/por'
 
-const adapterContext = { name: NAME }
+export const adapter = new PoRAdapter({
+  defaultEndpoint: balance.name,
+  name: 'LOTUS',
+  config,
+  endpoints: [balance],
+})
```
amit-momin, line 11: Realized we may want a custom input validation here that returns a 400 error if the `addresses` input param array is of length 0. Looks like we have that validation in v2

packages/sources/lotus/src/transport/balance.ts
```diff
@@ -0,0 +1,141 @@
...
+    const addresses = param.addresses
+
+    const providerDataRequestedUnixMs = Date.now()
+
+    const balances = await Promise.all(
+      addresses.map((addr, index) => this.getBalance(addr.address, index)),
+    )
+
+    const result = balances
+      .reduce((sum, balance) => sum.add(balance.result as unknown as BigNumber), BigNumber.from(0))
```
amit-momin, line 83: We can maybe just initialized the BigNumber here to make this cleaner. If that gives you an error, you may need to turn `balance.result` into a string first
```suggestion
      .reduce((sum, balance) => sum.add(BigNumber(balance.result)), BigNumber.from(0))
```

amit-momin (changes requested):

packages/sources/lotus/src/index.ts
karen-stepanyan: Agree, but i think this should be handled by the framework.  I'm not sure if [here](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/validation/input-params.ts#L327) we need to check if the `length >=0`, instead of `length > 0`

packages/sources/lotus/src/transport/balance.ts
karen-stepanyan: The requester is actually used in this one.

amit-momin: Ah you're right, was thinking the transport already exposed it but was mixing it up with response cache

packages/sources/lotus/src/index.ts
amit-momin: We may not want to add it to the framework in case an empty array is a valid input like in the case of `validatorStatus` in `eth-beacon`. Unless you were thinking about adding it just for `PoRBalanceEndpoint` and `PoRTotalBalanceEndpoint` specifically for the `addresses` input param. Otherwise this validation is specific to only a few adapters in which case we'd want `length > 0` ([example](https://github.com/smartcontractkit/external-adapters-js/blob/39d4cd3f1dc6072d082856da05cee5f3ecd9245c/packages/sources/lotus/src/endpoint/balance.ts#L49-L55)).

karen-stepanyan: Oh i see. Forgot the we have such case (empty array as a valid input param). I thought about having `customInputValidation` function in PoR adapters that will automatically check this condition but it will overwrite if such function exists on endpoint, so i think for now it would be better just to add  `customInputValidation` to every por balance EA.

amit-momin: Sounds good to me 👍

PR #2922, opened 2023-08-10
fix response providerDataRequestedUnixMs for eth-beacon

karen-stepanyan:
There was a bug from migration where the `providerDataRequestedUnixMs` and `providerDataReceivedUnixMs` had the same value


amit-momin (approved):
Good catch!

PR #2937, opened 2023-08-18
add validation for addresses input param (polkadot-balance, avalanche-platform, eth-beacon)

karen-stepanyan:
Added `customInputValidation` for `polkadot-balance`, `avalanche-platform` and `eth-beacon` that checks and throws an error when `addresses` input array is empty.

### DIFF
--- a/packages/sources/polkadot-balance/test/integration/adapter.test.ts
+++ b/packages/sources/polkadot-balance/test/integration/adapter.test.ts
@@ -134,5 +134,6 @@ describe('Balance Endpoint', () => {
   it('should return error (empty addresses)', async () => {
     const response = await testAdapter.request({ addresses: [] })
     expect(response.statusCode).toEqual(400)
+    expect(response.json()).toMatchSnapshot()
   })
 })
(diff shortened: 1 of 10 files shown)


packages/sources/polkadot-balance/test/integration/adapter.test.ts
```diff
@@ -130,4 +130,9 @@ describe('Balance Endpoint', () => {
     const response = await testAdapter.request({})
     expect(response.statusCode).toEqual(400)
   })
+
+  it('should return error (empty addresses)', async () => {
+    const response = await testAdapter.request({ addresses: [] })
+    expect(response.statusCode).toEqual(400)
+  })
```
amit-momin, line 137: I noticed it in the other two but could you add the snapshot check here too
```suggestion
  it('should return error (empty addresses)', async () => {
    const response = await testAdapter.request({ addresses: [] })
    expect(response.statusCode).toEqual(400)
    expect(response.json()).toMatchSnapshot()
  })
```

amit-momin (changes requested):

amit-momin (approved):

PR #2939, opened 2023-08-21
upgrade por-indexer to v3 framework

karen-stepanyan:
Removed por-indexer composite EA
Added new por-indexer source EA with already upgraded v3 version. 
Added `ion.au` to main tsconfig files as readme generation was failing because of it.

## Steps to Test

yarn test packages/sources/por-indexer


packages/sources/por-indexer/src/endpoint/balance.ts
```diff
@@ -0,0 +1,108 @@
...
+    )
+    const btcTestnet = addresses.some(
+      (a) => `${a.network}_${a.chainId}`.toUpperCase() === 'BITCOIN_TESTNET',
+    )
+    const dogeMainnet = addresses.some(
+      (a) => `${a.network}_${a.chainId}`.toUpperCase() === 'DOGECOIN_MAINNET',
+    )
+    const dogeTestnet = addresses.some(
+      (a) => `${a.network}_${a.chainId}`.toUpperCase() === 'DOGECOIN_TESTNET',
+    )
```
amit-momin, line 80: I think in reality most of these requests if not all of them just have one network/chainId per request. So 3 of these loops would be going through the address set without finding anything. Could we change this to find all of the unique network/chainId combos with just a single pass through the address set? I think address sets can get pretty large so it might make a difference especially since this runs on every request.

amit-momin (changes requested):

PR #2975, opened 2023-09-07
Include the network in L2 Sequencer Health logs

alecgard:
* Currently debugging can be challenging if an EA is checking the health for multiple networks, as you can't tell which network a log line is for. This commit adds a prefix to all logs to include the network, for easier debugging.

### DIFF
--- a/packages/sources/layer2-sequencer-health/src/adapter.ts
+++ b/packages/sources/layer2-sequencer-health/src/adapter.ts
@@ -34,7 +34,7 @@ export const makeExecute: ExecuteFactory<ExtendedConfig, endpoints.TInputParamet
   // Disclaimer on startup
   Object.keys(HEALTH_ENDPOINTS).forEach((network) => {
     if (!HEALTH_ENDPOINTS[network as Networks]?.endpoint)
-      Logger.info(`Health endpoint not available for network: ${network}`)
+      Logger.info(`[${network}] Health endpoint not available for network: ${network}`)
   })
   return async (request, context) => execute(request, context, config || makeConfig())
 }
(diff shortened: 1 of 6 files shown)


packages/sources/layer2-sequencer-health/src/adapter.ts
```diff
@@ -34,7 +34,7 @@ export const makeExecute: ExecuteFactory<ExtendedConfig, endpoints.TInputParamet
   // Disclaimer on startup
   Object.keys(HEALTH_ENDPOINTS).forEach((network) => {
     if (!HEALTH_ENDPOINTS[network as Networks]?.endpoint)
-      Logger.info(`Health endpoint not available for network: ${network}`)
+      Logger.info(`[${network}] Health endpoint not available for network: ${network}`)
```
mmcallister-cll, line 37: we're always guaranteed to have a valid `network` here, right?

packages/sources/layer2-sequencer-health/src/network.ts
```diff
@@ -49,7 +49,7 @@ export const checkSequencerHealth: NetworkHealthCheck = async (
   })
   const isHealthy = !!Requester.getResult(response.data, HEALTH_ENDPOINTS[network]?.responsePath)
   Logger.info(
-    `Health endpoint for network ${network} returned a ${
+    `[${network}] Health endpoint for network ${network} returned a ${
```
mmcallister-cll, line 52: is there a difference with `[${network}]` vs `${network}` or is this just for formatting reasons?

mmcallister-cll (approved):

packages/sources/layer2-sequencer-health/src/adapter.ts
alecgard: Yep! This uses `HEALTH_ENDPOINTS`, which is configured in the EA so we know it has networks. 

For the other logs, we use the input param `network` which is a required param.

packages/sources/layer2-sequencer-health/src/network.ts
alecgard: Just formatting, figured the prefix `[...]` will make it easier to discern the different logs when grepping through.

PR #2989, opened 2023-09-13
Disable validation on select envars for wBTC address set EA

alejoberardino:


packages/sources/wbtc-address-set/src/index.ts
```diff
@@ -18,4 +19,10 @@ export const adapter = new PoRAdapter({
   },
 })
 
+// Remove validations for these specific env vars.
+// This is a very ugly workaround, and might not work with slightly breaking changes in the framework.
```
alecgard, line 23: > ... might not work with slightly breaking changes in the framework

Optional: We could we add a logger.warning if these two configs aren't `undefined`, to give us a heads up if the framework changes.

alecgard (approved):

amit-momin (approved):

packages/sources/wbtc-address-set/src/index.ts
alejoberardino: Added this warning 👍

amit-momin (approved):

alecgard (approved):

PR #3006, opened 2023-09-21
Upgrade dydx from v2 to v3 framework

karen-stepanyan:
[PDI-2256](https://smartcontract-it.atlassian.net/browse/PDI-2256)
## Description

According to the ticket above dYdX EA was not working with v2 version of the framework. Apparently the issue was with input parameters. This PR migrates dYdX to v3 version of the framework and also introduces some breaking changes. 
......

## Changes

- Migrated EA to v3 version of the framework. Main logic is untouched, it's the same as in v2 and even prior to that. 
- In the logic, deprecated JS functions were replaced  
- **[Breaking]** Input parameters are changed. The initial version of the EA  had following input parameters definition 
`const customParams = {
  dataPath: false,
  result: false,
  asset: true
}
`
`dataPath` was used to identify where to look for the price if not in `result`. 
With this PR, input params are  `result` and `asset`. The reason why `dataPath` is removed is because 
a. v3 framework doesn't support dynamic input params. Technically it might be possible with `requestTransform` from v3 framework but there is a tradeoff, also considering the next point. 
b. checking the job specs it seems that the configuration uses only the `result` path anyway. 
- **[Breaking]** The `result` input parameter in older versions of the EA was configured to be of _any_ type. For the same two reasons as the point above, the `result` input param is changed to be _number_ only

## Steps to Test

`yarn test packages/targets/dydx-stark`

## Quality Assurance

[PDI-2256]: https://smartcontract-it.atlassian.net/browse/PDI-2256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/targets/dydx-stark/src/endpoint/send.ts
```diff
@@ -1,63 +1,47 @@
...
+    asset: {
+      description: 'Required asset name (of your choice, per asset. for example "BTCUSD")',
+      type: 'string',
+      required: true,
+    },
+    result: {
+      aliases: ['price'],
+      required: true,
+      type: 'number',
+      description: 'Price data that will  be sent to dydx',
```
amit-momin, line 17: nit: 
```suggestion
      description: 'Price data that will be sent to dydx',
```

packages/targets/dydx-stark/src/transport/send.ts
```diff
@@ -0,0 +1,96 @@
...
+  async initialize(
+    dependencies: TransportDependencies<DyDxTransportTypes>,
+    _adapterSettings: DyDxTransportTypes['Settings'],
+    _endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    this.responseCache = dependencies.responseCache
+    this.requester = dependencies.requester
+    this.name = transportName
+  }
```
amit-momin, line 36: I think `name` and `responseCache` should already be accessible. We should also initialize here.
```suggestion
  requester!: Requester

  async initialize(
    dependencies: TransportDependencies<DyDxTransportTypes>,
    _adapterSettings: DyDxTransportTypes['Settings'],
    _endpointName: string,
    transportName: string,
  ): Promise<void> {
    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
    this.requester = dependencies.requester
  }
```

packages/targets/dydx-stark/src/transport/send.ts
```diff
@@ -0,0 +1,96 @@
...
+      requestConfig,
+    )
+
+    const response = {
+      data: {
+        result: providerResponse.response.data,
+      },
+      statusCode: 200,
+      // see the endpoint file for explanation why the type cast is needed
+      result: providerResponse.response.data as unknown as string,
```
amit-momin, line 84: Since the expected response is `{ market: string; price: string }`. Should we return the price here?
```suggestion
      result: providerResponse.response.data.price,
```

amit-momin (changes requested):

packages/targets/dydx-stark/src/transport/send.ts
karen-stepanyan: The class is not inheriting from other class. It just implements the `Transport` interface. The main reason for this was to use `foregroundExecute` since this EA posts data to DP. Not sure if it makes sense to convert it to backgroundExecute, subscription based transport.

packages/targets/dydx-stark/src/transport/send.ts
karen-stepanyan: I wanted to keep the same logic here. The current EA also returns the whole response. Furthermore i'm not sure if `{ market: string; price: string }` is the right or complete response. I took it from current readme's output example (i wasn't able to actually get successful response).

amit-momin: Normally we just return null for complex results like in PoR but not sure if this field is used. I think it'd conflict with v3 expectations though so we should keep the full response in the `data` field but just the price in the external `result`. If the purpose of the adapter is to post the answer, the response from the adapter may not be too important other than the status code. Also I think we can trust the README for now since we're sort of going into this migration blind

packages/targets/dydx-stark/src/transport/send.ts
amit-momin: Ah sorry, I definitely skimmed over that part. This is good then

amit-momin (approved):

PR #3008, opened 2023-09-22
include root package in framework version-bump script

karen-stepanyan:
Updated `@chainlink/external-adapter-framework` dev dependency in the root package.json to the latest version.

Updated framework version-bump script to include target adapters and root package.json


alecgard (dismissed):
👍 

Would be great if we could update https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/scripts/src/framework/version-bump.sh to include this automatically.

alecgard:
Thanks for updating the script! 👍

alecgard (approved):

amit-momin (approved):

PR #3088, opened 2023-11-14
Update test-payload.json for EAs. pt 1

karen-stepanyan:
Part 1 of [DF-18523](https://smartcontract-it.atlassian.net/browse/DF-18523)

Although this is just the first part, there are lot of file changes, but changes are repetitive. In particular each EA folder has 3 changes 
1.  new entries are added into `test-payload.json`. These new entries are mostly the same payloads but with endpoint names and aliases. This makes sure that for every endpoint+aliases we have configuration for soak testing. 
2. new unit test is added (either in existing file or in a new file) that ensures that test-payload.json always contains requests with all endpoints+aliases. If new endpoint is added/updated and test-paylaod is not updated, unit test will fail
3. tsconfig.test.json is updated to include test-payload.json so that the test file can import test-payload.json

[DF-18523]: https://smartcontract-it.atlassian.net/browse/DF-18523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


austinborn:
The code itself looks great, it's clear you were careful with the set of endpoints and payloads for each adapter. Would you be able to resolve the conflicts, then let the soak test re-run and I can ✅  after that?

karen-stepanyan:
hey @austinborn . soak testing isn't even running as there are too many changed adapters and Deploy phase is timing out.

alecgard (approved):

mxiao-cll:
Felt there is a lot of repetitive code here, not sure if there is a way to extract this and have this test automatically run on each adapter

karen-stepanyan:
See the ticket in the description

PR #3092, opened 2023-11-15
Update test-payload.json for EAs. pt 2

karen-stepanyan:
Part 2 of [DF-18523](https://smartcontract-it.atlassian.net/browse/DF-18523)

Every EA folder has 3 changes

1. new entries are added into test-payload.json. These new entries are mostly the same payloads but with endpoint names and aliases. This makes sure that for every endpoint+aliases we have configuration for soak testing.
2. new unit test is added (either in existing file or in a new file) that ensures that test-payload.json always contains requests with all endpoints+aliases. If new endpoint is added/updated and test-paylaod is not updated, unit test will fail
3. tsconfig.test.json is updated to include test-payload.json so that the test file can import test-payload.json

Some EAs are not updated
 EAs that are in soak testing blacklist
 EAs that don't have test-payload.json (mostly composites and targets)

[DF-18523]: https://smartcontract-it.atlassian.net/browse/DF-18523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alecgard (approved):

karen-stepanyan:
See the ticket in the description

PR #3101, opened 2023-11-24
Changed ncfx forex WS auth flow

karen-stepanyan:
[DF-19295](https://smartcontract-it.atlassian.net/browse/DF-19295)

1. Changed forex WS default endpoint.
2. Changed forex WS auth flow. We used to generate the api key via encoding the username and password. Now the api key should be directly provided as env var.
3. Added `customInputValidation` for crypto and forex endpoints.

### DIFF
--- a/.changeset/violet-dogs-prove.md
+++ b/.changeset/violet-dogs-prove.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ncfx-adapter': minor
+---
+
+FOREX_WS_USERNAME and FOREX_WS_PASSWORD env vars replaced with FOREX_WS_API_KEY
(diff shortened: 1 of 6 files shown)


alecgard (approved):
Change looks good 👍 

Do we know if NCFX will support both endpoints & auth methods in parallel, or if we'll have a switchover moment where NOPs need to upgrade their EA?

karen-stepanyan:
hey @alecgard . according to the [comments](https://smartcontract-it.atlassian.net/browse/DF-19295?focusedCommentId=179440) in the ticket, NCFX will support both endpoints until all NOPs switch to the new EA

PR #3175, opened 2024-02-05
Fixed alongside integration tests

karen-stepanyan:
The integration test of Alongside EA somehow didn't have a mock for `infura-ipfs` api. The test was actually calling the real API and getting the data. Presumably the data was always the same therefore the test snapshots were not failing. 

However, in a different PR i noticed alongside integration tests failing and after checking it, it seems that the API that was actually running in the test is now throwing 401 errors. This PR adds the missing mocked response.


alecgard (approved):

alecgard:
Nice catch!

PR #3176, opened 2024-02-05
Add Coinbase Prime adapter

amit-momin:
## Description

Added an adapter to interface with the Coinbase Prime API.

## Changes

- Added `coinbase-prime` adapter

## Steps to Test

1. yarn test packages/sources/coinbase-prime/test

## Quality Assurance

- [X] The branch naming follows git flow (`feature/x`, `chore/x`, `release/x`, `hotfix/x`, `fix/x`) or is created from Jira.
- [X] This is related to a maximum of one Jira story or GitHub issue.
- [X] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types).

[PDI-2304]: https://smartcontract-it.atlassian.net/browse/PDI-2304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/coinbase-prime/src/transport/balance.ts
```diff
@@ -0,0 +1,101 @@
...
+      return params.map((param) => {
+        return {
+          params: param,
+          response: {
+            errorMessage: `The data provider did not return a proper response for Portfolio: ${param.portfolio}, Balance Type: ${param.type}, Symbol: ${param.symbol}`,
+            statusCode: 502,
+          },
+        }
+      })
+    }
```
austinborn, line 77: Would it be too much to split out these error cases? It may be useful if the errorMessage specifies `data.balances not present` or `data.balances.length is not 1`

austinborn (changes requested):

packages/sources/coinbase-prime/src/endpoint/balance.ts
```diff
@@ -0,0 +1,45 @@
...
+      options: ['total', 'vault', 'trading'],
+    },
+  },
+  [
+    {
+      portfolio: 'abcd1234-123a-1234-ab12-12a34bcd56e7',
+      symbol: 'BTC',
+      type: 'total',
+    },
+  ],
```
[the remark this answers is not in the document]
amit-momin: I believe it's been around for a bit. It's what's used to generate example requests for the README.

packages/sources/coinbase-prime/src/transport/balance.ts
amit-momin: Good call! I've updated the code in the latest commit. I also cleaned up all of the `params.map` loops and moved all of the logic into one. The way the code was set up previously doesn't really optimize anything since we're not batching.

austinborn: Cool, I'll approve then

packages/sources/coinbase-prime/src/transport/balance.ts
```diff
@@ -0,0 +1,119 @@
...
+            balance_type: `${param.type.toUpperCase()}_BALANCES`,
+          },
+        },
+      }
+    })
+  },
+  parseResponse: (params, response) => {
+    return params.map((param) => {
+      if (!response.data) {
+        return {
```
austinborn, line 68: Nice! Yeah this works

austinborn (approved):

.yarn/cache/@types-node-npm-16.11.51-02764a2cd3-b4193d3f9d.zip
mmcallister-cll, line 1: could probably update this to 16.11.68 like in the other adapters

packages/sources/coinbase-prime/src/endpoint/balance.ts
```diff
@@ -0,0 +1,45 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { config } from '../config'
+import { httpTransport } from '../transport/balance'
+
+export const inputParameters = new InputParameters(
```
mmcallister-cll, line 7: I guess post-v2 adapters we don't have any standard input params for balance endpoints?

packages/sources/coinbase-prime/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const adapter = new Adapter({
+  defaultEndpoint: balance.name,
+  name: 'COINBASE-PRIME',
+  config,
+  endpoints: [balance],
+  rateLimiting: {
+    tiers: {
+      default: {
+        rateLimit1s: 25,
```
mmcallister-cll, line 14: feels maybe a little overkill, any idea how often we'd expect updates on the coinbase side?

packages/sources/coinbase-prime/src/endpoint/balance.ts
```diff
@@ -0,0 +1,45 @@
...
+    symbol: {
+      required: true,
+      type: 'string',
+      description: 'The symbol to return the balance for',
+    },
+    type: {
+      type: 'string',
+      description: 'The balance type to return',
+      default: 'total',
+      options: ['total', 'vault', 'trading'],
```
mmcallister-cll, line 23: maybe not relevant, but if `TOTAL` all caps was passed in, would this fail on invalid option?

packages/sources/coinbase-prime/src/transport/balance.ts
```diff
@@ -0,0 +1,119 @@
...
+          response: {
+            errorMessage: `The data provider returned non-numeric balance: ${response.data.balances[0].amount}`,
+            statusCode: 502,
+          },
+        }
+      }
+      return {
+        params: param,
+        response: {
+          result: result,
```
mmcallister-cll, line 111: i'm unclear when we should be using SingleNumberResponse vs something like PoRBalanceResponse, any tips?

packages/sources/coinbase-prime/src/index.ts
amit-momin: It probably is for a single asset in a single portfolio but we usually configure whatever the data provider's doc mention. If for any reason in the future there are requests for other assets or portfolios, all of the requests would share this rate limit so this helps future-proof the adapter too.

packages/sources/coinbase-prime/src/endpoint/balance.ts
amit-momin: I believe options are case-sensitive so it would fail. Either GTM or EngOps usually test requests against the adapter before creating the job so a bad request should be caught ahead of time.

packages/sources/coinbase-prime/src/transport/balance.ts
amit-momin: The latter is what we'd use for a balance adapter part of the `proof-of-reserves` composite stack. Most of those PoR types were made for adapters used with the `proof-of-reserves` composite adapter. They help standardize the input and output the composite expects. With adapters like these that use a 3rd party API, we'll usually use `SingleNumberResponse` like a price feed adapter since they're standalone.

PR #3225, opened 2024-03-14
Added new Finalto EA

karen-stepanyan:
## Description

Added new websocket EA

## Steps to Test

yarn test packages/sources/finalto

## Quality Assurance

[DF-19492]: https://smartcontract-it.atlassian.net/browse/DF-19492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/finalto/src/transport/forex.ts
```diff
@@ -0,0 +1,161 @@
...
+      const value = subscriptionIdMap[symbol].toString()
+      delete subscriptionIdMap[symbol]
+      return value
+    }
+    return subscriptionIdMap[symbol].toString()
+  }
+  subscriptionIdMap[symbol] = subscriptionId
+  subscriptionId++
+  return subscriptionIdMap[symbol].toString()
+}
```
alecgard, line 41: Splitting this into two functions for getting and deleting might make it clearer what is going on - at first glance I wasn't sure what `deleteOnGet` meant

```suggestion
const getSubscriptionId = (symbol: string): string => {
  if (subscriptionIdMap[symbol]) {
    return subscriptionIdMap[symbol].toString()
  }
  subscriptionIdMap[symbol] = subscriptionId
  subscriptionId++
  return subscriptionIdMap[symbol].toString()
}

const deleteSubscriptionId = (symbol: string): string => {
      const value = subscriptionIdMap[symbol].toString()
      delete subscriptionIdMap[symbol]
      return value
}
```

packages/sources/finalto/src/transport/forex.ts
```diff
@@ -0,0 +1,161 @@
...
+            },
+          },
+        ]
+      },
+    },
+    builders: {
+      subscribeMessage: (params) => {
+        const symbol = `${params.base}${params.quote}`.toUpperCase()
+        wsTransport.setReverseMapping(symbol, params)
+        const id = getSubscriptionId(symbol)
```
alecgard, line 136: What happens if multiple EA's use the same API key, and attempt to use the same subscription ID? 

Could we instead make this a UUID rather than an incrementing integer? If not, some explicit error handling for the conflicting subscription ID case would be useful.

karen-stepanyan: @alecgard AFAIK each set of credentials have a unique WS address that allows only one connection, but i think you are right, that UUID might be better, as we won't think about collision.

alecgard: That makes sense, just seeing your comment about having a subscription id of `1` causing errors made me suspicious that the same credential may be being used in multiple places

packages/sources/finalto/src/transport/forex.ts
karen-stepanyan: Actually now i don't think we even need `deleteSubscriptionId` function. Updated the code so that we delete the key in unsubscribeMessage directly after getting it.

alecgard (approved):

PR #3252, opened 2024-04-02
Changed API method for eth-beacon

karen-stepanyan:
[DF-19699](https://smartcontract-it.atlassian.net/browse/DF-19699)

Updated eth-beacon balance transport to use POST method instead of GET when querying balances of addresses. 

Functional tests were performed as well

[DF-19699]: https://smartcontract-it.atlassian.net/browse/DF-19699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/giant-avocados-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/eth-beacon-adapter': patch
```
alecgard, line 2:
```suggestion
'@chainlink/eth-beacon-adapter': minor
```

alecgard (approved):

alecgard (dismissed):

.changeset/giant-avocados-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/eth-beacon-adapter': major
+---
+
+Changed API method from GET to POST for getting balances. RPC_URLs need to be updated to use POST method.
```
mmcallister-cll, line 5: Since these changesets are really the release notes, I think we need to add something here like
> strict requirement on Eth Beacon Node API >= v2.5.0

otherwise, this EA will fail with incompatible RPCs.

mmcallister-cll: Thinking about it, we might actually want to add this info into the README somewhere (known issues?)

packages/sources/eth-beacon/src/transport/balance.ts
```diff
@@ -200,13 +199,14 @@ export class BalanceTransport extends SubscriptionTransport<BalanceTransportType
 
   private async queryBeaconChain(
     url: string,
-    addresses: string,
-    statusList?: string,
+    addresses: string[],
+    statusList?: string[],
   ): Promise<StateResponseSchema> {
```
mmcallister-cll, line 204: I guess the other option here if we don't want to require a specific version/ensure backward compatibility would be to add a feature flag somewhere here and toggle between GET/POST payloads based on that. I think we're OK with a strict cutover though as long as it's clear and documented somewhere (and GTM/opdata/DON Ops know to specify the RPC requirements).

mmcallister-cll (approved):
Please explicitly mention v2.5.0 Beacon API in the changeset and README. Otherwise LGTM!

mmcallister-cll (changes requested):
Please explicitly mention v2.5.0 Beacon API in the changeset and README. Otherwise LGTM!

mmcallister-cll:
OK to merge this despite EA framework check failure -- new EA framework version just went in (should be addressed in a PR that updates all ea-framework versions)

PR #3275, opened 2024-04-23
Added new DLC-BTC-POR EA

karen-stepanyan:
## Description

Added new DLC-BTC-POR source EA. The EA doesn't take any input params. It fetches a list of vaults from a contract and for each vault verifies the vault deposit balance and sums all verified balances.

## Steps to Test

yarn test packages/sources/dlc-btc-por

## Quality Assurance

[DF-19884]: https://smartcontract-it.atlassian.net/browse/DF-19884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/dlc-btc-por/src/config/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      default: 10_000,
+    },
+  },
+  {
+    envDefaultOverrides: {
+      BACKGROUND_EXECUTE_TIMEOUT: 180_000,
+      API_TIMEOUT: 60_000,
+      CACHE_MAX_AGE: 360_000,
+    },
+  },
```
alecgard, line 53: Can we leave a comment explaining why we need these overrides?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,220 @@
...
+        if (isVerified) {
+          return vault.valueLocked.toNumber()
+        }
+      } catch (e) {
+        logger.error(e, `Error while verifying Deposit for Vault: ${vault.uuid}. ${e}`)
+      }
+      return 0
+    })
+
+    // totalPoR represents total proof of reserves value in sataoshis
```
alecgard, line 91:
```suggestion
    // totalPoR represents total proof of reserves value in satoshis
```

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  createMultiSigTransaction(
+    publicKeyA: string,
+    publicKeyB: string,
+    vaultUUID: string,
+    bitcoinNetwork: BitcoinNetwork,
+  ) {
+    const TAPROOT_UNSPENDABLE_KEY_STR =
+      '50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0'
```
alecgard, line 191: Is this an un-accessible public key, similar to the EVM `0x0000000000000000000000000000000000000000 ` address?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    const tweakedUnspendableTaprootKey = taprootTweakPubkey(
+      TAPROOT_UNSPENDABLE_KEY,
+      Buffer.from(vaultUUID),
+    )[0]
+
+    const multisigPayment = p2tr_ns(2, [hex.decode(publicKeyA), hex.decode(publicKeyB)])
+
+    const multisigTransaction = p2tr(tweakedUnspendableTaprootKey, multisigPayment, bitcoinNetwork)
+    multisigTransaction.tapInternalKey = tweakedUnspendableTaprootKey
```
alecgard, line 202: Could you leave a comment explaining what this is doing? Will be super useful for future reference, especially for those not too familiar with Bitcoin (like me)

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
alecgard: If so, could we just import `TAPROOT_UNSPENDABLE_KEY` from `@scure/btc-signer`? https://github.com/paulmillr/scure-btc-signer/blob/main/src/utils.ts#L95

packages/sources/dlc-btc-por/src/transport/utils.ts
```diff
@@ -0,0 +1,126 @@
...
+export const getBitcoinNetwork = (networkName: string): BitcoinNetwork => {
+  switch (networkName) {
+    case 'mainnet':
+      return bitcoin
+    case 'testnet':
+      return testnet
+    case 'regtest':
+      return regtest
+    default:
+      return bitcoin
```
alecgard, line 124: Might be clearer to throw an error here, rather than default to Bitcoin (even though currently the EA config validation should have caught this before here).

alecgard (commented):
This looks good! Feedback is mostly just for some extra context to be added in comments

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: Oh nice, didn't know that `@scure/btc-signer` already exports that.

alecgard (dismissed):

packages/sources/dlc-btc-por/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+This document was generated automatically. Please see [README Generator](../../scripts#readme-generator) for more info.
+
+## Environment Variables
+
+| Required? |         Name          |                                        Description                                        |  Type  |             Options             |                   Default                    |
+| :-------: | :-------------------: | :---------------------------------------------------------------------------------------: | :----: | :-----------------------------: | :------------------------------------------: |
+|    ✅     |        RPC_URL        |                          The RPC URL to connect to the EVM chain                          | string |                                 |                                              |
+|    ✅     |       CHAIN_ID        |                              The EVM chain id to connect to                               | number |                                 |                   `42161`                    |
```
mmcallister-cll, line 12: debating whether we should default this env var or not as it should be in line with the RPC_URL 🤔

packages/sources/dlc-btc-por/README.md
```diff
@@ -0,0 +1,55 @@
...
+There are no input parameters for this endpoint.
+
+### Example
+
+Request:
+
+```json
+{
+  "data": {
+    "endpoint": "proof-of-reserves"
```
mmcallister-cll, line 48: What do you think about renaming the endpoint `"reserves"` similar to the `proof-of-reserves` composite EA?

packages/sources/dlc-btc-por/README.md
```diff
@@ -0,0 +1,55 @@
...
+|    ✅     |       CHAIN_ID        |                              The EVM chain id to connect to                               | number |                                 |                   `42161`                    |
+|    ✅     |     DLC_CONTRACT      |                        Contract address to fetch all funded vaults                        | string |                                 | `0x20157DBAbb84e3BBFE68C349d0d44E48AE7B5AD2` |
+|    ✅     |    BITCOIN_RPC_URL    |                                THE RPC URL of bitcoin node                                | string |                                 |                                              |
+|    ✅     |    BITCOIN_NETWORK    |                                   Bitcoin network name                                    |  enum  | `mainnet`, `regtest`, `testnet` |                  `mainnet`                   |
+|    ✅     |     CONFIRMATIONS     |                      The number of confirmations to query data from                       | number |                                 |                     `6`                      |
```
mmcallister-cll, line 16: Not sure if some of these should be env vars vs input params 🤔 
Particularly CONFIRMATIONS, I think it is typically a request param. I'd say lean towards whatever we do in other EAs

packages/sources/dlc-btc-por/src/config/dlc-manager-abi.json
mmcallister-cll, line 1: maybe silly question, do/should we normally include the entire ABI or only the functions we are using?

packages/sources/dlc-btc-por/src/config/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      type: 'string',
+      required: true,
+    },
+    CHAIN_ID: {
+      description: 'The EVM chain id to connect to',
+      type: 'number',
+      required: true,
+      default: 42161,
+    },
+    DLC_CONTRACT: {
```
mmcallister-cll, line 16: similar comments as the README, I don't think there's anything wrong with this, but do you think we it would be worth making these configurable at the input param level?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,219 @@
...
+  RawVault,
+} from './utils'
+
+const logger = makeLogger('dlcBTC PoR')
+
+export type TransportTypes = BaseEndpointTypes
+
+export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
```
mmcallister-cll, line 26: [low priority] do we need this line or is this taken care of in the parent class?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,219 @@
...
+      }
+    }
+    await this.responseCache.write(this.name, [{ params: [], response }])
+  }
+
+  async _handleRequest(): Promise<AdapterResponse<TransportTypes['Response']>> {
+    const providerDataRequestedUnixMs = Date.now()
+
+    // Get all vault data. Filter placeholder values.
+    const vaultData: RawVault[] = (await this.dlcManagerContract.getFundedDLCs(0, 10_000)).filter(
```
mmcallister-cll, line 74: this means we are capped at 10k vaults, and 1 address can have multiple vaults, right?
I guess the question is dependent on Joshua's answer to [this](https://smartcontract-it.atlassian.net/browse/DF-19830?focusedCommentId=229460), whether we should preemptively add pagination here or maybe even ask them to maintain a numFundedDLCs.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,219 @@
...
+          return vault.valueLocked.toNumber()
+        }
+      } catch (e) {
+        logger.error(e, `Error while verifying Deposit for Vault: ${vault.uuid}. ${e}`)
+      }
+      return 0
+    })
+
+    // totalPoR represents total proof of reserves value in satoshis
+    const totalPoR = (await Promise.all(depositsPromise)).reduce((sum, deposit) => sum + deposit, 0)
```
mmcallister-cll, line 92: this is a really pretty solution, but assuming a case of 10000 DLCs, is it possible this would send 10000 parallel requests and overwhelm the RPC node?

packages/sources/dlc-btc-por/README.md
karen-stepanyan: I don't have any strong preference here. Obviously we can't use default for `RPC_URL` but it is known that the Arbitrum mainnet should be the destination chain.

packages/sources/dlc-btc-por/README.md
karen-stepanyan: You're right, vast majority of the EAs use `CONFIRMATIONS` input param, instead of config. There are however few reasons why i didn't want to have input params at all. 
1. In case of `CONFIRMATIONS` it makes sense to have it as request param if there are also addresses provided in the same request. For example having a generic indexer adapter that takes different addresses and returns a balance. In this case each address list that is given to indexer may require different confirmation number, so makes sense to also provide confirmations in the request. DLC.BTC  is not generic and fetches the addresses (vaults) internally. We still could have it in input params for consistency, but there are other concerns as well
2. In case of other properties like `DLC_CONTRACT`. Again, a lot of EAs have `contract` as a generic input param, but i think in this particular case it makes sene to have it in config. Generally i wanted to avoid input params because providing them with different values will just worsen the performance. For example in  a scenario when  Nops want to run one EA that supports both testnet and mainnet.  If we support `contract` or `confirmations` as input params, so that both chains can be used, then providing different values for those (two requests, one has `contract: MAINNET_CONTRACT`, the other has `contract: TESTNET_CONTRACT`) will actually slow down and possibly timeout the EA, as it will take much longer to process both params.  However, having those in config, now the Nops have to run another instance of EA for testnet. In this case each instance will deal maximum with 10000 requests, instead of 20000 and there is a less chance of EA timeout. Otherwise if we make `contract` as input param we would need to make `rpc_url` an input param as well.

We can still move `confirmations`, `contract` and probably others from config to input params, assuming some conditions are always true. Happy to read your suggestions or ideas.

packages/sources/dlc-btc-por/src/config/dlc-manager-abi.json
karen-stepanyan: Usually we include all the contract ABI as during the development it's hard to know which contract methods will be used. Also in most cases contract abi we use is in JSON and it is much easier to copy/paste the whole ABI, rather than parse it and copy the only methods that the EA is using. Lastly we won't need to touch it in the future when doing fixes or improvements.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: Good point, we don't need that 👍.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: Yes, we are capped at 10k vaults, tho now i think maybe it makes sense to makes this as a config as well. I think they do track the number of [vaults](https://github.com/DLC-link/dlc-solidity/blob/dev/contracts/DLCManager.sol#L267) but it's a private property.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: I think right now this is the only thing that concern me. (not necessarily right now, but maybe we face this issue in the next 6 month. again, depends on Joshua's answer). I lean towards adding a batch request support.

packages/sources/dlc-btc-por/README.md
mmcallister-cll: Given Joshua's latest comments about changes they are making, it's probably worth thinking about tradeoffs between
- A single EA that can handle RPC requests across multiple EVM chains (with multiple XXX_RPC_URLs)
- Deploying 1 EA per chain

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
mmcallister-cll: Right, so there are currently 2 vaults in mainnet, right?
Can we just multiply each by 100x or 1000x here and send to one of our RPCs and see if it falls over? 😂

mmcallister-cll (approved):

mmcallister-cll:
Looks good, I think the main worries are batching/upper limit testing and further chain expansion, but these shouldn't be immediate issues. Worst case we can discuss & handle in future sprints.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,219 @@
...
+          providerDataRequestedUnixMs: 0,
+          providerDataReceivedUnixMs: 0,
+          providerIndicatedTimeUnixMs: undefined,
+        },
+      }
+    }
+    await this.responseCache.write(this.name, [{ params: [], response }])
+  }
+
+  async _handleRequest(): Promise<AdapterResponse<TransportTypes['Response']>> {
```
mmcallister-cll, line 70: maybe not necessary, but should we be using `calculateHttpRequestKey` somewhere here?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: I'll try that, but i will implement batching anyway

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: Depending on how the RPC handles the load we might need to ask them to expose it and send paginated requests in batch to get the full list of vaults

packages/sources/dlc-btc-por/README.md
karen-stepanyan: I think in terms of performance it would be better to run a new instance of the EA for every new chain. In that case i don't think we need to change something now or in the future. Running a single instance means that the EA needs to handle potentially multiple chains at the same time which will heavily increase the load. For example a single EA that supports arbitrum, ethereum and optimism. Considering that in all of those chains there are 10000 vaults, during a background execute cycle the EA has to fetch all the vaults, fetch 10000 BTC transactions and verify them, multiplied by 3 times as there are 3 chains. My concern is that a single background execute cycle might fail or timeout.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: It can be used in `requester.request` to create a unique request key to coalesce HTTP requests. Since we have one endpoint/transport with no input params, the `txid` is used as a unique identifier, so i would say we don't actually need this.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
mmcallister-cll: Probably best to ask asap then since it will require redeploying contracts

karen-stepanyan:
Did some upper limit tests
10000 transaction checks (5 different transaction duplicated 2000 times)
70 concurrent requests to BITCOIN_RPC
took 2 minutes 16 seconds

5000 transaction checks (5 different transaction duplicated 1000 times)
30 concurrent requests to BITCOIN_RPC
took 2 minutes 24 seconds

5000 transaction checks (5 different transaction duplicated 1000 times)
50 concurrent requests to BITCOIN_RPC
took 1 minute 29 seconds

From my testing the RPC was able to handle 70 concurrent requests, depending on RPC and network latency maybe this is not the limit. In terms of overall time execution we also need to have some space for a EVM RPC call to get the vaults

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -0,0 +1,232 @@
...
+    ).filter(
+      (v: RawVault) =>
+        v.uuid != '0x0000000000000000000000000000000000000000000000000000000000000000',
+    )
+
+    // Get the Attestor Public Key
+    const attestorPublicKey = await this.dlcManagerContract.attestorGroupPubKey()
+
+    let totalPoR = 0
+    const concurencyGroupSize =
```
mmcallister-cll, line 83: [low priority] can this just be 
`const concurencyGroupSize = this.settings.GROUP_SIZE || vaultData.length`?

mmcallister-cll (dismissed):

alecgard (approved):

PR #3298, opened 2024-05-21
Remove dependency on flux-emulator dataProviders

cawthorne:
## Description

This patch edits the flux-emulator script in packages/scripts/flux-emulator/

**Previous Functionality:**

The flux emulator script queries a wei-watchers url and uses that to help configure all the feeds for a given input EA. This is used in our soak testing, as well as in another mode where it talks to a server url hardcoded to https://adapters.qa.stage.cldev.sh/fluxconfig.

The hardcoded wei-watchers url can be overridden with the environment variable WEIWATCHER_SERVER.

**New Functionality:**

This change removes the hardcoded wei-watchers url and the WEIWATCHER_SERVER environment variable, replacing it with only an environment variable named MASTER_SERVER.

This achieves the goal of removing on the dependency of on the wei-watchers service, while providing some level of backwards compatibility.

The intended behaviour is that the old behaviour can be achieved by setting MASTER_SERVER accordingly.

The k6payload option for flux-emulator still attempts to extract test cases from integration tests, as well as test-payload.json files as before.

## Changes

Removes the hardcoded wei-watchers url.
Removes the WEIWATCHER_SERVER environment variable.
Adds a MASTER_SERVER environment variable for backwards compatibility.
Updates related README files and some integration tests to support the new environment variable.
This patch also updates and fixes many of the miss-configured tests.

Previous test issues:
-Stale snapshots due to output formatting changing over time as dependencies change (mainly RxJS v6 -> v7+)
-ShellJS not running when in the test environment due to improper mocking of the fs function
-Floating ids in the test output changing between runs breaking snapshot matching.
-Input test data no longer being valid as new fields were added to the data feed configs

## Steps to Test

To run the tests for this tool you can run:
`yarn test flux-emulator`

You can run the flux-emulator k6 payload generation directly, outside of the test environment with, for example:
`yarn qa:flux:configure k6payload icap empty`

Before this patch 34 tests were failing, mainly due to stale snapshot data, and depencies changing, breaking the config of the tests.

After this patch all tests are passing for this script module, when ran with `yarn test flux-emulator`


alecgard (approved):
LGTM!

Nice job figuring out what was going on here, and thanks for fixing up the tests!

PR #3308, opened 2024-05-29
Feature/df 20015

cawthorne:
## Closes #DF-20015

N.B. useful Starknet constants can be found here:
https://github.com/starknet-io/starknet.js/blob/66a5c0341eccfef0dcdf1312c15627b7d4f6b675/src/constants.ts#L43

## Description

Adds Starknet Sepolia + Mainnet support to the view-function-multi-chain EA. Using the starknet.js library

## Changes

-Created a separate flow to detect and handle Starknet chains based on the chainId.
-Starknet chainIds can be larger than the javascript Number type can handle. Therefore this patch uses BigNumber.js to handle chainIds now.
-Adds new integration tests + test-payload.json test cases.
-Previous EVM functionality is aimed to be unchanged

## Steps to Test

The correct environment variables must be defined for the Starknet RPC and chainID e.g.:
export STARKNET_SEPOLIA_RPC_URL=https://starknet-sepolia.public.blastapi.io/rpc/v0_7
export STARKNET_SEPOLIA_CHAIN_ID=0x534e5f5345504f4c4941

export export STARKNET_MAINNET_RPC_URL=https://starknet-mainnet.public.blastapi.io/rpc/v0_7
export STARKNET_MAINNET_CHAIN_ID=0x534e5f4d41494e 

**For integration tests:**
yarn test view-function-multi-chain/test/integration

**The EA can also be ran outside of testings with:**
yarn && yarn setup && cd packages/sources/view-function-multi-chain && yarn start

Then in another process something like:
curl --location 'localhost:8080' \                                                  
--header 'Content-Type: application/json' \
--data '{
        "data": {
  "contract": "0x0228128e84cdfc51003505dd5733729e57f7d1f7e54da679474e73db4ecaad44",
  "function": "round_data",
  "inputParams": ["340282366920938463463374607431768254460"],
  "network": "STARKNET_SEPOLIA"
 }
}

### DIFF
--- a/packages/sources/view-function-multi-chain/src/transport/function.ts
+++ b/packages/sources/view-function-multi-chain/src/transport/function.ts
@@ -65,21 +65,30 @@ export class MultiChainFunctionTransport extends SubscriptionTransport<MultiChai
 
     const rpcUrl = process.env[networkEnvName]
 
-    // The Starknet ChainIds are too large to fit into the JS Number type
-    const chainId = new BigNumber(process.env[chainIdEnvName] ?? NaN)
+    const chainIdStr = process.env[chainIdEnvName] || NaN
+    let chainIdBN
 
-    if (!rpcUrl || chainId.isNaN()) {
+    if (chainIdStr === 'SN_MAIN') {
+      chainIdBN = new BigNumber(constants.StarknetChainId.SN_MAIN)
+    } else if (chainIdStr == 'SN_SEPOLIA') {
+      chainIdBN = new BigNumber(constants.StarknetChainId.SN_SEPOLIA)
+    } else {
+      // The Starknet ChainIds are too large to fit into the JS Number type
+      chainIdBN = new BigNumber(chainIdStr)
+    }
+
+    if (!rpcUrl || chainIdBN.isNaN()) {
       throw new AdapterInputError({
         statusCode: 400,
-        message: `Missing '${networkEnvName}': '${rpcUrl}' or '${chainIdEnvName}': '${chainId.toString()}' environment variables.`,
+        message: `Missing '${networkEnvName}': '${rpcUrl}' or '${chainIdEnvName}': '${chainIdBN.toString()}' environment variables.`,
       })
     }
 
     if (
-      !chainId.equals(constants.StarknetChainId.SN_SEPOLIA) &&
-      !chainId.equals(constants.StarknetChainId.SN_MAIN)
+      !chainIdBN.equals(constants.StarknetChainId.SN_SEPOLIA) &&
+      !chainIdBN.equals(constants.StarknetChainId.SN_MAIN)
     ) {
-      return this._handleRequestEVM(param, rpcUrl, chainId.toNumber())
+      return this._handleRequestEVM(param, rpcUrl, chainIdBN.toNumber())
     } else {
       return this._handleRequestStarknet(param, rpcUrl)
     }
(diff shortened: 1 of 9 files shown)


packages/sources/view-function-multi-chain/src/transport/function.ts
```diff
@@ -52,25 +54,78 @@ export class MultiChainFunctionTransport extends SubscriptionTransport<MultiChai
...
         statusCode: 400,
-        message: `Missing '${networkEnvName}' or '${chainIdEnvName}' environment variables.`,
+        message: `Missing '${networkEnvName}': '${rpcUrl}' or '${chainIdEnvName}': '${chainId.toString()}' environment variables.`,
       })
     }
 
+    if (
+      !chainId.equals(constants.StarknetChainId.SN_SEPOLIA) &&
+      !chainId.equals(constants.StarknetChainId.SN_MAIN)
+    ) {
```
alecgard, line 81: Can we use `networkName` rather than `chainId` for determining the network?

Reasoning is that `networkName` comes from the input params, and so are guaranteed to be consistent across nodes running the same job spec. `chainId` comes from the EA config, and so may differ depending on how the deployer has configured the EA. 

Happy path should have the same effect, but using `networkName` should eliminate a class of tough to debug issues.

packages/sources/view-function-multi-chain/test/integration/adapter.test.ts
```diff
@@ -22,6 +25,16 @@ describe('execute', () => {
     process.env.ETHEREUM_GOERLI_RPC_URL =
       process.env.ETHEREUM_GOERLI_RPC_URL ?? 'http://localhost:8554'
     process.env.ETHEREUM_GOERLI_CHAIN_ID = process.env.ETHEREUM_GOERLI_CHAIN_ID ?? '5'
+
+    process.env.SARKNET_MAINNET_RPC_URL =
```
alecgard, line 29: I'm not sure how this test is currently working as the `env` var wouldn't match the `networkName` 🤔

```suggestion
    process.env.STARKNET_MAINNET_RPC_URL =
```

packages/sources/view-function-multi-chain/src/transport/function.ts
cawthorne: We can use the network name yes, and just have two special cases for 'STARKNET_SEPOLIA' and 'STARKNET_MAINNET'.

packages/sources/view-function-multi-chain/test/integration/adapter.test.ts
cawthorne: Ah yes, will fix that typo. I think it worked for me as I had the STARKNET_MAINNET_RPC_URL env variable defined already.

packages/sources/view-function-multi-chain/src/transport/function.ts
alecgard: Apologies for the churn but I actually take back the above - if we rely on `networkName` then the EA has to know what network name will be used by the job spec & deployed EA, which don't want to hardcode.

augustbleeds: Throwing this out there ~ can't we have a new input parameter like `"chainType": "evm"` or `"chainType": "starknet"`? Would help us test more than one

augustbleeds: If we go with the chain id approach can we use the ASCII string representation? So `SN_SEPOLIA` and `SN_MAIN` vs the hexadecimal. I think the hexadecimal, for a node operator is more error prone. We can convert it to the hexadecimal representation in the code here

augustbleeds: (the starknet chainId is a hexadecimal representation of the network name)

PR #3313, opened 2024-06-07
DF-20083 s3 reader EA initial commit

mmcallister-cll:
## Description
Adds EA that is capable of reading CSV files from S3

## Changes
- Added new s3reader EA
- Uses the [AWS v3 S3 client lib](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/)

## Steps to Test
1. yarn test s3reader (from root)
2. sanity test with file in S3

sample test body:
```json
{
    "data":{
        "endpoint": "csv",
        "bucket": "test",
        "key": "test.csv",
        "row": "1",
        "column": "A"
    }
}
```

Note: may require an `aws sso login --profile <profile>` for testing locally

## Quality Assurance

[DF-20083]: https://smartcontract-it.atlassian.net/browse/DF-20083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    await this.responseCache.write(this.name, [{ params: param, response }])
+  }
+
+  async _handleRequest(param: RequestParams): Promise<AdapterResponse<TransportTypes['Response']>> {
+    const { bucket, key, row, column } = param
+    const providerDataRequestedUnixMs = Date.now()
+
+    const csvFileAsStr = await this.getFileFromS3(bucket, key)
+    const answer = this.parseCSV(csvFileAsStr, row, column)
```
alecgard, line 65: I'm slightly concerned that looking up by a fixed row won't be enough, as the sorting or contents of the file could potentially change.

Is there instead an ID or Name we can use to identify the row, and then select the column for that row?

cawthorne, line 65: Could we define a schema definition for the expected csv, and log or throw and error if the csv doesn't match the schema? Or just an assertion that certain columns exist and there are no missing entries if required etc.

Could help catch a class of bugs where the csv is corrupt, or the wrong one is sent.

packages/sources/s3reader/src/config/index.ts
```diff
@@ -0,0 +1,34 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  // AWS_CONFIG_FILE: {
```
alecgard, line 4: Can we break this out to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` explicitly? AWS Config file can have a few different auth methods, and this would make it clear how auth is being handled by the EA.

mmcallister-cll: We don't seem to use `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` elsewhere for our monitoring stack, opting to go for what I think is policy-based access, so I've removed these configs for now.
Testing requires an `aws sso login --profile <profilename>` to be done in the shell before running the EA.

packages/sources/s3reader/src/transport/csv.ts
mmcallister-cll: Yeah, added some handling for headers, matchers, and resultField (column). Does this cover it?

mmcallister-cll: [copied from Alec thread] Added some handling for headers, matchers, and resultField (column). Does this cover it?

packages/sources/s3reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,66 @@
...
+      required: true,
+      type: 'string',
+      description: 'The path of the file stored in S3',
+    },
+    headerRow: {
+      required: true,
+      type: 'number',
+      description: 'The 1-indexed row of the CSV file that contains the column headers',
+    },
+    matcherField: {
```
alecgard, line 26:
```suggestion
    matcherColumn: {
```

packages/sources/s3reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,66 @@
...
+      required: true,
+      type: 'string',
+      description: 'The column field to compare with the matcherValue',
+    },
+    matcherValue: {
+      required: true,
+      type: 'string',
+      description: 'The value to match with matcherField',
+    },
+    resultField: {
```
alecgard, line 36:
```suggestion
    resultColumn: {
```

packages/sources/s3reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,66 @@
...
+      required: true,
+      type: 'string',
+      description: 'The row of the CSV file to query',
```
alecgard, line 39: I think this is a column rather than row
 
```suggestion
      description: 'The column of the CSV file to return a result for, where the row matches matcherColumn and matcherValue',
```

packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,118 @@
...
+    matcherValue: string,
+    resultField: string,
+  ): number {
+    // from_line is 1-indexed
+    // columns: true sets first line as object fields rather than 2d arrays
+    const parser = parse(csvFileAsStr, { columns: true, from_line: headerRow })
+
+    // validate CSV contains headers matcherField and resultField
+    if (!(matcherField in parser[0])) {
+      throw new Error(`CSV file does not contain column header ${matcherField}`)
```
alecgard, line 99: Would be useful to include the headers we _did_ receive, for debugging

packages/sources/s3reader/src/transport/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  s3Client: S3Client,
+  bucket: string,
+  key: string,
+): Promise<string> {
+  const command = new GetObjectCommand({
+    Bucket: bucket,
+    Key: key,
+  })
+
+  const response = await s3Client.send(command)
```
alecgard, line 13: I think `s3Client.send` can throw an exception, which we likely want to handle, e.g. for a missing file

Example here: https://github.com/aws/aws-sdk-js-v3/blob/bc2396fd286a5836d6645cf25ba41139da345a52/clients/client-s3/src/waiters/waitForBucketExists.ts#L9-L18

packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+    if (!(resultField in parser[0])) {
+      throw new Error(`CSV file does not contain column header ${resultField}`)
+    }
+
+    // find correct row using matcher
+    const row = parser.find((row: { [x: string]: string }) => row[matcherField] == matcherValue)
```
alecgard, line 106: Let's throw an explicit error if we find multiple matching rows

packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (!row) {
+      throw new Error(`CSV file does not contain row where ${matcherField} == ${matcherValue}`)
+    }
+    return row[resultField]
```
alecgard, line 110: I think `csv-parse` returns all values as strings, so we'll want to convert this to a number

packages/sources/s3reader/src/transport/csv.ts
alecgard: Handled in added commit

packages/sources/s3reader/src/transport/csv.ts
alecgard: Handled in added commit

packages/sources/s3reader/test-payload.json
```diff
@@ -0,0 +1,6 @@
+{
+ "requests": [{
```
cawthorne, line 2: Is this a valid test-payload.json for this EA?

packages/sources/s3reader/src/endpoint/csv.ts
alecgard: Handled in pushed commit

packages/sources/s3reader/src/endpoint/csv.ts
alecgard: Handled in pushed commit

packages/sources/s3reader/src/endpoint/csv.ts
alecgard: Handled in pushed commit

packages/sources/s3reader/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  return {
+    ...originalModule,
+    S3Client: jest.fn().mockImplementation(() => {
+      return {
+        send: jest.fn().mockImplementation(async (command: GetObjectCommand) => {
+          if (command.input.Bucket == 's3_bucket' && command.input.Key == 'correct/path/file.csv') {
+            return {
+              Body: {
+                transformToString: () =>
+                  'Date,01/01/01,\nName,Value,Other\nRowOne,123,Other1\nRowTwo,456,Other2\nRowThree,789,Other3',
```
cawthorne, line 18: Can we have a test with a more 'realistic' csv?

alecgard:
On EA naming, `s3-csv-reader` may be more apt

packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,131 @@
...
+      throw new Error(`CSV file does not contain column header ${resultColumn}`)
+    }
+
+    // find correct row using matcher
+    const matchingRows = parsed.filter((row) => row[matcherColumn] == matcherValue)
+    if (matchingRows.length === 0) {
+      throw new Error(`CSV file does not contain row where ${matcherColumn} == ${matcherValue}`)
+    }
+    if (matchingRows.length > 1) {
+      throw new Error(`CSV file contains multiple rows where ${matcherColumn} == ${matcherValue}`)
```
cawthorne, line 115: Can we add a test, with a different csv to trigger this error?

packages/sources/s3reader/src/transport/utils.ts
alecgard: Handled in pushed commit

packages/sources/s3reader/src/transport/csv.ts
```diff
@@ -0,0 +1,131 @@
...
+      },
+    }
+  }
+
+  // csvFileAsStr: CSV file as string as received from S3
+  // headerRow: 1-indexed row of the CSV file to use as the header row
+  // matcherColumn: field to match with `matcherValue` to find the answer row
+  // matcherValue: value of field `matcherColumn` used to find the answer row
+  // resultColumn: header field containing the answer in matcher row
+  parseCSV(
```
cawthorne, line 86: Maybe pedantic, but I am not sure this is the best name for this function, as it returns a result, it isn't just about parsing/transforming the CSV. Maybe parseAndExtractCSV, extractFromCSV or fetchCSVCells?

packages/sources/s3reader/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,164 @@
...
+          if (command.input.Bucket != 's3_bucket') {
+            throw new Error('Error: The specified bucket does not exist.')
+          }
+          if (command.input.Key != 'correct/path/file.csv') {
+            throw new Error('Error: The specified key does not exist.')
+          }
+          return {
+            Body: {
+              transformToString: () =>
+                'Date,01/01/01,\nName,Value,Other\nRowOne,123,Other1\nRowTwo,456,Other2\nRowThree,789,Other3',
```
cawthorne, line 23: I think in the past I have worked with csvs, which are formatted like:

'ColHeader1,ColHeader2,ColHeader3\nRow1Value1,Row1Value2,Row1Value3\nRow2Value1,Row2Value2,Row2Value3...'

Can we support and test for this?

.changeset/nasty-jars-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/s3reader-adapter': major
```
cawthorne, line 2: Could the name of the EA have csv in it somewhere, if it is purely about csv parsing from s3?

packages/sources/s3reader/test/integration/adapter.test.ts
cawthorne: I see that is the format actually. The first row of the csv threw me off!

cawthorne: great!

packages/sources/s3reader/src/transport/csv.ts
alecgard: Added in https://github.com/smartcontractkit/external-adapters-js/pull/3313/commits/d06074b1a5b4d875daa0238bab4b2573d4247297#diff-a2f9928e24af6aedfb27b418061e6990f5a0141eb9c7cc0ffc2f1046abe1790cR251

packages/sources/s3reader/src/transport/csv.ts
alecgard: Renamed to `findValueInCSV`

packages/sources/s3-csv-reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: SingleNumberResultResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'csv',
+  aliases: ['nav'], // TODO: remove or switch?
```
alecgard, line 63: Vote remove

packages/sources/s3reader/test/integration/adapter.test.ts
cawthorne: Or just larger csvs, and also some which have weird/incorrect formatting.

packages/sources/s3-csv-reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,67 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { config } from '../config'
+import overrides from '../config/overrides.json'
+import { transport } from '../transport/csv'
+
+export const inputParameters = new InputParameters(
```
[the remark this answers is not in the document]
mmcallister-cll: Yeah it's definitely a good idea and something we discussed a few days back. There's a rough S3 fetching cost analysis in the design excalidraw amounting to ~2$/month USD for fetching this file every 10s across 10 NOPs.

packages/sources/s3reader/test-payload.json
mmcallister-cll: Added this to the soaktest blacklist so this file is not used for now.
Reason being we need some policies setup for AWS that make it not possible (again for now)

packages/sources/s3-csv-reader/src/endpoint/csv.ts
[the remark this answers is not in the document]
mmcallister-cll: We'll definitely need to do this eventually if we plan on accessing multiple "cells" from the same file.

packages/sources/s3-csv-reader/src/transport/csv.ts
```diff
@@ -0,0 +1,146 @@
...
+      throw new Error(`CSV file does not contain column header ${resultColumn}`)
+    }
+
+    // find correct row using matcher
+    const matchingRows = parsed.filter((row) => row[matcherColumn] == matcherValue)
+    if (matchingRows.length === 0) {
+      throw new Error(`CSV file does not contain row where ${matcherColumn} == ${matcherValue}`)
+    }
+    if (matchingRows.length > 1) {
+      throw new Error(`CSV file contains multiple rows where ${matcherColumn} == ${matcherValue}`)
```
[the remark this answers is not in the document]
mmcallister-cll: yeah, the point-at-which-to-stop is a little hard to land on
At this point I'm leaning towards as simple as possible for this iteration since we know we may need to rebuild a CSV solution from scratch for distribution across NOPs.

alecgard (approved):

packages/sources/s3-csv-reader/src/endpoint/csv.ts
```diff
@@ -0,0 +1,67 @@
...
+      required: true,
+      type: 'string',
+      description: 'The S3 bucket to query',
+    },
+    keyPrefix: {
+      aliases: ['pathPrefix'],
+      required: true,
+      type: 'string',
+      description:
+        'The path prefix of the file stored in S3. <Date.csv> is appended to search for older files.',
```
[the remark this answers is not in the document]
mmcallister-cll: will try to figure out how to make it clearer, but the idea is 
- provide `<bucket>` eg: data-feeds-s3-bucket
- provide `<path_prefix>` eg: path/to/file excluding -01-01-2024.csv
- `-01-01-2024.csv` (current date) gets appended to the path_prefix

so the resulting filename is `path/to/file-01-01-2024.csv` for example

I also thought about adding an `APPEND_DATE_CSV` env var defaulting to true in case we wanted to reuse this without dates, what do you think?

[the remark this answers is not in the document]
austinborn: I recommend specifying that we expect the file name to have the format `<name>-dd-MM-yyyy.csv`. Could we simplify this to say that we just pass in a file path to a folder, and each file has to be named just the `dd-MM-yyyy.csv`?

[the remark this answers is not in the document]
mmcallister-cll: yeah let's do this in a follow up, there will be a change on the infra side to make as well.

austinborn (approved):

mmcallister-cll:
Merging as other spot checked adapters have not had their EA FW version updated either.

PR #3314, opened 2024-06-07
Changed vault validation logic for DLC.BTC 

karen-stepanyan:
[DF-20050](https://smartcontract-it.atlassian.net/browse/DF-20050)

DLC team has changed the format of attestor public key. As a result the previous validation was not working and the EA was returning 0 as a result. This PR changes the logic for getting the attestor pub key and vault validation. Some new packages are added and some functions are moved to a utils file. 

[DF-20050]: https://smartcontract-it.atlassian.net/browse/DF-20050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/dlc-btc-por/src/transport/utils.ts
```diff
@@ -126,3 +130,78 @@ export const getBitcoinNetwork = (networkName: string): BitcoinNetwork => {
 }
 
 export const FUNDED_STATUS = 1
+
+const bip32 = BIP32Factory(ellipticCurveCryptography)
+
+// Gets the derived public key from the extended public key.
+export const getDerivedPublicKey = (extendedPublicKey: string, bitcoinNetwork: Network): Buffer => {
```
mmcallister-cll, line 137: I think we still need to account for extended and compressed keys
1. for testnet (probably possible with just a length check here -- last I checked they hadn't updated testnet to extended)
2. on the off chance mainnet changes again

karen-stepanyan: It seems that they have updated the testnet as well and they use a new key format there as well, so now there is only one key format (extended)

mmcallister-cll (approved):

PR #3315, opened 2024-06-07
Change set-token-index EA to return wei balances with order of magnitude matching the underlying tokens decimals.

cawthorne:
## Description

Fixes the set-token-index EA to provide token balances which are properly scaled to the correct size.

## Changes

Scale down the balance of a token allocations by 10^(18-tokenDecimals), to account for the adapter contract normalising all token balances to 18 decimals:

## Steps to Test
 
yarn && yarn setup && yarn test set-token-index

### DIFF
--- a/packages/composites/set-token-index/src/endpoint/allocations.ts
+++ b/packages/composites/set-token-index/src/endpoint/allocations.ts
@@ -105,6 +105,7 @@ export const execute: ExecuteWithConfig<Config> = async (input, context, config)
         return {
           balance: new BigNumber(balances[i].toString())
             .shiftedBy(-(18 - token.decimals))
+            .toFixed(0)
             .toString(),
           decimals: token.decimals,
           symbol: token.symbol,
--- a/packages/composites/set-token-index/src/endpoint/allocations.ts
+++ b/packages/composites/set-token-index/src/endpoint/allocations.ts
@@ -11,6 +11,7 @@ import {
 import { Config } from '../config'
 import * as TA from '@chainlink/token-allocation-adapter'
 import { makeExecute } from '../adapter'
+import BigNumber from 'bignumber.js'
 
 export const supportedEndpoints = ['allocations']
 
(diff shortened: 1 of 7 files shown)


packages/composites/set-token-index/test/integration/__snapshots__/adapter.test.ts.snap
alecgard, line 1: Can we add a token with non-18 decimals to this integ test?

packages/composites/set-token-index/src/endpoint/allocations.ts
```diff
@@ -95,10 +96,19 @@ export const execute: ExecuteWithConfig<Config> = async (input, context, config)
...
+          context,
+          jobRunID,
+          address,
+        )) as unknown as TA.types.TokenAllocation
+
         return {
-          balance: balances[i].toString(),
-          ...token,
+          balance: new BigNumber(balances[i].toString())
+            .shiftedBy(-(18 - token.decimals))
```
alecgard, line 107: Can we also handle tokens with greater than 18 decimals? Not sure how common it is, but found at least one example: https://etherscan.io/token/0xaba8cac6866b83ae4eec97dd07ed254282f6ad8a#readContract

alecgard (dismissed):
Have we confirmed that the `adapter` contract will always return 18 decimals? I don't have the context on that contract. 

Also looks like you need to add a changeset

packages/composites/set-token-index/test/integration/__snapshots__/adapter.test.ts.snap
cawthorne: The adapter contract starts with a number which is tokenDecimals wei, then multiplies by 10^18, then divides by 10^tokenDecimals, which gives an 18 decimals wei number always

Here is a stack trace of the execution:
https://www.tdly.co/shared/simulation/8615d5ed-8b38-4b4f-9608-2e9616997cf7

I have added a non 18 decimal test, by altering the mocked rpc responce to return non 18 decimals in the test.

packages/composites/set-token-index/src/endpoint/allocations.ts
cawthorne: This should correctly handle tokens with decimals greater than 18 decimals, as in that case it will scale up rather than down, by -(18-tokenDecimals), tokenDecimals=24, which would give +6. So it would scale the provided 18 decimal number up to the correct 24 decimal value.

packages/composites/set-token-index/test/integration/__snapshots__/adapter.test.ts.snap
cawthorne: Will add the changset now

packages/composites/set-token-index/src/endpoint/allocations.ts
alecgard: Ahh yep that makes sense, nice!

alecgard (approved):

PR #3325, opened 2024-06-18
Fix clear-bank tests

alecgard:
- Tests were failing due to not mocking the data before making the request


.changeset/tasty-deers-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/clear-bank-adapter': patch
```
karen-stepanyan, line 2: Do we need to bump the version here since it is only a change in a test file ?

karen-stepanyan (approved):

PR #3357, opened 2024-07-17
DF-20045 - Separate LWBA & price endpoints

mjk90:
## Description
A follow up to the LWBA validation changes. This PR separates LWBA & price endpoints in:
- Elwood
- NCFX

and updates GSR to use a custom Adapter implementation, allowing LWBA validation while using the same transport & cache keys for price & LWBA endpoints. This was necessary because GSR doesn't allow multiple WS connections and doesn't seem to handle multiple subscription requests from the same connection well.

......

## Steps to Test

- yarn test:unit
- yarn test:integration

## Quality Assurance

### DIFF
--- a/packages/sources/elwood/src/endpoint/crypto-lwba.ts
+++ b/packages/sources/elwood/src/endpoint/crypto-lwba.ts
@@ -1,0 +1,27 @@
+import {
+  LwbaEndpoint,
+  LwbaResponseDataFields,
+  lwbaEndpointInputParametersDefinition,
+} from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import { transport } from '../transport/lwba'
+
+export const inputParameters = new InputParameters(lwbaEndpointInputParametersDefinition, [
+  {
+    base: 'ETH',
+    quote: 'USD',
+  },
+])
+
+export type BaseEndpointTypesLwba = {
+  Parameters: typeof inputParameters.definition
+  Settings: typeof config.settings
+  Response: LwbaResponseDataFields
+}
+
+export const cryptoLwbaEndpoint = new LwbaEndpoint({
+  name: 'crypto-lwba',
+  transport,
+  inputParameters,
+})
--- a/packages/sources/elwood/src/endpoint/crypto.ts
+++ b/packages/sources/elwood/src/endpoint/crypto.ts
@@ -1,15 +1,11 @@
 import {
   CryptoPriceEndpoint,
-  DEFAULT_LWBA_ALIASES,
-  LwbaResponseDataFields,
   priceEndpointInputParametersDefinition,
-  validateLwbaResponse,
 } from '@chainlink/external-adapter-framework/adapter'
 import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
 import { InputParameters } from '@chainlink/external-adapter-framework/validation'
 import { config } from '../config'
 import { transport } from '../transport/crypto'
-import { AdapterLWBAError } from '@chainlink/external-adapter-framework/validation/error'
 
 const inputParameters = new InputParameters(priceEndpointInputParametersDefinition, [
   {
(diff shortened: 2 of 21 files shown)


packages/sources/elwood/src/endpoint/crypto.ts
```diff
@@ -18,27 +14,14 @@ const inputParameters = new InputParameters(priceEndpointInputParametersDefiniti
   },
 ])
 
-type OmitResultFromLwba = Omit<LwbaResponseDataFields, 'Result'>
-
 export type BaseEndpointTypes = {
   Parameters: typeof inputParameters.definition
   Settings: typeof config.settings
-  Response: OmitResultFromLwba & SingleNumberResultResponse
+  Response: SingleNumberResultResponse
```
mxiao-cll, line 20: This is not backward compatible so should be a major version bump?

packages/sources/elwood/src/endpoint/crypto-lwba.ts
```diff
@@ -0,0 +1,27 @@
...
+  },
+])
+
+export type BaseEndpointTypesLwba = {
+  Parameters: typeof inputParameters.definition
+  Settings: typeof config.settings
+  Response: LwbaResponseDataFields
+}
+
+export const cryptoLwbaEndpoint = new LwbaEndpoint({
```
karen-stepanyan, line 23: Since this is a LWBA endpoint, should we have `customOutputValidation` here ?

karen-stepanyan: Oh nevermind, LwbaEndpoint already has that internally.

packages/sources/elwood/src/transport/lwba.ts
```diff
@@ -0,0 +1,165 @@
...
+import { WebSocketTransport } from '@chainlink/external-adapter-framework/transports'
+import { makeLogger } from '@chainlink/external-adapter-framework/util'
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
+import axios from 'axios'
+
+const logger = makeLogger('ElwoodWsLwba')
+
+const DEFAULT_TRANSPORT_NAME = 'default_single_transport'
+
+export type SubscribeRequest = {
```
karen-stepanyan, line 11: It seems that this and bellow types are identical with `crypto` transport as well. Maybe we can move them to a shared `utils.ts` in transport folder ?

packages/sources/elwood/src/transport/lwba.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export type ResponseMessage = PriceResponse | HeartbeatResponse
+
+export type WsTransportTypes = BaseEndpointTypesLwba & {
+  Provider: {
+    WsMessage: ResponseMessage
+  }
+}
+
+export const transport: WebSocketTransport<WsTransportTypes> =
```
karen-stepanyan, line 49: How is this different from a transport used in `crypto` endpoint ? Is it only the top level `result` response that is null in this transport ? Do we want to do this to fix the alert issues with missing/wrong ask/bid/mid prices ? If so can me have the logic in one place and reuse it somehow ?

packages/sources/gsr/src/adapter.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (!includesMap[from]) {
+      includesMap[from] = {}
+    }
+    includesMap[from][to] = includes[0]
+  }
+
+  return includesMap
+}
+
+export class GSRPriceAdapter<
```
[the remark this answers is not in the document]
mjk90: You're right, we can just extend from PriceAdapter. For some reason, I had the idea that we couldn't override validateOutput using PriceAdapter but it looks like we can. I'll update this 🙏

packages/sources/elwood/src/endpoint/crypto.ts
mjk90: Good point. I don't think it's necessary to actually make this change so I'll revert the response to it's original format

mxiao-cll (approved):

PR #3359, opened 2024-07-18
Updated balancer EA READMEs with descriptions for MAX_PAYLOAD_SIZE_LIMIT

karen-stepanyan:
[DF-20246](https://smartcontract-it.atlassian.net/browse/DF-20246)

Added `known issues` section in README files for balancer EAs that explains the connection between `MAX_PAYLOAD_SIZE_LIMIT` env var the EA supports and Nginx config. 

Note for reviewer. 
Ideally this should apply to all EAs, although is currently relevant to balancer EAs as they accept an array of any size as an input parameter. I'm open to include this description in the [ea-framework](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/config/index.ts#L165) itself, but since the ea-framework configs are not linked to EA READMEs, I'm also ok having those explicitly in balancer EAs READMEs for now

[DF-20246]: https://smartcontract-it.atlassian.net/browse/DF-20246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sources/amberdata/docs/known-issues.md
+++ b/packages/sources/amberdata/docs/known-issues.md
@@ -1,9 +1,5 @@
 ## Known Issues
 
-### ETH Beacon API version
-
-Starting from version 3.0.0, the eth-beacon EA is compatible with the ETH Beacon API(`ETH_CONSENSUS_RPC_URL`) version 2.5.0 or later. If you are using an older version of the ETH Beacon API, you will need to upgrade it to 2.5.0 or later to use the eth-beacon EA.
-
 ### MAX_PAYLOAD_SIZE_LIMIT configuration
 
 The `MAX_PAYLOAD_SIZE_LIMIT` environment variable is used for controlling the maximum size of the incoming request body that the EA can handle. If you decide to customize this value it's essential to ensure that any reverse proxy or web server in front of the EA, such as Nginx, is also configured with a corresponding limit. This alignment prevents scenarios where Nginx rejects a request for exceeding its payload size limit before it reaches the EA.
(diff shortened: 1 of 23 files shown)


mxiao-cll:
Can we just put content in one file and point to that file?

karen-stepanyan:
@mxiao-cll As I mentioned in the PR description, ideally we should have the content in a ea-framework repo and just point to it from EA READMEs, there is a separate ticket for that. Until that is done we can have the content here for each README

mxiao-cll (approved):

packages/sources/amberdata/docs/known-issues.md
```diff
@@ -0,0 +1,5 @@
+## Known Issues
```
mmcallister-cll, line 1: README update missing here

mmcallister-cll:
are all of these EAs related to growing address lists and/or PoR feeds?

packages/sources/amberdata/docs/known-issues.md
austinborn: What do you mean? That the `amberdata` README should point to this?

austinborn (approved):

karen-stepanyan:
@mmcallister-cll Yes. It seems that i've missed couple but don't think they are used that much.

mmcallister-cll (approved):

mmcallister-cll:
Yeah I think we're good with this set, can always move to the others as well

PR #3383, opened 2024-08-09
fix ICAP heartbeat cache TTL refresh

karen-stepanyan:
ICAP EA was not refreshing its cache entries on a heartbeat from DP because of either a wrong check or DP changing heartbeat response schema.  This PRs removes the check so any heartbeat coming from TP or ICAP streams will refresh the cache TTL for corresponding EAs.


karen-stepanyan:
EA framework dependency check can be ignored

mmcallister-cll (approved):

PR #3389, opened 2024-08-14
DLC-BTC. Remove filtering from vault fetching logic, add network support

karen-stepanyan:
## Description
Fixed BTC vault fetching logic. Previously we filtered out placeholder values, but DLC.LINK team now recommends to fetch all the vault and skip the filtering. 
Additionally added support for different networks. Currently the EA was deployed to Aribtirum and to deploy to a new chain, a new instance is needed as the config options supported one chain. PR includes changes so that different chains and RPCs can be provided.

## Changes
- Removed the filter function from transport when fetching vault info
- Added support for new chains
- Replaced DLC_CONTRACT env variable with 'dlcContract' input parameter.
- Replaced RPC_URL, CHAIN_ID env variables with network-specific env variables ( {NETWORK}\_RPC_URL, {NETWORK}\_CHAIN_ID ).
- Added the EA to readme-generator blacklist as several config options are added to readme manually that are not part of config.ts

## Steps to Test

 `yarn test packages/sources/dlc-btc-por/`

## Quality Assurance

[DF-20315]: https://smartcontract-it.atlassian.net/browse/DF-20315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


karen-stepanyan:
I would suggest to ignore the ea-framework dependency check until we figure out what to do

packages/sources/dlc-btc-por/src/config/index.ts
```diff
@@ -3,23 +3,22 @@ import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
...
-      description: 'The RPC URL to connect to the EVM chain',
+      description:
+        '[DEPRECATED]. Please use {NETWORK_NAME}_RPC_URL instead. The RPC URL to connect to the EVM chain',
       type: 'string',
-      required: true,
     },
     CHAIN_ID: {
-      description: 'The EVM chain id to connect to',
+      description:
+        '[DEPRECATED]. Please use {NETWORK_NAME}_CHAIN_ID instead.The EVM chain id to connect to',
```
mmcallister-cll, line 12: [nit] space after the second period here

packages/sources/dlc-btc-por/src/config/index.ts
```diff
@@ -3,23 +3,22 @@ import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
...
       type: 'number',
-      required: true,
     },
     DLC_CONTRACT: {
-      description: 'Contract address to fetch all funded vaults',
-      required: true,
+      description:
+        '[DEPRECATED]. Please use {NETWORK_NAME}_DLC_CONTRACT instead.Contract address to fetch all funded vaults',
```
mmcallister-cll, line 17: here too

packages/sources/dlc-btc-por/README.md
```diff
@@ -1,22 +1,23 @@
...
-|           | BACKGROUND_EXECUTE_MS  | The amount of time the background execute should sleep before performing the next request | number |                                 |  `10000`  |
+| Required? |          Name           |                                             Description                                              |  Type  |             Options             |  Default  |
+| :-------: | :---------------------: | :--------------------------------------------------------------------------------------------------: | :----: | :-----------------------------: | :-------: |
+|           |         RPC_URL         |     [DEPRECATED]. Please use {NETWORK}\_RPC_URL instead. The RPC URL to connect to the EVM chain     | string |                                 |           |
+|           |        CHAIN_ID         |         [DEPRECATED]. Please use {NETWORK}\_CHAIN_ID instead.The EVM chain id to connect to          | number |                                 |           |
+|           |      DLC_CONTRACT       | [DEPRECATED]. Please use {NETWORK}\_DLC_CONTRACT instead.Contract address to fetch all funded vaults | string |                                 |           |
```
mmcallister-cll, line 11: here too

packages/sources/dlc-btc-por/src/endpoint/proof-of-reserves.ts
```diff
@@ -1,16 +1,33 @@
 import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
-import { EmptyInputParameters } from '@chainlink/external-adapter-framework/validation/input-params'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation/input-params'
 import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
 import { config } from '../config'
 import { porTransport } from '../transport/proof-of-reserves'
 
+export const inputParameters = new InputParameters(
```
mmcallister-cll, line 7: more of a philosophical comment, but rather than each NOP setting the `{NETWORK}_DLC_CONTRACT`, which would be the same across all NOPs, I'd lean towards adding the contract address as an input param. Useful in the unlikely event of a contract change.

Not sure if it's entirely safe to generalize, but let's restrict env vars here to individual configs (such as RPC URLs) that differ between NOPs, and use input params for configs across all NOPs.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -66,18 +69,63 @@ export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
...
+
+    let networkEnvName = `RPC_URL`
+    let chainIdEnvName = `CHAIN_ID`
+    let dlcContractEnvName = `DLC_CONTRACT`
+
+    let rpcUrl = this.settings.RPC_URL
+    let chainId = this.settings.CHAIN_ID
+    let dlcContractAddress = this.settings.DLC_CONTRACT
+
+    // If network is provided, use the network specific environment variables. We still support the old environment variables for backward compatibility.
```
mmcallister-cll, line 87: thanks! 👍

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -66,18 +69,63 @@ export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
...
+    }
+
+    if (!network) {
+      logger.warn(
+        `Network is not provided. Reading from RPC_URL, CHAIN_ID and DLC_CONTRACT environment variables. Please use the network parameter to specify the network in the environment variables and in the request.`,
+      )
+    }
+
+    if (!this.providers[networkName]) {
+      this.providers[networkName] = new ethers.providers.JsonRpcProvider(rpcUrl, chainId)
```
mmcallister-cll, line 113: maybe a silly question, but is there any lag on the first request to a new chain because we have to set up this object?

packages/sources/dlc-btc-por/README.md
mxiao-cll: Why are we changing readme in PR? We have a GH that update these automatically after merge

mmcallister-cll (changes requested):
small changes: contract as input param & related rework, typos
LGTM overall

packages/sources/dlc-btc-por/README.md
mmcallister-cll: this EA was added to the readme generator blacklist: packages/scripts/src/generate-readme/readmeBlacklist.json

these env var changes are similar to your additions for harris-and-trotter in https://github.com/smartcontractkit/external-adapters-js/commit/5e53f62b47c9c523b30bad8e0edc402b2adad8b9

packages/sources/dlc-btc-por/src/config/index.ts
```diff
@@ -3,23 +3,22 @@ import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
 export const config = new AdapterConfig(
   {
     RPC_URL: {
```
mxiao-cll, line 5: Since we are doing a major version bump, should we just delete these?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: There will be a lag on the very first request because it needs an active subscription set to execute the code. On a new network request whether there will be a lag completely depends on the timing. The object setup is synchronous and super fast, but each background cycle runs every 10s, so if a new requests comes right before a new background cycle has started, a response will be immediate, otherwise worst case scenario there can be a lag of maximum 10s

packages/sources/dlc-btc-por/src/config/index.ts
karen-stepanyan: We do the major bump because of the logic change and not RPC change. In theory we could also remove those but deprecating them first adds extra safety as NOPs are infamous for not reading the changelogs or not following them. I would suggest to make sure all the NOPs use network specific env vars and we can remove legacy one with the next release of the EA.

mxiao-cll: I see

"make sure all the NOPs use network specific env vars" -- How do we usually do this?

packages/sources/dlc-btc-por/src/endpoint/proof-of-reserves.ts
karen-stepanyan: Agree, on the other hand this will make even more confusion between legacy env vars and new input params. I'm kinda thinking now that it worth just removing old env vars, make all those breaking changes and ensure NOPs configure everything properly before running new images. Otherwise there will be very strange behavior which can lead to errors. wdyt ?

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -66,18 +66,35 @@ export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
...
 
-  async _handleRequest(): Promise<AdapterResponse<TransportTypes['Response']>> {
+  async _handleRequest(param: RequestParams): Promise<AdapterResponse<TransportTypes['Response']>> {
+    const { network, dlcContract } = param
+
+    const networkName = network.toUpperCase()
+    // ${networkName}_RPC_URL and ${networkName}_CHAIN_ID are already validated in customInputValidation
+    const rpcUrl = process.env[`${networkName}_RPC_URL`] as string
+    const chainId = Number(process.env[`${networkName}_CHAIN_ID`])
+    const dlcContractAddress = dlcContract
```
mxiao-cll, line 79: Why do we need this?

mxiao-cll (dismissed):

mxiao-cll:
> I would suggest to ignore the ea-framework dependency check until we figure out what to do

You can simply bump this one and the error will go away

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
mxiao-cll: I mean line 79

mxiao-cll (approved):

packages/sources/dlc-btc-por/src/endpoint/proof-of-reserves.ts
```diff
@@ -1,16 +1,60 @@
...
 
 export const endpoint = new AdapterEndpoint({
   name: 'reserves',
+  inputParameters,
   transport: porTransport,
+  customInputValidation: (
+    req: AdapterRequest<typeof inputParameters.validated>,
+  ): AdapterInputError | undefined => {
+    const { network } = req.requestContext.data
+    const networkName = network.toUpperCase()
```
mmcallister-cll, line 46: the network specific env vars are pretty common now, do you think it's worth getting in a follow up ticket to formalize them in a utility function?

karen-stepanyan: Yeah I think it makes sense. Either find a way to support this in config, or like you suggested

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
karen-stepanyan: It will just try again in every background execution loop

mmcallister-cll (approved):

PR #3393, opened 2024-08-15
Bump framework version to 1.3.1

martin-cll:


mxiao-cll:
Need a change set

Did you generate this by hand or via packages/scripts/src/framework/version-bump.sh ?

martin-cll:
> Need a change set
> 
> Did you generate this by hand or via packages/scripts/src/framework/version-bump.sh ?

Added a changeset. I just find and replaced the versions in package.json files.

austinborn:
@martin-cll for future reference you can also use this script to do the same thing (incl. the changeset): https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/scripts/src/framework/version-bump.sh

mxiao-cll (approved):

karen-stepanyan:
I am against merging this as it will break the github pipeline for Release (due to big amount of changes). Details are [here](https://smartcontract-it.atlassian.net/browse/DF-20336)

PR #3395, opened 2024-08-16
New market status composite adapter

martin-cll:
New composite adapter for market status building on top of the TradingHours (primary) and NXFC (secondary) adapters:

1. Return primary's market status if it is open or closed.
2. If primary is unknown or errors, then fall back to secondary.
3. Return secondary's market status if it is open or closed.
4. If secondary is unknown or errors, then return closed.


packages/composites/market-status/src/adapter.ts
```diff
@@ -0,0 +1,44 @@
+import { AdapterRequest, AdapterResponse, Execute, InputParameters } from '@chainlink/ea-bootstrap'
+import { Validator } from '@chainlink/ea-bootstrap'
```
mxiao-cll, line 2: I believe these are from the v2 framework? We should be using v3 framework now?

karen-stepanyan: @martin-cll You're right. Looks like this is created as v2 EA. I would also suggest to create a v3 EAs using ea-framework as v2 is deprecated and we don't support it or fix bugs there.

martin-cll: Hey @mxiao-cll @karen-stepanyan I have updated this PR to use the v3 EA. For reference, the reason I did not opt to do so in the first place (sorry I did not include this in the PR description) is that the v3 transports (e.g. http, ws) have a strong assumption that there is only 1 upstream data provider. This makes sense for regular adapters, but maybe less so for composite adapters such as this one which uses multiple underlying providers (i.e. other adapters).

The simplest way I could see to use the existing transports without implementing a new one was to use the http transport with a modified `Requester`. The modification is to intercept the `axios.request` call with a modified function (see `overrideAxiosRequest`) that calls the multiple underlying adapters. Unfortunately, because most of the internal impl of the `Requester` class is declared `private`, I've had to copy paste the entire class to make this work.

This isn't a great final solution, but I felt that it was better than implementing a transport class from scratch in which case I would've lost out on all the goodies that both the HTTP transport and requester provided. Perhaps a future version of the framework could redeclare the Requester's internals as `protected` instead of make inheritance easier?

karen-stepanyan: hey @martin-cll . Thank you very much for detailed explanation. Actually there is no need to use workarounds with `Requester` or `axios` as the ea-framework provides all the components, tho I agree that we might need more comprehensive documentation and guides. 
You are absolutely right that `http` and `ws` transports are not a good fit here, but for any non standard data fetching/processing, you can create a custom transport. We have many examples in monorepo of custom transport EAs, and we actually have a custom composite transport (gm-token) that uses 3 different source EAs - https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/gm-token/src/transport/price.ts#L139 in its logic. 
When creating a new EA with `yarn new` you can choose `Custom (subscription based)` which will generate the most of boilerplate code. I think it will be much simpler if we use custom, subscription based transport approach for this EA. Let me know if you have concerns or questions and we can go through this privately

martin-cll: @karen-stepanyan Makes sense, thanks for showing me the custom transport! I have updated this PR to use a similar transport.

mjk90 (approved):

packages/composites/market-status/src/transport/requester.ts
```diff
@@ -0,0 +1,332 @@
+import { AxiosError, AxiosRequestConfig, AxiosResponse } from 'axios'
```
karen-stepanyan, line 1: Do we still need this fie ?

packages/composites/market-status/src/transport/market-status.ts
```diff
@@ -0,0 +1,139 @@
...
+  ): Promise<MarketStatusResult> {
+    const baseURL = context.adapterSettings[`${adapterName}_ADAPTER_URL`]
+    const data = {
+      data: {
+        endpoint: 'market-status',
+        market,
+      },
+    }
+    try {
+      const resp = await axios.post(baseURL, data, { timeout: 30_000 })
```
karen-stepanyan, line 118: We can initialize a requester instance and use it instead of axios
https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/gm-token/src/transport/price.ts#L57
It internally handles error retires, rate limits and request coalescing

.changeset/good-points-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/market-status-adapter': minor
```
karen-stepanyan, line 2:
```suggestion
'@chainlink/market-status-adapter': major
```

packages/composites/market-status/src/transport/market-status.ts
martin-cll: Update the PR!

packages/composites/market-status/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@types/node": "16.18.96",
+    "@types/sinonjs__fake-timers": "8.1.5",
+    "@types/supertest": "2.0.16",
+    "mock-socket": "9.3.1",
+    "nock": "13.5.4",
+    "supertest": "6.2.4",
+    "typescript": "5.0.4"
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "1.3.1",
```
mxiao-cll, line 42: This will actually break readMe so better to use 1.3.0 for now

packages/composites/market-status/src/config/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  NCFX_ADAPTER_URL: {
+    description: 'URL of the NCFX adapter',
+    type: 'string',
+    required: true,
+  },
+  BACKGROUND_EXECUTE_MS: {
+    description:
+      'The amount of time the background execute should sleep before performing the next request',
+    type: 'number',
+    default: 1_000,
```
mxiao-cll, line 18: I think we usually do 10s here, why 1s?

packages/composites/market-status/src/transport/market-status.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  async _handleRequest(
+    context: EndpointContext<BaseEndpointTypes>,
+    param: RequestParams,
+  ): Promise<MarketStatusResult> {
+    const market = param.market
+    if (!market) {
+      throw new Error(`Missing market in params: ${market}`)
+    }
```
mxiao-cll, line 86: We should validate this on layer up. either in InputParameters or use customInputValidation

karen-stepanyan: `market` is [required](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/adapter/market-status.ts#L25) input param. The validation for required input params are handled internally in the ea-framework, so this if statement can be removed

packages/composites/market-status/package.json
karen-stepanyan: @mxiao-cll Why would it break the readme ?

packages/composites/market-status/src/config/index.ts
martin-cll: This setting effectively determines how stale the data can be right? 10s feels like a long time in that context. Is there an issue with using 1s? That's also how frequently NCFX sends us a WS message for example.

mxiao-cll: Yes, the lower the number -> more request are being sent to APi providers. If they are fine with load we are totally cool

mxiao-cll (approved):

mxiao-cll:
Might want to check why soak test is failing

PR #3417, opened 2024-09-04
MERC-6221 Enable full symbol overrides for TP EA

martin-cll:
https://smartcontract-it.atlassian.net/browse/MERC-6221

Enable full symbol overrides for the TP EA. For example, sending the request `{"data":{"endpoint":"forex","from":"EGP","to":"JPY","overrides":{"tp":{"EGP":"FXSPTEGPJPYSPT:GBL.BIL.QTE.RTM!TP"}}}}` would return the exact price for the symbol `FXSPTEGPJPYSPT:GBL.BIL.QTE.RTM!TP` rather than any other prices cached for the base `EPG`. This is relevant in cases where there are multiple instruments (e.g. different expiring forwards) for the same base/quote pair.

This is simply implemented by returning two provider results, the default one with the parsed base and an additional one where the base is the full symbol. Note that I refactored the parsing to use string splitting because forward symbols (e.g. `CEFWDXAUUSDSPT06M:LDN.BIL.QTE.RTM!TP`) can have more characters than spot symbols (e.g. `FXSPTUSDAEDSPT:GBL.BIL.QTE.RTM!TP`)

Also added a `commodities` endpoint that's an alias of the existing `prices` / `forex` endpoint.

### DIFF
--- a/packages/sources/tp/src/endpoint/price.ts
+++ b/packages/sources/tp/src/endpoint/price.ts
@@ -1,5 +1,5 @@
 import {
-  ForexPriceEndpoint,
+  PriceEndpoint,
   priceEndpointInputParametersDefinition,
   PriceEndpointInputParametersDefinition,
 } from '@chainlink/external-adapter-framework/adapter'
@@ -20,6 +20,7 @@ export type GeneratePriceOptions = {
   streamName: 'TP' | 'IC'
   sourceOptions?: string[]
 }
+
 export const generateInputParams = (
   generatePriceOptions: GeneratePriceOptions,
 ): InputParameters<PriceEndpointInputParametersDefinition> =>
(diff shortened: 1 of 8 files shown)


packages/sources/tp/src/endpoint/price.ts
```diff
@@ -47,14 +48,14 @@ export const generateInputParams = (
 const tpOptions: GeneratePriceOptions = {
   sourceName: 'tpSource',
   streamName: 'TP',
-}
+} as const
 
 const inputParameters = generateInputParams(tpOptions)
 const wsTransport = generateTransport(tpOptions)
 
-export const priceEndpoint = new ForexPriceEndpoint({
```
karen-stepanyan, line 55: What's the reason of this change ?

karen-stepanyan (changes requested):
LGTM. Can you add integration test that covers the override part ?

packages/sources/tp/src/endpoint/price.ts
martin-cll: So right now we have the `price` and `forex` aliases with the `ForexPriceEndpoint` class being the same thing as `PriceEndpoint`. Since we want to add new `commodities` alias as well, I thought it would make more sense to use the base `PriceEndpoint` class rather than continue to use the `ForexPriceEndpoint`.

martin-cll:
> LGTM. Can you add integration test that covers the override part ?

Added some integration tests for overrides + WTI mapping.

karen-stepanyan (approved):

PR #3420, opened 2024-09-06
Added new Anchorage EA

karen-stepanyan:
## Description
Added new Anchorage EA that will be a used as a new custodian for ARKB and CETH feeds.


packages/sources/anchorage/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Wallet Endpoint
+
+`wallet` is the only supported name for this endpoint.
+
+### Input Params
+
+| Required? |  Name   |  Aliases  |          Description          |  Type  |       Options        |  Default  | Depends On | Not Valid With |
+| :-------: | :-----: | :-------: | :---------------------------: | :----: | :------------------: | :-------: | :--------: | :------------: |
+|    ✅     | vaultId | `vaultID` |          customerId           | string |                      |           |            |                |
+|           | chainId |           | The ID of the chain to return | string | `mainnet`, `testnet` | `mainnet` |            |                |
```
mxiao-cll, line 38: Should we call this chainName? I think we use numbers for id throughout other places

packages/sources/anchorage/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+This document was generated automatically. Please see [README Generator](../../scripts#readme-generator) for more info.
+
+## Environment Variables
+
+| Required? |         Name          |                                        Description                                        |  Type  | Options | Default |
+| :-------: | :-------------------: | :---------------------------------------------------------------------------------------: | :----: | :-----: | :-----: |
+|    ✅     |        API_KEY        |                                 An API key for Anchorage                                  | string |         |         |
+|    ✅     |     API_ENDPOINT      |                               An API endpoint for Anchorage                               | string |         |         |
+|           | BACKGROUND_EXECUTE_MS | The amount of time the background execute should sleep before performing the next request | number |         | `40000` |
```
mxiao-cll, line 13: API_LIMIT is not here

packages/sources/anchorage/src/transport/wallet.ts
```diff
@@ -0,0 +1,162 @@
+import { TransportDependencies } from '@chainlink/external-adapter-framework/transports'
+import { ResponseCache } from '@chainlink/external-adapter-framework/cache/response'
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import { AdapterResponse, makeLogger, sleep } from '@chainlink/external-adapter-framework/util'
+import { SubscriptionTransport } from '@chainlink/external-adapter-framework/transports/abstract/subscription'
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/wallet'
+import { AdapterError } from '@chainlink/external-adapter-framework/validation/error'
+
+const logger = makeLogger('WalletTransport')
```
mxiao-cll, line 10: Use this name needs to be unique? I see another WalletTransport exist already

karen-stepanyan: Not at all. On adapter level this should be unique between transports, but since there is one transport it can be anything.

packages/sources/anchorage/test-payload.json
```diff
@@ -0,0 +1,3 @@
+{
+ "requests": [{}]
```
mxiao-cll, line 2: If we do this? Will the soak test be skipped? Or it will run with empty input and fail?

packages/sources/anchorage/test/integration/fixtures.ts
```diff
@@ -0,0 +1,140 @@
...
+            subaccountId: '0ec207d4-9b4e-4428-9a6e-bfd52cdb0d0c',
+            type: 'WALLET',
+            vaultId: 'b0bb5449c1e4926542ce693b4db2e883',
+            vaultName: 'ETH ETF',
+            walletId: '4f7a79a18ba97f6df336d3a5bf5112ba',
+            walletName: 'CETH Wallet 3',
+          },
+        ],
+        page: {
+          next: null,
```
mxiao-cll, line 126: Can we put something here? to test the loop?

packages/sources/anchorage/test-payload.json
karen-stepanyan: No, soak tests will be skipped only if we add the EA to the blacklist. This means it will try to send a an empty request to an EA. The soak tests are failing now for this EA anyway, I think we need deployment files first. I am ok skipping the soak tests now, but don't want to add the EA to blacklist so that after the deployments we can configure it. Also i need proper input data for soak tests.

mxiao-cll: You don't need deployment files, you will in proper input as well as setting up the env var in that secrets repo

packages/sources/anchorage/src/transport/wallet.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    while (hasNext) {
+      const reqKey = requestConfig.baseURL + requestConfig.url
+      const response = await this.requester.request<WalletResponse>(reqKey, requestConfig)
+      wallets.push(...response.response.data.data)
+      hasNext = response.response.data.page.next !== null
+      if (response.response.data.page.next) {
+        // Remove the '/v2' prefix from the URL as it's already part of the baseURL
+        requestConfig.url = response.response.data.page.next.replace('/v2', '')
+      }
```
mxiao-cll, line 152: Is there a way to get around this or make it more flexible? In a sense that if this changes from v2 to v3 we can avoid a change?

packages/sources/anchorage/test-payload.json
karen-stepanyan: Yeah the latter ones are not ready yet, so I'm fine with this now

packages/sources/anchorage/src/transport/wallet.ts
karen-stepanyan: I think if they change to v3, regardless if we handle this specific case or not, we would have to change the EA anyways (API changes, schema changes, workflow changes).

packages/sources/anchorage/src/config/index.ts
```diff
@@ -0,0 +1,28 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+import { validator } from '@chainlink/external-adapter-framework/validation/utils'
+
+export const config = new AdapterConfig({
+  API_KEY: {
```
mmcallister-cll, line 5: do we need to support multiple API keys?

packages/sources/anchorage/test/integration/fixtures.ts
karen-stepanyan: Changed the test case to use pagination

mxiao-cll (dismissed):

packages/sources/anchorage/src/config/index.ts
karen-stepanyan: Most likely no, at least not for this custodian. Still verifying that information

karen-stepanyan: Confirmed, one API_KEY for Anchorage

packages/sources/anchorage/src/transport/wallet.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    while (hasNext) {
+      const reqKey = requestConfig.baseURL + requestConfig.url
+      const response = await this.requester.request<WalletResponse>(reqKey, requestConfig)
+      wallets.push(...response.response.data.data)
+      hasNext = response.response.data.page.next !== null
+      if (response.response.data.page.next) {
+        // Remove the '/v2' prefix from the URL as it's already part of the baseURL
+        requestConfig.url = response.response.data.page.next.replace('/v2', '')
```
mmcallister-cll, line 148: why don't we just get rid of it from the base API then, and add it above to the `requestConfig.url`?

packages/sources/anchorage/test-payload.json
mmcallister-cll, line 2: add sandbox creds to the secrets repo?

mxiao-cll (approved):

austinborn (approved):

PR #3421, opened 2024-09-06
Added new Bitgo EA

karen-stepanyan:
[DF-20417](https://smartcontract-it.atlassian.net/browse/DF-20417)

Added new Bitgo EA that will be a used as a new custodian for ARKB and CETH feeds.

[DF-20417]: https://smartcontract-it.atlassian.net/browse/DF-20417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/bitgo/src/config/index.ts
```diff
@@ -0,0 +1,15 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  API_LIMIT: {
+    description: 'The default maximum number of results to request from the API',
+    type: 'number',
+    default: 100,
+  },
+  BACKGROUND_EXECUTE_MS: {
```
austinborn, line 9: Sucks that this doesn't translate as well at the [framework level]( https://github.com/smartcontractkit/ea-framework-js/blob/main/src/config/index.ts#L327-L345), but this seems fine.

austinborn (dismissed):

packages/scripts/src/generate-readme/readmeBlacklist.json
```diff
@@ -61,6 +61,7 @@
     "view-function-multi-chain",
     "xsushi-price",
     "coinbase-prime",
-    "harris-and-trotter"
+    "harris-and-trotter",
+    "bitgo"
```
mmcallister-cll, line 65: @mxiao-cll let's remove since we made the change in adapter-secrets?

packages/sources/bitgo/src/transport/wallet.ts
```diff
@@ -0,0 +1,230 @@
...
+          prevId: '',
+        },
+        headers: {
+          Authorization: `Bearer ${apiKey}`,
+        },
+      }
+
+      let hasNext = true
+      while (hasNext) {
+        const reqKey =
```
mmcallister-cll, line 188:
```suggestion
        const reqKey = `${requestConfig.baseURL}${requestConfig.url}${JSON.stringify(requestConfig.params)}`
```

mmcallister-cll, line 188: what about
```
const reqKey = `${requestConfig.baseURL}${requestConfig.url}${JSON.stringify(requestConfig.params)}`
```

packages/sources/bitgo/src/transport/wallet.ts
```diff
@@ -0,0 +1,230 @@
...
+  async fetchWalletAddresses(
+    coin: string,
+    walletIds: string[],
+    apiKey: string,
+    apiEndpoint: string,
+    apiLimit: number,
+  ) {
+    const addresses: string[] = []
+
+    for (const walletId of walletIds) {
```
mmcallister-cll, line 173: can we prep all of the requests and fire them at the same time to reduce latency?

Example expectation afaik is ~50 wallets https://chainlink-core.slack.com/archives/C07KYEAS84S/p1725648502950229

packages/sources/bitgo/src/transport/wallet.ts
```diff
@@ -0,0 +1,230 @@
...
+import { getApiInfo } from './utils'
+
+const logger = makeLogger('BitgoTransport')
+
+type RequestParams = typeof inputParameters.validated
+
+export type WalletTransportTypes = BaseEndpointTypes
+
+interface WalletResponse {
+  wallets: {
```
mmcallister-cll, line 18: are we guaranteed to have wallets in this array or could it be empty?

mxiao-cll: I think being empty is a valid case

packages/sources/bitgo/src/transport/wallet.ts
mxiao-cll: I am gonna parallel this for loop but not sure how to deal with the hasNext

mmcallister-cll (approved):

mmcallister-cll (approved):

packages/sources/bitgo/src/transport/wallet.ts
mmcallister-cll: not convinced this would be an issue, so let's just note and fix in a follow up if we notice any weirdness

PR #3423, opened 2024-09-08
Added new MULTI_ADDRESS_LIST EA

karen-stepanyan:
## Description
Added new `multi_address_list` composite EA, that uses Anchorage, Bitgo and Coinbase Prime source EAs to create a single address list from those source EAs. The composite EA calls underlying source EAs via scheduler, at a specified time. 

## Changes

- Added new `multi_address_list` composite EA
- Modified the `proof-of-reserves` EA to support `multi_address_list` as a new protocol
- Modified `anchorage`, `bitgo` and `coinbase-prime` EAs to export input params of their `wallet` endpoint

## Steps to Test

1. Run the underlying source EAs (coinbase prime is already deployed so you can use it instead of running a local instance)
2. Provide required env variables - `ANCHORAGE_ADAPTER_URL` , `BITGO_ADAPTER_URL` and `COINBASE_PRIME_ADAPTER_URL` 
3. Set `SCHEDULER_HOUR` , `SCHEDULER_MINUTES` and `SCHEDULER_TIMEZONE` env vars to appropriate values so you can test the function execution
4. Run the `multi_address_list` EA
5. Make a request to the `multi_address_list` EA before the scheduled execution
6. Wait until executions starts. If underlying EAs are running locally and don't have entries in cache, the cold start might take about a minute, but eventually it will return a value. 

[DF-20419]: https://smartcontract-it.atlassian.net/browse/DF-20419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/composites/multi-address-list/README.md
```diff
@@ -0,0 +1,70 @@
...
+|           | coinbasePrime.batchSize |           |                                                    The number of addresses to fetch at a time                                                     |  number  |                                                 |   `100`   |            |                |
+|           |  coinbasePrime.apiKey   |           | Alternative api keys to use for this request, {$apiKey}_ACCESS_KEY {$apiKey}\_PASSPHRASE {$apiKey}\_SIGNING_KEY required in environment variables |  string  |                                                 |           |            |                |
+|           | coinbasePrime.endpoint  |           |                                                        Endpoint name to make a request to                                                         |  string  |                                                 | `wallet`  |            |                |
+
+### Example
+
+There are no examples for this endpoint.
```
mxiao-cll, line 66: Can we build an example?

packages/composites/multi-address-list/src/endpoint/address-list.ts
```diff
@@ -0,0 +1,90 @@
...
+      type: {
+        ...coinbasePrimeParams.definition,
+        endpoint: {
+          type: 'string',
+          description: 'Endpoint name to make a request to',
+          default: 'wallet',
+        },
+      },
+      description: 'Input parameters for the Coinbase Prime EA',
+    },
```
austinborn, line 49: That's a lot of required inputs, will all 3 always be used?

austinborn (approved):

packages/composites/multi-address-list/src/endpoint/address-list.ts
karen-stepanyan: So right now all 3 custodians will be used. The future is unknown, maybe it will be 2 or 4 :)

packages/sources/bitgo/src/index.ts
```diff
@@ -11,3 +11,4 @@ export const adapter = new PoRAdapter({
 })
 
 export const server = (): Promise<ServerInstance | undefined> => expose(adapter)
+export const walletParameters = wallet.inputParameters
```
mmcallister-cll, line 14: not sure if this is the right way to do this (also applies to the other EAs) but i also don't have any better suggestions 🤔

packages/composites/multi-address-list/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -0,0 +1,61 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`execute address-list endpoint should return success 1`] = `
```
mmcallister-cll, line 3: This doesn't feel representative of the request that would be made from proof-of-reserves. The `network` and `chainId` at least should be the same across all feeds (since the `proof-of-reserves` input params would be passed through to each)

Shouldn't we fail when we receive different networks from the different custodians?

packages/sources/bitgo/src/index.ts
karen-stepanyan: You mean the idea or implementation ? I thought there should be a way to import the params without exporting them from index file. That ended up with lot of errors and didn't want to waste time there. 
There is really two ways, either this composite supports exact same params as underlying sources, or nothing and we forward the raw request body. 
Copy pasting source EA params is not good as any change will require an update to composite EA

packages/composites/multi-address-list/test/integration/__snapshots__/adapter.test.ts.snap
karen-stepanyan: I think this is a fixture issue, but you are right, currently the composite EA doesn't do any validation. In theory prod source EAs should not return assets for different network,  as long as the input data sent to them is correct. I agree that we can add some checks here for `chainId` and `network`.

packages/sources/bitgo/src/index.ts
mxiao-cll: How about we don't do any definition, allow user to pass simply a json blob, this allows us to add and remove additional EAs in multi-address-list without code change

mxiao-cll (approved):

packages/composites/multi-address-list/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Example
+
+Request:
+
+```json
+{
+  "data": {
+    "endpoint": "address-list",
+    "anchorage": {
```
mmcallister-cll, line 72: maybe silly question since i haven't looked at the code yet, but is the idea that for any given request, only the EAs listed in the request will be called?

packages/composites/multi-address-list/README.md
```diff
@@ -0,0 +1,101 @@
...
+      "vaultId": "b0bb2439c1e4926542ce693b4db2e883",
```
mmcallister-cll, line 73: maybe put a different vaultId :P

packages/composites/multi-address-list/src/endpoint/address-list.ts
```diff
@@ -0,0 +1,90 @@
...
+import { addressListTransport } from '../transport/address-list'
+import {
+  PoRAddressEndpoint,
+  PoRAddressResponse,
+} from '@chainlink/external-adapter-framework/adapter/por'
+import { walletParameters as anchorageParams } from '@chainlink/anchorage-adapter'
+import { walletParameters as coinbasePrimeParams } from '@chainlink/coinbase-prime-adapter'
+import { walletParameters as bitGoParams } from '@chainlink/bitgo-adapter'
+
+export const inputParameters = new InputParameters(
```
mxiao-cll, line 12: I felt there is just too much hard-coding here, adding another EA here would require us to make changes but we are essentially just doing a pass-through here

packages/composites/multi-address-list/src/endpoint/address-list.ts
mmcallister-cll: would it work if we set the top level ones to `required: false` and only requested from the ones present? I think this would probably be better for future expansion

packages/composites/multi-address-list/src/endpoint/address-list.ts
```diff
@@ -0,0 +1,90 @@
...
+)
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: PoRAddressResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new PoRAddressEndpoint({
+  name: 'address-list',
```
mmcallister-cll, line 87: I don't feel strongly about this, but in the other address list EAs, we've normally called this `address`

packages/composites/multi-address-list/README.md
```diff
@@ -0,0 +1,101 @@
...
+|           |      bitgo.chainId      |           |                                                           The ID of the chain to return                                                           |  string  |              `mainnet`, `testnet`               | `mainnet` |            |                |
+|           |      bitgo.network      |           |                                                               The network to return                                                               |  string  |                                                 | `bitcoin` |            |                |
+|    ✅     |      bitgo.reserve      |           |                     Used to select {$reserve}_API_KEY {$reserve}\_API_ENDPOINT {$reserve}\_API_LIMIT in environment variables                     |  string  |                                                 |           |            |                |
```
mmcallister-cll, line 52: I'm not a fan of this description carry over, similar for cbp

But for the sake of this PR, let's ignore it

packages/composites/multi-address-list/src/config/index.ts
```diff
@@ -0,0 +1,51 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  ANCHORAGE_ADAPTER_URL: {
+    description: 'URL of Anchorage EA',
+    type: 'string',
+    required: true,
```
mmcallister-cll, line 7: if the specific input params become optional, let's make these optional too

packages/composites/multi-address-list/src/config/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  },
+  SCHEDULER_MINUTES: {
+    description: 'Minute to run scheduler [0-59]',
+    type: 'number',
+    default: 1,
+  },
+  SCHEDULER_TIMEZONE: {
+    description: 'Timezone to run scheduler',
+    type: 'string',
+    default: 'America/New_York',
```
mmcallister-cll, line 32: 😬

packages/sources/bitgo/src/index.ts
mmcallister-cll: the implementation, it's fine for now.

packages/composites/multi-address-list/src/config/index.ts
mmcallister-cll: can we make sure to get sign off for daylight savings at least? there will be an hour gap/overlap that day and i'd rather call it out super early

packages/composites/multi-address-list/src/transport/address-list.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      const response = await this.requester.request<PoRAdapterResponse>(
+        JSON.stringify(requestConfig),
+        requestConfig,
+      )
+      return response.response.data.data.result
+    })
+
+    const addresses = await Promise.all(promises)
```
mmcallister-cll, line 138: didn't we want to fail if one custodian wasn't responding?

packages/composites/multi-address-list/src/endpoint/address.ts
```diff
@@ -0,0 +1,87 @@
...
+    Object.keys(sources).forEach((source) => {
+      const envName = `${source.toUpperCase()}_ADAPTER_URL` as keyof typeof adapterSettings
+      const params = sources[source as keyof typeof sources]
+      if (params && !adapterSettings[envName]) {
+        throw new AdapterInputError({
+          statusCode: 400,
+          message: `Error: missing environment variable ${envName}`,
+        })
+      }
+      return
```
mxiao-cll, line 83: Is this needed?

mxiao-cll (approved):

packages/composites/multi-address-list/README.md
karen-stepanyan: I changed this part. The EA will request only to sources that are in the request

packages/composites/multi-address-list/test/integration/__snapshots__/adapter.test.ts.snap
karen-stepanyan: Fixed. chainId and network are not top level EA params

packages/composites/multi-address-list/src/endpoint/address-list.ts
karen-stepanyan: done

karen-stepanyan:
Tested the `proof-of-reserves` locally with coinbase-prime only, as we don't have prod data yet, works fine. 
```json
{
    "data":  {
       "coinbase_prime": {
        "endpoint": "wallet",
        "batchSize": 100,
        "portfolio": "COINBASE_PRIME_PORTFOLIO",
        "symbols": [
            "BTC"
          ]
       },
       "indexer": "por_indexer",
       "protocol": "multi_address_list",
       "network": "bitcoin",
       "chainId": "mainnet"
    }
}
```

mxiao-cll (dismissed):

packages/composites/multi-address-list/README.md
```diff
@@ -0,0 +1,82 @@
...
+`address` is the only supported name for this endpoint.
+
+### Input Params
+
+| Required? |  Name   | Aliases |          Description          |  Type  |       Options        |  Default  | Depends On | Not Valid With |
+| :-------: | :-----: | :-----: | :---------------------------: | :----: | :------------------: | :-------: | :--------: | :------------: |
+|           | chainId |         | The ID of the chain to return | string | `mainnet`, `testnet` | `mainnet` |            |                |
+|           | network |         |     The network to return     | string |                      | `bitcoin` |            |                |
+
+**Please note that this EA requires additional input parameters for each of the underlying EAs. Please see the READMEs for the individual EAs for more information.**
```
mmcallister-cll, line 46:
```suggestion
**Note that this EA requires additional input parameters for each of the underlying EAs. See their respective READMEs for more information.**
```

packages/composites/multi-address-list/src/config/index.ts
```diff
@@ -41,6 +41,7 @@ export const config = new AdapterConfig({
     description: 'The number of times to retry when execution fails.',
     type: 'number',
     default: 5,
+    validate: validator.integer({ min: 1, max: 10 }),
```
mmcallister-cll, line 47: nit: might be worth allowing for >10

packages/composites/multi-address-list/src/transport/address.ts
```diff
@@ -113,27 +113,31 @@ export class AddressListTransport extends SubscriptionTransport<AddressListTrans
...
-      )
-      return response.response.data.data.result
-    })
+    const { chainId, network, ...sources } = params
+
+    const promises = Object.entries(sources)
+      .filter(([_, sourceParams]) => sourceParams)
+      .map(async ([sourceName, sourceParams]) => {
+        // customInputValidation ensures that if the source EA is present in the input params, the corresponding env variable is also present
+        const envName = `${sourceName.toUpperCase()}_ADAPTER_URL` as keyof typeof this.settings
```
mmcallister-cll, line 122: can change to `adapterUrl` for clarity

mmcallister-cll (approved):

mxiao-cll (dismissed):

mmcallister-cll (approved):

mxiao-cll (approved):

PR #3431, opened 2024-09-12
Removed BITGO multi api key support, add enterprise id support

karen-stepanyan:


packages/sources/bitgo/src/transport/wallet.ts
```diff
@@ -122,19 +121,16 @@ export class WalletTransport extends SubscriptionTransport<WalletTransportTypes>
   ): Promise<AdapterResponse<WalletTransportTypes['Response']>> {
     const { chainId, network } = params
```
mxiao-cll, line 122: Can get enterpriseId here

mxiao-cll (approved):

mxiao-cll (approved):

PR #3433, opened 2024-09-13
[DO NOT MERGE] Refactor Anchorage input params

karen-stepanyan:
Added `coin` input param to Anchorage EA in case we confirm that Anchorage returns multiple assets for a single vault. Merge when it's confirmed


mxiao-cll:
Can we write a test for this? Also wondering if we just use coin to select API_KEYs

karen-stepanyan:
@mxiao-cll What exactly you want a test for ?

mxiao-cll:
That we do the filtering properly, I guess we need to update fixture to return mixed coins? Or is that already set up that way

mxiao-cll (approved):

PR #3452, opened 2024-09-28
elwood: check subscriptions before sending subscribe requests

martin-cll:
Slack context: https://chainlink-core.slack.com/archives/C0290SGU1KR/p1727513673689309

All Elwood connections sharing the same API key will have the same set of subscriptions, e.g. if there are multiple EAs using the same key. To avoid overloading Elwood's servers, we should only add new subscriptions if the current set of subscriptions does not have it. Similarly, we should stop unsubscribing as there is no way to know if another EA needs the asset or not.

This PR adds an API call to first fetch the set of subscribed items and filters the subscribe messages for new ones. It also ignores unsubscribe messages.

Example logs from local server:

Screenshot 2024-09-28 at 23 33 33


packages/sources/elwood/src/transport/crypto.ts
```diff
@@ -90,16 +97,37 @@ export const transport: WebSocketTransport<WsTransportTypes> =
     override async sendMessages(
       context: EndpointContext<WsTransportTypes>,
       subscribes: SubscribeRequest[],
-      unsubscribes: SubscribeRequest[],
     ): Promise<void> {
-      const messages = subscribes.concat(unsubscribes)
+      // All connections that share an API key have the same set of subscribed symbols so if there
+      // are multiple EAs that share the same key then we should only subscribe to new symbols. For
+      // the same reason, it is not safe to unsubscribe otherwise another EA expecting the symbol
```
alecgard, line 103: An EA will unsubscribe from an asset if it the subscription is stale, i.e. if the EA hasn't received a request for that subscription in X time: https://github.com/smartcontractkit/ea-framework-js/blob/main/src/transports/websocket.ts#L402-L404

How does this interact with the above comment - Will this result in EA's unsubscribing from assets that other EA's are still using? Is it common to share a single set of API credentials across Elwood EAs?

alecgard (dismissed):

packages/sources/elwood/src/transport/crypto.ts
martin-cll: > Will this result in EA's unsubscribing from assets that other EA's are still using? Is it common to share a single set of API credentials across Elwood EAs?

Yes currently that is the behaviour which is obviously not great. Hence this PR makes it so that we never unsubscribe. This does mean that the EA will do extra work receiving data for assets that are not needed, but it seems better than the alternative.

martin-cll: Sharing credentials is probably just a Chainlink only problem.

alecgard (approved):

PR #3486, opened 2024-10-17
DF-20626 add bitgo-reserves EA

mmcallister-cll:
## Description
Adding new EA for bitgo reserves
**NOTE** the default endpoint is a test endpoint and not meant to be used in production
**NOTE** no secrets

## Steps to Test
1. yarn test bitgo-reserves
2. sanity test reserves endpoint EA request

## Quality Assurance

[DF-20626]: https://smartcontract-it.atlassian.net/browse/DF-20626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sources/bitgo-reserves/src/config/index.ts
+++ b/packages/sources/bitgo-reserves/src/config/index.ts
@@ -1,0 +1,9 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  API_ENDPOINT: {
+    description: 'An API endpoint for Data Provider',
+    type: 'string',
+    default: 'https://d2nz84bmzl209r.cloudfront.net',
+  },
+})
--- a/packages/sources/bitgo-reserves/src/config/index.ts
+++ b/packages/sources/bitgo-reserves/src/config/index.ts
@@ -4,6 +4,6 @@ export const config = new AdapterConfig({
   API_ENDPOINT: {
     description: 'An API endpoint for Data Provider',
     type: 'string',
-    default: 'https://d2nz84bmzl209r.cloudfront.net',
+    default: 'http://por.usdstandard-test.com',
   },
 })
--- a/packages/sources/bitgo-reserves/src/endpoint/reserves.ts
+++ b/packages/sources/bitgo-reserves/src/endpoint/reserves.ts
@@ -1,0 +1,19 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { config } from '../config'
+import { httpTransport } from '../transport/reserves'
+
+export const inputParameters = new InputParameters({})
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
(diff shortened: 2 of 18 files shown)


packages/sources/bitgo-reserves/src/config/index.ts
```diff
@@ -0,0 +1,9 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  API_ENDPOINT: {
+    description: 'An API endpoint for Data Provider',
+    type: 'string',
+    default: 'https://d2nz84bmzl209r.cloudfront.net',
```
mxiao-cll, line 7: I think we should not set the testing endpoint as default?

packages/sources/bitgo-reserves/src/endpoint/reserves.ts
```diff
@@ -0,0 +1,19 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
```
mxiao-cll, line 1: We can use PoRProviderResponse and PoRProviderEndpoint here

packages/sources/bitgo-reserves/src/transport/reserves.ts
```diff
@@ -0,0 +1,57 @@
...
+          response: {
+            errorMessage: `The data provider didn't return any value`,
+            statusCode: 502,
+          },
+        }
+      })
+    }
+
+    return params.map((param) => {
+      const result = response.data.totalReserve
```
mxiao-cll, line 45: We can read time here and send as part of PoRProviderResponse

packages/sources/bitgo-reserves/src/config/index.ts
mmcallister-cll: so set the real one and overwrite with the test one? or just hold until the real one is ready?

packages/sources/bitgo-reserves/src/transport/reserves.ts
mmcallister-cll: good spot

packages/sources/bitgo-reserves/src/endpoint/reserves.ts
mmcallister-cll: Debatable if there's any value. The advantage of PoRProviderResponse is to force returning a `ripcord` field, which this API+endpoint does/will not have (as far as we know at this point).

Given that we'd just end up forcing ripcord=false, what do you think?

mxiao-cll: I am fine with both

packages/sources/bitgo-reserves/src/config/index.ts
mmcallister-cll: rather set the real one than blank

packages/sources/bitgo-reserves/src/endpoint/reserves.ts
mmcallister-cll: alright, i'm leaning towards without PoRProviderResponse. Happy to revisit if the API changes and they decide to add a ripcord in the future.

mxiao-cll (approved):

PR #3487, opened 2024-10-18
Add "id" field and logging to DLC-BTC-POR BTC RPC request

cawthorne:
"id" is a required field in the jsonrpc spec. We should include it in the request to support all RPC specs.

https://www.jsonrpc.org/specification#notification


packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -210,7 +210,12 @@ export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
       txId,
       requestConfig,
     )
-    return response.data.result
+
+    const result = response.data.result
+
+    if (!result) console.log('BITCOIN_RPC_URL - getrawtransaction failed: ', response)
```
karen-stepanyan, line 216: We need to use a `logger` instance

cawthorne: Done! went with the logger.error method, but open to a different method.

mxiao-cll:
Need a changeset

mmcallister-cll:
would prob be best if it was just 1 log line, but OK as is i guess

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
```diff
@@ -210,7 +210,15 @@ export class DLCBTCPorTransport extends SubscriptionTransport<TransportTypes> {
       txId,
       requestConfig,
     )
-    return response.data.result
+
+    const result = response.data.result
+
+    if (!result) {
+      logger.error('BITCOIN_RPC_URL - getrawtransaction failed, requestConfig: ', requestConfig)
```
karen-stepanyan, line 217: We don't need the `requestConfig` but maybe only `txId`

cawthorne:
> Need a changeset

I intentionally didn't add a changeset, as I was not currently planning on merging this patch, until it was conceptually approved.

packages/sources/dlc-btc-por/src/transport/proof-of-reserves.ts
cawthorne: In a current exchange with a NOP on this, printing out the `requestConfig` might have highlighted the situation faster. However, it isn't a huge value add either.

karen-stepanyan: It looks like they have issues with their RPC. Do you still think this change will add a value ?

cawthorne: I do think it adds, value as their RPC actually follows the jsonrpc spec perfectly, and it wasn't an uptime issue. So it is technically an issue with the EA not with them. The changes they are making now to accommodate the missing id field in their RPC is a work around. Even though most RPCs can handle a missing id field, we should probably support all RPCs which technically follow the spec.

karen-stepanyan (approved):

PR #3498, opened 2024-10-23
market-status: Use NCFX as primary for forex and metals markets

martin-cll:
Use NCFX as primary for forex and metals markets only. All other markets, including future ones, will continue to use TradingHours as primary.

### DIFF
--- a/.changeset/fair-goats-carry.md
+++ b/.changeset/fair-goats-carry.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/market-status-adapter': minor
+---
+
+Use NCFX as primary for forex and metals markets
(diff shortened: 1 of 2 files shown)


dav009 (approved):
ticking but we need to reflect this change in RDD

martin-cll:
> ticking but we need to reflect this change in RDD

cc @dav009 Despite the resemblance, the order of the `secondaryApis` field in RDD ([example](https://github.com/smartcontractkit/reference-data-directory/blob/master/mercury-feeds/feeds/0x000465640194893c34dcc38679866b78bdce1e9f4d4b873195a02eecc6f00e91.json#L15-L18)) does not determine or correspond to the order of the order of the EAs used by this composite adapter. The RDD fields record the fact that these 2 underlying EAs are used, but not in which order.

PR #3506, opened 2024-10-24
EA Telemetry: Include source EA name in composite EA response

cawthorne:
We need the composite EAs to return the source EA name in the response for the EA Telemetry, as this is how it is represented in RDD:

e.g.
```
"meta": { "adapterName:" "SET_TOKEN_INDEX?SOURCE=TIINGO" }
```


packages/composites/crypto-volatility-index/src/config/index.ts
```diff
@@ -2,7 +2,7 @@ import { Requester } from '@chainlink/ea-bootstrap'
 import { Config } from '@chainlink/ea-bootstrap'
 
 export const DEFAULT_NETWORK = 'ETHEREUM'
-export const NAME = 'CRYPTO_VOLATILITY_INDEX'
+export const NAME = 'CVI'
```
mxiao-cll, line 5: Is this safe?

packages/core/bootstrap/src/lib/metrics/index.ts
```diff
@@ -74,9 +74,30 @@ export const withMetrics =
...
+      try {
+        if (
+          input?.data?.source &&
+          typeof input?.data?.source === 'string' &&
+          input?.data?.source !== ''
+        ) {
+          const sourceName = input?.data?.source
+          sourceString = '?SOURCE=' + sourceName.toUpperCase()
+        }
+      } catch (e: any) {
```
mxiao-cll, line 88: What could go wrong here?

packages/composites/crypto-volatility-index/src/config/index.ts
cawthorne: It makes it match what we have in RDD, to help us with EA Telemetry alerting. I am not aware of where this would cause an issue personally.

packages/core/bootstrap/src/lib/metrics/index.ts
cawthorne: Ah yes, I was doing a JSON.parse in there earlier, but I removed that, so try catch is no longer needed.

mxiao-cll (approved):

PR #3514, opened 2024-10-29
Fix build failure

mxiao-cll:
https://github.com/smartcontractkit/external-adapters-js/actions/runs/11574813658/job/32220485779

Removed dangling references


karen-stepanyan:
I think the issue is that the NEXUS_KILN has a reference to enzyme EA in its tsconfig. Why is it needed ? 
I don't think we should change this command

alecgard (approved):

PR #3518, opened 2024-10-29
Upgrade yarn

mxiao-cll:
Bump to 4.5.1 for ts upgrade

Tried build and test a couple of EAs seems to be fine


mmcallister-cll (approved):

mmcallister-cll:
maybe run yarn setup?

PR #3528, opened 2024-10-31
Allow override on cfbenchmarks2

mxiao-cll:
When SECONDARY_API is true, override adapter name to be cfb2.

This allows
 - metrics to show adapter as cfb2
 - Override will try to match cfb2 first, then cfb

### DIFF
--- a/packages/sources/cfbenchmarks/src/endpoint/crypto.ts
+++ b/packages/sources/cfbenchmarks/src/endpoint/crypto.ts
@@ -30,6 +30,11 @@ export const inputParameters = new InputParameters(
       description: 'The symbol of the currency to convert to',
       required: false,
     },
+    adapterNameOverride: {
+      type: 'string',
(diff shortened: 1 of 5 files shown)


packages/sources/cfbenchmarks/src/endpoint/crypto.ts
```diff
@@ -30,6 +30,11 @@ export const inputParameters = new InputParameters(
       description: 'The symbol of the currency to convert to',
       required: false,
     },
+    adapterNameOverride: {
+      type: 'string',
+      description: 'Alternate adapter name to use for override and metrics',
```
karen-stepanyan, line 35: This won't take any effect if set be the user right ? Maybe we can clarify that in the description

mxiao-cll: Good idea, user should not touch this

PR #3537, opened 2024-11-04
Fix underscore in token allocation

mxiao-cll:
Currently `blocksize-capital` is not accepted as input because we expect it to be `blocksize_capital`

To fix this, we simply replace `-` in the input to be `_`


.changeset/gentle-gorillas-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/token-allocation-adapter': minor
+---
+
+Fix blocksize-captial as source
```
mmcallister-cll, line 5: typo

mmcallister-cll (approved):

PR #3583, opened 2024-11-25
Fix bedrock

mxiao-cll:
## Description

Fix bedrock address issue

## Changes

The roles of vault and token are reversed

## Steps to Test

Tested with token-balance EA

## Quality Assurance

### DIFF
--- a/packages/sources/por-address-list/src/transport/bedrockUniBTC.ts
+++ b/packages/sources/por-address-list/src/transport/bedrockUniBTC.ts
@@ -89,14 +89,32 @@ const getAddresses = (
           chainId: 'mainnet',
         }))
         .sort()
-    case 'tokens':
-      return Object.entries(data.evm)
-        .map(([_, value]) => ({
-          chainId: value.chain_id.toString(),
-          contractAddress: value.vault,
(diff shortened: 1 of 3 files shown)


packages/sources/por-address-list/src/transport/bedrockUniBTC.ts
```diff
@@ -89,14 +89,32 @@ const getAddresses = (
...
+    case 'tokens': {
+      const list = Object.entries(data.evm)
+        .flatMap(([_, value]) =>
+          value.tokens.map((token) => ({
+            chainId: value.chain_id.toString(),
+            contractAddress: token,
+            wallet: value.vault,
+          })),
+        )
+        .filter((item) => item.contractAddress != VAULT_PLACEHOLDER)
```
mmcallister-cll, line 101: why not just end it here with `wallets` instead of `wallet`?

mxiao-cll: Since this is a reverse mapping, we are mapping
{
1, [2, 3]
}

into 
{
[2, 1]
[3, 1]
}

So need to flatten the whole thing first

mmcallister-cll (approved):

PR #3587, opened 2024-11-26
Remove NCFX RON/USD Includes Config

cawthorne:
Removes NCFX RON/USD Includes config, as we don't have that feed as a Forex feed (which it was originally added for), only a Crypto feed (which we don't need it for).

As it is effecting RON/USD on the crypto endpoint.

Discussion here:
https://chainlink-core.slack.com/archives/C02UT77UL6R/p1732641514765829

### DIFF
--- a/.changeset/light-walls-look.md
+++ b/.changeset/light-walls-look.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/ncfx-adapter': minor
+---
+
+Remove NCFX RON/USD Includes Config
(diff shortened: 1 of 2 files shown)


mmcallister-cll (dismissed):

mxiao-cll (approved):

.changeset/light-walls-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ncfx-adapter': minor
```
mmcallister-cll, line 2: prob patch since this wasn't used before but this is fine too

mmcallister-cll (approved):

PR #3599, opened 2024-12-03
API Key scan

mxiao-cll:
Scan for leaked API keys during pre commit


karen-stepanyan (approved):

mxiao-cll:
Does not work on public repo

PR #3669, opened 2025-02-03
fix up test

akuzni2:
Fixes test result - the mock getGlvTokenPrice function was missing `longTokenPrice: number[]` so the minimized value was never returning


akuzni2:
have a larger refactor coming so want to get this test fixed so that refactor review shows that this test result doesn't change

mxiao-cll (approved):

PR #3675, opened 2025-02-14
Add timeout + default setting to Arbitrum RPC requests

akuzni2:
## Description

Expose RPC timeout setting as environment variable

## Quality Assurance


mxiao-cll:
Why do we need this?

akuzni2:
> Why do we need this?

@mxiao-cll the default timeout is 2 min in the current ethersjs version ([source](https://github.com/ethers-io/ethers.js/blob/c7c07f5dea5a91d8bc246f6c4f3ac66dc7318300/packages/web/src.ts/index.ts#L122))

It probably might not ever hit this if the background execute timeout is shorter - but if we don't set something reasonable isn't it possible for the background execute to freeze up?

mxiao-cll:
> > Why do we need this?
> 
> @mxiao-cll the default timeout is 2 min in the current ethersjs version ([source](https://github.com/ethers-io/ethers.js/blob/c7c07f5dea5a91d8bc246f6c4f3ac66dc7318300/packages/web/src.ts/index.ts#L122))
> 
> It probably might not ever hit this if the background execute timeout is shorter - but if we don't set something reasonable isn't it possible for the background execute to freeze up?

Q1: What kind of heavy request are we doing here that may exceed the default timeout
Q2: What do you mean `background execute to freeze up` ? If background execute timeout the request will be cancelled AFAIK

PR #3684, opened 2025-02-17
feat: expand l2ep to new chains

KodeyThomas:
## Description

Expands L2EP to 4 new chains (Ink, Mantle, Unichain & Soneium)

......

## Quality Assurance

### DIFF
--- a/.changeset/fast-cobras-sell.md
+++ b/.changeset/fast-cobras-sell.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/layer2-sequencer-health-adapter': minor
+---
+
+Added support for Ink, Mantle, Unichain and Soneium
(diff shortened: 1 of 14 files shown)


mohamed-mehany (approved):
LGTM!
A few non blocking things to follow up on:

1. Can we please check with the teams for those chains if we have sequencer health endpoints? If they already have  it, it's usually a much stronger signal for indication the chain health.
2. Do NOPs already have dedicated RPCs for those chains? I assume yes, but we need to check as we prefer NOPs to use dedicated RPCs as opposed to public / 3P RPCs for L2EP. 
3. We can proceed to make the infra-k8s env changes for testnet with internal RPCs if we have them but not a hard requirement for testnet.

Issue #3705, opened 2025-02-28
[Bug]:🐛Prototype-polluting assignment


odaysec:
https://github.com/smartcontractkit/external-adapters-js/blob/c242f89745186dc922bba0f0911d1adfcc9e4354/packages/core/bootstrap/src/lib/modules/overrider.ts#L119-L119

Most JavaScript objects inherit the properties of the built-in `Object.prototype` object. Prototype pollution is a type of vulnerability in which an attacker is able to modify `Object.prototype`. Since most objects inherit from the compromised `Object.prototype` object, the attacker can use this to tamper with the application logic, and often escalate to remote code execution or cross-site scripting.

One way to cause prototype pollution is by modifying an object obtained via a user-controlled property name. Most objects have a special `__proto__` property that refers to `Object.prototype`. An attacker can abuse this special property to trick the application into performing unintended modifications of `Object.prototype`.

## POC
In the vulnerable below, the untrusted value `req.params.id` is used as the property name `req.session.todos[id]`. If a malicious user passes in the ID value `__proto__`, the variable `items` will then refer to `Object.prototype`. Finally, the modification of `items` then allows the attacker to inject arbitrary properties onto `Object.prototype`.
```ts
let express = require('express');
let app = express()

app.put('/todos/:id', (req, res) => {
    let id = req.params.id;
    let items = req.session.todos[id];
    if (!items) {
        items = req.session.todos[id] = {};
    }
    items[req.query.name] = req.query.text;
    res.end(200);
});
```
One way to fix this is to use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) objects to associate key/value pairs instead of regular objects, as shown below:
```ts
let express = require('express');
let app = express()

app.put('/todos/:id', (req, res) => {
    let id = req.params.id;
    let items = req.session.todos.get(id);
    if (!items) {
        items = new Map();
        req.sessions.todos.set(id, items);
    }
    items.set(req.query.name, req.query.text);
    res.end(200);
});
```
Another way to fix it is to prevent the `__proto__` property from being used as a key, as shown below:
```ts
let express = require('express');
let app = express()

app.put('/todos/:id', (req, res) => {
    let id = req.params.id;
    if (id === '__proto__' || id === 'constructor' || id === 'prototype') {
        res.end(403);
        return;
    }
    let items = req.session.todos[id];
    if (!items) {
        items = req.session.todos[id] = {};
    }
[...]

PR #3706, opened 2025-02-28
fix: Prototype-polluting assignment

odaysec:
## Ticket 🎟️#3705

To fix the prototype pollution vulnerability, we should ensure that the keys used in the `combineOverrides` method cannot be used to modify the `Object.prototype`. One effective way to achieve this is by using a `Map` object instead of a plain object for `combinedOverrides`. This will prevent any prototype pollution since `Map` does not have the same prototype properties as plain objects.

## Quality Assurance


mxiao-cll:
Please run `yarn changeset`

PR #3748, opened 2025-03-25
Support multiplying in implied-price EA

dskloetc:
[DF-21193](https://smartcontract-it.atlassian.net/browse/DF-21193)

## Description

The `implied-price` EA can divide 2 prices from other EA(s).
We now want to be able to multiply 2 prices as well.

The current parameters (dividend* and divisor*) refer specifically to the division operation.
So we rename them to operand1* and operand2* while continuing to support the old parameters as well.

Note that while `dividendInput` and `divisorInput` were marked as optional parameters, their default values were not valid values so in practice those parameters were always required.

Now all parameters are marked as optional because you can choose which set of parameters to use. But without each set of parameters, the source and input parameters are effectively required.

One additional `operation` parameter determines whether to divide or multiply. This allows the new set of parameters to be used for all (including future) functionality of the EA.

## Changes

1. Mark all parameters as optional.
2. Remove default values from `*Input` parameters.
3. Add "legacy" description to old parameters.
4. Add custom validation to determine if parameter sets are not mixed and required parameters of the chosen set are present.
5. Transform the request data to always have the new parameters.
6. Change the code to use the new parameters.
7. Use `operation` to determine whether to divide or multiply.
8. Add additional payload to `test-payload.json`.
9. Add integration tests for using new parameters, multiplying and mixing parameters.
10. Update README.

## Steps to Test

1. Run `aleno` EA (any should work but this is the one I used because I'm familiar with it)
2. Run `implied-price` EA
```
$ curl -X POST http://localhost:8082 -H "Content-Type: application/json" -d '{
  "id": "1",
  "data": {
    "dividendSources": ["aleno"],
    "divisorSources": ["aleno"],
    "dividendInput": {
      "from": "FRAX",
      "to": "USD"
    },
    "divisorInput": {
      "from": "LBTC",
      "to": "USD"
    }
  }
}
'

{"jobRunID":"1","result":"0.00001137192088438718021","providerStatusCode":200,"statusCode":200,"data":{"result":"0.00001137192088438718021"},"meta":{"adapterName":"IMPLIED_PRICE"},"metricsMeta":{"feedId":"{\"data\":{\"dividendSources\":[\"aleno\"],\"divisorSources\":[\"aleno\"],\"dividendInput\":{\"from\":\"FRAX\",\"to\":\"USD\"},\"divisorInput\":{\"from\":\"LBTC\",\"to\":\"USD\"}}}"}}

$ curl -X POST http://localhost:8082 -H "Content-Type: application/json" -d '{
  "id": "1",
  "data": {
    "operand1Sources": ["aleno"],
    "operand2Sources": ["aleno"],
    "operand1Input": {
      "from": "FRAX",
      "to": "USD"
    },
    "operand2Input": {
      "from": "LBTC",
      "to": "USD"
    },                      
    "operation": "MULTIPLY"
  }
}
'
{"jobRunID":"1","result":"87745.994466934539848","providerStatusCode":200,"statusCode":200,"data":{"result":"87745.994466934539848"},"meta":{"adapterName":"IMPLIED_PRICE"},"metricsMeta":{"feedId":"{\"data\":{\"operand1Sources\":[\"aleno\"],\"operand2Sources\":[\"aleno\"],\"operand1Input\":{\"from\":\"FRAX\",\"to\":\"USD\"},\"operand2Input\":{\"from\":\"LBTC\",\"to\":\"USD\"},\"operation\":\"MULTIPLY\"}}"}}
```

## Quality Assurance

[DF-21193]: https://smartcontract-it.atlassian.net/browse/DF-21193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/composites/implied-price/src/endpoint/impliedPrice.ts
+++ b/packages/composites/implied-price/src/endpoint/impliedPrice.ts
@@ -1,4 +1,11 @@
-import { AdapterResponseInvalidError, Requester, util, Validator } from '@chainlink/ea-bootstrap'
+import {
+  AdapterError,
+  AdapterInputError,
+  AdapterResponseInvalidError,
+  Requester,
+  util,
+  Validator,
+} from '@chainlink/ea-bootstrap'
 import type {
   AdapterRequest,
   ExecuteWithConfig,
@@ -14,6 +21,13 @@ export const supportedEndpoints = ['impliedPrice']
 export type SourceRequestOptions = { [source: string]: AxiosRequestConfig }
 
 export type TInputParameters = {
+  operand1Sources: string | string[]
+  operand1MinAnswers?: number
+  operand1Input?: AdapterRequest
+  operand2Sources: string | string[]
+  operand2MinAnswers?: number
+  operand2Input?: AdapterRequest
+  operation?: string
   dividendSources: string | string[]
   dividendMinAnswers?: number
   dividendInput?: AdapterRequest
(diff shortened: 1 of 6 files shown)


packages/composites/implied-price/src/endpoint/impliedPrice.ts
```diff
@@ -23,83 +37,214 @@ export type TInputParameters = {
...
-    dividendInput,
-    dividendMinAnswers,
+    operand1Urls,
+    operand1Input,
+    operand1MinAnswers,
     config,
   )
-  if (dividendResult.isZero()) {
+  if (operand1Result.isZero()) {
     throw new AdapterResponseInvalidError({ message: 'Dividend result is zero' })
```
mxiao-cll, line 218: This is only for when we are dividing right? if mutiplication then we can accept 0?

mxiao-cll (approved):

packages/composites/implied-price/src/endpoint/impliedPrice.ts
dskloetc: Good point but I'm not so sure.
Note that it was already checking for zero for both the dividend and the divisor, even though it's perfectly valid divide 0 by a non-zero number.
I'm not sure why it was checking both but I guess you could argue that while valid, multiplying by 0 is also a degenerate case with an answer that's not really meaningful.

mxiao-cll: I see, can keep it then, but need to update the error message

dskloetc:
I made a new PR to do it as a separate endpoint: https://github.com/smartcontractkit/external-adapters-js/pull/3751

PR #3796, opened 2025-04-09
Fail workflow if test compilation fails

dskloetc:
## Description

In http://github.com/smartcontractkit/external-adapters-js/actions/runs/14356649047/job/40247734286 there were test compilation issues, but the workflow passed anyway.
This is because the step runs [multiple commands](https://github.com/smartcontractkit/external-adapters-js/blob/b607cdeb3c319e2c89ef76cdea7a38c587bdb4d4/.github/workflows/checks.yml#L77-L86) and in this case it only fails if the last command fails.
`set -e` makes it fail as soon as any command fails.

## Changes

1. Add `set -e` before compiling tests in the workflow.

## Steps to Test

1. Steps
3. to
4. test

## Quality Assurance


dskloetc:
Actually, I was confuse and `set -e` is not necessary.
I didn't realized `/core/bootstrap/` was already whitelisted.

PR #3842, opened 2025-05-07
Adding market status endpoint to Finnhub EA

thomjg:
## Description

Adding market status endpoint to Finnhub EA. Will return a 1 if the market is open (`regular` session for market) or a 0 if the market is closed (`pre-market`, `null`, `post-market`).
https://smartcontract-it.atlassian.net/browse/MERC-6946

## Quality Assurance


packages/sources/finnhub/src/endpoint/market-status.ts
```diff
@@ -0,0 +1,31 @@
...
+  MarketStatusResultResponse,
+} from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+
+import { config } from '../config'
+import { markets, transport } from '../transport/market-status'
+
+const inputParameters = new InputParameters({
+  market: {
+    aliases: [],
```
mmcallister-cll, line 12: you can remove this line

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
+import { MarketStatus } from '@chainlink/external-adapter-framework/adapter'
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { makeLogger } from '@chainlink/external-adapter-framework/util'
+
+import type { BaseEndpointTypes } from '../endpoint/market-status'
+
+export const markets = [
```
mmcallister-cll, line 7: could move this out into a separate defs file but here is also ok

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+      return {
+        params: [param],
+        request: requestConfig,
+      }
+    })
+  },
+  parseResponse: (params, res) => {
+    return params.map((param) => {
+      let marketStatus = MarketStatus.CLOSED
```
mxiao-cll, line 135: I think there is also an MarketStatus.unknown?

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+
+const logger = makeLogger('FinnhubMarketStatusEndpoint')
+
+export const transport = new HttpTransport<HttpEndpointTypes>({
+  prepareRequests: (params, config) => {
+    return params.map((param) => {
+      const market = param.market
+      if (!isMarket(market)) {
```
mmcallister-cll, line 116: I think you could do this check in a `customInputValidation` block in the `endpoint/market-status` file to fail fast (throw an AdapterError).

Example: https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/sources/nav-generic/src/endpoint/nav.ts#L34-L37

Reason you'd want to fail fast is to avoid sending a param to the DP you know will be rejected. And the reason you'd want to do it in the endpoint file vs in this `prepareRequests` block is if you return an error here, it will retry outside of the built-in rate limiter, which can be confusing.

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
...
+        logger.warn(`Invalid market in params: ${market}`)
+      }
+      const requestConfig = {
+        baseURL: `${config.API_ENDPOINT}/stock/market-status`,
```
mmcallister-cll, line 120: You could try something like this (from coinmetrics price-http transport)
```js
        baseURL: config.API_ENDPOINT,
        url: `/timeseries/asset-metrics`,
```

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
...
+      return {
+        params: [param],
+        request: requestConfig,
+      }
+    })
+  },
+  parseResponse: (params, res) => {
+    return params.map((param) => {
+      let marketStatus = MarketStatus.CLOSED
+      if (res.data.session !== null) {
```
mmcallister-cll, line 136: you may want to think about the behaviour if you receive a missing data payload from the DP, eg: in a non-200 response.

Otherwise, this looks according to spec (closed if not "regular"). I'm not even sure you need these 4 lines or if they could be compacted to 
```js
const marketStatus = parseMarketStatus(res.data?.session)
```

maybe with `parseMarketStatus(marketStatus: string | null)`

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      return {
+        params: param,
+        response: {
+          result: marketStatus,
+          data: {
+            result: marketStatus,
+          },
+          timestamps: {
+            providerIndicatedTimeUnixMs: new Date(res.data.t).getTime(),
```
mmcallister-cll, line 148: maybe check if `res.data.t` exists first because

```
> new Date(undefined).getTime()
NaN
> new Date(null).getTime()
0
```

packages/sources/finnhub/src/transport/market-status.ts
mmcallister-cll: spec says to default to closed if not "regular" 🤷

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Reduced, no longer necessary

packages/sources/finnhub/src/transport/market-status.ts
thomjg: No longer necessary due to validation on which markets can be supplied

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Done!

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Done!

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Done!

packages/sources/finnhub/src/endpoint/market-status.ts
mmcallister-cll: could add `default` to this input param, but also makes sense as is

packages/sources/finnhub/src/endpoint/market-status.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: MarketStatusResultResponse
+  Settings: typeof config.settings
+}
+
+export const marketStatusEndpoint = new MarketStatusEndpoint({
+  name: 'market-status',
+  aliases: [],
```
mmcallister-cll, line 28: can remove this line

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+export const transport = new HttpTransport<HttpEndpointTypes>({
+  prepareRequests: (params, config) => {
+    return params.map((param) => {
+      const market = param.market
+      const requestConfig = {
+        baseURL: config.API_ENDPOINT,
+        url: `/stock/market-status`,
```
mmcallister-cll, line 33: this can have regular quotes

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,71 @@
+import { MarketStatus } from '@chainlink/external-adapter-framework/adapter'
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+
+import type { BaseEndpointTypes } from '../endpoint/market-status'
+
+export const markets = ['NYSE'] as const
+
+const marketToExchange = new Map<string, string>([['NYSE', 'US']])
```
mmcallister-cll, line 8: This works well, but I think something like tradinghours https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/sources/tradinghours/src/transport/market-status.ts#L11-L15
is more readable

packages/sources/finnhub/test/integration/__snapshots__/adapter-market-status.test.ts.snap
```diff
@@ -0,0 +1,27 @@
...
+{
+  "data": {
+    "result": 2,
+  },
+  "result": 2,
+  "statusCode": 200,
+  "timestamps": {
+    "providerDataReceivedUnixMs": 1641035471111,
+    "providerDataRequestedUnixMs": 1641035471111,
+    "providerIndicatedTimeUnixMs": 1697018041,
```
mmcallister-cll, line 24: Looks like this time is in seconds, may need to *1000

packages/sources/finnhub/src/transport/market-status.ts
```diff
@@ -0,0 +1,71 @@
...
+      const marketStatus = parseMarketStatus(res.data?.session)
+
+      return {
+        params: param,
+        response: {
+          result: marketStatus,
+          data: {
+            result: marketStatus,
+          },
+          ...(res.data?.t && {
```
mmcallister-cll, line 57: lol feel free to do this outside of the return statement too for clarity

thomjg: Updated!

packages/sources/finnhub/test/integration/__snapshots__/adapter-market-status.test.ts.snap
thomjg: Agreed, updated.

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Updated.

packages/sources/finnhub/src/transport/market-status.ts
thomjg: Updated.

mxiao-cll (approved):

PR #3849, opened 2025-05-09
Fix/DF-21299 tiingo includes

mmcallister-cll:
## Description
Update tiingo forex endpoint to correctly recognize includes file

## Changes
- Update forex Endpoint obj from generic AdapterEndpoint to PriceEndpoint ([includes file is only recognized by PriceEndpoints](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/adapter/price.ts#L120) & children): 
- Added ZAR/USD to the includes file (to fetch inverse)
- Tests were failing due to inclusion of apy-finance-test (apy-finance package has a dependency on tiingo and yarn test * matches on glob rather than exact), so added a script to add a trailing `/` to any `yarn test *` command. Open to changing this for a different solution or fixing only for the workflow step if people are strongly against.

## Steps to Test
1. `yarn test tiingo`
2. With prod instance this req body constantly 504s because the forward pair isn't provided by tiingo, fixed in this PR
```json
{
    "data": {
        "base": "ZAR",
        "quote": "USD",
        "endpoint": "forex"
    }
}
```

## Quality Assurance

[DF-21299]: https://smartcontract-it.atlassian.net/browse/DF-21299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mxiao-cll (dismissed):

mxiao-cll (approved):

packages/scripts/src/run-test/run-test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
+
+echo "running EA_PORT=0 METRICS_ENABLED=false jest $@/ ..."
+EA_PORT=0 METRICS_ENABLED=false jest $@/
```
dskloetc, line 4: Was there a reason to add the `/` at the end?
It breaks my workflow for running a specific test:
```
$ yarn test packages/sources/token-balance/test/unit/xrpl.test.ts
No tests found, exiting with code 1
Run with `--passWithNoTests` to exit with code 0
In /Users/dskloet/dev/external-adapters-js/tree7
  7156 files checked.
  testMatch: **/__tests__/**/*.[jt]s?(x), **/?(*.)+(spec|test).[tj]s?(x) - 482 matches
  testPathIgnorePatterns: /node_modules/, /dist/ - 3793 matches
  testRegex:  - 0 matches
Pattern: packages/sources/token-balance/test/unit/xrpl.test.ts - 0 matches
```

dskloetc, line 4: If you want to pass parameters to `jest` exactly as they were passed to the script, you need to put quotes around `"$@"`. Otherwise a single argument containing one or more spaces becomes multiple arguments.

PR #3865, opened 2025-05-14
remove run-test.sh

mmcallister-cll:
## Changes
- removing run-test.sh
- TODO: add slash back if necessary in `Run non-unit, non-integration tests for changed packages` failing step from previous PR

## Quality Assurance


mxiao-cll (commented):
So this is no longer failing?

mmcallister-cll:
@mxiao-cll need to get the workflow to fail to try another fix directly in it. @dskloetc mentioned the addition of run-test.sh breaks his local test setup.

dskloetc:
For context, this was added in https://github.com/smartcontractkit/external-adapters-js/pull/3849

I didn't realize the issue was with the workflow rather than manual invocation.

For unit and integration tests we already add `/test/unit` or `/test/integration` [here](https://github.com/smartcontractkit/external-adapters-js/blob/f2914721ac722dbbd3d52a03c1cd9d0bf75dd3bc/.github/workflows/checks.yml#L109) and [here](https://github.com/smartcontractkit/external-adapters-js/blob/f2914721ac722dbbd3d52a03c1cd9d0bf75dd3bc/.github/workflows/checks.yml#L127).
Do we just need to add `/` in the same way for non-unit-non-integration tests [here](https://github.com/smartcontractkit/external-adapters-js/blob/f2914721ac722dbbd3d52a03c1cd9d0bf75dd3bc/.github/workflows/checks.yml#L150)?

Also, if the problem was that another non-changed package was tested, and that failed, does that mean that that package actually has broken tests? Can we fix those as well?

dskloetc:
> and their package dependencies not being built

Ah, thanks for clarifying.

dskloetc:
> I think this below comment is correct

> > Do we just need to add / in the same way for non-unit-non-integration tests [here](https://github.com/smartcontractkit/external-adapters-js/blob/f2914721ac722dbbd3d52a03c1cd9d0bf75dd3bc/.github/workflows/checks.yml#L150)?

Do you plan to do that in this PR?

PR #3897, opened 2025-06-02
Log warning for missing XRPL_RPC_URL

dskloetc:
## Description

We currently log an error when `token-balance` starts up and `XRPL_RPC_URL` is not set on the environment.
But this env var is only needed when the `xrpl` endpoint is used and not all NOPs will be receiving requests on this endpoint.
So logging this as an error can be confusing and it's better to log it as a warning.

## Changes

Change the message about missing `XRPL_RPC_URL` from an error to a warning.

## Steps to Test

1. Unit tests were adjusted.
2. Tested manually by starting the EA and seeing `"level":"warn"` in
```
{"level":"warn","time":1748849605640,"isoTime":"2025-06-02T07:33:25.640Z","pid":65575,"hostname":"MB-LYCC91HP2M","layer":"Token Balance - XRPL","msg":"Environment variable XRPL_RPC_URL is missing"}
```

## Quality Assurance

### DIFF
--- a/packages/sources/token-balance/src/transport/xrpl.ts
+++ b/packages/sources/token-balance/src/transport/xrpl.ts
@@ -56,7 +56,7 @@ export class XrplTransport extends SubscriptionTransport<BaseEndpointTypes> {
     this.requester = dependencies.requester
 
     if (!adapterSettings.XRPL_RPC_URL) {
-      logger.error('Environment variable XRPL_RPC_URL is missing')
+      logger.warn('Environment variable XRPL_RPC_URL is missing')
     }
   }
 
--- a/.changeset/tender-pigs-shake.md
+++ b/.changeset/tender-pigs-shake.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/token-balance-adapter': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/sources/token-balance/src/transport/xrpl.ts
```diff
@@ -56,7 +56,7 @@ export class XrplTransport extends SubscriptionTransport<BaseEndpointTypes> {
     this.requester = dependencies.requester
 
     if (!adapterSettings.XRPL_RPC_URL) {
-      logger.error('Environment variable XRPL_RPC_URL is missing')
+      logger.warn('Environment variable XRPL_RPC_URL is missing')
```
karen-stepanyan, line 59: Can't we determine if the endpoint is XRPL and error ?

dskloetc: I'm not sure what you mean? If you make a request on the `xrpl` endpoint if `XRPL_RPC_URL` is not set, it will throw an error on line 143.

This log is just here to warn you in advance (in initialization) before you even make a request.

mxiao-cll (approved):

PR #3898, opened 2025-06-03
finalto: Support overrides for stock symbols

martin-cll:
This PR adds support for stock assets for Finalto. Unlike FX, Finalto's stock symbols (e.g. `AAPL.xnas`) have a suffix that is market / exchange specific.

Changes:
* Converts the forex endpoint to a more general quote endpoint.
* Added `forex` (for back compat) and `stock` as aliases of this quote endpoint.
* If the base symbol contains a `.` then use it as the symbol, ignoring the quote.
* Additionally return the `bid`, `mid`, and `ask` fields, similar to crypt LWBA EAs.

### DIFF
--- a/packages/sources/finalto/src/endpoint/quote.ts
+++ b/packages/sources/finalto/src/endpoint/quote.ts
@@ -1,11 +1,11 @@
 import {
-  ForexPriceEndpoint,
+  PriceEndpoint,
   priceEndpointInputParametersDefinition,
 } from '@chainlink/external-adapter-framework/adapter'
-import { InputParameters } from '@chainlink/external-adapter-framework/validation'
 import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
 import { config } from '../config'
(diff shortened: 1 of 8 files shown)


packages/sources/finalto/src/endpoint/quote.ts
```diff
@@ -25,9 +25,9 @@ export type BaseEndpointTypes = {
   Settings: typeof config.settings
 }
 
-export const endpoint = new ForexPriceEndpoint({
-  name: 'forex',
-  aliases: ['fx', 'commodities'],
+export const endpoint = new PriceEndpoint({
```
mxiao-cll, line 28: Is ForexPriceEndpoint -> PriceEndpoint safe?

mxiao-cll (approved):

packages/sources/finalto/src/endpoint/quote.ts
martin-cll: Seems so, ForexEndpoint just wraps PriceEndpoint with no changes: https://github.com/smartcontractkit/ea-framework-js/blob/main/src/adapter/price.ts#L221

PR #3911, opened 2025-06-16
DF-21398 OnRe NAV EA

Subarna-Singh:
## Closes #DF-21398
https://smartcontract-it.atlassian.net/browse/DF-21398 

## Description
OnRe EA to fetch the latest Net Asset Value (NAV) from Onre's endpoint and publish it to a Chainlink Custom Data Feed onchain

## Changes
Build a EA with http transport to query client API and pull latest NAV value.

## Steps to Test
yarn test packages/source/onre

## Quality Assurance


mmcallister-cll:
update all of the dependencies to the ones used in all other EAs
```json
  "devDependencies": {
    "@types/jest": "27.5.2",
    "@types/node": "16.18.119",
    "nock": "13.5.5",
    "typescript": "5.6.3"
  },
  "dependencies": {
    "@chainlink/external-adapter-framework": "2.6.0",
    "tslib": "2.4.1"
  }
```

packages/sources/onre/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const adapter = new Adapter({
+  defaultEndpoint: nav.name,
+  name: 'ONRE',
+  config,
+  endpoints: [nav],
+  rateLimiting: {
+    tiers: {
+      default: {
+        rateLimit1m: 30,
```
mmcallister-cll: this means you're going to make 30 req/min

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+export const httpTransport = new HttpTransport<HttpTransportTypes>({
+  prepareRequests: (params, config) => {
+    return params.map((param) => {
+      return {
+        params: [param],
+        request: {
+          baseURL: config.API_ENDPOINT,
```
mmcallister-cll, line 32: may be worth just sticking the /nav into the default endpoint

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,105 @@
...
+          net_asset_value_date: new Date(year, month - 1, day).getTime(),
+        }
+      })
+
+      const sortedData = dataWithDateMs.sort(
+        (a, b) => b.net_asset_value_date - a.net_asset_value_date,
+      )
+      const latestData = sortedData[0]
+
+      const result_data: ResultType = {
```
mmcallister-cll: camelCase this

packages/sources/onre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      fundId: 1,
+      reportValue: NetAssetValue,
+    },
+  ],
+)
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: SingleNumberResultResponse
```
mmcallister-cll: Try defining the response schema here

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,105 @@
...
+        navPerShare: Number(latestData.net_asset_value),
+        navDate: String(latestData.net_asset_value_date),
+        currency: null,
+        fundId: latestData.fund_id,
+      }
+
+      let result
+      if (param.reportValue === NetAssetValue) {
+        result = Number(latestData.net_asset_value)
+      } else if (param.reportValue === AssetsUnderManagement) {
```
mmcallister-cll: Do we lose precision with the Number() casts?

Separate, I don't think this selector is bad per se, but if we're going to be responding with both values in a single then maybe we don't need this, worth asking SA expectations on how this will be used in the future.

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,105 @@
...
+          params: param,
+          response: {
+            errorMessage: `The data provider didn't return any value for fundId: ${param.fundId}`,
+            statusCode: 502,
+          },
+        }
+      }
+
+      const dataWithDateMs = filteredData.map((item) => {
+        const [month, day, year] = item.net_asset_value_date.split('/').map(Number)
```
mmcallister-cll: is it worth checking some of these fields exist?

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      const sortedData = dataWithDateMs.sort(
+        (a, b) => b.net_asset_value_date - a.net_asset_value_date,
+      )
+      const latestData = sortedData[0]
+
+      const result_data: ResultType = {
+        navPerShare: Number(latestData.net_asset_value),
+        navDate: String(latestData.net_asset_value_date),
+        currency: null,
```
mmcallister-cll: useless to set null

packages/sources/onre/src/transport/nav.ts
Subarna-Singh: I think it's safe to assume. But no harm being extra precautious. 
I'll add the checks.

packages/sources/onre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,52 @@
...
+    {
+      fundId: 1,
+      reportValue: NetAssetValue,
+    },
+  ],
+)
+
+export type NavResultResponse = {
+  Result: number
+  Data: {
```
mmcallister-cll, line 33: I'm wondering if we need another Result field internal to data to support the case where this may need to be a standard single value answer data feed in the future (which may auto pull from `data.result`). In that case you may be able to do something like

```js
export type NavResultResponse = SingleNumberResultResponse & {
  Data: {...
  }
}
```

May be overkill, so I think it's probably OK like this too.

packages/sources/onre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const inputParameters = new InputParameters(
+  {
+    fundId: {
+      description: 'Fund id',
+      type: 'number',
+      required: true,
+    },
+    reportValue: {
+      description: 'Which value to report on',
```
mmcallister-cll: If we're leaving this for the sole purpose of the top level result field then we should be more descriptive here, otherwise think about removing altogether
```suggestion
      description: 'Which value to return as top-level result',
```

packages/sources/onre/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  describe('nav endpoint', () => {
+    it('should return success - nav', async () => {
+      const data = {
+        fundId: 1,
+        reportValue: 'net_asset_value',
+      }
+      mockResponseSuccess()
+      const response = await testAdapter.request(data)
+      expect(response.statusCode).toBe(200)
+      console.log(response.json())
```
mmcallister-cll: remove

packages/sources/onre/src/transport/nav.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+
+      const dataWithDateMs = filteredData.map((item) => {
+        return {
+          ...item,
+          net_asset_value_date: parseDateToTimestamp(item.net_asset_value_date),
+        }
+      })
+
+      dataWithDateMs[1].net_asset_value_date = null
```
mmcallister-cll: i don't understand this line

packages/sources/onre/src/transport/nav.ts
mmcallister-cll: Ok to answer the question, it looks like we may lose precision when we cast to Number. eg an aum value from the test data:
```js
> Number(15172254.24630367655644876519110000)
15172254.246303676
```

In the updated code, we have
```
navPerShare: Number(latestData.net_asset_value),
```
do we have to cast this or can this be a string in the response schema?

packages/sources/onre/src/transport/nav.ts
Subarna-Singh: Ignore. That should not be there. I was testing something.

packages/sources/onre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: NavResultResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'reserves',
+  aliases: ['por', 'nav'],
```
mmcallister-cll:
```suggestion
  aliases: ['nav'],
```

packages/sources/onre/src/transport/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+export function parseDateToTimestamp(dateString: string): number | null {
+  const [month, day, year] = dateString.split('/').map(Number)
+
+  if (![month, day, year].every(Number.isFinite)) return null
+
+  const timestamp = Date.UTC(year, month - 1, day, 0, 0, 0)
+  return Number.isFinite(timestamp) ? timestamp : null
+}
+
+export function sanityCheckData(data: FundNavDataWithTimestamp) {
```
mmcallister-cll, line 12: this could be 1 line

packages/sources/onre/src/endpoint/nav.ts
mmcallister-cll:
```suggestion
  aliases: ['nav'],
```

mmcallister-cll (approved):

PR #3916, opened 2025-06-24
Feat/ice

vlfig:
## Description

This introduces the adapter for ICE Fx price feeds, via NetDania.

## Changes

- This, for now, makes use of a Typescriptified version of the `jsApi` client library, which is wrapped by our `netdania/StreamingClient` which contains all references to the client library and serves as a façade for future reimplementation or refactoring.

- The ICE adapter follows a LWBA response type. Because the server sends only partial updates, and the cache provided by the framework does not allow for in-place updates (i.e. cache.update taking a closure) nor reads, we keep a local cache so that we can produce valid updates by coalescing partial updates with previous data. 

## Steps to Test

1. `yarn test packages/sources/ice/test/` on project root 

## Quality Assurance


.changeset/tough-hornets-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ice-adapter': patch
```
karen-stepanyan:
```suggestion
'@chainlink/ice-adapter': major
```

packages/sources/ice/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@chainlink/ice-adapter",
+  "version": "0.1.0",
```
karen-stepanyan:
```suggestion
  "version": "0.0.0",
```

packages/sources/ice/src/endpoint/price.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  AdapterEndpoint,
+  LwbaResponseDataFields,
+} from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import overrides from '../config/overrides.json'
+import { transport } from '../transport/price'
+
+// Input parameters define the structure of the request expected by the endpoint. The second parameter defines example input data that will be used in EA readme
```
karen-stepanyan: Can we remove those auto generated comments ?

packages/sources/ice/src/transport/netdania/index.ts
```diff
@@ -0,0 +1,332 @@
...
+      password: cfg.NETDANIA_PASSWORD,
+    })
+
+    this.connection.addListener(Events.ONINIT, function (info: string) {
+      logger.info('ONINIT:' + info + '\n')
+    })
+
+    this.connection.addListener(Events.ONCONNECTED, (sessionId: string) => {
+      // Upon successful connection and receipt of a session ID, instruments can be requested.
+      logger.debug('ONCONNECTED: ' + sessionId)
```
karen-stepanyan:
```suggestion
      logger.info('ONCONNECTED: ' + sessionId)
```

packages/sources/ice/src/transport/netdania/index.ts
```diff
@@ -0,0 +1,332 @@
...
+    })
+
+    this.connection.addListener(Events.ONRECONNECTED, (sessionId: string) => {
+      logger.debug('ONRECONNECTED: ' + sessionId)
```
karen-stepanyan:
```suggestion
      logger.info('ONRECONNECTED: ' + sessionId)
```

NetDania-License.txt
```diff
@@ -0,0 +1,9 @@
+NetDania License Terms
```
karen-stepanyan, line 1: should this be in the root ?

packages/sources/ice/src/transport/price.ts
```diff
@@ -0,0 +1,141 @@
...
+    dependencies: TransportDependencies<BaseEndpointTypes>,
+    adapterSettings: BaseEndpointTypes['Settings'],
+    endpointName: string,
+    transportName: string,
+  ) {
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+    this.client = new StreamingClient(adapterSettings)
+    this.client.on('price', async (update: InstrumentPartialUpdate) => {
+      // get base and quote from the instrument name in the requestIdToInstrument map
+      const base = update.instrument.substring(0, 3)
```
karen-stepanyan, line 41: this limits to 3 chars tickers only

packages/sources/ice/src/transport/price.ts
```diff
@@ -0,0 +1,141 @@
...
+      this.localCache.drop(instruments)
+    }
+  }
+
+  private subscribe(fresh: { base: string; quote: string }[]) {
+    if (fresh.length <= 0) {
+      return []
+    }
+
+    logger.info(`Subscribing to pairs ${fresh}`)
```
karen-stepanyan:
```suggestion
    logger.info(`Subscribing to pairs ${JSON.stringify(fresh)}`)
```
same above

packages/sources/ice/src/transport/netdania/index.ts
```diff
@@ -0,0 +1,332 @@
...
+    })
+
+    // triggered when the client is disconnected from the server, followed by an automatic reconnection attempt.
+    this.connection.addListener(Events.ONDISCONNECTED, (msg: string, url: string) => {
+      // Log the disconnection message and URL for reference.
+      logger.warn('ONDISCONNECTED: ' + msg + ' ' + url + '\n')
```
karen-stepanyan: in all of those cases where we log the url, it will contain the encoded credentials in the queries.

NetDania-License.txt
vlfig: Yes, they specifically asked for that.

packages/sources/ice/src/transport/price.ts
vlfig: Yes. All of them are like so. Maybe we should add a custom config validation rule instead.

packages/sources/ice/src/transport/netdania/index.ts
vlfig: I'd love to be able to say that it will only happen after connecting, which can only happen when the url necessarily contains the `sessid` parameter and does not contain the `h` parameter.

Let me dig a bit to see if we can be sure of this. If not, we'll ensure that we either redact or drop it.

karen-stepanyan: I don't think logging URL is needed, considering the host is also a secret.

vlfig: ended up just redacting the h param, in case

vlfig: Sorry, I missed your last message before writing mine.

I think I'd keep the rest of the url for ops visibility. The restriction for host names is in terms of source code distribution, not completely a secret per se.

karen-stepanyan (dismissed):

packages/sources/ice/src/endpoint/price.ts
```diff
@@ -0,0 +1,42 @@
+import { LwbaEndpoint, LwbaResponseDataFields } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import overrides from '../config/overrides.json'
+import { transport } from '../transport/price'
+
+export const inputParameters = new InputParameters(
```
karen-stepanyan: neat
```suggestion
export const inputParameters = new InputParameters(
  lwbaEndpointInputParametersDefinition,
  [
    {
      base: 'EUR',
      quote: 'USD',
    },
  ],
)
```

karen-stepanyan (approved):
one small comment, not blocker

packages/sources/ice/src/config/overrides.json
mmcallister-cll, line 1: can remove for now

PR #3919, opened 2025-06-25
Merc6933 GMCI EA -  Data Link Feed

Subarna-Singh:
## Description
GMCI EA 
- to get price value and rebalnce status of GMCI30 Index
- uses websocket transportation
- EA subscribes to two types of messages 
   - price [[LINK](https://api.gmci.co/docs#:~:text=%7B-,%22topic%22%3A%20%22price%22%2C,-%22data%22%3A%20%5B%0A%20%20%20%20%20%20%20%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22symbol%22%3A%20%22GML2)]
   - rebalance_status [[LINK](https://api.gmci.co/docs#:~:text=%22topic%22%3A%20%22rebalance_status%22%2C)]
 - The EA pricesses the incoming messages and stores them in the cache. When data of oth types of messages are available in the cache, the EA combines them to form the EA response.

(https://smartcontract-it.atlassian.net/browse/MERC-6933)

## Changes
- Create New Adapter for GMCI EA
- Override buildConnectionHandlers to allow cacheing price and rebalance_status response, and combine the responses to compose the EA response.

## Steps to Test
yarn test packeages/source/gmci

## Quality Assurance

- [ x] If a new adapter was made, or a new endpoint was added, update the `test-payload.json` file with relevant requests.


.changeset/itchy-pianos-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/gmci-adapter': minor
```
mmcallister-cll: major?

packages/sources/gmci/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "devDependencies": {
+    "@sinonjs/fake-timers": "9.1.2",
+    "@types/jest": "27.5.2",
```
mmcallister-cll: update these to what is used in the other EAs so you don't have to merge these extra older yarn cache files

packages/sources/gmci/src/config/overrides.json
mmcallister-cll, line 1: can remove?

packages/sources/gmci/src/endpoint/price.ts
```diff
@@ -0,0 +1,45 @@
...
+)
+
+export type GMCIResultResponse = {
+  Result: number
+  Data: {
+    status?: string
+    end_time?: string
+    start_time?: string
+    symbol?: string
+    result?: number
```
mmcallister-cll: are we ok returning if these all don't exist?

packages/sources/gmci/src/endpoint/price.ts
```diff
@@ -0,0 +1,45 @@
...
+  Response: GMCIResultResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'price',
+  aliases: [],
+  transport: transport,
+  inputParameters,
+  overrides: overrides['gmci'],
```
mmcallister-cll: remove this line if the overrides file is removed, and can remove the empty aliases line too

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,175 @@
...
+            endpointName: 'price',
+            adapterSettings: config.settings,
+          }),
+        )
+
+        const rebalance_data = await this.responseCache.cache.get(
+          calculateCacheKey({
+            transportName: this.name,
+            data: { index: data.symbol, type: 'rebalance_status' },
+            adapterName: 'GMCI',
```
mmcallister-cll: you might be able to import adapter from the top level index and use `adapter.name` instead of hardcoding this

packages/sources/gmci/test/integration/fixtures.ts
```diff
@@ -0,0 +1,20 @@
+import { MockWebsocketServer } from '@chainlink/external-adapter-framework/util/testing-utils'
+
+export const mockWebsocketServer = (URL: string): MockWebsocketServer => {
+  const mockWsServer = new MockWebsocketServer(URL, { mock: false })
+  mockWsServer.on('connection', (socket) => {
+    socket.on('message', (message) => {
+      return socket.send(
+        JSON.stringify({
```
mmcallister-cll, line 8: Is this the response we're expecting from the WS API?

packages/sources/gmci/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { mockWebsocketServer } from './fixtures'
+
+describe('websocket', () => {
+  let mockWsServer: MockWebsocketServer | undefined
+  let testAdapter: TestAdapter
+  const wsEndpoint = 'ws://localhost:9090'
+  let oldEnv: NodeJS.ProcessEnv
+
+  const dataPrice = {
+    base: 'ETH',
```
mmcallister-cll: are we going to be querying this EA for ETH/USD?

packages/sources/gmci/test/integration/fixtures.ts
Subarna-Singh: I am yet to add testing. Ignore the test files for now.

packages/sources/gmci/package.json
Subarna-Singh: Done

packages/sources/gmci/src/endpoint/price.ts
Subarna-Singh: Done

packages/sources/gmci/src/config/overrides.json
Subarna-Singh: Done.

.changeset/itchy-pianos-wave.md
Subarna-Singh: FIXED

packages/sources/gmci/src/endpoint/price.ts
Subarna-Singh: Fixed

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: Done

packages/sources/gmci/test/integration/fixtures.ts
Subarna-Singh: Added.

packages/sources/gmci/test/integration/adapter.test.ts
Subarna-Singh: Added Testing

.changeset/itchy-pianos-wave.md
dskloetc: Just in case you don't know. You can just edit these changeset files. You don't need to run the command again. Because now you have 2 changeset files which seems unnecessary for a single PR.
Can you remove one of them and make sure the other one is correct?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,189 @@
...
+              data: {
+                result: item.price,
+                symbol: item.symbol,
+              },
+              timestamps: {
+                providerIndicatedTimeUnixMs: convertTimetoUnixMs(item.last_updated),
+              },
+            },
+          }
+          transport.price_cache.set(item.symbol, result, 90000)
```
dskloetc: Since this is async, there is no guarantee that this result is in the cache by the time you call `processMessage`.
You should probably just move the entire body here inside `processMessage`.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,189 @@
...
+    WsMessage: WSResponse
+  }
+}
+
+export class GMCIWebsocketTransport extends WebSocketTransport<WsTransportTypes> {
+  price_cache = new LocalCache<CachedPriceResponse>(10)
+  rebalance_status_cache = new LocalCache<CachedRebalanceResponse>(10)
+}
+
+export async function processMessage(symbols: Array<string>, transport: GMCIWebsocketTransport) {
```
dskloetc: Can you make this a method on `GMCIWebsocketTransport`.
It's a bit unfortunate that this function is using `transport` as a global variable.
If we move it in as a message, it has a cleaner access to everything inside the transport.
Then we still refer to the global `transport` where the method is called, but at least then it's a single isolated reference.

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  const context = {} as EndpointContext<WsTransportTypes>
+
+  it('builds correct unsubscribe message for a symbol', () => {
+    const message = builders.unsubscribeMessage!({ index: 'GMCI30' }, context) as any
+    expect(message.op).toBe('unsubscribe')
+    expect(message.args).toContain('price.gmci30')
+    expect(message.args).toContain('rebalance_status.gmci30')
+    expect(message.args.length).toBe(2)
+  })
+})
```
dskloetc, line 134: Also add unit tests for the transport itself.

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: Done.

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: Done

packages/sources/gmci/test/unit/price.test.ts
Subarna-Singh: Done

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -43,3 +50,136 @@ describe('builders.unsubscribeMessage', () => {
...
+    }
+
+    const rebalanceMsg: RebalanceMessage = {
+      symbol: 'GMCI30',
+      start_time: '2025-07-14T12:00:00Z',
+      end_time: '2025-07-14T13:00:00Z',
+      status: 'open',
+    }
+
+    transport.price_cache.set(
```
dskloetc: Can you mock the websocket connection and get this into the cache by sending the corresponding message over the mock websocket?

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -43,3 +50,136 @@ describe('builders.unsubscribeMessage', () => {
...
+      'GMCI30',
+      {
+        params: { index: 'GMCI30', type: 'price' },
+        response: {
+          result: 184,
```
dskloetc: declare variables for the values (price, symbol, timestamps, etc.) so that it's clear that the values you start with are the values you end up with.

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -43,3 +50,136 @@ describe('builders.unsubscribeMessage', () => {
...
+      topic: 'rebalance_status',
+      data: [rebalanceMsg],
+    }
+
+    await transport.processMessage(transport, priceWsMessage)
+    await transport.processMessage(transport, rebalanceWsMessage)
+
+    expect(mockWrite).toHaveBeenCalledTimes(2)
+    const [callArg] = mockWrite.mock.calls[0]
+    expect(callArg).toBe(transport.name)
```
dskloetc: Do you know `expect(mockWrite).toHaveBeenNthCalledWith(...)`?
Then you don't need to reach into the inner workings of the mock.

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -43,3 +50,136 @@ describe('builders.unsubscribeMessage', () => {
...
+
+    const [result] = mockWrite.mock.calls[0][1]
+    expect(result.params.index).toBe('GMCI30')
+    expect(result.response.data.symbol).toBe('GMCI30')
+    expect(result.response.data.status).toBe('open')
+    expect(result.response.data.result).toBe(184)
+    expect(result.response.timestamps.providerIndicatedTimeUnixMs).toBe(1752494400000)
+    expect(result.response.timestamps.providerDataReceivedUnixMs).toBe(now)
```
dskloetc: I think it's better to expect the value of the entire argument, rather than expect specific individual fields.
That way you know you're not missing something you didn't think of or which was added later.

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -43,3 +50,136 @@ describe('builders.unsubscribeMessage', () => {
...
+  })
+
+  it('skips writing to cache if any data is missing', async () => {
```
dskloetc: Also add a test where rebalance status is set but price is not.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -121,44 +160,7 @@ export const options: WebSocketTransportConfig<WsTransportTypes> = {
         return
```
dskloetc, line 57: Do we expect unsuccessful messages that we don't care about?
Or should we log something?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
+            timestamps: {
+              providerIndicatedTimeUnixMs: convertTimetoUnixMs(item.last_updated),
+            },
           },
-          timestamps: {
-            providerDataStreamEstablishedUnixMs: transport.providerDataStreamEstablished,
-            providerDataReceivedUnixMs: now,
-            providerIndicatedTimeUnixMs: price_data.response.timestamps.providerIndicatedTimeUnixMs,
+        }
+        transport.price_cache.set(item.symbol, result, 90000)
```
dskloetc: `transport` is the same as `this`, right?
Can you just remove the `transport` parameter from the method?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
-        params: { index: symbol },
-        response: {
-          result: price_data.response.result,
-          data: {
-            result: price_data.response.result,
-            status: rebalance_data.response.data.status,
-            end_time: rebalance_data.response.data.end_time,
-            start_time: rebalance_data.response.data.start_time,
-            symbol,
+  async processMessage(transport: GmciWebsocketTransport, message: WsResponse) {
```
dskloetc: This method is really long. Let's try to bring some structure to it.
First of all, it's a bit confusing that one message can contain multiple message.
Would it make sense to use a different name for the inner "messages"? Maybe we can call them "updates"?
So assuming we'll call the inner messages "updates", how about something like this:

```
async processMessage(message: WsResponse): Promise<void> {
  if (message.topic === 'price') {
    return this.processPriceMessage(message)
  } else if (message.topic === 'rebalance_status') {
    return this.processRebalanceStatusMessage(message)
  }
}
```

Then we can have
```
async processPriceMessage(message: WsResponse) {
  await Promise.all(message.data.map(this.processPriceUpdate.bind(this)))
  await this.updateResponseCache(message)
}
```

etc.

packages/sources/gmci/src/transport/price.ts
dskloetc: We should await the cache update to make sure it's updated before we read from it in order to update the response cache.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
+              symbol: item.symbol,
+            },
           },
-        },
+        }
+        transport.rebalance_status_cache.set(item.symbol, result, 90000)
       }
+    }
+
+    if (message.topic === 'price' || message.topic === 'rebalance_status') {
```
dskloetc: Do we expect other topics as well or should we throw/log an error if we get a different topic?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
+      const symbols: Array<string> = (message.data as any[]).map((d) => d.symbol)
+
+      for (const symbol of symbols) {
+        const [cached_price_data, cached_rebalance_data] = await Promise.all([
+          transport.price_cache.get(symbol),
+          transport.rebalance_status_cache.get(symbol),
+        ])
+
+        if (cached_price_data != undefined && cached_rebalance_data != undefined) {
```
dskloetc: Use `!==` instead of `!=`, especially when dealing with falsy values.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
+          const now = Date.now()
+
+          const price_data = cached_price_data as CachedPriceResponse
```
dskloetc: Why do we need the cast? Isn't something coming from `price_cache.get` always `undefined` or `CachedPriceResponse`?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
 export class GmciWebsocketTransport extends WebSocketTransport<WsTransportTypes> {
   price_cache = new LocalCache<CachedPriceResponse>(10)
```
dskloetc: What is the 10 here?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -68,41 +68,80 @@ export type WsTransportTypes = BaseEndpointTypes & {
...
+                status: rebalance_data.response.data.status,
+                end_time: rebalance_data.response.data.end_time,
+                start_time: rebalance_data.response.data.start_time,
+                symbol,
+              },
+              timestamps: {
+                providerDataStreamEstablishedUnixMs: transport.providerDataStreamEstablished,
+                providerDataReceivedUnixMs: now,
+                providerIndicatedTimeUnixMs:
+                  price_data.response.timestamps.providerIndicatedTimeUnixMs,
```
dskloetc: What is conceptually the difference between `providerDataReceivedUnixMs` and `providerIndicatedTimeUnixMs`?
And why do we use the price_data timestamp as opposed to the rebalance_data timestamp for `providerIndicatedTimeUnixMs`?

Would it make sense to use the maximum of both timestamps for `providerDataReceivedUnixMs`, rather than creating a new `now` timestamp?

packages/sources/gmci/src/transport/price.ts
dskloetc: If either of them is `undefined`, return early.
Then you avoid the extra indentation in the rest of the method.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,191 @@
...
+      end_time: string
+      start_time: string
+      symbol: string
+    }
+  }
+}
+
+export interface WsResponse {
+  success: boolean
+  data: Array<PriceMessage | RebalanceMessage>
```
dskloetc: We never mix types of message in a single array, right?
If so `data: Array<PriceMessage> | Array<RebalanceMessage>` would be more accurate.
Or with the following, the compiler could even deduce the type of array from the value of `topic`:
```
interface WsPriceResponse {
  success: boolean
  data: Array<PriceMessage>
  topic: 'price'
}

interface WsRebalanceResponse {
  success: boolean
  data: Array<RebalanceMessage>
  topic: 'rebalance_status'
}

export type WsResponse = WsPriceResponse | WsPriceResponse
```

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,191 @@
...
+import { convertTimetoUnixMs } from './util'
+
+export interface PriceMessage {
+  last_updated: string
+  price: number
+  symbol: string
+}
+
+export interface CachedPriceResponse {
+  params: {
```
dskloetc: I don't see where these `params` are used.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,191 @@
...
+    index: string
+    type: string
+  }
+  response: {
+    result: number
+    data: {
+      result: number
```
dskloetc: This is just an intermediate data object. We don't need to put the same information multiple times in it, right?

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: We should expect unsuccessful message. Added log for this.

Subarna-Singh:
@dskloetc I have a question. 
For the WS_API_ENDPOINT, do we keep a default? Since the same EA can now be used for GMCI and Wintermute.

packages/sources/gmci/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
...
+    description: 'An API key for Data Provider',
+    type: 'string',
+    required: true,
+    sensitive: true,
+  },
+  WS_API_ENDPOINT: {
+    description: 'WS endpoint for GMCI Data Provider',
+    type: 'string',
+    // default: 'wss://api.gmci.co/private',
+    default: 'wss://generic-alb-ty.api.wintermute-rfq.xyz:15494/private',
```
dskloetc: Is this for testing?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,88 @@
...
+      if (message.success === false) {
+        return
+      }
+
+      const results = []
+
+      if (message.topic === 'price') {
+        for (const item of message.data as PriceMessage[]) {
+          results.push({
+            params: { index: item.symbol },
```
dskloetc: Why do we call it `index` if it's the same as `symbol`? Would it make sense to call it `symbol` everywhere?

packages/sources/gmci/src/transport/price.ts
dskloetc: I don't see a log.

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,88 @@
...
+            response: {
+              result: item.price,
+              data: {
+                result: item.price,
+                symbol: item.symbol,
+              },
+              timestamps: {
+                providerIndicatedTimeUnixMs: convertTimetoUnixMs(item.last_updated),
```
dskloetc, line 73: Weren't there multiple different timestamps before? How do we decide which timestamps to provide?

packages/sources/gmci/src/transport/price.ts
```diff
@@ -0,0 +1,88 @@
...
+  handlers: {
+    message(message) {
+      if (message.success === false) {
+        return
+      }
+
+      const results = []
+
+      if (message.topic === 'price') {
+        for (const item of message.data as PriceMessage[]) {
```
dskloetc: I think if we do [this](https://github.com/smartcontractkit/external-adapters-js/pull/3919/files#r2210349712), we can avoid a cast here.

packages/sources/gmci/test/unit/price.test.ts
dskloetc: Did you remove it again?

packages/sources/gmci/src/config/index.ts
Subarna-Singh: https://github.com/smartcontractkit/external-adapters-js/pull/3919#issuecomment-3083261780

Thats my question. We should remove the defaultaccording to me. Forces us to set endpoint for GMCI and wintermute separately.

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: Yup. Makes sense.

packages/sources/gmci/src/transport/price.ts
Subarna-Singh: The others get written [here](https://github.com/smartcontractkit/ea-framework-js/blob/a618c5d6242cb69650bad6c92d57652524635e3a/src/transports/websocket.ts#L172C3-L172C26). 
Since we had write to the cache ourselves, we have to this bit of work in the previous version.

packages/sources/gmci/src/config/index.ts
dskloetc: If there is no obvious single default then it makes sense not to have a default at all.

packages/sources/gmci/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Price Endpoint
+
+`price` is the only supported name for this endpoint.
+
+### Input Params
+
+| Required? | Name  | Aliases | Description |  Type  | Options | Default | Depends On | Not Valid With |
+| :-------: | :---: | :-----: | :---------: | :----: | :-----: | :-----: | :--------: | :------------: |
+|    ✅     | index |         | Index name  | string |         |         |            |                |
```
dskloetc: Can you recreate the README now that the name of the parameter is changed?

packages/sources/gmci/test/unit/price.test.ts
```diff
@@ -35,9 +35,82 @@ describe('builders.unsubscribeMessage', () => {
...
+      ],
+    }
+
+    const results = handlers.message(mockMessage, context)
+
+    expect(results).toHaveLength(2)
+    expect(results?.[0].params.symbol).toBe('GMCI30')
+    expect(results?.[1].params.symbol).toBe('GML2')
+  })
+})
```
dskloetc, line 134: Can you add a test that rebalance messages don't cause problems?

PR #3921, opened 2025-06-26
DF21398 OnRe NAV EA

Subarna-Singh:
## Description
Change API endpoint for OnRe and add Ripcord

## Quality Assurance


packages/sources/onre/src/transport/nav.ts
```diff
@@ -100,6 +104,9 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
         response: {
           result: result,
           data: resultData,
+          ripcord: Number(response.data.ripcord),
+          ripcordValue: response.data.ripcord,
+          ripcordDettails: response.data.ripcordDetails,
```
alecgard:
```suggestion
          ripcordDetails: response.data.ripcordDetails,
```

alecgard (changes requested):

packages/sources/onre/src/transport/nav.ts
alecgard: `ripcordDetails` will be an `object` - do we want to extract the reason? Else consumers will need to be aware of the details object schema

Subarna-Singh: We have no data on what it looks like. We have just been told, it'll be a string. At the moment it is null

alecgard: > We have just been told, it'll be a string

Then should it be `string | null` in the type? https://github.com/smartcontractkit/external-adapters-js/blob/DF-21398-OnRe-NAV/packages/sources/onre/src/transport/nav.ts#L16

alecgard (approved):

PR #3924, opened 2025-07-03
 Adds LibreCapital nav EA 

mohamed-mehany:
## Changes

- Adds a new external adapter to retrieve and return the latest NAV for a specified LibreCapital fund

## Steps to Test

1. Integration tests
2. Via local server
Test payload
```
POST http://127.0.0.1:8080
{
  "data": {
    "endpoint": "nav",
    "globalFundID": 1234

  }
}

```

## Quality Assurance

[DF-21483]: https://smartcontract-it.atlassian.net/browse/DF-21483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DF-21507]: https://smartcontract-it.atlassian.net/browse/DF-21507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/nav-libre/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "devDependencies": {
+    "@types/crypto-js": "^4",
+    "@types/jest": "27.5.2",
+    "@types/node": "16.18.119",
+    "nock": "13.5.5",
+    "typescript": "5.6.3"
```
mmcallister-cll: could you update these again to whatever is present across the other EAs? 🙏

packages/sources/nav-libre/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "dependencies": {
+    "@chainlink/external-adapter-framework": "2.6.0",
+    "crypto-js": "^4.2.0",
+    "dayjs": "^1.11.13",
```
mmcallister-cll: not sure if this is a new package, but we usually push the yarn cache package too

packages/sources/nav-libre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,53 @@
...
+    globalFundID: {
+      required: true,
+      type: 'number',
+      description: 'The global fund ID for the Libre fund',
+    },
+    fromDate: {
+      required: false,
+      type: 'string',
+      description: 'Start date in MM-DD-YYYY format (defaults to 7 days ago)',
+      default: dayjs().subtract(7, 'day').format('MM-DD-YYYY'),
```
mmcallister-cll: interesting, this looks nice! does it default to UTC?

packages/sources/nav-libre/package.json
mohamed-mehany: Fixed, thanks!

packages/sources/nav-libre/package.json
mohamed-mehany: removed that one in the latest revision

packages/sources/nav-libre/src/endpoint/nav.ts
mohamed-mehany: Had to remove this one too unfortunately

packages/sources/nav-libre/src/transport/fund.ts
```diff
@@ -0,0 +1,66 @@
...
+  }[]
+}
+
+export const getFund = async (
+  globalFundID: number,
+  fromDate: string,
+  toDate: string,
+  baseURL: string,
+  apiKey: string,
+  secret: string,
```
dskloetc: With so many parameters of the same type, it's easy to confuse the order and the compiler won't catch it.
How about using named parameters instead of positional parameters?

(Also consider for other functions)

packages/sources/nav-libre/src/transport/nav.ts
```diff
@@ -0,0 +1,113 @@
...
+      this.config.API_KEY,
+      this.config.SECRET_KEY,
+      this.requester,
+    )
+
+    // Find the latest NAV entry by Accounting Date
+    const latest = fund.reduce((a, b) => {
+      return new Date(a['Accounting Date']) > new Date(b['Accounting Date']) ? a : b
+    })
+    const [month, day, year] = latest['Accounting Date'].split('-').map(Number)
```
dskloetc: You already import `date-fns`. Why not use that instead of doing manual date parsing here?

packages/sources/nav-libre/src/transport/nav.ts
```diff
@@ -0,0 +1,113 @@
...
+      toDateString(to),
+      this.config.API_ENDPOINT,
+      this.config.API_KEY,
+      this.config.SECRET_KEY,
+      this.requester,
+    )
+
+    // Find the latest NAV entry by Accounting Date
+    const latest = fund.reduce((a, b) => {
+      return new Date(a['Accounting Date']) > new Date(b['Accounting Date']) ? a : b
```
dskloetc: If `a['Accountint Date']` fails to parse as a date, this will just result in `Invalid Date` which then is not considered `>` so it will just not be considered the latest date.
Should we do some validation?

dskloetc: Can you declare a constant for `'Accounting Date'` instead of repeating it in multiple places?
You can do the same for the other strings that are used as keys, such as `'NAV Per Share'`.

packages/sources/nav-libre/src/transport/fund.ts
```diff
@@ -0,0 +1,66 @@
...
+  requester: Requester,
+) => {
```
dskloetc: Can you declare the function return type?

packages/sources/nav-libre/src/transport/fund-dates.ts
```diff
@@ -0,0 +1,39 @@
...
+  ToDate: string
+}
+
+export const getFundDates = async (
+  globalFundID: number,
+  baseUrl: string,
+  apiKey: string,
+  secret: string,
+  requester: Requester,
+) => {
```
dskloetc: Can you declare the function return type? Even if the compiler can infer it, it's helpful in understanding the code.

packages/sources/nav-libre/src/transport/date-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function parseDateString(dateStr: string): Date {
+  const parsed = parse(dateStr, DATE_FORMAT, new Date())
+  if (!isValid(parsed)) {
+    throw new Error(`date must be in ${DATE_FORMAT} format: got "${dateStr}"`)
+  }
+  return parsed
+}
+
+/** Ensure the window is <= N business days. Returns the (possibly adjusted) from-date. */
```
dskloetc: Even with reading the code and this comment, I have difficulty understanding what this function is trying to accomplish.
Maybe an example (in the function documentation) would be helpful.

packages/sources/nav-libre/src/config/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  MAX_RETRIES: {
+    description: 'Maximum attempts of sending a request',
+    type: 'number',
+    default: 3,
+  },
+  BACKGROUND_EXECUTE_MS: {
+    description:
+      'The amount of time the background execute should sleep before performing the next request',
+    type: 'number',
+    default: 120_000, // one call per two minute
```
mxiao-cll: nav consulting has serious rate limiting problem so we have some custom logic to get around that.

What's this DP's rate limiter?

packages/sources/nav-libre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,37 @@
...
+  {
+    globalFundID: {
+      required: true,
+      type: 'number',
+      description: 'The global fund ID for the Libre fund',
+    },
+  },
+  [
+    {
+      globalFundID: 139767,
```
mxiao-cll, line 16: This is public so need to confirm with SA on if we can disclose this or not

packages/sources/nav-libre/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Adapter } from '@chainlink/external-adapter-framework/adapter'
+import { config } from './config'
+import { nav } from './endpoint'
+
+export const adapter = new Adapter({
+  defaultEndpoint: nav.name,
+  name: 'NAV_LIBRE',
+  config,
+  endpoints: [nav],
+  rateLimiting: {
```
mxiao-cll: SubscriptionTransport does not use this

packages/sources/nav-libre/src/config/index.ts
mohamed-mehany: @mxiao-cll The limit that was share with us was 60 per minute. However, we (@mmcallister-cll and SA team) discussed earlier that since this limit is shared across, we should lower it to at least 1 per minute. I set it here to 1 per two minutes for extra safety. 
Do you think we should add more guarantees?

packages/sources/nav-libre/src/index.ts
mohamed-mehany: Leftover from before the HttpTransport refactor, will remove, thanks!

packages/sources/nav-libre/src/config/index.ts
mxiao-cll: Got it, consider increase cache if we are calling it less frequent

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: Nice catch, thanks!

packages/sources/nav-libre/src/transport/fund.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/transport/fund-dates.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/transport/date-utils.ts
mohamed-mehany: Rephrased the comments and renamed the function, hopefully more understandable.

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/config/index.ts
mohamed-mehany: Not familiar yet with the EA cache parameters, do you mean the `CACHE_MAX_AGE`?

mxiao-cll: Yes, we should have something similar set up on nav consulting

mohamed-mehany: Done, thanks!

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: Added, thx

packages/sources/nav-libre/src/transport/authentication.ts
```diff
@@ -0,0 +1,27 @@
...
+  const contentHash = CryptoJS.SHA256(body).toString(CryptoJS.enc.Base64)
+  const stringToSign = [apiKey, path, method, utcNow, nonce, contentHash].join(';')
+
+  // Compute the HMAC-SHA256 signature, Base64-encoded
+  const signature = CryptoJS.HmacSHA256(stringToSign, secret).toString(CryptoJS.enc.Base64)
+
+  return {
+    'x-date': utcNow,
+    'x-content-sha256': contentHash,
+    'x-hmac256-signature': `${apiKey};${nonce};${signature}`,
```
dskloetc, line 31: Is it documented somewhere (on navfundservices.com?) that these are the required headers?
If so, it would be useful to link to that documentation.

packages/sources/nav-libre/src/transport/date-utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function parseDateString(dateStr: string): Date {
+  const parsed = parse(dateStr, DATE_FORMAT, new Date())
+  if (!isValid(parsed)) {
+    throw new Error(`date must be in ${DATE_FORMAT} format: got "${dateStr}"`)
+  }
+  return new Date(Date.UTC(parsed.getFullYear(), parsed.getMonth(), parsed.getDate()))
+}
+
+/** Returns a start date that is at most `maxBusinessDays` before `endDate`. */
```
dskloetc: I still don't get it.
I don't see `endDate` anywhere.
But also, there must be other requirements because simply `endDate` itself is at most `maxBusinessDays` before `endDate` but you don't want it to just return `endDate`.
I still think an example in the comments would be really useful.

packages/sources/nav-libre/src/transport/fund.ts
dskloetc: Did you miss this comment?

packages/sources/nav-libre/src/transport/nav.ts
```diff
@@ -0,0 +1,116 @@
...
+    const { FromDate, ToDate } = await getFundDates(
+      param.globalFundID,
+      this.config.API_ENDPOINT,
+      this.config.API_KEY,
+      this.config.SECRET_KEY,
+      this.requester,
+    )
+    let from = parseDateString(FromDate)
+    const to = parseDateString(ToDate)
+    from = clampStartByBusinessDays(from, to)
```
dskloetc: If the `from` date comes from the API itself, why do we need to change it? What requirement is this satisfying?

packages/sources/nav-libre/src/transport/nav.ts
```diff
@@ -0,0 +1,116 @@
...
+import { SubscriptionTransport } from '@chainlink/external-adapter-framework/transports/abstract/subscription'
+import { AdapterResponse, makeLogger, sleep } from '@chainlink/external-adapter-framework/util'
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { clampStartByBusinessDays, parseDateString, toDateString } from './date-utils'
+const logger = makeLogger('NavLibreTransport')
+
+type RequestParams = typeof inputParameters.validated
+
+export class NavLibreTransport extends SubscriptionTransport<BaseEndpointTypes> {
```
dskloetc, line 16: Will you also add a unit test for the transport?

packages/sources/nav-libre/src/transport/authentication.ts
mohamed-mehany: AFAIK these are not public.
The only documentation I know of exists in the linked ticket.

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: Because the from data can be much further back than 7 days, think months or even years for example and we don't need that much data. It would actually exceed the max limit of the response size allowed by the provider and the provider would error.

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: For this one I don't see the need for unit tests given that the integration covers this component. Also mocking all its dependencies was a bit tricky. Happy to discuss if you think otherwise.

packages/sources/nav-libre/src/endpoint/nav.ts
mmcallister-cll: Worth double checking but I'm guessing it's fine since there are creds

packages/sources/nav-libre/src/endpoint/nav.ts
```diff
@@ -0,0 +1,37 @@
...
+      globalFundID: 1234,
+    },
+  ],
+)
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: {
+    Result: number
+    Data: {
+      navPerShare: number
```
mmcallister-cll, line 25: no harm in having this, but if this is for a feed, only the top level result (or in its absence, the data.result) will be used directly by the core node querying it.

packages/sources/nav-libre/src/transport/fund.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/transport/date-utils.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/endpoint/nav.ts
mohamed-mehany: Good to know, thanks! I was following other EAs for this, but I assume this would be only needed for other usecases?

packages/sources/nav-libre/src/transport/nav.ts
mohamed-mehany: I've added unit tests for this for completeness.

packages/sources/nav-libre/src/config/index.ts
```diff
@@ -0,0 +1,40 @@
...
+      type: 'string',
+      required: true,
+      sensitive: true,
+    },
+    API_ENDPOINT: {
+      description: 'An API endpoint for Data Provider',
+      type: 'string',
+      default: 'https://api.navfundservices.com',
+    },
+    MAX_RETRIES: {
```
dskloetc: This doesn't seem to be used anywhere?

packages/sources/nav-libre/test/unit/nav.test.ts
```diff
@@ -0,0 +1,149 @@
...
+// requester stub that we'll control per‑test
+const requester = makeStub('requester', { request: jest.fn() })
+const responseCache = makeStub('responseCache', { write: jest.fn() })
+const dependencies = makeStub('dependencies', {
+  requester,
+  responseCache,
+  subscriptionSetFactory: { buildSet: jest.fn() },
+} as unknown as TransportDependencies<any>)
+
+beforeEach(async () => {
```
dskloetc: I didn't know `beforeEach` can be used outside a `describe` block.
But since you only have 1 top-level `describe` block I think it would be clearer to have it inside anyway.
Can you put the `beforeEach` inside the `describe` block?

packages/sources/nav-libre/test/unit/nav.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      timestamps: expect.objectContaining({
+        providerDataRequestedUnixMs: expect.any(Number),
+        providerDataReceivedUnixMs: expect.any(Number),
+        providerIndicatedTimeUnixMs: expect.any(Number),
+      }),
+    })
+
+    expect(requester.request).toHaveBeenCalledTimes(2)
+    const fundDatesCall = requester.request.mock.calls[0]
+    expect(fundDatesCall[1]).toMatchObject({
```
dskloetc: Why not use `expect(requester.request).toHaveBeenNthCalledWith`?

packages/sources/nav-libre/test/unit/nav.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    ]
+    const fundRes = makeStub('fundRes', { response: { data: { Data: fundRows } } })
+    requester.request.mockResolvedValueOnce(fundRes)
+
+    const param = makeStub('param', { globalFundID: FUND_ID } as typeof navInputParams.validated)
+
+    await transport.handleRequest({ adapterSettings } as any, param)
+
+    const fundCallCfg = requester.request.mock.calls[1][1]
+    expect(fundCallCfg.url).toContain('fromDate=06-28-2025')
```
dskloetc: Use `toHaveBeenNthCalledWith`?

packages/sources/nav-libre/test/unit/fund.test.ts
```diff
@@ -17,7 +17,11 @@ describe('getFund', () => {
   let getRequestHeadersStub: jest.SpyInstance
   beforeEach(() => {
     jest.clearAllMocks()
-    getRequestHeadersStub = jest.spyOn(authModule, 'getRequestHeaders').mockReturnValue({})
+    getRequestHeadersStub = jest.spyOn(authModule, 'getRequestHeaders').mockReturnValue({
```
dskloetc: Why does this need to be mocked?

packages/sources/nav-libre/src/transport/authentication.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Generate the necessary headers for calling the NAV API with a 5-minute-valid signature.
+ */
+export const getRequestHeaders = (
+  method: string,
+  path: string,
+  body: string,
+  apiKey: string,
+  secret: string,
```
dskloetc: Can you use named parameters?

packages/sources/nav-libre/src/transport/fund-dates.ts
```diff
@@ -0,0 +1,45 @@
...
+  const url = `/navapigateway/api/v1/ClientMasterData/GetAccountingDataDates?globalFundID=${globalFundID}`
+  const requestConfig = {
+    baseURL: baseURL,
+    url: url,
+    method: method,
+    headers: getRequestHeaders(method, url, '', apiKey, secret),
+  }
+
+  const sourceResponse = await requester.request<FundDatesResponse>(
+    JSON.stringify(requestConfig),
```
dskloetc, line 40: This might be fine but `calculateHttpRequestKey` seems to exist for this purpose.
Since `nav-consulting` also uses `JSON.stringify` I guess it's probably fine but just wanted to point this out.

packages/sources/nav-libre/src/config/index.ts
mohamed-mehany: Good catch! removed, thanks!

packages/sources/nav-libre/test/unit/nav.test.ts
mohamed-mehany: sure, added.

packages/sources/nav-libre/test/unit/nav.test.ts
mohamed-mehany: sure, added.

packages/sources/nav-libre/src/transport/authentication.ts
mohamed-mehany: Added

packages/sources/nav-libre/src/transport/fund-dates.ts
mohamed-mehany: Good to know, thanks!

packages/sources/nav-libre/test/unit/nav.test.ts
mohamed-mehany: Sure, FWIW, I find it easier to read outside the describe, especially if we need to add other describe blocks later.
But happy to change if you think it's clearer.

packages/sources/nav-libre/test/unit/fund.test.ts
mohamed-mehany: getRequestHeaders does crypto work and pulls in CryptoJS.
The transport tests aren’t about header-signing; they just need the request to go out.
Mocking the helper keeps the unit test lightweight, slightly more deterministic, and  might also avoid failures when the dummy secret isn’t valid.
Happy to change it since it can also do without it.

dskloetc: I usually only mock things that have to be mocked, such as network connections.
If the test is noticeably slower otherwise, I guess it also makes sense.
Anyway, thanks for explaining.

packages/sources/nav-libre/test/unit/nav.test.ts
dskloetc: If you put `beforeEach` in a nested `describe` block, then it will only run before the tests in that block.
So each level of `describe` can have its own `beforeEach` that applies to it.
So conceptually it would make sense that the outer `beforeEach` is inside the outer `describe`.

PR #3928, opened 2025-07-09
Update/df 21571 solv trading por address list

mmcallister-cll:
## Description
Updating solv to allowing for new type TRADING. This is meant as a replacement for type: ENA, but ENA is left in order to maintain previous functionality.

## Changes
- Added new type option 'TRADING'
- Updated URL for this type

## Steps to Test
1. `yarn test por-address-list`
2. with payload using the new type
```json
{
    "data": {
        "endpoint": "solvBtcAddress",
        "type": "TRADING"
    }
}

```

## Quality Assurance

[DF-21571]: https://smartcontract-it.atlassian.net/browse/DF-21571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/por-address-list/src/config/index.ts
```diff
@@ -44,10 +44,10 @@ export const config = new AdapterConfig({
     type: 'string',
     default: 'https://por.sft-api.com/x-solv-btc-addresses.json',
   },
-  SOLVBTC_ENA_API_ENDPOINT: {
-    description: 'An API endpoint for SolvBTC.ENA native BTC wallet address',
+  SOLVBTC_TRADING_API_ENDPOINT: {
```
dskloetc, line 52: This is a breaking change, right?
How about keeping the old config and continue using it for type `ENA` and use the new URL for type `TRADING`.
Then we can control when we switch by changing the job spec.

Or if that's too much work, then we should at least have a major version change, right?

packages/sources/por-address-list/src/transport/solvBTC.ts
dskloetc, line 1: How come there is no test changed or added?

packages/sources/por-address-list/src/config/index.ts
mxiao-cll: +1 on major

mmcallister-cll: This is not a breaking change? It will, however, break when this old ENA endpoint is taken down.

Using the existing job spec will get you the updated answer using the new TRADING endpoint. If the ENA endpoint is overridden (which, no one would have) it will be ignored correctly since it is deprecated.

But I'll make that change, it's more correct. Do we agree on minor for just the addition then?

packages/sources/por-address-list/src/transport/solvBTC.ts
mmcallister-cll: filled in some missing tests per type for this endpoint

packages/sources/por-address-list/src/config/index.ts
dskloetc: Yes, now it is minor.

PR #3944, opened 2025-07-29
Mock new fields for secure mint

mxiao-cll:
Mock new fields in EA to unblock plugin and core development
- Add new `nativeMint` `totalAborts` `mintablePacked` fields
- Read `ripcord` from reserve, support read `ripcord` from both `success` response as well as `error` response
- Fix reserve timestamp bug, use `providerIndicatedTimeUnixMs` which is when DP updated reserve time

### DIFF
--- a/.changeset/empty-forks-love.md
+++ b/.changeset/empty-forks-love.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/secure-mint-adapter': patch
+---
+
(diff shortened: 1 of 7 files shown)


dskloetc:
Can you provide some context in the PR description?

alecgard (approved):

mxiao-cll:
Added link to design

PR #3949, opened 2025-07-31
Fix README generator with different framework version

dskloetc:
## Description

`yarn generate:readme` doesn't work for packages which have a different version of the framework than the generate script.
This is because we [use instanceof](https://github.com/smartcontractkit/external-adapters-js/blob/83c779664ede47d9504284bc041688f31bc8e716/packages/scripts/src/generate-readme/generator.ts#L477) which doesn't work if the class of the object is from a different version of the library than the class we check against.

## Changes

Check if the object constructor and the class have the same properties instead of literally checking `instanceof`.

## Steps to Test

1. I upgraded Aleno to framework 2.7.0 and checked that this didn't break generating a README for it.
2. I verified that while all packages are at framework version 2.6.0, `instanceof` and `objectHasShapeOfClass` currently give the same answer for all existing packages.

## Quality Assurance

### DIFF
--- a/packages/scripts/src/generate-readme/generator.ts
+++ b/packages/scripts/src/generate-readme/generator.ts
@@ -6,17 +6,17 @@ import { InputParametersDefinition } from '@chainlink/external-adapter-framework
 import fs from 'fs'
 import path from 'path'
 import process from 'process'
-import * as generatorPack from '../../package.json'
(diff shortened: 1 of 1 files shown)


alecgard:
Nice find and fix 🙌

PR #3952, opened 2025-08-04
Add View Function to Proof of Reserves

Subarna-Singh:
## Closes #OPDATA-3647 
https://smartcontract-it.atlassian.net/browse/OPDATA-3647

## Description
Add view-function capability for this [dedicated contract](https://basescan.org/address/0xa69b964a597435A2F938cc55FaAbe34F2A9AF278#readContract#F4) and retrieve the getPending() return value to reflect withdrawal queue balance. 

View function is not supported by Proof of Reserves. 
This PR adds view function multi chain as an indexer

## Changes
- Adding view function multi chain EA to supported indexers on proof of reserves

## Quality Assurance

- [] All code changes have 100% unit and integration test coverage. If testing is not applicable or too difficult to justify doing, the reasoning should be documented explicitly in the PR.


packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -158,6 +158,23 @@ export const execute: ExecuteWithConfig<Config> = async (input, context, config)
     validator.validated.data.indexerEndpoint,
     validator.validated.data.indexerParams,
   )
+
+  // convert hex result to number
+  if (indexer === 'VIEW_FUNCTION_MULTI_CHAIN') {
+    if (typeof balanceOutput.result === 'string') {
+      const result = String(
+        parseInt(
```
dskloetc: We should use `BigInt` because these numbers can be very large and we don't want to lose precision.
`BigInt` already supports parsing hex so you don't even need to trim the `"0x"`.

Please also add a test that would have caught this.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -74,6 +75,8 @@ export const runReduceAdapter = async (
           message: `ETHEREUM_CL_INDEXER ripcord is true: ${JSON.stringify(input.data)}`,
         })
       }
+    case viewFunctionMultiChain.name:
+      return returnParsedUnits(input.jobRunID, input.data.result as string, 18, true)
```
dskloetc: Would it be possible to do the hex parsing here instead of inside `execute`?

packages/composites/proof-of-reserves/src/utils/balance.ts
```diff
@@ -53,6 +54,7 @@ export const adaptersV3: v3AdapterImplementation[] = [
   porIndexer,
   tokenBalance,
   ceffu,
+  viewFunctionMultiChain,
```
dskloetc, line 57: What is this for? How is this used?

dskloetc (commented):
Tests are missing?

mxiao-cll:
You'll also need to add it as an option for indexer see https://github.com/smartcontractkit/external-adapters-js/commit/6ac4280db8c3f500baa3d4e659c1353afce638eb#diff-328e63cc0edd75ade1baf135858f80f472a672e4d49b1520f2fd294a672b2bee

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
Subarna-Singh: Caught loss of precision?

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Moved

packages/composites/proof-of-reserves/src/utils/balance.ts
Subarna-Singh: this is used in endpoint->reserves.ts to list available indexers.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -75,7 +75,7 @@ export const runReduceAdapter = async (
           message: `ETHEREUM_CL_INDEXER ripcord is true: ${JSON.stringify(input.data)}`,
         })
       }
-    case viewFunctionMultiChain.name:
+    case viewFunctionMultiChain.name: {
```
dskloetc: Are you sure these braces are necessary? The previous case block doesn't have them.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -114,6 +115,11 @@ const inputParameters: InputParameters<TInputParameters> = {
     type: 'string',
     description: 'end time for scheduleWindow in UTC [Format HHMM]',
   },
+  decimals: {
+    required: false,
+    type: 'number',
+    description: 'The decimal precision of the value returned by the indexer.',
```
dskloetc: Several indexers also return `decimals` themselves.
What happens if both are present and can you document that?

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -76,7 +76,16 @@ export const runReduceAdapter = async (
...
       }
     case viewFunctionMultiChain.name:
-      return returnParsedUnits(input.jobRunID, input.data.result as string, 18, true)
+      const result = parseHexToBigInt(input.data.result)
+      input.data.result = result.toString()
+
+      return returnParsedUnits(
+        input.jobRunID,
+        input.data.result as string,
+        18 - (input.data.decimals as number),
```
dskloetc: `input.data.decimals` is optional. What happens if it's not present?

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -159,22 +165,10 @@ export const execute: ExecuteWithConfig<Config> = async (input, context, config)
...
-            : balanceOutput.result ?? '0',
-          16,
-        ),
-      )
-      balanceOutput.result = balanceOutput.data.result = result
-    } else {
-      throw new Error(`Not supported data type for result: ${typeof balanceOutput.result}`)
-    }
+  if (validator.validated.data.decimals) {
+    balanceOutput.data.decimals = validator.validated.data.decimals
```
dskloetc: Changing the `balanceOutput` will be confusing later because one would reasonably expect `balanceOutput` to contain the output from the balance adapter.
Please pass the validated data separately to `runReduceAdapter` and put the logic for which decimals to use there.

packages/composites/proof-of-reserves/test/integration/adapter.test.ts
```diff
@@ -190,6 +192,40 @@ describe('execute', () => {
         .expect(200)
       expect(response.body).toMatchSnapshot()
     })
+
+    it('view-function-multi-chain should return success', async () => {
+      const data: AdapterRequest = {
+        id: '1',
+        data: {
+          endpoint: 'multiReserves',
```
dskloetc: You don't need `multiReserves` with a single input, right?

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -114,6 +115,11 @@ const inputParameters: InputParameters<TInputParameters> = {
     type: 'string',
     description: 'end time for scheduleWindow in UTC [Format HHMM]',
   },
+  decimals: {
```
dskloetc: `decimals` is a confusing name. It could easily be misinterpreted as the number of decimals we expect in the result.
Please use something more specific.

packages/composites/proof-of-reserves/test/integration/adapter.test.ts
```diff
@@ -190,6 +192,40 @@ describe('execute', () => {
...
+              addresses: [''],
+              indexer: 'view_function_multi_chain',
+              indexerEndpoint: 'function',
+              indexerParams: {
+                signature: 'function getPending() public view returns (uint256)',
+                address: '0xa69b964a597435A2F938cc55FaAbe34F2A9AF278',
+                network: 'BASE',
+              },
+              disableDuplicateAddressFiltering: true,
+              decimals: 6,
```
dskloetc: Why 6? If you intend to use it with 18 for `view-function-multi-chain` it could avoid confusion by using a realistic number in the test.

packages/composites/proof-of-reserves/test/integration/fixtures.ts
```diff
@@ -1,5 +1,22 @@
 import nock from 'nock'
 
+export const mockViewFunctionMultiChainSuccess = (): nock.Scope => {
+  return nock('https://view-function-multi-chain-adapter.com')
+    .post('/')
+    .reply(200, {
+      jobRunID: '1',
+      result: '0x0000000000000000000000000000000000000000000000000000322f64b31522',
```
dskloetc: Use a larger number so that the test actually fails if it uses `parseInt` for the conversion.

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Yes. It fails the lint workflow otherwise

packages/composites/proof-of-reserves/test/integration/adapter.test.ts
Subarna-Singh: 6 is the decimal of the result returned by view function multi chain for USDO feed.

Subarna-Singh: I think its prudent to change the name of this field.

packages/composites/proof-of-reserves/src/utils/reduce.ts
dskloetc: I was confused because the other blocks don't need it. So I tried it and the error is:
```
   79:7   error    Unexpected lexical declaration in case block  no-case-declarations
```
It's explained here: https://eslint.org/docs/latest/rules/no-case-declarations
The problem is with the `const` which applies in a larger scope than you might think. And the other blocks don't use `const` so that's why that don't need this.

So this is fine. But an alternative (also fine) solution would be to assign `parseHexToBigInt(input.data.result).toString()` directly to `input.data.result` without `const result`.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -75,7 +75,7 @@ export const runReduceAdapter = async (
           message: `ETHEREUM_CL_INDEXER ripcord is true: ${JSON.stringify(input.data)}`,
         })
       }
-    case viewFunctionMultiChain.name:
+    case viewFunctionMultiChain.name: {
       const result = parseHexToBigInt(input.data.result)
       input.data.result = result.toString()
```
dskloetc: Do you actually need to assign to `input.data.result`?
Can you just pass `result.toString()` instead of `input.data.result` on line 84?

Changing the input could be confusing, especially if not necessay.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -31,7 +31,7 @@ export type TInputParameters = {
   description?: string
   startUTC?: string
   endUTC?: string
-  decimals?: number
+  contractAnswerDecimals?: number
```
dskloetc: How about `indexerResultDecimals`?

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
dskloetc: Can you address this please?

packages/composites/proof-of-reserves/src/utils/reduce.ts
dskloetc: Can you address this please?

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -79,10 +80,13 @@ export const runReduceAdapter = async (
...
 
+      if (!contractDecimal) {
+        throw new Error('contractDecimal is a required parameter')
```
dskloetc: The parameter is declared as `contractDecimal?: number` so it's not actually required.
And all existing feeds don't specify this so it can't be required.

If this is specifically and only required for `view-function-multi-chain`, then we should make that clear from the name (and in the error message).

But I think it might make sense to not make it specific only to `view-function-multi-chain`, what do you think?

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
Subarna-Singh: replace decimals with indexerResultDecimals. 
indexerResultDecimals is exclusive to view-function-multi-chain. It will not impact any other indexer.

I think I should add this in the description to make it visible to developers.

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Absence of indexerResultDecimals will throw an error. 
There is a test case for it as well.

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
Subarna-Singh: Passing indexerResultDecimals as an argument to runReduceAdapter

packages/composites/proof-of-reserves/test/integration/adapter.test.ts
Subarna-Singh: Fixed the testcase to query reserves endpoint

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
Subarna-Singh: replaced with indexerResultDecimals

packages/composites/proof-of-reserves/test/integration/fixtures.ts
Subarna-Singh: using 0x000000000000000000000000000000000000000000000000FFFFFFFFFFFFFFFF
BigInt returns 18446744073709551615n
parseInt returns 18446744073709552000

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
Subarna-Singh: Already added.

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Made the change.

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: That would allow us to manually override the decimals in other indexer. Do we want to have that option?

Subarna-Singh: I think this is already a security risk as is.

Subarna-Singh: I will highlight in today's stand up that we many feeds using view function are not querying the decimals method on contract. This is big risk.

dskloetc: > That would allow us to manually override the decimals in other indexer. Do we want to have that option?

If we want it for `view-function-multi-chain`, isn't it reasonable to assume we might want it for other indexers? (this is not a rhetorical question)

To make it safer, we could throw if the indexer also returns `decimals` with a different value. Then it can only be used for indexers that don't specify `decimals` already.

If you don't want to do this now, then I repeat that the name of the EA input parameter should make it clear this is only for `view-function-multi-chain`. And the error message here as well.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -77,15 +77,12 @@ export const runReduceAdapter = async (
         })
       }
     case viewFunctionMultiChain.name: {
```
dskloetc: You can remove the curly brackets now, right?

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -36,6 +37,7 @@ export const runReduceAdapter = async (
   indexer: string,
   context: AdapterContext,
   input: AdapterResponse,
+  contractDecimal?: number,
```
dskloetc: Fix the name here as well.

packages/composites/proof-of-reserves/src/utils/balance.ts
```diff
@@ -179,3 +181,15 @@ function buildCeffuRequest(input: AdapterResponse) {
     },
   }
 }
+
+export function parseHexToBigInt(value: unknown): bigint {
```
dskloetc: Why is this function defined here, even though it's only used in `reduce.ts`?

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Most contracts support getDecimal or similar methods. 
In future, I would want to see this option removed for view function, and use decimal method on contracts to pull decimals value. 

So, I am against giving this option for other indexers.

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: fixed.

packages/composites/proof-of-reserves/src/utils/balance.ts
Subarna-Singh: moved the function

packages/composites/proof-of-reserves/src/utils/reduce.ts
Subarna-Singh: Yes. Workflow not failing.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -76,18 +87,17 @@ export const runReduceAdapter = async (
           message: `ETHEREUM_CL_INDEXER ripcord is true: ${JSON.stringify(input.data)}`,
         })
       }
-    case viewFunctionMultiChain.name: {
-      if (!contractDecimal) {
-        throw new Error('contractDecimal is a required parameter')
+    case viewFunctionMultiChain.name:
+      if (!viewFunctionIndexerResultDecimals) {
+        throw new Error('viewFunctionIndexerResultDecimals is a required parameter')
```
dskloetc: please add `... when using the 'view-function-multi-chain' indexer.`

Subarna-Singh: Done

dskloetc (approved):

packages/composites/proof-of-reserves/src/endpoint/reserves.ts
```diff
@@ -31,6 +31,7 @@ export type TInputParameters = {
   description?: string
   startUTC?: string
   endUTC?: string
+  viewFunctionIndexerResultDecimals?: number
```
mxiao-cll, line 34: I don't think we should do this, this will quickly get very ugly as we add params for each indexer/protocal pair

Why can't we have the indexer return the decimal in it's response and then take it from there

Subarna-Singh: @mxiao-cll This is a temporary work around. 

The idea is to add additional functionality to view-function to query decimal method on contract, so the view function indexer can return the decimal value. 

The only other way to do it right now would be to overwrite the balanceOutput.data.decimal with a value, but that miss represents what the indexer is doing.

PR #3965, opened 2025-08-19
OPDATA-3708 USDO add Solana reserves to token balance

Subarna-Singh:
## Closes #OPDATA-3708
https://smartcontract-it.atlassian.net/browse/OPDATA-3708

## Description
USDO PoR is being extended to include Solana-based reserves, beginning with TBILL tokens that back USDO on Solana.
The token-balance EA should:

- Query the TBILL (SOL) token account at 4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6 to fetch the balance.
- Retrieve the TBILL price from the on-chain price oracle (Ethereum or Arbitrum).
- Multiply the balance by the oracle price to calculate the USD value of Solana reserves.
- Contribute this USD value to the composite USDO PoR output so it reflects Solana reserves alongside existing assets.

## Changes
Added new endpoint usdoSolana.

## Steps to Test
yarn test packes/source/token-balance/test/unit
yarn test packes/source/token-balance/test/integration

## Quality Assurance


packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+import { config } from '../config'
+import { usdoSolanaTransport } from '../transport/usdoSolana'
+
+export const inputParameters = new InputParameters(
+  {
+    addresses: {
+      required: true,
+      type: {
+        network: {
+          aliases: ['chain'],
```
dskloetc: Do we need an alias? Having different names for the same thing can lead to confusion.
Or do we already have similar input parameters with the same alias so people might be used to it?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+          description: 'Chain ID of the network',
+        },
+        address: {
+          required: true,
+          type: 'string',
+          description: 'Address of token contract',
+        },
+      },
+      array: true,
+      description: 'Array of wallets to sum balances',
```
dskloetc: The other input fields have the `description` before the `type`.
Can you try to be consistent in the field ordering?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+    },
+    tokenMint: {
+      required: true,
+      description: 'token mint',
```
dskloetc: I don't know what `token mint` is. A more useful description would be good.

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+      type: {
+        token: {
+          required: false,
+          type: 'string',
+          default: 'TBILL',
```
dskloetc: Does it really make sense to have a specific token as the default?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+          description: 'Token symbol',
+        },
+        contractAddress: {
+          required: true,
+          type: 'string',
+          description: 'Address of token mint address',
```
dskloetc: Does the address have an address? Or is one "address" redundant?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+        address: {
+          required: true,
+          type: 'string',
+          description: 'Address of token contract',
+        },
+      },
+      array: true,
+      description: 'Array of wallets to sum balances',
+    },
+    tokenMint: {
```
dskloetc: It looks like the purpose of this is to group 2 things together:
1. A token symbol
2. A contract address

If I understand that correctly, would it make sense to call this field `token` and call the "token" field below `symbol`?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,102 @@
...
+          chainId: '8453',
+        },
+      ],
+      tokenMint: {
+        token: 'TBILL',
+        contractAddress: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6 ',
+      },
+      priceOracle: {
+        contractAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
+        chainId: '1',
```
dskloetc: Can this `chainId` be different from the one on line 71?
It has to match for us to be able to look up the price, right?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -16,7 +16,7 @@ export const getToken = async (
 
   const response = await Promise.all(
     addresses
-      .filter((a) => a.token?.toLowerCase() == token)
+      .filter((a) => a.token?.toLowerCase() == token.toLowerCase())
```
dskloetc: Is this related to this PR?
If so, can you cover it in a unit test?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,168 @@
...
+        ? this.arbProvider
+        : (() => {
+            throw new AdapterInputError({
+              message: `Unsupported chainId: ${priceOracle.chainId} for price oracle`,
+              statusCode: 400,
+            })
+          })()
+
+    // 2. Fetch TBILL price ONCE from oracle contract
+    const priceFeed = await getRate(priceOracle.contractAddress, evmProvider)
```
dskloetc: Have you seen this [getTokenPrice](https://github.com/smartcontractkit/external-adapters-js/blob/1d8efca52325b55b0ffd214003366c796638d5ff/packages/sources/token-balance/src/transport/priceFeed.ts#L29) function?
It abstracts away getting the evmProvider for the right network.
The only difference is that the network is specified by name instead of chain ID.
Is it an interface requirement that you get the network by chain ID in the input params?

packages/sources/token-balance/src/transport/utils.ts
```diff
@@ -99,3 +101,49 @@ export const getNetworkEnvVar = (network: string, suffix: string): string => {
   }
   return envVar
 }
+
+export const getTokenBalance = async (
```
dskloetc: This is a Solana specific function, right?
Could you reuse (some of) [this function](https://github.com/smartcontractkit/external-adapters-js/blob/1d8efca52325b55b0ffd214003366c796638d5ff/packages/sources/token-balance/src/transport/solana.ts#L5)?

If not, would it at least make sense to add this function in that file?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: Best do away with it. I reused it from somewhere

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: noted

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: added.

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: no

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: redundant

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: No. TokenMint is an account on Solana that generates tokens. This is the correct representation of that in my opinion.

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: Yes it can be different. 
In fact 71 is wrong. That is an account on Solana. I have to change input params on por-addr-list to reflect that correctly.

price can be fetched from ethereum or arbitrum contracts,

packages/sources/token-balance/src/transport/utils.ts
Subarna-Singh: I looked into this. However, it expects address to have wallets which does not fit inti the address structure in this case. 

I will make some changes to make the function modular to allow using parts of it in future.

Subarna-Singh: Move function to Solana and made it modular. 
I will not be making any changes to getToken in this PR.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Removed it from this PR.

packages/sources/token-balance/src/endpoint/index.ts
```diff
@@ -2,4 +2,5 @@ export { endpoint as etherFi } from './etherFi'
 export { endpoint as evm } from './evm'
 export { endpoint as solvJlp } from './solvJlp'
 export { endpoint as tbill } from './tbill'
+export { endpoint as usdoSolana } from './usdoSolana'
```
dskloetc: Does the endpoint do anything specific to USDO? Or would it make sense to give it a more generic name?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,106 @@
...
+export const inputParameters = new InputParameters(
+  {
+    addresses: {
+      required: true,
+      description: 'Array of wallets to sum balances',
+      type: {
+        network: {
+          required: false,
+          type: 'string',
+          description: 'Network of the contract',
```
dskloetc: Which contract? Is this always going to be "solana" or can it be something else?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,106 @@
...
+        },
+        chainId: {
+          required: false,
+          type: 'string',
+          description: 'Chain ID of the network',
+        },
+        address: {
+          required: true,
+          type: 'string',
+          description: 'Address of token contract',
```
dskloetc: If this is the address of the token contract, where is the address of the wallet?

packages/sources/token-balance/src/endpoint/usdoSolana.ts
```diff
@@ -0,0 +1,106 @@
...
+    addresses: {
+      required: true,
+      description: 'Array of wallets to sum balances',
+      type: {
+        network: {
+          required: false,
+          type: 'string',
+          description: 'Network of the contract',
+        },
+        chainId: {
```
dskloetc: From a quick search I get the impression that Solana doesn't have chain IDs. Do we need this?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -1,7 +1,21 @@
-import { Connection, PublicKey } from '@solana/web3.js'
+import { makeLogger } from '@chainlink/external-adapter-framework/util'
 import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { Connection, PublicKey } from '@solana/web3.js'
 import { inputParameters } from '../endpoint/solvJlp'
 
+const logger = makeLogger('Solana helper functions')
+
+type tokenMint = {
```
dskloetc: Don't we start custom type names with an upper case?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
     formattedResponse,
   }
 }
+
+function sanitizeInputs(addresses: { address: string }[], tokenMint: { contractAddress: string }) {
+  return {
+    mint: new PublicKey(tokenMint.contractAddress.trim()),
+    normalizedAddresses: addresses.map((a) => a.address.trim()),
```
dskloetc: Why do we need this? Do we expect addresses to be padded with whitespace? Is it useful to allow that? Why specifically this as opposed to other sanitizations? (lower case, adding `0x`, removing other invalid characters or removing white space from inside rather than just front and back are just some examples that come to mind)

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+  }
+}
+
+async function fetchTokenAccounts(address: string, mint: PublicKey, connection: Connection) {
```
dskloetc: Just for my understanding, is there a difference between an address and a token account?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+  const response = await connection.getParsedTokenAccountsByOwner(new PublicKey(address), { mint })
```
dskloetc: The type of `response` here is [RpcResponseAndContext<Array<{pubkey: PublicKey; account: AccountInfo<ParsedAccountData>}>](https://github.com/solana-foundation/solana-web3.js/blob/c26c13bf841821d9bbf83bb476c567fe29d13821/src/connection.ts#L3486C5-L3487C74).
But in the end you only use the `.value` from it and never the `.context`.
How about returning `response.value` from this function?
Then it's easier to be explicit about the return type and we can avoid declaring types as `Awaited<ReturnType<typeof fetchTokenAccounts>>` which requires a lot of investigation to find out what it really is.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+    normalizedAddresses.map((address) => fetchTokenAccounts(address, mint, connection)),
+  )
+
+  // Flatten and normalize results
+  const results = responses.flatMap((res) => {
+    if (res.status === 'fulfilled') {
+      return extractBalances(res.value, cachedDecimals)
+    } else {
+      logger.error(`Failed to fetch token accounts: ${res.reason}`)
+      return []
```
dskloetc: Do we really just want to exclude balances that we failed to fetch? Shouldn't that cause an error so we don't write a wrong value on chain?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+
+  return response.value.map((v) => ({
+    token: v.account.data.parsed.info.mint,
+    wallet: v.account.data.parsed.info.owner,
+    value: BigInt(v.account.data.parsed.info.tokenAmount.amount),
+    decimals: Number(v.account.data.parsed.info.tokenAmount.decimals ?? cachedDecimals),
+  }))
+}
+
+export const getTokenBalance = async (
```
dskloetc: At a glance, `getToken` (line 19) seems to do mostly the same steps. Can you explain the difference?

packages/sources/token-balance/src/transport/utils.ts
dskloetc: > I will not be making any changes to getToken in this PR.

Why not? If the logic is the same but only the input type is different, can we just have a wrapper function that does the conversion instead of have 2 full implementations of the same logic?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+import { getTokenBalance } from './solana'
+
+const logger = makeLogger('Token Balance - Tbill Solana')
+
+type RequestParams = typeof inputParameters.validated
+
+const RESULT_DECIMALS = 18
+
+export class USDOSolanaTransport extends SubscriptionTransport<BaseEndpointTypes> {
+  ethProvider!: ethers.JsonRpcProvider
```
dskloetc: Is this unused?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+  arbProvider!: ethers.JsonRpcProvider
```
dskloetc: Is this unused?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  async _handleRequest(
+    param: RequestParams,
+  ): Promise<AdapterResponse<BaseEndpointTypes['Response']>> {
+    const addresses = param.addresses
+    const tokenMint = param.tokenMint
+    const providerDataRequestedUnixMs = Date.now()
+
+    // 1. Fetch TBILL price ONCE from oracle contract
```
dskloetc: The logic has abstracted away whether this is TBILL, right?
Just stating in the comment that it's TBILL is confusing if it can be any token.

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+    const priceFeed = await getTokenPrice({
```
dskloetc: What are you trying to indicate by putting "feed" in the name?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+      },
+    }
+  }
+
+  async calculateTokenSharesUSD(
+    addresses: typeof inputParameters.validated.addresses,
+    tokenMint: typeof inputParameters.validated.tokenMint,
+    priceFeed: { value: bigint; decimal: number },
+  ): Promise<bigint> {
+    //1. Fetch TBILL token balances for the given address on Solana
```
dskloetc: Again "TBILL".

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+    const { result: balances } = await getTokenBalance(addresses, tokenMint, this.connection)
+
+    // 2. Multiply Solana token balances by oracle price and normalize decimals to RESULT_DECIMALS
+    let totalSharesUSD = BigInt(0)
```
dskloetc: What is "shares" indicating here. We don't typically speak of USD amounts in terms of "shares", right?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+    for (const bal of balances) {
+      // Normalize token balance decimals to RESULT_DECIMALS
+      const normalizedShares = bal.value * BigInt(10 ** (RESULT_DECIMALS - bal.decimals))
```
dskloetc: Does this work if `bal-decimals > RESULT_DECIMALS`?
I think it's safer to multiply and divide: `bal.value * 10n ** BigInt(RESULT_DECIMALS) / 10n ** BitInt(bal.decimals)`

But I think it's even safer to first add all the values together and then do the decimal conversion. And this should be fine since each balance should have the same decimals, right?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+  addresses: address[],
+  tokenMint: tokenMint,
+  connection?: Connection,
+) => {
```
dskloetc: Can you declare a return type explicitly?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+    logger.warn(`No token account found for address: ${address} and mint: ${mint.toBase58()}`)
+  }
+
+  return response
+}
+
+function extractBalances(
+  response: Awaited<ReturnType<typeof fetchTokenAccounts>>,
+  cachedDecimals: number,
+) {
```
dskloetc: Can you declare a return type explicitly?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+        decimals: cachedDecimals,
+      },
+    ]
+  }
+
+  return response.value.map((v) => ({
+    token: v.account.data.parsed.info.mint,
+    wallet: v.account.data.parsed.info.owner,
+    value: BigInt(v.account.data.parsed.info.tokenAmount.amount),
+    decimals: Number(v.account.data.parsed.info.tokenAmount.decimals ?? cachedDecimals),
```
dskloetc: Can `v.account.data.parsed.info.tokenAmount.decimals` be present and different from `cachedDecimals`?
Can `v.account.data.parsed.info.tokenAmount.decimals` be absent?

Based on [this documentation](https://solana.com/docs/rpc/json-structures#token-balances)
> decimals: <number> - Number of decimals configured for token's mint.

it's always equal to the decimals for the token.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +65,82 @@ export const getToken = async (
...
+        wallet: '',
+        value: BigInt(0),
+        decimals: cachedDecimals,
+      },
+    ]
+  }
+
+  return response.value.map((v) => ({
+    token: v.account.data.parsed.info.mint,
+    wallet: v.account.data.parsed.info.owner,
```
dskloetc: Why include `token` and `wallet` while they are unused?

packages/sources/token-balance/src/transport/usdoSolana.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      // Multiply balance by oracle price and normalize again
+      totalSharesUSD +=
+        (normalizedShares * priceFeed.value * BigInt(10 ** (RESULT_DECIMALS - priceFeed.decimal))) /
+        BigInt(10 ** RESULT_DECIMALS)
```
dskloetc: You're both multiplying and dividing by `10 ** RESULT_DECIMALS` here. Doesn't that cancel out.

packages/sources/token-balance/src/endpoint/index.ts
Subarna-Singh: A generic name is a good idea.

Subarna-Singh: Using solanaTokenBalance as new name

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: Added better descriptions

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: No we don't. Removed it.

packages/sources/token-balance/src/endpoint/usdoSolana.ts
Subarna-Singh: Fixed the description

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Fixed.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: makes sense. I am removing this, seems unnecessary

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Address is the wallet address of the owner. 
Each wallet has tokenAccounts associated which store balance of a token. 

Image

packages/sources/token-balance/src/endpoint/index.ts
dskloetc: token-balance is already the name of the EA. And other endpoints also don't have "TokenBalance" in the name.
Is there something specific about solana that the endpoint needs to be distinguished as a "token balance" endpoint in particular?

One thing that might be worth distinguishing (although I failed to do that with `xrpl`) is that the result is converted with a price contract. Or is that done for all endpoints aready?

For now I think the endpoint could just be called `solana`?

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: agreed. Fixed it.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Done

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Agreed.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: I agree on this point. removed it.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Done.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: This was something I was contemplating. 
Do we want the feeds to fail or do we want to underreport and log the issue. 

What is your take?

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: 1. getToken is doing a lot of things in one place. Its not modular in my opinion.
2. I did not agree with the address structure used. I would rather pass tokenMint and accountOwner values as param than restructure inputs to fit getToken.

There must be enough flexibility in these functions to be used by any new upcoming EA or endpoints.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: yes.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: yes. removed

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: removed

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: fixed.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: tokenPrice is more intuitive. renamed it.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: fixed

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: AUM is better word than Shares. 
Using that instead.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: Yes.

packages/sources/token-balance/src/transport/usdoSolana.ts
Subarna-Singh: no. 
1. normalizedBalance is 18 digits.
2. tokenPrice decimal might be different from balance decimal so we scale it to 18. 
3. multiplying the two results in 36 decimals
4. need to divide it to bring it to 18.

packages/sources/token-balance/src/endpoint/index.ts
Subarna-Singh: solana already exists in the transport folder. Hence I cant use it.

Subarna-Singh: I dont understand what you mean here 
'One thing that might be worth distinguishing (although I failed to do that with xrpl) is that the result is converted with a price contract. Or is that done for all endpoints aready?'

dskloetc: > solana already exists in the transport folder. Hence I cant use it.

That file is just an internal utility file, right? If we think `solana` is the best name for the endpoint, we can easily move or rename the file in the `transport` directory.

> I dont understand what you mean here
'One thing that might be worth distinguishing (although I failed to do that with xrpl) is that the result is converted with a price contract. Or is that done for all endpoints aready?'

I was trying to think of any reason why we shouldn't call this endpoint just `solana`. The only reason I could think of is that one might think an endpoint called `solana` would just give token balances on Solana. But it doesn't just do that. It also converts the balances using a price oracle contract. If we anticipate also having a straight-forward Solana endpoint that just gives balances without converting, we might consider using an endpoint name that makes it clear that the result is converted. But we already have several (or all? I don't know) endpoints that do price conversion where we didn't consider this. And I didn't consider it when I implemented XRPL which also does price conversion.

packages/sources/token-balance/src/transport/solana.ts
dskloetc: OK, but you shouldn't just duplicate the code. Either fix the existing code (in a separate PR) or use the existing code. Otherwise we get a confusing collection of function that all do the same thing slightly differently.

packages/sources/token-balance/src/transport/solana.ts
dskloetc: This is a question for the SA.

In general my opinion is that it's better to fail obviously than to fail silently.

packages/sources/token-balance/src/transport/usdoSolana.ts
dskloetc: > 2. tokenPrice decimal might be different from balance decimal so we scale it to 18.

When multiplying numbers, they don't need to have the same number of decimals.
You anyway don't end up with the same number of decimals, as you can see from the fact that you get 36 decimals rather than 18 decimals.

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import { solanaTokenBalanceTransport } from '../transport/solanaTokenBalance'
+
+export const inputParameters = new InputParameters(
+  {
+    addresses: {
+      required: true,
+      description:
+        'List of wallet accounts to query. The balances of all provided wallets will be retrieved and summed together.',
```
dskloetc: What are "wallet accounts"? I thought we have "wallet addresses" and each wallet can have multiple "token accounts".

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+      type: {
+        network: {
+          required: false,
+          type: 'string',
+          description: 'Blockchain network where the wallet resides.',
```
dskloetc: I still don't understand. Is it possible for this to be something other than "SOLANA"?
If not, why is this here?

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+      required: true,
+      description:
+        'Configuration of the on-chain price oracle that provides real-time token valuations.',
+      type: {
+        contractAddress: {
+          required: true,
+          type: 'string',
+          description: 'Contract address of the price oracle used to fetch token price data.',
+        },
+        chainId: {
```
dskloetc: Is this used? The `getTokenPrice` function that you're reusing doesn't require it and instead expects the chain ID to be defined in the environment.

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+    // let totalAumUSD = BigInt(0)
+    // for (const bal of balances) {
+    //   // Normalize token balance decimals to RESULT_DECIMALS
+    //   const normalizedAum = bal.value * BigInt(10 ** (RESULT_DECIMALS - bal.decimals))
+
+    //   // Multiply balance by oracle price and normalize again
+    //   totalAumUSD +=
+    //     (normalizedAum * tokenPrice.value * BigInt(10 ** (RESULT_DECIMALS - tokenPrice.decimal))) /
+    //     BigInt(10 ** RESULT_DECIMALS)
+    // }
```
dskloetc: Should this commented out code be here?

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -51,3 +70,63 @@ export const getToken = async (
...
+
+function extractBalances(
+  accounts: Awaited<ReturnType<typeof fetchTokenAccounts>>,
+): ExtractedBalance[] {
+  if (accounts.length === 0) {
+    return [
+      {
+        value: BigInt(0),
+        decimals: 0,
+      },
```
dskloetc, line 110: Can we just return an empty array?

packages/sources/token-balance/src/transport/solana.ts
dskloetc: And my main reason for asking is that if we do want to throw, then we can use `Promise.all` instead of `Promise.allSettled` which is cleaner.

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+  async initialize(
+    dependencies: TransportDependencies<BaseEndpointTypes>,
+    adapterSettings: BaseEndpointTypes['Settings'],
+    endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+
+    if (!adapterSettings.SOLANA_RPC_URL) {
+      logger.error('SOLANA_RPC_URL is missing')
```
dskloetc: We can't do anything without it, can we? Might as well throw here.

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+    // 1. Fetch token price ONCE from oracle contract
+    const tokenPrice = await getTokenPrice({
+      priceOracleAddress: param.priceOracle.contractAddress,
+      priceOracleNetwork: param.priceOracle.network,
+    })
+
+    // 2. Fetch balances for each Solana wallet and calculate their USD value using the SINGLE tokenPrice
+    const totalTokenUSD = await this.calculateTokenAumUSD(addresses, tokenMint, tokenPrice)
+
+    // 4. Build adapter response object
```
dskloetc: 4 after 2. What happened to 3?

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+    //   // Normalize token balance decimals to RESULT_DECIMALS
+    //   const normalizedAum = bal.value * BigInt(10 ** (RESULT_DECIMALS - bal.decimals))
+
+    //   // Multiply balance by oracle price and normalize again
+    //   totalAumUSD +=
+    //     (normalizedAum * tokenPrice.value * BigInt(10 ** (RESULT_DECIMALS - tokenPrice.decimal))) /
+    //     BigInt(10 ** RESULT_DECIMALS)
+    // }
+
+    // 1. Sum raw balances (all balances are for the same mint, so same decimals)
```
dskloetc: Fetching the balances was step 1. This is step 2, right?

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+    let totalRaw = BigInt(0)
+    let tokenDecimals = 0
+
+    for (const bal of balances) {
+      totalRaw += bal.value
+      tokenDecimals = bal.decimals // safe because same token
```
dskloetc: Let's throw if it's not the same, to be sure.

```
    let tokenDecimals = undefined
    for (const bal of balances) {
      totalRaw += bal.value
      if (!bal.decimals) {
        throw new AdapterError('Missing decimals on balance response');
      }
      if (tokenDecimals !== undefined && bal.decimals !== tokenDecimals) {
        throw new AdapterError(`Inconsistent balance decimals: ${tokenDecimals} != ${bal.decimals}`);
      }
    }
    tokenDecimals ??= RESULT_DECIMALS
```

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // 2. Normalize once to RESULT_DECIMALS
+    const normalizedBalance = totalRaw * BigInt(10 ** (RESULT_DECIMALS - tokenDecimals))
```
dskloetc: This is still not safe if `tokenDecimals > RESULT_DECIMALS`.
Safest is to first multiply everything and only then divide.

`totalRaw * tokenPrice.value * 10n ** BigInt(RESULT_DECIMALS) / 10n ** BigInt(tokenDecimals) / 10n ** BigInt(tokenPrice.decimal)`

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  let transport: SolanaTokenBalanceTransport
+
+  beforeEach(() => {
+    transport = new SolanaTokenBalanceTransport()
+    transport.connection = {} as any
+    jest.clearAllMocks()
+  })
+
+  it('fetches price and balances, calculates correct USD result', async () => {
+    ;(getTokenPrice as jest.Mock).mockResolvedValue({ value: 1000n, decimal: 18 })
```
dskloetc: What are these semi-colons at the start of the line? Why does this not fail on the formatter?
Is this some special syntax I'm not aware of?

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    ;(getTokenBalance as jest.Mock).mockResolvedValue({
```
dskloetc: I think you can do `jest.mocked(getTokenBalance)` instead of `(getTokenBalance as jest.Mock)` to avoid a cast.

https://jestjs.io/docs/mock-function-api#jestmockedsource-options

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    })
+
+    const resp = await transport._handleRequest({
+      addresses: [{ address: 'wallet1', network: 'SOLANA' }],
+      tokenMint: { token: 'TBILL', contractAddress: 'mint1' },
+      priceOracle: { contractAddress: 'oracle1', chainId: '1', network: 'ethereum' },
+    })
+
+    expect(getTokenPrice).toHaveBeenCalledWith({
+      priceOracleAddress: 'oracle1',
```
dskloetc: Use variables for these magic values ('oracle1', 'ethereum', 'mint1', etc.) so it's clearer where you need the same value and harder to make typos.

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
dskloetc: Also add a test where the token price has a different number of decimals to make sure the scaling is done correctly.

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      priceOracleNetwork: 'ethereum',
+    })
+    expect(getTokenBalance).toHaveBeenCalledWith(
+      [{ address: 'wallet1', network: 'SOLANA' }],
+      { token: 'TBILL', contractAddress: 'mint1' },
+      transport.connection,
+    )
+    expect(resp.statusCode).toBe(200)
```
dskloetc: Also check that the actual result in the response is correct.

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
Subarna-Singh: It is addresses. 
Lack of clarity of thought on my part.

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
Subarna-Singh: Should not be. Removed it.

packages/sources/token-balance/src/endpoint/solanaTokenBalance.ts
Subarna-Singh: Redundant. Removed

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Yes.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Pending: asked SA on what result is expected in this case

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: I would prefer opening a separate PR for this.

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
Subarna-Singh: Added Error

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
Subarna-Singh: I left this by mistake. Need to remove it.

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
Subarna-Singh: fixing numbering

packages/sources/token-balance/src/transport/solanaTokenBalance.ts
Subarna-Singh: fixing numbering

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
Subarna-Singh: added

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
Subarna-Singh: added

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
Subarna-Singh: done

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
Subarna-Singh: okay.

packages/sources/token-balance/test/unit/solanaTokenBalance.test.ts
Subarna-Singh: Because I was using (getTokenPrice as jest.Mock), JavaScript required semicolon to distinguish it as a separate line. 
It is a weird issue. I didnt fully understand why its happening.

packages/sources/token-balance/src/transport/solana-utils.ts
```diff
@@ -51,3 +70,58 @@ export const getToken = async (
...
+  const responses = await Promise.allSettled(
+    addresses.map((addr) => fetchTokenAccounts(addr.address, mint, connection)),
+  )
+
+  // Flatten and normalize results
+  const results = responses.flatMap((res) => {
+    if (res.status === 'fulfilled') {
+      return extractBalances(res.value)
+    } else {
+      logger.error(`Failed to fetch token accounts: ${res.reason}`)
```
dskloetc: Actually, I think silently underreporting is dangerous default behavior.
We can still ask the SA if they'd want to underreport rather than error (though I would argue against it) but that behavior should then be controlled with a function parameter.
Doing it by default would be surprising.

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        contractAddress: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6',
+      },
+      priceOracle: {
+        contractAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
+        network: 'ethereum',
+      },
+    })
+
+    expect(getTokenPrice).toHaveBeenCalledWith({
+      priceOracleAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
```
dskloetc: > > Use variables for these magic values ('oracle1', 'ethereum', 'mint1', etc.) so it's clearer where you need the same value and harder to make typos.
>
> done

It doesn't look done. You changed `'oracle1'` to `'0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40'`. It's still a magic literal value. What I meant is to declare a variable
```
const priceOracleAddress = '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40'
```
at the top of the test and use that here and on line 36 so that it's obvious that those 2 values are intended to be the same and it's caught by the compiler if there is a typo.

Same for the other literal values.

packages/sources/token-balance/src/transport/solana.ts
dskloetc: You can do that but then it should be merged first. We should not have 2 functions that do the same and look different.

packages/sources/token-balance/src/endpoint/solana.ts
```diff
@@ -0,0 +1,92 @@
...
+    },
+    tokenMint: {
+      required: true,
+      description:
+        'Token mint information. A mint is the canonical on-chain account that defines the token’s metadata (name, symbol, supply rules).',
+      type: {
+        token: {
+          required: false,
+          type: 'string',
+          description: 'Readable token symbol (e.g., USDC, TBILL). Used for reference only.',
```
dskloetc: I don't see where this is used.
If it's not used and we decide to remove it, then `tokenMint` will have only 1 field, so then it would make sense to collapse it into a single `tokenMintContractAddress` field.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -0,0 +1,148 @@
...
+    // 1. Fetch token price ONCE from oracle contract
+    const tokenPrice = await getTokenPrice({
+      priceOracleAddress: param.priceOracle.contractAddress,
+      priceOracleNetwork: param.priceOracle.network,
+    })
+
+    // 2. Fetch balances for each Solana wallet and calculate their USD value using the SINGLE tokenPrice
+    const totalTokenUSD = await this.calculateTokenAumUSD(addresses, tokenMint, tokenPrice)
+
+    // 4. Build adapter response object
```
dskloetc: >> 4 after 2. What happened to 3?
>
> fixing numbering

I still see step 4 after step 2.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -0,0 +1,148 @@
...
+  async calculateTokenAumUSD(
+    addresses: typeof inputParameters.validated.addresses,
+    tokenMint: typeof inputParameters.validated.tokenMint,
+    tokenPrice: { value: bigint; decimal: number },
+  ): Promise<bigint> {
+    // 1. Fetch token balances for the given address on Solana
+    const { result: balances } = await getTokenBalance(addresses, tokenMint, this.connection)
+
+    // 2. Sum raw balances (all balances are for the same mint, so same decimals)
+    let totalRaw = BigInt(0)
```
dskloetc: nit: `BigInt(0)` can be written as `0n`.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -0,0 +1,148 @@
...
+    dependencies: TransportDependencies<BaseEndpointTypes>,
+    adapterSettings: BaseEndpointTypes['Settings'],
+    endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+
+    if (!adapterSettings.SOLANA_RPC_URL) {
+      logger.error('SOLANA_RPC_URL is missing')
+      throw new AdapterInputError({
```
dskloetc: Sorry, this is my mistake!
After seeing the need to change the existing integration tests, I realized that not all NOPs that run `token-balance` might need to be able to handle the solana endpoint. Only those that get a job that needs it.
So we should definitely remove the error and log only a warning. Not an error.

Otherwise this would definitely be a major breaking change requiring all NOPs to support solana immediately.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  Connection: class {
+    async getParsedTokenAccountsByOwner() {
+      return {
+        value: [
+          {
+            account: {
+              data: {
+                parsed: {
+                  info: {
+                    mint: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6',
```
dskloetc: Use (and reuse) a constant for the mint address.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+                    owner: 'G7v3P9yPtBj1e3JN7B6dq4zbkrrW3e2ovdwAkSTKuUFG',
```
dskloetc: Use a constant.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+                    tokenAmount: {
+                      amount: '1000000000', // 1e9
```
dskloetc: You can write `String(1e9)`. Then you don't need the comment and you don't need to trust that the comment is correct.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+                      decimals: 6,
```
dskloetc: Use a constant.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+              },
+            },
+          },
+        ],
+      }
+    }
+    async getBalance() {
+      return 0
+    }
+    async getParsedAccountInfo() {
```
dskloetc: Is this used? Do we need it?
If yes, does it need to return the same values as above?
If yes, use the same constants.
If no, use different values.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+  },
+}))
+
+jest.mock('ethers', () => {
+  return {
+    ethers: {
+      JsonRpcProvider: jest.fn().mockImplementation(() => ({} as JsonRpcProvider)),
+      Contract: jest.fn().mockImplementation(() => ({
+        decimals: jest.fn().mockResolvedValue(8),
```
dskloetc: Use a constant.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        latestAnswer: jest.fn().mockResolvedValue(150000000n), // 1.5 USD
```
dskloetc: Use a constant.
You can declare it as
```
const tokenPriceValue = BigInt(1.5 * 10 ** tokenPriceDecimals)
```
so the value is then clear without a need to comment.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        latestRoundData: jest.fn().mockResolvedValue({
```
dskloetc: Do we need both `latestRoundData` and `latestAnswer`? Are both used?
If yes, do they need to return the same values?
If yes, reuse the same constants.
If no, return different values.

packages/sources/token-balance/test/unit/solana.test.ts
dskloetc, line 1: Please use constants in this file as well.

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    expect(resp.statusCode).toBe(200)
+    expect(resp.result).toBe('44896700000000000000000')
+    expect(resp.data).toEqual({
+      result: '44896700000000000000000',
+      decimals: 18,
+    })
+  })
+
+  it('test scaling of calculates correct USD result', async () => {
+    jest.mocked(getTokenPrice).mockResolvedValue({ value: 900000000n, decimal: 8 })
```
dskloetc: FYI, you can write numbers like this to make them easier to read: `900_000_000n`

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      priceOracleNetwork: 'ethereum',
+    })
+    expect(getTokenBalance).toHaveBeenCalledWith(
+      [{ address: 'G7v3P9yPtBj1e3JN7B6dq4zbkrrW3e2ovdwAkSTKuUFG' }],
+      { token: 'TBILL', contractAddress: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6' },
+      transport.connection,
+    )
+
+    expect(resp.statusCode).toBe(200)
+    expect(resp.result).toBe('90000000000000000000')
```
dskloetc: It's not immediately obvious that this is the correct result.
If you declare some values like this at the top, it's clearer:
```
const tokenPrice = 9.0;
const tokenPriceDecimals = 8;
const tokenBalance = 1.0;
const tokenBalanceDecimals = 6;
const expectedUsdBalance = tokenBalance * tokenPrice;
jest.mocked(getTokenPrice).mockResolvedValue({ value: BigInt(tokenPrice * 10 ** tokenPriceDecimals), decimal: tokenPriceDecimals })
jest.mocked(getTokenBalance).mockResolvedValue({
  result: [{ value: BigInt(tokenBalance * 10 ** tokenBalanceDecimals), decimals: tokenBalanceDecimals }],
})
```
and then here
```
expect(resp.result).toBe(String(expectedUsdBalance * 10 ** RESULT_DECIMALS))
```

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    expect(resp.data).toEqual({
+      result: '90000000000000000000',
+      decimals: 18,
+    })
+  })
```
dskloetc, line 145: Please also add a test with multiple balances that need to be added together.

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -0,0 +1,131 @@
...
+          token: 'TBILL',
+          contractAddress: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6',
+        },
+        priceOracle: {
+          contractAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
+          network: 'ethereum',
+        },
+      }),
+    ).rejects.toThrow('balance fail')
+  })
```
dskloetc, line 188: Also add a test with different balances with different decimals which should cause an error.

packages/sources/token-balance/src/endpoint/solana.ts
Subarna-Singh: Noted and Fixed

packages/sources/token-balance/src/transport/solana-utils.ts
Subarna-Singh: I have informed the SA. He has asked for time to think over it.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Reverted.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Fixed

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: Done.

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: done.

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: done

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: dont

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: done

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: its not used. Removed.

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: Done

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: Okay

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: Need only one. latestAnswer

packages/sources/token-balance/test/unit/solana.test.ts
Subarna-Singh: understood. Fixed it.

packages/sources/token-balance/test/unit/solana.test.ts
Subarna-Singh: Understood. Fixed it.

packages/sources/token-balance/test/unit/solana.test.ts
Subarna-Singh: Added.

Subarna-Singh:
@dskloetc I have fixed the issues based on your comment. 
Including 

1. using getToken [will fix the function in another PR, as you suggested]
2. Added error handling, if getParsedTokenAccountsByOwner returns no value. 

Cases where value will be empty ([]):

- Wallet never received that token
- Token account closed
- Wrong mint PublicKey
- Wrong wallet PublicKey
- RPC node lag / data not yet finalized (rare)

packages/sources/token-balance/src/transport/solana.ts
dskloetc: It doesn't look done.

packages/sources/token-balance/test/integration/solana.test.ts
```diff
@@ -112,12 +92,12 @@ describe('execute', () => {
...
           },
         ],
         tokenMint: {
-          token: 'TBILL',
-          contractAddress: '4MmJVdwYN8LwvbGeCowYjSx7KoEi6BJWg8XXnW4fDDp6',
+          token: token,
+          contractAddress: mintAddress,
         },
         priceOracle: {
           contractAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
```
dskloetc: Use the constant you declared.

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -20,112 +26,283 @@ describe('solanaTransport._handleRequest', () => {
...
         priceOracle: {
-          contractAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
+          contractAddress: priceOracleAddress,
           network: 'ethereum',
         },
       }),
     ).rejects.toThrow('balance fail')
   })
+
+  it('test scaling with multiple balances, calculates correct aggregated USD result', async () => {
```
dskloetc, line 190: Nice test!

packages/sources/token-balance/test/unit/solana.test.ts
```diff
@@ -20,112 +26,283 @@ describe('solanaTransport._handleRequest', () => {
...
-    jest.mocked(getTokenBalance).mockResolvedValue({
-      result: [{ value: 1000000000n, decimals: 6 }],
+  const tokenBalanceValue = 1000
+  const tokenPriceValue = 44.8967
+  const tokenPriceDecimals = 8
+  const tokenBalanceDecimals = 6
+  const expectedUsdBalance = tokenBalanceValue * tokenPriceValue
+  jest
+    .mocked(getTokenPrice)
+    .mockResolvedValue({
```
dskloetc: Never define mock behavior directly in the `describe` block.
It's very unintuitive when this actually get executed.
If the behavior is for all tests, put it in `beforeEach`. If it's for one test, put it in the test itself.
It looks like you defined the behavior individual tests so I think you can just remove this?
But if you need to define some default behavior for tests that don't define the mock behavior, you can put it in `beforeEach`.

packages/sources/token-balance/src/transport/solana-utils.ts
```diff
@@ -30,12 +30,18 @@ export const getToken = async (
       ),
   )
 
-  const result = response
-    .flatMap((r) => r.value)
-    .map((v) => ({
+  const result = response.flatMap((r, i) => {
+    if (!r.value || r.value.length === 0) {
```
dskloetc: What prompted you to add this extra check?

Did you check that all existing feeds that use this utility function don't have wallet addresses with zero token accounts.
Previously those would just be ignored but now they cause an error.

I think it might be a good idea to have this error, since it's likely a mistake when someone uses a wallet address without accounts, but we need to consider not breaking existing feeds.

Also, this new behavior should be covered by a unit test.

packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+    await this.responseCache.write(this.name, [{ params: param, response }])
+  }
+
+  async _handleRequest(
+    param: RequestParams,
+  ): Promise<AdapterResponse<BaseEndpointTypes['Response']>> {
+    const addresses = param.addresses
+    const tokenMint = param.tokenMint
```
dskloetc: Just FYI, as this is a matter of taste.
But you can write these 2 lines as
```
const { addresses, tokenMint } = param
```

packages/sources/token-balance/src/transport/solana-utils.ts
Subarna-Singh: Should I remove it from this PR then? Instead open a separate PR and release incase it breaks something we can revert.

packages/sources/token-balance/src/transport/solana.ts
Subarna-Singh: cleaner for sure

packages/sources/token-balance/src/transport/solana-utils.ts
Subarna-Singh: I was thinking about what would break this particular piece of code. Searched about solana accounts and found out that it would not return any value in the cases list earlier by me. 

I will open a separate PR for this to isolate the change incase it breaks existing feeds.

dskloetc: Sounds good. So it wasn't triggered by an actual issue but more something you anticipated?
Even as a separate PR I would recommend looking at existing feeds that use this to see if it might affect them.

packages/sources/token-balance/test/unit/solana.test.ts
dskloetc: Can you please address this?

packages/sources/token-balance/test/integration/solana.test.ts
Subarna-Singh: Used.

packages/sources/token-balance/src/transport/solana-utils.ts
Subarna-Singh: It is anticipated. 
I'll look into existing feeds.

packages/sources/token-balance/test/unit/solana.test.ts
Subarna-Singh: Addressed

dskloetc (approved):

PR #3968, opened 2025-08-21
Support porBalance endpoint of ethereum-cl-indexer in proof-of-reserves

dskloetc:

## Description

When `proof-of-reserves` is told to use `ethereum-cl-indexer` it assumes the `etherFiBalance` endpoint and implements special handling logic for it.

`ethereum-cl-indexer` now also has a generic `porBalance` endpoint, which doesn't require special handling, except for checking that the results are valid.

## Changes

1. Pass `indexerEndpoint` to runReduceAdapter.
2. In `runReduceAdapter`, if the indexer is `ethereum-cl-indexer`, use different logic depending on the `indexerEndpoint`.
3. For the `porBalance` endpoint, check validity and then use the default handling.

## Steps to Test

1. Unit tests added.
2. Manually tested end-to-end with this request to `proof-of-reserves`:
```
    {
      "protocol": "por_address_list",
      "protocolEndpoint": "virtune",
      "accountId": "VIRSETHS",
      "network": "ethereum",
      "chainId": "1",
      "indexer": "ethereum_cl_indexer",
      "indexerEndpoint": "porBalance"
    }
```

## Quality Assurance


packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -71,6 +72,26 @@ export const runReduceAdapter = async (
...
+        if (hasInvalidResults) {
+          throw new AdapterError({
+            statusCode: 400,
+            message: 'ETHEREUM_CL_INDEXER endpoint porBalance ripcord is true',
+          })
+        }
+        // If all results are valid, use default processing below the
+        // switch block.
+        break
+      } else if (indexerEndpoint !== 'etherFiBalance') {
```
mxiao-cll: Can we just use if else without break? maybe turn the if below to an else if

dskloetc: Sounds good. I kept the block below unchanged to make it easier to review.
But I've now rearranged it into a more natural if/else flow.

packages/composites/proof-of-reserves/src/utils/reduce.ts
```diff
@@ -71,20 +72,40 @@ export const runReduceAdapter = async (
...
+        )
+        if (hasInvalidResults) {
+          throw new AdapterError({
+            statusCode: 400,
+            message: 'ETHEREUM_CL_INDEXER endpoint porBalance ripcord is true',
+          })
+        }
+        // If all results are valid, use default processing below the
+        // switch block.
+        break
```
mxiao-cll: So this can be removed?

dskloetc: Yes, it's now redundant. Removed.

dskloetc: Actually, this is necessary.
It could be moved to below line 110 or it could be kept here.
Any preference?

mxiao-cll: I think we should either use if else or use break, using both is just confusing

mxiao-cll: Yes below 110 is better

dskloetc: Done

PR #3977, opened 2025-08-26
Feat/OPDATA-4030 solactive

mmcallister-cll:
## Description
Initial commit for solactive EA

## Changes
- added price/nav endpoint
- tests

## Steps to Test
1. yarn test solactive
2. with payload
```

```

## Quality Assurance

### DIFF
--- a/packages/sources/solactive/docs/known-issues.md
+++ b/packages/sources/solactive/docs/known-issues.md
@@ -1,0 +1,3 @@
+## Variable env vars
+
+- The `clientIdPassword` input param maps the request's `clientId` field to the wildcard env var `{clientIdPassword}_PASSWORD`
--- a/packages/sources/solactive/src/endpoint/price.ts
+++ b/packages/sources/solactive/src/endpoint/price.ts
@@ -1,0 +1,55 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import {
+  AdapterRequest,
+  SingleNumberResultResponse,
+} from '@chainlink/external-adapter-framework/util'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+import { getPasswordFromEnvVar, httpTransport } from '../transport/price'
+
+export const inputParameters = new InputParameters(
+  {
+    clientId: {
+      required: true,
+      type: 'string',
+      description: 'The client ID associated with the fund',
+    },
+    isin: {
+      required: true,
+      type: 'string',
+      description: 'The ISIN identifying the fund',
+    },
+    clientIdPassword: {
+      required: true,
+      type: 'string',
+      description: 'The mapped client ID to the password env var',
+    },
+  },
+  [
+    {
+      clientId: 'abc123',
+      isin: 'A0B1C2D3',
+      clientIdPassword: 'clientId1',
+    },
+  ],
+)
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: SingleNumberResultResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'price',
+  aliases: ['nav'],
+  transport: httpTransport,
+  inputParameters,
+  customInputValidation: (
+    req: AdapterRequest<typeof inputParameters.validated>,
+  ): AdapterInputError | undefined => {
+    getPasswordFromEnvVar(req.requestContext.data.clientIdPassword)
+    return
+  },
+})
--- a/packages/sources/solactive/src/transport/price.ts
+++ b/packages/sources/solactive/src/transport/price.ts
@@ -1,0 +1,109 @@
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { BaseEndpointTypes } from '../endpoint/price'
+
+// Example Response
+// {
+//   "indexId": "<index ID>",
+//   "timestamp": 1755686255764,
+//   "level": "48.3499",
+//   "levelHigh": "48.4283",
+//   "levelLow": "47.4286",
+//   "yearLevelHigh": "49.2585",
+//   "yearLevelLow": "47.4286",
+//   "lastClosingLevel": "49.0050",
+//   "differencePercentage": "-1.34",
+//   "differencePercentage4p": "-1.3368",
+//   "differenceAbsolute": "-0.66"
+// }
+export interface ResponseSchema {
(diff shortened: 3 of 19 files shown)


packages/sources/solactive/src/endpoint/price.ts
```diff
@@ -0,0 +1,55 @@
...
+      required: true,
+      type: 'string',
+      description: 'The client ID associated with the fund',
+    },
+    isin: {
+      required: true,
+      type: 'string',
+      description: 'The ISIN identifying the fund',
+    },
+    clientIdPassword: {
```
mmcallister-cll: I _think_ this is a bad field name, open to suggestions, or i'll try to think of something else. It's a mapping for a client ID used to grab the corresponding password via env var. Thought of using this sort of mapping rather than prefixing the env var with the clientId because it is a long string of chars, but we could also use clientId directly for simplicity.

mmcallister-cll: ended up removing in favour of using clientId directly

packages/sources/solactive/docs/known-issues.md
```diff
@@ -0,0 +1,3 @@
+## Variable env vars
+
+- The `clientId` input param maps the request's `clientId` field to the wildcard env var `PASSWORD_{clientId}`
```
dskloetc: Is this an "issue"? Or why is this in a file called `known-issues.md`?

dskloetc: I have trouble understanding this sentence. What is the difference between "The `clientId` input param" and "the request's `clientId` field"?

packages/sources/solactive/src/endpoint/price.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: SingleNumberResultResponse
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'price',
+  aliases: ['nav'],
```
dskloetc: Are these aliases used? Is there something that will try to connect to `nav` instead of `price`?
Otherwise this just seems like a way to cause confusion.

packages/sources/solactive/src/transport/price.ts
```diff
@@ -0,0 +1,108 @@
...
+export const httpTransport = new HttpTransport<HttpTransportTypes>({
+  prepareRequests: (params, config) => {
+    return params.map((param) => {
+      return {
+        params: [param],
+        request: {
+          baseURL: config.API_ENDPOINT,
+          url: `/${param.clientId}/${param.isin}/performance`,
+          auth: {
+            username: param.clientId,
```
dskloetc: `clientId` comes from the request so that's something we'll commit to RDD, right?
But it's also the username for the NOP's API account?
I guess this might work if all NOPs use the same account.
But even then it seems weird to have an API username in RDD.
And this way it will be hard to later switch to each NOP having their own API account.

Would the same EA be used to connect with different usernames to the same `API_ENDPOINT`?
Otherwise, why isn't `CLIENT_ID` (or rather `API_USERNAME`?) just another environment variable like `API_ENDPONT`?

packages/sources/solactive/test/integration/adapter.test.ts
dskloetc, line 1: Please also add some unit tests.

packages/sources/solactive/docs/known-issues.md
mmcallister-cll: Yeah, good question. You are correct, it is not an issue, but this is a simple way to
1. Be able to use the readme generator without having to overwrite anything
2. Add a note about custom env vars

Originally this is was only used for reporting known issues, but i think it should be expanded. Here's where it's used in the generator:
https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/scripts/src/generate-readme/generator.ts#L136

Here's the ticket to allow for filenames other than `known-issues.md`: https://smartcontract-it.atlassian.net/browse/OPDATA-4084

With this context, OK for now/good to resolve?

mmcallister-cll: They are the same. Do you prefer one over the other?

packages/sources/solactive/src/endpoint/price.ts
mmcallister-cll: yeah, happy to only retain 1 name, though typically for nav EAs we use a nav endpoint. I'll remove price unless you feel strongly otherwise

packages/sources/solactive/src/transport/price.ts
mmcallister-cll: There are incomplete notes in the ticket, but the TLDR is client ID is used in the URL as a path param and also as the basic auth username. So yes, commit to RDD.

From a DM: 
> the ClientID might be used for multiple indexID (ISIN)

Due to the way the API works, they are only providing 1 [client ID & password] for use across NOPs. There will not be "each NOP having their own API account"

In the future, yes different client ID + password for different funds using the same API endpoint.

packages/sources/solactive/test/integration/adapter.test.ts
mmcallister-cll: For which functions?

dskloetc: For the transport. For example :https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/sources/token-balance/test/unit/xrpl.test.ts

packages/sources/solactive/src/endpoint/price.ts
dskloetc: Just `nav` is also fine.

packages/sources/solactive/docs/known-issues.md
dskloetc: Thanks for clarifying.

dskloetc: If you refer to the same thing twice, it helps make this clear by using the same phrase both times.
But the sentence says that one does a thing with the other. So if they are the same thing I don't understand what it is saying.
Using the same phrase both times, it becomes:
> The `clientId` input param maps the `clientId` input param to the wildcard env var `PASSWORD_{clientId}`

mmcallister-cll: OK I'll change it to 
> The `clientId` input param maps to an env var `PASSWORD_{clientId}` used for auth

packages/sources/solactive/test/integration/adapter.test.ts
mmcallister-cll: 🤔 i don't see any additional value in unit testing the transport for simple EAs, _maybe_ for the helper function in the transport file, and agree for any transport/utils.ts file or similar

dskloetc: OK, let's talk next week.

dskloetc (approved):

PR #3979, opened 2025-08-28
[OPDATA-3889] Full SFTP EA for FTSE

chray-zhang:
## Description
- Added sftp endpoint and sftp transport
- Input will be the Index name from a list of possible supported indicies

- Have screenshots of the return values and responses underneath
- Only supports FTSE100, RUSSELL1000, RUSSELL2000, and RUSSELL3000 as of right now
- CSV Parsing has a strict structure that it follows or expects 
- `DD` and `mm` Turn over happens at 4 PM GMT and has a 3 day fallback if file not found. We add this because the file population is varied and there is no concrete schedule that we can follow. See `downloadFile` on line `195` of `transport/sftp.ts`

## Steps to Test
Example curl commands:
```
curl -s \
  --header "Content-Type: application/json" \
  --request POST \
  --data '{
    "data": {
      "endpoint": "sftp",
      "instrument": "Russell3000INDEX"
    }
  }' \
  http://localhost:8080/ | jq
```

## Screenshots

Screenshot 2025-09-02 at 2 09 27 PM
Screenshot 2025-09-02 at 2 09 16 PM
Screenshot 2025-09-02 at 2 09 07 PM
Screenshot 2025-09-02 at 2 08 55 PM

## Quality Assurance

- [ x] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.
- [ x] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `adapter-secrets` configuration file or update the [soak testing blacklist](/packages/scripts/src/get-changed-adapters/soakTestBlacklist.ts).
- [ x] If a new adapter was made, or a new endpoint was added, update the `test-payload.json` file with relevant requests.
- [ x] This is related to a maximum of one Jira story or GitHub issue.
- [ x] Types are safe (avoid TypeScript/TSLint features like any and disable, instead use more specific types).
- [ x] All code changes have 100% unit and integration test coverage. If testing is not applicable or too difficult to justify doing, the reasoning should be documented explicitly in the PR.


packages/sources/ftse-sftp/src/config/index.ts
```diff
@@ -0,0 +1,32 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const NAME = 'SFTP_GENERIC'
```
mmcallister-cll: ?

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+  }
+
+  private async connectToSftp(): Promise<void> {
+    // Check if already connected
+    if (this.isConnected) {
+      return
+    }
+
+    if (!this.config.SFTP_HOST) {
```
mmcallister-cll, line 113: may not need this if the config is `required: true` but ok to leave it too

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+    } else {
+      throw new AdapterInputError({
+        statusCode: 400,
+        message: 'SFTP_PASSWORD must be provided',
+      })
+    }
+
+    try {
+      // Add a timeout wrapper around the connection
+      const connectPromise = this.sftpClient.connect(connectConfig)
```
mmcallister-cll: Rather than Promise.race you might be able to add `readyTimeout: 30000` into the sftpConfig to achieve the same goal with less LOC. You might get a more descriptive (from base lib) error as well.

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  private async disconnectFromSftp(): Promise<void> {
+    try {
+      await this.sftpClient.end()
+      this.isConnected = false
+      logger.debug('Disconnected from SFTP server')
+    } catch (error) {
+      this.isConnected = false
+      logger.warn(error, 'Error while disconnecting from SFTP server')
```
mmcallister-cll, line 162: could this cause some sort of memory leak?

packages/sources/ftse-sftp/src/util.ts
mmcallister-cll, line 1: can remove if unused

packages/sources/ftse-sftp/src/endpoint/sftp.ts
```diff
@@ -0,0 +1,62 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import { sftpTransport } from '../transport/sftp'
+
+export const inputParameters = new InputParameters(
+  {
+    operation: {
```
mmcallister-cll: If you think it's appropriate, you can add an array of valid options to this, the validation will take place at the framework level rather than manually doing it in the transport.processFiles(). Alternatively, if you don't foresee us needing anything other than download then you can consider removing this input param altogether

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+  }
+
+  private async processFiles(param: RequestParams): Promise<any> {
+    if (param.operation === 'download') {
```
mmcallister-cll: may be worth using guards here rather than nested ifs

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+    const now = new Date()
+    const londonTime = new Date(now.toLocaleString('en-US', { timeZone: 'Europe/London' }))
+
+    // Check if it's before 4 PM London time (16:00)
+    const isBeforeFileGeneration = londonTime.getHours() < 16
+
+    // Start with current London date, but go back one day if before 4 PM
+    const targetDate = new Date(londonTime)
+    if (isBeforeFileGeneration) {
+      targetDate.setDate(londonTime.getDate() - 1)
```
mmcallister-cll: is there any flux in the published file times or are they guaranteed to always be there by 4PM UK?

packages/sources/ftse-sftp/src/transport/sftp.ts
```diff
@@ -0,0 +1,303 @@
...
+      })
+    }
+  }
+
+  private buildFilePath(remotePath: string, instrument: string): string {
+    const filePathTemplate = this.getInstrumentFilePath(instrument)
+
+    // Get current date in London time
+    const now = new Date()
+    const londonTime = new Date(now.toLocaleString('en-US', { timeZone: 'Europe/London' }))
```
mmcallister-cll, line 276: double checking cause we've had weird behaviour with timezones and daylight savings in the past (who hasn't lol). Publishing is at 4PM London time, not a specific UTC time, right?

packages/sources/ftse-sftp/test/integration/__snapshots__/adapter.test.ts.snap
```diff
@@ -0,0 +1,147 @@
...
+  "error": {
+    "message": "[Param: operation] param is required but no value was provided",
+    "name": "AdapterError",
+  },
+  "status": "errored",
+  "statusCode": 400,
+}
+`;
+
+exports[`execute sftp endpoint should handle missing required parameters 1`] = `
```
mmcallister-cll, line 14: i think the previous tests already cover this?

packages/sources/ftse-sftp/src/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+export interface SftpResponse {
+  operation: 'list' | 'download'
+  timestamp: number
+  [key: string]: any
+}
+
+export interface SftpListResponse extends SftpResponse {
```
mmcallister-cll: do we use list anywhere?

packages/sources/ftse-sftp/src/transport/sftp.ts
chray-zhang: Added a finally statement to make sure we catch this

packages/sources/ftse-sftp/src/transport/sftp.ts
chray-zhang: This is still a WIP. We don't actually have a schedule cadence for the file, so I'll add a fallback instead to the previous days file for now

packages/sources/ftse-sftp/src/transport/sftp.ts
chray-zhang: Same as above. No guarantee, I added a fallback method

packages/sources/ftse-sftp/src/endpoint/sftp.ts
chray-zhang: You're right. I gave it some thought and I don't think we need this parameter. I'll remove it for now and add it back if we need it in the future

mxiao-cll:
Re-open if still relevent

PR #3984, opened 2025-09-01
Add reserve endpoint and test to TNF adapter

cl-mayowa:
## Description

The Network Firm provides reserve/health data via client-specific endpoints under a single base URL. The current the-network-firm External Adapter (EA) requires a code change for every new client endpoint. This limits scalability when onboarding new feeds.

Blocker: 
- The reserve endpoint test block currently fails. The only test that passed from the block is `should return success for uranium`. For some reason I cannot explain yet, the nock does not intercept the request from the transport. 
......

## Changes

- Added a new reserve endpoint
- Added tests for the new reserve endpoint

## Steps to Test

- This change can be tested by running the command `yarn test the-network-firm/test/integration/adapter.test.ts --testNamePattern="reserve endpoint"`

## Quality Assurance

[OPDATA-4094]: https://smartcontract-it.atlassian.net/browse/OPDATA-4094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/the-network-firm/src/endpoint/reserve.ts
```diff
@@ -0,0 +1,48 @@
+// import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
```
dskloetc: Should this line be here?

packages/sources/the-network-firm/src/endpoint/reserve.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+    resource: {
+      type: 'string',
+      required: true,
+      description: 'The resource of the TNF to consume',
+    },
+    apiKey: {
+      type: 'string',
+      description:
+        'Alternative api key to use for this request, {$apiKey}_API_KEY  required in environment variables',
```
dskloetc: Alternative to what? Is there a default API key?
I think this should have a better name.
What does the API key belong to?
Does a NOP have a separate API key per client? If so, can we derive the environment variable name from the client parameter instead?
Or if it's an API key per <something else>, perhaps that something else would suggest a better naming for this parameter.

packages/sources/the-network-firm/src/endpoint/reserve.ts
```diff
@@ -0,0 +1,48 @@
...
+      default: '',
+    },
+  },
+  [
+    {
+      client: 'acme',
+      resource: 'reserve',
+      apiKey: '',
```
dskloetc: It would be helpful to have a realistic value for `apiKey` in the example request.

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,105 @@
...
+      throw new AdapterInputError({
+        statusCode: 400,
+        message: `Missing '${apiKeyName}' environment variables.`,
+      })
+    }
+
+    return apiKeyValue
+  }
+
+  return config.API_KEY
```
dskloetc: Is it plausible that the default `API_KEY` (which is presumable configured for other endpoints?) would work for this endpoint? Or should we just make this input parameter required?

packages/sources/the-network-firm/src/config/index.ts
```diff
@@ -25,4 +25,11 @@ export const config = new AdapterConfig({
     sensitive: true,
     default: '',
   },
+  API_KEY: {
```
Subarna-Singh: Would remove default value from API_ENDPOINT. 
This is throw an error if not set, else it will assume the same default value for all clients.

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const httpTransport = new HttpTransport<HttpTransportTypes>({
+  prepareRequests: (params, config) => {
+    const client = params[0].client
+    const resource = params[0].resource
+    return {
+      params,
+      request: {
+        baseURL: config.ALT_API_ENDPOINT,
```
Subarna-Singh, line 41: Could I get clarification?
Why use the ALT_API_ENDPOINT and not API_ENDPOINT?

packages/sources/the-network-firm/src/endpoint/reserve.ts
```diff
@@ -0,0 +1,48 @@
...
+import { httpTransport } from '../transport/reserve'
+
+export const inputParameters = new InputParameters(
+  {
+    client: {
+      type: 'string',
+      required: true,
+      description: 'The name of the TNF client to consume from',
+    },
+    resource: {
```
Subarna-Singh: What is the resource?
Need a clearer description and maybe an example

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,105 @@
+import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { BaseEndpointTypes } from '../endpoint/reserve'
+
+export interface ResponseSchema {
```
Subarna-Singh, line 5: Will all clients adhere to this schema?
Have you tested it with an existing client?

packages/sources/the-network-firm/src/endpoint/reserve.ts
cl-mayowa: @dskloetc I already derive the environment variable name from the client parameter in the `transport/reserve.ts` file. Regarding the spit, the ticket says: the EA should `Supports API key authentication (global or per-client).`. I added the `apiKey` input parameter to indicate that the request does not use a client api key, but a global api key. 

Do you have any recommendations here for a better implementation?

packages/sources/the-network-firm/src/endpoint/reserve.ts
cl-mayowa: Alright, I will get this from Michael Imperiale

packages/sources/the-network-firm/src/endpoint/reserve.ts
dskloetc: > I already derive the environment variable name from the client parameter in the transport/reserve.ts file.

Can you point to and copy the exact line? Because I don't see it.

cl-mayowa: > > I already derive the environment variable name from the client parameter in the transport/reserve.ts file.
> 
> Can you point to and copy the exact line? Because I don't see it.

https://github.com/smartcontractkit/external-adapters-js/pull/3984/files#diff-c083d0e8dc1c5cabe1087c5f4a2fe975b965d5560df12606086a0fa1334c4d6cR21

getApiKey function, line 2, transport/reserve.ts

packages/sources/the-network-firm/src/transport/reserve.ts
cl-mayowa: This is something that I need to confirm with Michael. Will be back to this when I get a response.

packages/sources/the-network-firm/src/transport/reserve.ts
cl-mayowa: ALT_API_ENDPOINT is the new api endpoint for TNF. That is why I am using it.

packages/sources/the-network-firm/src/endpoint/reserve.ts
cl-mayowa: resource is reserve, ripcord, etc.

packages/sources/the-network-firm/src/transport/reserve.ts
cl-mayowa: Michael mentioned that all clients will adhere to this schema. I will do some tests with existing clients and include the report in the PR

packages/sources/the-network-firm/src/config/index.ts
cl-mayowa: Alright, removed it and left the client generated API_KEY

.changeset/real-mangos-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/the-network-firm-adapter': minor
+---
+
+Update The Network Firm EA with a new client-agnostic endpoint called
```
dskloetc: called...?

Is the sentence unfinished?

packages/sources/the-network-firm/src/endpoint/reserve.ts
dskloetc: Discussed on Slack. It wasn't already derived from the client parameter but it is now.

packages/sources/the-network-firm/src/config/index.ts
dskloetc: Looks like it's still here. Did you forget to push?

packages/sources/the-network-firm/src/endpoint/reserve.ts
mxiao-cll: Maybe a better name is path?

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,101 @@
...
+        url: `/v1/${client}/${resource}`,
+        headers: {
+          apikey: getApiKey(params[0].client),
+        },
+      },
+    }
+  },
+  parseResponse: (params, response) => {
+    return params.map((param) => {
+      const ripcord = response.data.ripcord || response.data.ripcord.toString() === 'true'
```
mxiao-cll: .toLowerCase

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,101 @@
...
+          },
+        }
+      }
+
+      const client = params[0].client
+      if (!response.data.name || response.data.name !== client) {
+        return {
+          params: param,
+          response: {
+            errorMessage: 'Provider did not return resource for client',
```
mxiao-cll: Better to log both response.data.name and client here

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export type HttpTransportTypes = BaseEndpointTypes & {
+  Provider: {
+    RequestBody: never
+    ResponseBody: ResponseSchema
+  }
+}
+
+export const getApiKey = (client: string) => {
```
mxiao-cll, line 21: You need to use this in endpoint file to validate input before it reaches transport layer, the difference is that if request is rejected in transport it will loop again and again wasting resources. See coinbase-prime for example

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -0,0 +1,94 @@
...
+  return apiKeyValue
+}
+
+export const httpTransport = new HttpTransport<HttpTransportTypes>({
+  prepareRequests: (params, config) => {
+    const client = params[0].client
+    console.log(
+      'request hit here',
+      config.ALT_API_ENDPOINT,
+      getApiKey(params[0].client),
```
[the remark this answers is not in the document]
Subarna-Singh: Please don't log API_Key.

packages/sources/the-network-firm/test/integration/adapter.test.ts
```diff
@@ -28,6 +31,9 @@ describe('execute', () => {
     process.env.ALT_API_ENDPOINT = 'http://test-endpoint-new'
     process.env.EMGEMX_API_KEY = 'api-key'
     process.env.URANIUM_API_KEY = 'api-key'
+    process.env.ACME_API_KEY = 'acme-api-key'
+    process.env.URANIUM_DIGITAL_QOHMMJQAF4JK_API_KEY = 'uranium-api-key'
+    process.env.EMGEMX_TDFKF3_API_KEY_API_KEY = 'emgemx-api-key'
```
Subarna-Singh: correct to EMGEMX_TDFKF3_API_KEY

packages/sources/the-network-firm/src/transport/reserve.ts
cl-mayowa: We have tested this with the new response schema for the following supported clients: `uranium-digital-qohmmjqaf4jk`,  `emgemx-tdfkf3`, `m0-stablecoin-inpd83`, `re-protocol-8tawlm`, and the API returns the same response schema for all of them. However older schema will remain on the legacy endpoint and their EA endpoint will remain unchanged.

packages/sources/the-network-firm/src/transport/reserve.ts
cl-mayowa: This will be implemented.

packages/sources/the-network-firm/test/integration/adapter.test.ts
cl-mayowa: removed, thanks

mxiao-cll (approved):

PR #3992, opened 2025-09-03
Add new transaction endpoint to superstate

mxiao-cll:
## Description

......

## Changes

- High level
- changes that
- you made

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance


packages/sources/superstate/src/config/index.ts
```diff
@@ -18,6 +26,12 @@ export const config = new AdapterConfig(
       type: 'number',
       default: 60000,
     },
+    BACKGROUND_EXECUTE_MS: {
```
mmcallister-cll, line 29: double checking.. we're sure this won't affect the other endpoints passively (through the framework), right?

packages/sources/superstate/src/endpoint/nav.ts
```diff
@@ -33,9 +33,11 @@ export type BaseEndpointTypes = {
   Settings: typeof config.settings
 }
 
+export const transport = new NavTransport()
```
mmcallister-cll: I think usual convention is export this in the transport? This is also fine though

packages/sources/superstate/src/endpoint/transactions.ts
```diff
@@ -0,0 +1,72 @@
...
+import { ResponseSchema } from '../transport/transactionUtils'
+
+export const inputParameters = new InputParameters(
+  {
+    fundId: {
+      description: 'Used to fetch Net Asset Value',
+      type: 'number',
+      required: true,
+    },
+    ticker: {
```
mmcallister-cll, line 19: should we add status here too as an input or are we ok with just pending below?

packages/sources/superstate/src/transport/transactionUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+  ticker: string
+  operation_type: string
+  status: string
+  share_amount: string
+  dollar_amount: string
+  notional_value: string
+}
+
+export const getTransactions = async (
+  apiKey: string | undefined,
```
mmcallister-cll, line 17: since it's a util, maybe force string here and do the || '' in the caller? 🤔 
I'm also fine with it as-is

packages/sources/superstate/src/transport/transactions.ts
```diff
@@ -0,0 +1,109 @@
...
+        transactions,
+        navPrice: nav,
+        decimals: param.decimals,
+      },
+      statusCode: 200,
+      result,
+      timestamps: {
+        providerDataRequestedUnixMs,
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: undefined,
```
mmcallister-cll, line 99: should we use one of `created_at, completed_at`?

packages/sources/superstate/src/config/index.ts
mxiao-cll: No, the other transport is not a subscription transport

packages/sources/superstate/src/endpoint/nav.ts
mxiao-cll: I will move

packages/sources/superstate/src/endpoint/transactions.ts
mxiao-cll: Yes we should add

packages/sources/superstate/src/transport/transactionUtils.ts
mxiao-cll: Actually I will throw here

packages/sources/superstate/src/transport/transactionUtils.ts
```diff
@@ -0,0 +1,63 @@
...
+    method: 'GET',
+    queryParams: {
+      transaction_status: transactionStatus,
+    },
+  })) as ResponseSchema[]
+
+  const filtered = (transactions ?? [])
+    .filter((t) => t && t.ticker && t.operation_type)
+    .filter((t) => t.ticker.toUpperCase() == ticker.toUpperCase())
+    .filter((t) =>
```
karen-stepanyan, line 40: why not to use one filter ? unless there are just few transactions

mxiao-cll: I feel like code looks cleaner this way.

We would only get about 48 hours of transaction here, I won't expect it to be more than 10

karen-stepanyan (approved):

PR #4000, opened 2025-09-05
Initial implementation of solana-functions adapter

fmunshi:
## Description

Adds a new EA for "solana functions" 
This is will be an EA for one off computations based on one or more attributes on a single solana program

The first of this is for exchange rate of EUSX/USX 
......

## Changes

- Adds a new EA for solana functions with a single endpoint `eusx-price`.

## Steps to Test

1. `export RPC_URL=https://rpcs.cldev.sh/solana/mainnet` && yarn build && yarn start
2.  `curl -s --header "Content-Type: application/json" --request POST --data '{ "data":  {"endpoint":"eusx-price"} }' http://localhost:8080 | jq`

## Quality Assurance

[OPDATA-4060]: https://smartcontract-it.atlassian.net/browse/OPDATA-4060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/solana-functions/src/utils/anchor.ts
```diff
@@ -0,0 +1,9 @@
+import type { Program } from '@coral-xyz/anchor'
+import * as anchor from '@coral-xyz/anchor'
+import * as YieldVaultIDL from '../idl/eusx_yield_vault.json'
+import type { YieldVault } from '../types/eusx_yield_vault'
+
+// Wrapper to create the yield vault anchor program
```
fmunshi: My thinking with this file is that we can initialize all future specific programs we use in here too 

This makes it easy to mock this file for tests.

Im sure there are ways to mock the program without creating an extra file like this with these wrapper functions but these seemed the easiest and cleanest way IMO 

Open to suggestions if anyone has any

augustbleeds (approved):
lgtm

packages/sources/solana-functions/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "clean": "rm -rf dist && rm -f tsconfig.tsbuildinfo",
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "devDependencies": {
+    "@coral-xyz/anchor": "^0.31.1",
+    "@solana/web3.js": "^1.98.4",
```
mxiao-cll: Can we use @solana/kit?

packages/sources/solana-functions/src/config/overrides.json
```diff
@@ -0,0 +1,3 @@
+{
```
mxiao-cll: Delete this file

packages/sources/solana-functions/src/index.ts
```diff
@@ -0,0 +1,13 @@
+import { expose, ServerInstance } from '@chainlink/external-adapter-framework'
+import { Adapter } from '@chainlink/external-adapter-framework/adapter'
+import { config } from './config'
+import { eusxPrice } from './endpoint'
+
+export const adapter = new Adapter({
+  defaultEndpoint: eusxPrice.name,
+  name: 'SOLANA_FUNCTIONS',
```
mxiao-cll, line 8: This should be EA name right?

packages/sources/solana-functions/test-payload.json
```diff
@@ -0,0 +1,5 @@
+{
```
mxiao-cll: Remove this file

packages/sources/solana-functions/src/utils/anchor.ts
mxiao-cll: If for testing purposes only, you can mock `new anchor.Program` as well

There is plenty of examples of how we mock new Contract for evm in the code base

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,147 @@
...
+  provider?: anchor.Provider
+  wallet?: anchor.Wallet
+
+  // Get Connection creates or returns an existing Solana connection
+  getConnection(): solanaWeb3.Connection {
+    if (this.connection) {
+      return this.connection
+    }
+    const rpcUrl = process.env.RPC_URL
+    const commitment = 'confirmed' as solanaWeb3.Commitment
```
mxiao-cll: Load this from env var as well

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!rpcUrl) throw new Error('RPC_URL not set')
+    const connection = new solanaWeb3.Connection(rpcUrl, commitment)
+    this.connection = connection
+    return connection
+  }
+
+  // Get Provider creates or returns an existing Anchor provider
+  getProvider(): anchor.Provider {
```
mxiao-cll: We can prob just do this once in the init?

mxiao-cll:
I don't see this as a generic Solana EA, prob re-name to `eusx` or generalize more

packages/sources/solana-functions/package.json
fmunshi: I think if i use solana kit for the RPC / connection i cant really use anchor - do we prefer to do everything with solana/kit?

Im not sure whats the preferred library internally

packages/sources/solana-functions/src/index.ts
fmunshi: yes the name is `solana-functions` so this is correct is it not?

packages/sources/solana-functions/src/utils/anchor.ts
fmunshi: yea not sure why i didnt do that - done

packages/sources/solana-functions/package.json
mxiao-cll: I was able to do it here: https://github.com/smartcontractkit/secure-mint-indexer/pull/46/files#diff-eaa88c9c9ab27fd7e933b70d84237ad681bb7fde8e5f518c44573e3eb0c2728eR1

There is no preferred library, asking just because kit is the newer version

packages/sources/solana-functions/src/index.ts
mxiao-cll: This is correct

augustbleeds (approved):

packages/sources/solana-functions/src/endpoint/eusx-price.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'eusx-price',
+  aliases: ['eusx', 'eusx-rate'],
+  transport: eUSXPriceTransport,
+  inputParameters,
+  overrides: {},
```
mxiao-cll: Delete this

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
```diff
@@ -0,0 +1,2160 @@
+{
+  "address": "eUSXyKoZ6aGejYVbnp3wtWQ1E8zuokLAJPecPxxtgG3",
```
mxiao-cll, line 2: I think this is a bad idea to hard-code address with-in the schema

Better to pass address in as a param?

packages/sources/solana-functions/src/shared/account_reader.ts
```diff
@@ -0,0 +1,31 @@
+import { BorshCoder, Idl } from '@coral-xyz/anchor'
+import { type Address } from '@solana/addresses'
+import { createSolanaRpc, type Rpc, type SolanaRpcApi } from '@solana/rpc'
+
+export class SolanaAccountReader {
```
mxiao-cll: Do we now use commitment level here?

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
fmunshi: Yea that makes more sense - this was the IDL they gave us and it had the address so I was parsing it from the IDL but we can make it an argument on the endpoint

packages/sources/solana-functions/src/shared/account_reader.ts
```diff
@@ -0,0 +1,31 @@
...
+  rpc: Rpc<SolanaRpcApi>
+
+  initSolanaRpc = (): Rpc<SolanaRpcApi> => {
```
mxiao-cll: I'd put Rpc in a separate class, and pass it in for each function call

packages/sources/solana-functions/src/shared/account_reader.ts
fmunshi: no longer needed - forgot to remove it

packages/sources/solana-functions/src/shared/account_reader.ts
fmunshi: not sure if theres much value there is in a separate class that just wraps the RPC class from solana/rpc

fmunshi: i could initialize it in a helper function and call it from the transport and then pass that into this account reader

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,150 @@
...
+    const vestingSchedule = await accountReader.fetchAccountInformation<VestingSchedule>(
+      vestingSchedulePda as Address,
+      VESTING_SCHEDULE_ACCOUNT_NAME,
+      YieldVaultIDL as Idl,
+    )
+    const yieldPool = await accountReader.fetchAccountInformation<YieldPool>(
+      yieldPoolPda as Address,
+      YIELD_POOL_ACCOUNT_NAME,
+      YieldVaultIDL as Idl,
+    )
```
mxiao-cll: 1. We should parallel these calls
2. The utility function should handle more, we should move `getProgramDerivedAddress` inside `fetchAccountInformation`

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // Calculate the EUSX price
+    const result = this.calcEusxPrice(
+      yieldPool.shares_supply.toNumber(),
+      yieldPool.total_assets.toNumber() - unvestedAmount,
+    )
+
+    return {
+      data: {
+        result,
```
mxiao-cll, line 148: Consider return more intermediate data for debugging purposes

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,150 @@
...
+      },
+      statusCode: 200,
+      result,
+      timestamps: {
+        providerDataStreamEstablishedUnixMs: Date.now(),
```
mxiao-cll: Where is data stream?

packages/sources/solana-functions/src/shared/account_reader.ts
fmunshi: @mxiao-cll i've decoupled the RPC from this class and moved it back into the transport (initialized separately to be re-used across transports)

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
fmunshi: done

packages/sources/solana-functions/src/shared/account_reader.ts
mxiao-cll: I maybe thinking ahead of their calls we might make to the Rpc later down the road

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: I have no idea what this is typically used for but removing it gives me a host of type errors - what should be done here?

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: return or log? - or do you mean split the endpoint into multiple endpoints?

mxiao-cll: return, for example return `yieldPool` and `vestingSchedule`

Anything you feel that might be helping when for example, 10 NOPs return 100 in their EA and 1 NOP is returning 110

packages/sources/solana-functions/src/transport/eusx-price.ts
mxiao-cll: Should use `providerDataRequestedUnixMs` instead, this is the time when you request data to DP

fmunshi: gotcha - fixed

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: gotcha - returned all intermediate values in this - it will indeed be helpful for debugging.

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: parallelized and pushed PDA generation inside of the fetchAccount call 
However im still generating the seeds inside this call

fmunshi:
well except for `chalk-template`... let me see if i can see whats importing the newest version of that to avoid it

mxiao-cll (dismissed):

packages/sources/solana-functions/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "prepack": "yarn build",
+    "build": "tsc -b",
+    "server": "node -e 'require(\"./index.js\").server()'",
+    "server:dist": "node -e 'require(\"./dist/index.js\").server()'",
+    "start": "yarn server:dist"
+  },
+  "devDependencies": {
+    "@types/bn.js": "^5.1.0",
+    "@types/jest": "27.5.2",
+    "@types/node": "16.18.119",
```
mmcallister-cll: update these to be in line with the other EAs, then you can remove some of the added dependencies

packages/sources/solana-functions/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
...
+    description: 'The RPC URL for the Solana cluster',
+    type: 'string',
+    required: true,
+    sensitive: true,
+  },
+  BACKGROUND_EXECUTE_MS: {
+    description:
+      'The amount of time the background execute should sleep before performing the next request',
+    type: 'number',
+    default: 1_000,
```
mmcallister-cll, line 14: i guess the reason to go this fast is the fast block times?

packages/sources/solana-functions/src/endpoint/eusx-price.ts
```diff
@@ -0,0 +1,47 @@
...
+      address: 'eUSXyKoZ6aGejYVbnp3wtWQ1E8zuokLAJPecPxxtgG3',
+    },
+  ],
+)
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: {
+    Data: {
+      result: number
```
mmcallister-cll, line 25: is this a standard reusable schema or a one off the customer is ok with? (or is the intent to just use the result and provide this additional metadata for future use?)

packages/sources/solana-functions/src/endpoint/eusx-price.ts
```diff
@@ -0,0 +1,47 @@
...
+      unvestedAmount: number
+    }
+    Result: number
+  }
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'eusx-price',
+  aliases: ['eusx', 'eusx-rate'],
```
mmcallister-cll: can probably get rid of these unless you expect them to ever be used?

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
mmcallister-cll: errrr, `address` is still here, is this correct?

packages/sources/solana-functions/src/shared/account-reader.ts
```diff
@@ -0,0 +1,42 @@
...
+  async fetchAccountInformation<T>(
+    rpc: Rpc<SolanaRpcApi>,
+    address: Address,
+    accountName: string,
+    idl: Idl,
+  ): Promise<T> {
+    const encoding = 'base64'
+    const resp = await rpc.getAccountInfo(address, { encoding }).send()
+    const value = resp.value
+    if (!value || !value.data || value.data[0] == null) {
```
mmcallister-cll: you might be able to use something like `!value?.data?.[0]` for this condition to collapse the 3 checks

packages/sources/solana-functions/src/shared/solana-rpc-factory.ts
```diff
@@ -0,0 +1,14 @@
+import { createSolanaRpc, type Rpc, type SolanaRpcApi } from '@solana/rpc'
+export class SolanaRpcFactory {
+  private readonly defaultUrl?: string
+
+  constructor(override?: string) {
+    this.defaultUrl = override ?? process.env.RPC_URL
+  }
+
+  create(url?: string): Rpc<SolanaRpcApi> {
```
mmcallister-cll, line 9: should we just default like (url: string = this.defaultUrl)?

mmcallister-cll: actually, do we really need a class for this? nothing wrong with it tho

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,180 @@
...
+
+const logger = makeLogger('EUSXPriceTransport')
+
+export type EUSXPriceTransportTypes = BaseEndpointTypes
+
+type RequestParams = typeof inputParameters.validated
+
+// Types for Solana accounts - snake_case due to using BorschCoder directly instead of Anchor
+type VestingSchedule = {
+  vesting_amount?: BN
```
mmcallister-cll, line 26: maybe silly q, is it possible to avoid this extra dep by using some built-in or does this BN type make it simpler/necessary?

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,180 @@
...
+      const totalAssets = yieldPool.total_assets?.toNumber?.()
+      if (sharesSupply == null || totalAssets == null) {
+        throw new Error('Invalid vesting schedule fields')
+      }
+
+      const lastVestingAmount = vestingAmount
+      const vestingDuration = end - start
+      const vestingEnd = end
+
+      const now = Math.floor(providerDataRequestedUnixMs / 1000)
```
mmcallister-cll: can specify this is nowSec or something

Bigger question: should we specify `vestingSchedule.[start|end]Sec` in the response payload to be super explicit?

packages/sources/solana-functions/src/transport/eusx-price.ts
```diff
@@ -0,0 +1,180 @@
...
+        timestamps: {
+          providerDataRequestedUnixMs,
+          providerDataReceivedUnixMs: Date.now(),
+          providerIndicatedTimeUnixMs: undefined,
+        },
+      }
+    } catch (err) {
+      const errorMsg = `Failed to calculate EUSX price: ${err}`
+      logger.error(errorMsg)
+      throw new Error(errorMsg)
```
mmcallister-cll: do we really need this try/catch and rethrow here? Isn't the try/catch in `handleRequest()` sufficient?

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
fmunshi: well this is the IDL as solstice labs provided it - i left the file as is but am not parsing the address from here 
I guess i can remove it from here too

packages/sources/solana-functions/src/config/index.ts
fmunshi: Im not sure - I just assumed 1s since its going to be used with streams?

packages/sources/solana-functions/src/endpoint/eusx-price.ts
fmunshi: yea the additional metadata was added for debugging purposes.

`result` is all we really need

packages/sources/solana-functions/src/endpoint/eusx-price.ts
fmunshi: yea dont expect usage

packages/sources/solana-functions/src/shared/solana-rpc-factory.ts
fmunshi: ` (url: string = this.defaultUrl)?` 
does not quite work since defaultUrl is optional

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: I dont think a type that can be used outside of BN here that makes sense

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: true and theres already a logger.error there - this makes no sense

packages/sources/solana-functions/src/transport/eusx-price.ts
fmunshi: specified seconds

vestingSchedule is really only used so the person making the request can understand what the values were that were used to calculate the price

packages/sources/solana-functions/src/shared/account-reader.ts
fmunshi: done

packages/sources/solana-functions/src/config/index.ts
fmunshi: If this should change lmk

packages/sources/solana-functions/src/shared/solana-rpc-factory.ts
fmunshi: a class is a bit overkill but its just a bit cleaner and easier to use than a single utility function and its more extensible for other env vars if we need to in the future

mmcallister-cll (approved):

packages/sources/solana-functions/src/idl/eusx_yield_vault.json
mmcallister-cll: IDL is directly as provided, leaving as is, will revisit solution if we ever need to.

PR #4001, opened 2025-09-05
Add wallet endpoint to ceffu

mxiao-cll:
## Description

Add a new wallet endpoint 

## Changes
 - New endpoint that fetches wallet first then assets then using price feed to convert to USD and sum

## Quality Assurance


packages/sources/ceffu/src/endpoint/wallet.ts
```diff
@@ -0,0 +1,97 @@
...
+          description: 'Name of the token',
+        },
+        address: {
+          required: true,
+          type: 'string',
+          description: 'Contract address',
+        },
+      },
+      array: true,
+      description: 'List of */USD price feeds on arbitrum',
```
dskloetc: Is the idea that this is only a backup mechanism for tokens that aren't listed in the config?
If so, I think it would be good to document that clearly.
Because now it's confusing why there is both the list in the config and in the input params.

packages/sources/ceffu/src/endpoint/wallet.ts
```diff
@@ -0,0 +1,97 @@
...
+  name: 'wallet',
+  transport: walletTransport,
+  inputParameters,
+  requestTransforms: [
+    (request) => {
+      request.requestContext.data.contracts.forEach((c) => {
+        c.token = c.token.toUpperCase()
+      })
+
+      const existingToken = new Set(request.requestContext.data.contracts.map((c) => c.token))
```
dskloetc: What does "existing" mean?
Are the input params more existing than the hardcoded config?

packages/sources/ceffu/src/transport/rate.ts
```diff
@@ -0,0 +1,58 @@
...
+        amount: c.amount,
+        rate: rate.value,
+        decimal: rate.decimal,
+        value,
+      }
+    }),
+  )
+}
+
+export const getUSDRate = async (contracts: string, provider: ethers.JsonRpcProvider) => {
```
dskloetc, line 36: Do we know that it's acceptable to have the additional delay that comes from reading conversion rates from a feed instead of directly?

I needed to make a similar consideration for the CMETH feed and Thomas answered this:
https://chainlink-core.slack.com/archives/C091MDPGF36/p1754600955736499?thread_ts=1754557987.284319&cid=C091MDPGF36

But I don't know to what extent that applies here.

packages/sources/ceffu/src/transport/mirrorX.ts
```diff
@@ -36,7 +36,7 @@ export const getAssetPositions = async (
       const params = {
         mirrorXLinkId: id,
         excludeZeroAmountFlag: true,
-        pageLimit: 25,
+        pageLimit: 500,
```
dskloetc, line 41: What's the reason for this change?

packages/sources/ceffu/src/transport/mirrorX.ts
```diff
@@ -1,9 +1,9 @@
-import { createPrivateKey, createSign } from 'crypto'
 import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import { AdapterError } from '@chainlink/external-adapter-framework/validation/error'
+import { Decimal } from 'decimal.js'
 import { HttpsProxyAgent } from 'https-proxy-agent'
 import { stringify } from 'querystring'
-import { Decimal } from 'decimal.js'
-import { AdapterError } from '@chainlink/external-adapter-framework/validation/error'
+import { sign } from './wallet/requester' // TODO: Refactor to use request instead
```
dskloetc: How would one use request instead?
Do you plan to do that soon?

packages/sources/ceffu/src/transport/rate.ts
```diff
@@ -0,0 +1,58 @@
...
+  return {
+    value,
+    decimal: Number(decimal),
+  }
+}
+
+const getContract = (coin: string, contracts: Record<string, string>) => {
+  if (!contracts[coin.toUpperCase()]) {
+    throw new AdapterInputError({
+      statusCode: 500,
```
dskloetc: Isn't this more a 400 than a 500?

packages/sources/ceffu/src/transport/rate.ts
```diff
@@ -0,0 +1,58 @@
...
+  ])
+
+  return {
+    value,
+    decimal: Number(decimal),
+  }
+}
+
+const getContract = (coin: string, contracts: Record<string, string>) => {
+  if (!contracts[coin.toUpperCase()]) {
```
dskloetc: If you start with converting `coin` to upper case, then you don't need to call `toUpperCase()` 3 times.

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -0,0 +1,88 @@
...
+    pageLimit: number
+  }
+  code: string
+  message: string
+}
+
+export const request = async <T>(
+  baseUrl: string,
+  url: string,
+  params: Record<string, string | number>,
```
dskloetc: Do we not know more precisely the params we will pass?

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -0,0 +1,88 @@
...
+  proxy: string,
```
dskloetc: It's a bit surprising that `proxy` is optional but that we use empty string instead of `undefined` to indicate its absence.
Can we make its type `string?`?
If it has to be like this, can you add a comment documenting it?

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -0,0 +1,88 @@
...
+      httpsAgent: proxy.length > 0 ? new HttpsProxyAgent(proxy) : null,
+      headers: {
+        'open-apikey': apiKey,
+        signature: sign(stringify(params), privateKey),
+      },
+      params,
+    }
+
+    const response = await requester.request<GeneralResponse<T>>(
+      JSON.stringify(requestConfig),
```
dskloetc, line 58: There is `calculateHttpRequestKey` which I suppose its purpose is exactly to use it here?
Do you know why it exists if `JSON.stringify` works?
Why is it even a separate parameter rather than `requester.request` just doing `JSON.stringify` itself?

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (!response || !response.response || !response.response.data) {
+      throw new AdapterError({
+        statusCode: 500,
+        message: `Ceffu wallet API ${url} does not return data`,
+      })
+    }
+
+    const data = response.response.data
+    if (data.code != '000000') {
```
dskloetc: What is `'000000'`? Is that something specific to the Ceffu API?
Let's declare a constant for it.

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -0,0 +1,88 @@
...
+      throw new AdapterError({
+        statusCode: 500,
+        message: `Ceffu wallet API ${url} failed, code: ${data.code}, message:${data.message}`,
+      })
+    }
+
+    loop = data.data.data.length != 0 && data.data.pageNo < data.data.totalPage
```
dskloetc: If we need more pages but there was no data in the current page, that should be an error, right?

packages/sources/ceffu/src/transport/wallet/wallet.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const getAssets = async (
+  walletId: string,
+  url: string,
+  proxy: string,
+  apiKey: string,
+  privateKey: string,
+  requester: Requester,
+) => {
```
dskloetc: Can you declare the return type explicitly?

packages/sources/ceffu/src/transport/wallet/wallet.ts
```diff
@@ -0,0 +1,78 @@
...
+  walletIdStr: string
+}
+
+export const getWallets = async (
+  url: string,
+  proxy: string,
+  apiKey: string,
+  privateKey: string,
+  requester: Requester,
+) => {
```
dskloetc: Can you declare the return type explicitly?

packages/sources/ceffu/src/transport/wallet/transport.ts
```diff
@@ -0,0 +1,114 @@
...
+    const results = await toUsd(
+      tokens.flatMap((t) => t),
+      Object.fromEntries(param.contracts.map(({ token, address }) => [token, address])),
+      this.provider,
+    )
+
+    const hex = results
+      .map((r) => scale(r.value, { from: r.decimal, to: param.decimals }))
+      .reduce((acc, v) => acc + v, 0n)
+      .toString(16)
```
dskloetc, line 83: Why convert the result to hex?

packages/sources/ceffu/src/transport/wallet/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+
+export const getApiKeys = (client: string) => {
+  const proxyKeyName = `WALLET_${client.toUpperCase()}_API_PROXY`
```
dskloetc: If you start with converting `client` to upper case, then you don't need to call `toUpperCase()` 3 times.

(I'm not worried about runtime, but I think it reduces the risk of forgetting to convert when changing the code later.)

packages/sources/ceffu/test/integration/fixtures.ts
```diff
@@ -15,3 +15,86 @@ export const mockResponseSuccess = (): nock.Scope =>
...
+  nock('https://open-api.ceffu.com')
+    .get('/open-api/v1/wallet/asset/list')
+    .query((q) => q.walletId === 'w1')
+    .reply(
+      200,
+      () => ({
+        data: {
+          data: [
+            { coinSymbol: 'BTC', amount: '100' },
+            { coinSymbol: 'ETH', amount: '100' },
```
dskloetc: Having different amounts for different coins makes it slightly more likely to catch a potential bug.

packages/sources/ceffu/test/integration/fixtures.ts
```diff
@@ -15,3 +15,86 @@ export const mockResponseSuccess = (): nock.Scope =>
...
+    .persist()
+    .get('/open-api/v1/wallet/asset/list')
+    .query((q) => q.walletId === 'w2')
+    .reply(
+      200,
+      () => ({
+        data: {
+          data: [
+            { coinSymbol: 'BTC', amount: '200' },
+            { coinSymbol: 'ETH', amount: '200' },
```
dskloetc: same

packages/sources/ceffu/test/unit/rate.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      JsonRpcProvider: function (): JsonRpcProvider {
+        return {} as JsonRpcProvider
+      },
+      Contract: function () {
+        return {
+          decimals: jest.fn().mockImplementation(() => {
+            return 8
+          }),
+          latestAnswer: jest.fn().mockImplementation(() => {
+            return 50n
```
dskloetc: By returning the same number regardless of the contract, you're not testing that the correct contract is being called.

packages/sources/ceffu/test/unit/rate.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      ]
+
+      const result = await toUsd(coins, mockContracts, mockProvider)
+
+      expect(result).toStrictEqual([
+        {
+          coin: 'BTC',
+          amount: '1.0',
+          rate: 50n,
+          decimal: 8,
```
dskloetc: If `decimal` is 8, shouldn't the value be` 5_000_000_000n`?
Where are the 8 decimals here?

packages/sources/ceffu/test/unit/requester.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          },
+        } as any)
+
+      const result = await request('', '', {}, '', '', '', mockRequester)
+
+      expect(result).toEqual([100, 200])
+      expect(mockRequester.request).toHaveBeenCalledTimes(2)
+    })
+
+    it('should stop pagination when response is empty', async () => {
```
dskloetc: Does this actually happen?

packages/sources/ceffu/src/endpoint/wallet.ts
mxiao-cll: Yes, updated description

packages/sources/ceffu/src/endpoint/wallet.ts
mxiao-cll: Renamed

packages/sources/ceffu/src/transport/rate.ts
mxiao-cll: Yes, we would eventually swap this with streams once they are done with their API

packages/sources/ceffu/src/transport/mirrorX.ts
mxiao-cll: The API provider bumped the max value of this, it would be better since we'll make less API requests. But we only have 2 here I believe so doesn't really matter

packages/sources/ceffu/src/transport/mirrorX.ts
mxiao-cll: Similar to how request is being used for wallet, I don't have a plan yet can create a ticket

packages/sources/ceffu/src/transport/wallet/requester.ts
mxiao-cll: The idea is for this to be shared with all ceffu APi endpoints, and I'd expect each would take different params, although there will be some shared values

packages/sources/ceffu/src/transport/wallet/requester.ts
mxiao-cll: Swap to undefined

packages/sources/ceffu/src/transport/wallet/requester.ts
mxiao-cll: I think `calculateHttpRequestKey` basically just append the EA endpoint name and transport name to the prefix of the cache

We should update the framework to have a requester.request that does that

packages/sources/ceffu/src/transport/wallet/requester.ts
mxiao-cll: Actually thinking about it twice, we should still loop even if current page is empty

packages/sources/ceffu/src/transport/wallet/wallet.ts
mxiao-cll: I'll do it but why?

packages/sources/ceffu/src/transport/wallet/transport.ts
mxiao-cll: Streams does not support string numbers in decimals only string numbers in hex

packages/sources/ceffu/test/unit/requester.test.ts
mxiao-cll: Removed since we updated code

packages/sources/ceffu/src/transport/wallet/requester.ts
dskloetc: How about having a `CeffuFooParams` type per endpoint and `type CeffuParams = ... | ... | ...` and then declare `params` here as `CeffuParams`?

packages/sources/ceffu/src/transport/mirrorX.ts
dskloetc: I don't understand. As is, I think this comment is not helpful because it's not clear what is intended.
If you think it's important we don't forget about it, yes, please create a ticket and link to the ticket from this code comment.

packages/sources/ceffu/src/transport/wallet/wallet.ts
dskloetc: Thanks. I find it helpful to understand what the function does and it wasn't obvious to me.

packages/sources/ceffu/src/transport/wallet/transport.ts
dskloetc: How about adding a comment explaining that it's for streams?

packages/sources/ceffu/src/transport/wallet/wallet.ts
mxiao-cll: I see, my IDE does it automatically for me

packages/sources/ceffu/src/transport/wallet/transport.ts
mxiao-cll: Sure

packages/sources/ceffu/src/transport/wallet/requester.ts
mxiao-cll: Yes

packages/sources/ceffu/src/transport/mirrorX.ts
mxiao-cll: Added

packages/sources/ceffu/src/transport/wallet/wallet.ts
dskloetc: You do code reviews in your IDE?

mxiao-cll: 😅

PR #4003, opened 2025-09-09
Feat/OPDATA-4101 decode response selector view-function-multi-chain

mmcallister-cll:
## Description
New endpoint for view-function-multi-chain: function-response-selector

## Changes
- Refactored function transport to add response processor param in constructor
- Added new endpoint function-response-selector that allows for a multi-arg return statement to be decoded and 1 field to be returned in the EA response as the result
- Added tests for new endpoint

## Steps to Test
1. yarn test view-function-multi-chain
2. regression payloads
3. with payload
```json
{
    "data":   {      
        "signature": "function lastPrice() external view returns (uint256 price, uint256 timestamp)",
        "address": "0xaE2364579D6cB4Bbd6695846C1D595cA9AF3574d",
        "network": "ethereum",
        "endpoint": "function-response-selector",
        "inputParams": [],
        "resultField": "price"
    }
}
```

## Quality Assurance

[OPDATA-4101]: https://smartcontract-it.atlassian.net/browse/OPDATA-4101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/loud-humans-jam.md
+++ b/.changeset/loud-humans-jam.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/token-balance-adapter': patch
+---
+
+Validate xrpl input in endpoint instead of transport
--- a/packages/sources/token-balance/src/endpoint/xrpl.ts
+++ b/packages/sources/token-balance/src/endpoint/xrpl.ts
@@ -1,7 +1,9 @@
 import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
 import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { AdapterError } from '@chainlink/external-adapter-framework/validation/error'
 import { config } from '../config'
 import { xrplTransport } from '../transport/xrpl'
+import { getXrplRpcUrl } from '../transport/xrpl-utils'
 
 export const inputParameters = new InputParameters(
   {
@@ -63,4 +65,8 @@ export const endpoint = new AdapterEndpoint({
   name: 'xrpl',
   transport: xrplTransport,
   inputParameters,
+  customInputValidation: (_request, settings): AdapterError | undefined => {
+    getXrplRpcUrl(settings)
+    return
+  },
 })
--- a/packages/sources/token-balance/src/transport/xrpl-utils.ts
+++ b/packages/sources/token-balance/src/transport/xrpl-utils.ts
@@ -1,0 +1,12 @@
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+
+export const getXrplRpcUrl = (settings: typeof config.settings) => {
+  if (!settings.XRPL_RPC_URL) {
+    throw new AdapterInputError({
+      statusCode: 400,
+      message: 'Environment variable XRPL_RPC_URL is missing',
+    })
+  }
+  return settings.XRPL_RPC_URL
+}
--- a/packages/sources/token-balance/src/transport/xrpl.ts
+++ b/packages/sources/token-balance/src/transport/xrpl.ts
@@ -10,6 +10,7 @@ import Decimal from 'decimal.js'
 import { BaseEndpointTypes, inputParameters } from '../endpoint/xrpl'
 import { getTokenPrice } from './priceFeed'
 import { SharePriceType } from './utils'
+import { getXrplRpcUrl } from './xrpl-utils'
 
 const logger = makeLogger('Token Balance - XRPL')
 
@@ -139,16 +140,11 @@ export class XrplTransport extends SubscriptionTransport<BaseEndpointTypes> {
     address: string
     tokenIssuerAddress: string
   }): Promise<Decimal> {
-    if (!this.config.XRPL_RPC_URL) {
-      throw new AdapterInputError({
-        statusCode: 400,
-        message: 'Environment variable XRPL_RPC_URL is missing',
-      })
-    }
+    const rpcUrl = getXrplRpcUrl(this.config)
 
     const requestConfig = {
       method: 'POST',
-      baseURL: this.config.XRPL_RPC_URL,
+      baseURL: rpcUrl,
       data: {
         method: 'account_lines',
         params: [
--- a/packages/sources/token-balance/test/integration/__snapshots__/xrpl-config-error.test.ts.snap
+++ b/packages/sources/token-balance/test/integration/__snapshots__/xrpl-config-error.test.ts.snap
@@ -1,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`execute xrpl endpoint should error if XRPL_RPC_URL is not set 1`] = `
+{
+  "error": {
+    "message": "Environment variable XRPL_RPC_URL is missing",
+    "name": "AdapterError",
+  },
+  "status": "errored",
+  "statusCode": 400,
+}
+`;
--- a/packages/sources/token-balance/test/integration/xrpl-config-error.test.ts
+++ b/packages/sources/token-balance/test/integration/xrpl-config-error.test.ts
@@ -1,0 +1,58 @@
+import {
+  TestAdapter,
+  setEnvVariables,
+} from '@chainlink/external-adapter-framework/util/testing-utils'
+import * as nock from 'nock'
+import { mockETHContractCallResponseSuccess, mockXrplResponseSuccess } from './fixtures'
+
+describe('execute', () => {
+  let spy: jest.SpyInstance
+  let testAdapter: TestAdapter
+  let oldEnv: NodeJS.ProcessEnv
+
+  beforeAll(async () => {
+    oldEnv = JSON.parse(JSON.stringify(process.env))
+    process.env.ETHEREUM_RPC_URL = 'http://localhost-eth-mainnet:8080'
+    process.env.ETHEREUM_RPC_CHAIN_ID = '1'
+    process.env.BACKGROUND_EXECUTE_MS = '10000'
+
+    const mockDate = new Date('2001-01-01T11:11:11.111Z')
+    spy = jest.spyOn(Date, 'now').mockReturnValue(mockDate.getTime())
+
+    const adapter = (await import('./../../src')).adapter
+    adapter.rateLimiting = undefined
+    testAdapter = await TestAdapter.startWithMockedCache(adapter, {
+      testAdapter: {} as TestAdapter<never>,
+    })
+  })
+
+  afterAll(async () => {
+    setEnvVariables(oldEnv)
+    await testAdapter.api.close()
+    nock.restore()
+    nock.cleanAll()
+    spy.mockRestore()
+  })
+
+  describe('xrpl endpoint', () => {
+    it('should error if XRPL_RPC_URL is not set', async () => {
+      const data = {
+        endpoint: 'xrpl',
+        tokenIssuerAddress: 'rJNE2NNz83GJYtWVLwMvchDWEon3huWnFn',
+        priceOracleAddress: '0xCe9a6626Eb99eaeA829D7fA613d5D0A2eaE45F40',
+        priceOracleNetwork: 'ethereum',
+        addresses: [
+          {
+            address: 'rGSA6YCGzywj2hsPA8DArSsLr1DMTBi2LH',
+          },
+        ],
+      }
+      mockETHContractCallResponseSuccess()
+      mockXrplResponseSuccess()
+
+      const response = await testAdapter.request(data)
+      expect(response.json()).toMatchSnapshot()
+      expect(response.statusCode).toBe(400)
+    })
+  })
+})


packages/sources/view-function-multi-chain/test/integration/adapter.test.ts
```diff
@@ -122,6 +122,66 @@ describe('execute', () => {
...
+        network: 'ethereum_mainnet',
+        resultIndex: 1,
+      }
+      mockETHMainnetContractCallResponseSuccess()
+      const response = await testAdapter.request(data)
+      expect(response.statusCode).toBe(200)
+      expect(response.json()).toMatchSnapshot()
+    })
+
+    it('should choose resultField over resultIndex if both specified', async () => {
```
dskloetc: Why would you want to specify both? Shouldn't this just be an error?

packages/sources/view-function-multi-chain/src/endpoint/function.ts
```diff
@@ -40,8 +56,38 @@ export type BaseEndpointTypes = {
...
+      throw new AdapterInputError({
+        statusCode: 400,
+        message: `resultField and resultIndex should not be set, use endpoint function-responseSelector instead`,
+      })
+    }
+    return
+  },
+})
+
+export const functionResponseSelectorEndpoint = new AdapterEndpoint({
```
dskloetc: I was confused because so far I've only seen 1 endpoint per file.
Are there other cases where we have multiple endpoints per file?
I think it would help avoid confusion by sticking to 1 endpoint per file.

packages/sources/view-function-multi-chain/src/endpoint/function.ts
```diff
@@ -40,8 +56,38 @@ export type BaseEndpointTypes = {
...
 
-export const endpoint = new AdapterEndpoint({
+export const functionEndpoint = new AdapterEndpoint({
   name: 'function',
   transport: multiChainFunctionTransport,
   inputParameters,
+  customInputValidation: (req): AdapterInputError | undefined => {
+    if (
+      req.requestContext.data.resultField != null ||
+      req.requestContext.data.resultIndex != null
```
dskloetc: Why add these parameters to the existing `function` endpoint if they shouldn't be used at all?

packages/sources/view-function-multi-chain/src/endpoint/function.ts
```diff
@@ -27,6 +31,18 @@ export const inputParameters = new InputParameters({
     description: 'RPC network name',
     type: 'string',
   },
+  resultField: {
```
dskloetc: `resultField` and `resultIndex` are basically the same thing.
Would it make sense to just have 1 parameter and allow it to be both a string or a number?
It's even the same syntax to get the value out of a `Record` or an `Array`.

Then you can make the parameter required and don't have issues with only 1 of them being valid/used.

packages/sources/view-function-multi-chain/src/transport/function.ts
```diff
@@ -35,7 +63,8 @@ export class MultiChainFunctionTransport extends SubscriptionTransport<MultiChai
   async handleRequest(param: RequestParams) {
     let response: AdapterResponse<MultiChainFunctionTransportTypes['Response']>
     try {
-      response = await this._handleRequest(param)
+      const onchainResponse = await this.getContractResponseHex(param)
```
dskloetc: Having `handleRequest` call `_handleRequest` and having the actual logic in `_handleRequest` is standard boilerplate we have in a lot of EAs.
Can we keep that and just put the new logic in `_handleRequest` to avoid confusion?

packages/sources/view-function-multi-chain/src/transport/function.ts
```diff
@@ -107,4 +128,60 @@ export class MultiChainFunctionTransport extends SubscriptionTransport<MultiChai
   }
 }
 
-export const multiChainFunctionTransport = new MultiChainFunctionTransport()
+/* eslint-disable @typescript-eslint/no-unused-vars */
+function createHexResultPostProcessor(
```
dskloetc: I would think that a "processor" is a function.
So for something called "create...Processor" I would expect it to return a function.
But looking at this, it just seems to be the processor itself, rather than a function that creates a processor.
Am I interpreting this wrong?

packages/sources/view-function-multi-chain/src/transport/function.ts
```diff
@@ -107,4 +128,60 @@ export class MultiChainFunctionTransport extends SubscriptionTransport<MultiChai
...
+    statusCode: 200,
+    result,
+    timestamps,
+  }
+}
+
+export const multiChainFunctionTransport = new MultiChainFunctionTransport(
+  createHexResultPostProcessor,
+)
+export const multiChainFunctionResponseSelectorTransport = new MultiChainFunctionTransport(
```
dskloetc: How about putting the share transport class in a separate file and then having a file per transport?
I think that would make it easier to navigate the EA and harder to miss that there are actually 2 transports here.

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: to avoid review comments about duplicate code :P
... but there's probably a better way to do this, i'll try some ideas

packages/sources/view-function-multi-chain/src/transport/function.ts
mmcallister-cll: sure!

packages/sources/view-function-multi-chain/src/transport/function.ts
mmcallister-cll: you are correct. will remove the `create`s from the processor names

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: sure, i stuck it in the same cause the transport file is the same but i'll separate them and the transports (below comment) for clarity

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: can we even make a single parameter an OR type? Will look into it, but either way if we get rid of these from the existing function endpoint it's not too bad.
Or we can just force the use of a single param (eg: force named param)

packages/sources/view-function-multi-chain/test/integration/adapter.test.ts
mmcallister-cll: yeah, good point. I will update the input validation to remove this possibility.

mmcallister-cll: Actually on further thought, I'm going to force the use of the `resultField` input param and completely remove `resultIndex` for simplicity (as mentioned as a possibility in https://github.com/smartcontractkit/external-adapters-js/pull/4003#discussion_r2334182261). As a byproduct this test and the individual test for resultIndex will be removed

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: As I'm going through these changes, I remember why I tried to keep everything the same and add customInputValidators now: it was just an attempt to avoid generics for the entire transport, which i find are much messier than having an optional input param.

packages/sources/view-function-multi-chain/src/endpoint/function-response-selector.ts
```diff
@@ -0,0 +1,32 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import { multiChainFunctionResponseSelectorTransport } from '../transport/function-response-selector'
+import { inputParamDefinition as functionInputParamDefinition } from './function'
+
+const inputParameters = new InputParameters({
+  ...functionInputParamDefinition,
+  resultField: {
+    required: false,
```
dskloetc: Why is this not required?

packages/sources/view-function-multi-chain/src/endpoint/function.ts
dskloetc: What do you think about keeping just a single endpoint and adding an optional parameter to it?
If the parameter isn't there, we keep the old behavior. If the parameter is there, we have the new behavior.

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
```diff
@@ -0,0 +1,38 @@
+import { AdapterResponse } from '@chainlink/external-adapter-framework/util'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { BaseEndpointTypes } from '../endpoint/function-response-selector'
+import { createMultiChainFunctionTransport, RawOnchainResponse } from './function-common'
+
+function decodedResultSelectorPostProcessor(
```
dskloetc: It's just a function. I like function names that start with a verb.
What do you think about `selectFieldFromDecodedResult`?

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
```diff
@@ -0,0 +1,38 @@
...
+  const result = BigInt(decodedResult[resultField]).toString()
+
+  return {
+    data: {
+      result,
+    },
+    statusCode: 200,
+    result,
+    timestamps,
+  }
```
dskloetc: Let's keep the creation of this structure in the shared code and only return the `result` from the post processor.

packages/sources/view-function-multi-chain/src/transport/function.ts
```diff
@@ -1,110 +1,24 @@
...
-  getSubscriptionTtlFromConfig(
-    adapterSettings: MultiChainFunctionTransportTypes['Settings'],
-  ): number {
-    return adapterSettings.WARMUP_SUBSCRIPTION_TTL
+import { AdapterResponse } from '@chainlink/external-adapter-framework/util'
+import { BaseEndpointTypes } from '../endpoint/function'
+import { createMultiChainFunctionTransport, RawOnchainResponse } from './function-common'
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export function hexResultPostProcessor(
```
dskloetc: If you move creating the response structure to the shared code, then this processor does nothing.
It could just be an inline function `(result) => result`, or the processor parameter could be optional and default to `(x) => x` if not specified.

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: I initially implemented like your suggestion but I felt like it would be too confusing downstream to have the endpoint return a hexEncoded string (that needs to be decoded downstream) and also a number string depending on the presence of the resultField (or resultIndex) param so separated it

mmcallister-cll: discussed offline, second endpoint makes sense due to hex vs number response expectation

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
mmcallister-cll: yeah, i tried returning just a string from the post processor but it was complaining about the different Result fields due to the BaseEndpointTypes, will try again another way, but I don't think it warrants a lot of time/effort.

packages/sources/view-function-multi-chain/src/transport/function.ts
mmcallister-cll: yup!

packages/sources/view-function-multi-chain/src/endpoint/function-response-selector.ts
mmcallister-cll: Good find TY, this is an error

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
mmcallister-cll: <3

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: Resolution was forcing named param. Open to other solutions but we'll go with this barring any complaints.

packages/sources/view-function-multi-chain/src/endpoint/function.ts
mmcallister-cll: moved specific endpoint & transport code to separate files & shared to common transport

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
```diff
@@ -0,0 +1,28 @@
...
+import { BaseEndpointTypes } from '../endpoint/function-response-selector'
+import { createMultiChainFunctionTransport, RawOnchainResponse } from './function-common'
+
+function selectFieldFromDecodedResult(
+  onchainResponse: RawOnchainResponse,
+  resultField?: string | undefined,
+): string {
+  if (!resultField) {
+    throw new AdapterInputError({
+      message: 'Missing resultField input param',
```
dskloetc, line 11: This is not possible now that the param is required, right?

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
mmcallister-cll: fixed

packages/sources/view-function-multi-chain/src/transport/function-response-selector.ts
mmcallister-cll: correct, I suppose the function could be used elsewhere at some point (but in practice never will be)

PR #4006, opened 2025-09-10
Re-factor ceffu

mxiao-cll:
## Description

......

## Changes

- High level
- changes that
- you made

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance


.changeset/funny-lemons-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/ceffu-adapter': patch
+---
+
+Code refactor
```
dskloetc, line 5: Changing retry behavior and exposing extra data in the response are not "refactoring".

packages/sources/ceffu/src/transport/mirrorX.ts
```diff
@@ -1,28 +1,14 @@
...
 import { AdapterError } from '@chainlink/external-adapter-framework/validation/error'
 import { Decimal } from 'decimal.js'
-import { HttpsProxyAgent } from 'https-proxy-agent'
-import { stringify } from 'querystring'
-// TODO: Refactor to use request instead
-// https://smartcontract-it.atlassian.net/browse/OPDATA-4224
-import { sign } from './wallet/requester'
+import { request } from './requester'
 
 interface ApiResponse {
```
dskloetc: Is `ApiResponse` still the right name? We don't control the API so refactoring the code can't really change the type of the response we get from the API.

packages/sources/ceffu/src/transport/mirrorX.ts
```diff
@@ -35,50 +21,33 @@ export const getAssetPositions = async (
...
+        privateKey,
+        requester,
+        proxy,
+      )
     }),
   )
 
   for (const response of responses) {
-    if (!response.response || !response.response.data) {
+    if (!response || !response.extra || response.extra.length == 0) {
```
dskloetc: Prefer `===` over `==`. For example `false == 0` is true which can be confusing.
Or optionally shorten this to
```
if (!response?.extra?.length) {
```

packages/sources/ceffu/src/transport/mirrorX.ts
```diff
@@ -35,50 +21,33 @@ export const getAssetPositions = async (
...
-
-    if (!response.response.data.data?.exchangeBalance) {
-      throw new AdapterError({
-        statusCode: 500,
-        message: `Ceffu API does not return exchangeBalance, code: ${response.response.data.code}, message:${response.response.data.message}`,
-      })
-    }
   }
 
+  const exchangeBalances = responses.flatMap(({ extra }) =>
```
dskloetc, line 45: What is "extra". Why do we need to access `extra` here while previously we could just use `response.data.data.exchangeBalance`?

packages/sources/ceffu/src/transport/requester.ts
dskloetc, line 1: Moving and changing the file in the same commit makes it hard to review.
Can you keep the file in the same place for now, or separate out the changes to this file?

packages/sources/ceffu/src/transport/wallet/wallet.ts
```diff
@@ -26,14 +26,14 @@ export const getWallets = async (
     proxy,
   )
 
-  if (!response || response.length == 0) {
+  if (!response || !response.data || response.data.length == 0) {
```
dskloetc: Use `===` or replace `response.data.length == 0` with `!response.data.length` depending on intent.

.changeset/funny-lemons-invent.md
mxiao-cll: Removed logic change

packages/sources/ceffu/src/transport/mirrorX.ts
mxiao-cll: renamed

packages/sources/ceffu/src/transport/mirrorX.ts
mxiao-cll: Because the exchangeBalance is now packed into the extra array, we can not fit this into the data array due to exchangeBalance sit outside of the data object in response.

packages/sources/ceffu/src/transport/requester.ts
mxiao-cll: Moved back

packages/sources/ceffu/test/unit/requester.test.ts
```diff
@@ -67,7 +67,7 @@ describe('requester.ts', () => {
 
       const result = await request('', '', {}, '', '', mockRequester)
 
-      expect(result).toEqual([100, 200])
+      expect(result).toEqual({ data: [100, 200], extra: [] })
```
dskloetc, line 70: Can you add a unit test with non-empty `extra`?

packages/sources/ceffu/src/transport/rate.ts
```diff
@@ -28,6 +28,7 @@ export const toUsd = async (
         rate: rate.value,
         decimal: rate.decimal,
         value,
+        data: c.data,
```
dskloetc: How is this extra field used?
Is it related to the refactoring or new functionality?

packages/sources/ceffu/src/transport/wallet/requester.ts
```diff
@@ -19,19 +24,29 @@ type EmptyParams = Record<string, never>
...
 }
+type MirrorXParams = {
+  mirrorXLinkId: string
+  excludeZeroAmountFlag: boolean
+}
 
 const successCode = '000000'
 
-export const request = async <T>(
+export const request = async <T, E = EmptyParams>(
```
dskloetc: Can you add some documenting comments explaining that `data` will have 1 array with all items across pages but that `extra` will only have 1 item per page?

packages/sources/ceffu/src/transport/rate.ts
dskloetc: If the goal is just to augment the `coins` array, would it make sense to do
```
      return {
        ...c,
        coin: c.coin.toUpperCase(),
        rate: rate.value,
        decimal: rate.decimal,
        value,
      }
```
instead? (add `...c,` and remove `amount: c.amount` and `data: c.data`)

mxiao-cll: It is currently not being used, we may use it later to funnel extra data through

PR #4009, opened 2025-09-11
Adds DBAG EA

mohamed-mehany:
## Changes

- Adds `deutsche-boerse` EA

## Steps to Test

1. Unit tests
2. Integration tests
3. Running EA locally

Note: 

To run or test locally use:

```
yarn build && API_KEY=secret WS_API_ENDPOINT=wss://sth.com yarn start
```

Request example:
```
POST http://127.0.0.1:8080/

{

  "data": {
    "isin": "IE00B53L3W79",
    "market": "md-xetraetfetp"
  }
}

```
## Quality Assurance

[OPDATA-2395]: https://smartcontract-it.atlassian.net/browse/OPDATA-2395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/deutsche-boerse/src/endpoint/lwba.ts
```diff
@@ -0,0 +1,53 @@
...
+  Result: number | null
+  Data: {
+    mid: number | null
+    bid: number | null
+    ask: number | null
+    latestPrice: number | null
+    quoteProviderIndicatedTimeUnixMs: number | null
+    tradeProviderIndicatedTimeUnixMs: number | null
+  }
+}
```
mxiao-cll, line 43: How come everything can be null in response?

packages/sources/deutsche-boerse/src/transport/lwba.ts
```diff
@@ -0,0 +1,209 @@
...
+          return []
+        }
+
+        // Enforce exactly ONE payload
+        const msgs = sm.messages ?? []
+        if (msgs.length !== 1) {
+          logger.warn({ count: msgs.length }, 'Expected exactly one message in StreamMessage')
+          return []
+        }
+        const market = 'md-xetraetfetp' // currently only one supported market/stream
```
mxiao-cll: Read this from param, and use options on param to enforce value

packages/sources/deutsche-boerse/src/transport/lwba.ts
```diff
@@ -0,0 +1,209 @@
...
+            params: { isin, market },
+            response: {
+              result: null,
+              data: {
+                mid: quote.mid ?? null,
+                bid: quote.bid ?? null,
+                ask: quote.ask ?? null,
+                latestPrice: quote.latestPrice ?? null,
+                quoteProviderIndicatedTimeUnixMs: quote?.quoteProviderTimeUnixMs ?? null,
+                tradeProviderIndicatedTimeUnixMs: quote?.tradeProviderTimeUnixMs ?? null,
```
mxiao-cll: Why don't we throw if these are null

packages/sources/deutsche-boerse/src/transport/lwba.ts
```diff
@@ -0,0 +1,209 @@
...
+      error: (errorEvent, _context) => {
+        logger.error({ errorEvent }, 'LWBA websocket error')
+      },
+      close: (closeEvent, _context) => {
+        const code = (closeEvent as any)?.code
+        const reason = (closeEvent as any)?.reason
+        const wasClean = (closeEvent as any)?.wasClean
+        logger.info({ code, reason, wasClean }, 'LWBA websocket closed')
+      },
+      message(buf) {
```
mxiao-cll, line 52: Can break this down a bit into several functions

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
```diff
@@ -0,0 +1,56 @@
+import Decimal from 'decimal.js'
+import type {
+  Data as DataProto,
+  Decimal as DecimalProto,
+  MarketData as MarketDataProto,
+} from '../gen/md_cef_pb'
+const MAX_SAFE = BigInt(Number.MAX_SAFE_INTEGER)
+
+export function decimalToNumber(decimal?: DecimalProto): number {
```
mxiao-cll, line 9: Why don't we use bigint here

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  return (typeof sym === 'string' && sym) || undefined
+}
+
+export function pickProviderTime(dat: DataProto): number {
+  return convertNsToMs(dat?.Tm)
+}
+
+export function isDecimalPrice(x?: DecimalProto): boolean {
+  return (
+    !!x && (typeof x.m === 'bigint' || typeof (x as any).m === 'number') && typeof x.e === 'number'
```
mxiao-cll: why we need x as any only for one

packages/sources/deutsche-boerse/src/transport/protobuf-wstransport.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  async sendMessages(
+    _context: EndpointContext<T>,
+    subscribes: unknown[],
+    unsubscribes: unknown[],
+  ): Promise<void> {
+    const messages = [...subscribes, ...unsubscribes]
+      .map((m) => this.toRawData(m))
+      .filter((m): m is Buffer => m !== null) // Filter out null values
+    for (const m of messages) this.wsConnection?.send(m)
```
mxiao-cll: use forEach

packages/sources/deutsche-boerse/src/transport/protobuf-wstransport.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  deserializeMessage(data: WebSocket.Data): T['Provider']['WsMessage'] {
+    if (Array.isArray(data) && data.every(Buffer.isBuffer)) {
+      return Buffer.concat(data as Buffer[]) as unknown as T['Provider']['WsMessage']
+    }
+    if (Buffer.isBuffer(data)) return data as unknown as T['Provider']['WsMessage']
+    if (data instanceof ArrayBuffer)
```
mxiao-cll: Use {}

Actually use {} everywhere

packages/sources/deutsche-boerse/src/transport/lwba.ts
```diff
@@ -0,0 +1,209 @@
...
+        if (!isin) {
+          logger.warn({ md }, 'Could not parse ISIN from MarketData.Instrmt.Sym')
+          return []
+        }
+
+        if (!cache.has(isin)) {
+          logger.debug({ isin }, 'Ignoring message for inactive instrument (not in cache)')
+          return []
+        }
+        const quote = cache.get(isin)! // exists due to has() guard above
```
mxiao-cll: Drop .has use .get and if (quote != null)

.yarn/cache/decimal.js-npm-10.6.0-a72c1b8a2f-c0d45842d4.zip
mmcallister-cll, line 1: Can we use 10.5.0 similar to token-balance to get rid of the extra package?

packages/sources/deutsche-boerse/src/endpoint/lwba.ts
mohamed-mehany: Because either bid/mid/ask will be set or the latestPrice would be set, depending on the order of messages.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Params aren't accessible here AFAIK.
I had it like this initially:
`const market = sm.subs` however due to the options field, it fails validation.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Because either bid/mid/ask will be set or the latestPrice would be set, depending on the order of messages.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mxiao-cll: If that is the case remove the param

packages/sources/deutsche-boerse/src/transport/lwba.ts
mxiao-cll: Do we have checks that throw if neither is set?

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: What do you mean remove the param? you mean not have it part of the request?
The reason I left it is that it will be later extended to have more markets and will be simpler to extend later.

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mohamed-mehany: Because the output can be a floating point / fractional number since the exponent can be negative.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mxiao-cll: Yes, remove as part of the request, if that is the case we should use param here

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mxiao-cll: You can represent floating using bigint, use a bigint and a number for decimals

mohamed-mehany: Ah but I would still need to convert it for the output. 
I checked with the streams team and they mentioned they prefer a decimal floating point number as opposed to having a mantissa and exponent or number and decimal point number.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Not sure if there's a construct that we can use in the framework but I can add a custom check at the end before returning to check and throw if neither is set.

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mxiao-cll: Then how do you know line 21 is not going to overflow

packages/sources/deutsche-boerse/src/transport/lwba.ts
mxiao-cll: I think a check would be good, how is the output being used though?

Does stream try to read one field if doesn't exist fallback to another feed?

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mohamed-mehany: It might overflow to Infinity when |value| > Number.MAX_VALUE (~1.7976931348623157e+308) or |value| < Number.MIN_VALUE (~5e-324) with loss in precision beyond 15-17 digits. 
Note that we're using decimal.js for everything up until the conversion. If we want to preserve precision further than that, the schema of the output would have to use strings which I'm not sure how compatible it is with streams, so I thought the range with that precision should be sufficient.

packages/sources/deutsche-boerse/src/transport/protobuf-wstransport.ts
mohamed-mehany: will do.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: This was a request from product to add both fields not sure about the exact implementation from the stream side though.

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mohamed-mehany: I think bigint should be enough, will remove the number condition.

packages/sources/deutsche-boerse/src/transport/protobuf-wstransport.ts
mohamed-mehany: sure.

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mxiao-cll: Stream does not work with decimal strings only hex strings.

That's why I was saying use bigint for all your internal calculations and do a conversation at the end

packages/sources/deutsche-boerse/src/transport/lwba.ts
mxiao-cll: We need to figure that out, it is better to enforce something on our side

.yarn/cache/decimal.js-npm-10.6.0-a72c1b8a2f-c0d45842d4.zip
mohamed-mehany: Sounds good.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Fixed

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Sounds good, added the validation.

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Added

packages/sources/deutsche-boerse/src/transport/lwba.ts
mohamed-mehany: Done

packages/sources/deutsche-boerse/src/transport/proto-utils.ts
mohamed-mehany: Validated with Product that 15 significant digits is sufficient so updated the logic to only handle up to that limit. Thanks for the flag.

packages/sources/deutsche-boerse/src/transport/instrument-quote-cache.ts
```diff
@@ -0,0 +1,49 @@
+import Decimal from 'decimal.js'
+
+export type Quote = {
+  bid?: number
+  ask?: number
+  mid?: number
+  latestPrice?: number
+  quoteProviderTimeUnixMs?: number
+  tradeProviderTimeUnixMs?: number
```
mxiao-cll, line 9: How can these be null?

mxiao-cll (approved):

packages/sources/deutsche-boerse/src/transport/instrument-quote-cache.ts
mohamed-mehany: If either the bid/ask/mid or latestPrice is set before the other.

PR #4020, opened 2025-09-15
Add solana-balance endpoint to token-balance EA

dskloetc:

## Description

We need to be able to get Solana balances for a PoR feed.

There is an existing `solana` endpoint. This can query balances of tokens on the Solana network.
But we need balances of the native Solana token on the same network so the new `solana-balance` endpoint queries just the native token.
This uses a different method on the the same `SOLANA_RPC_URL` endpoint.

Note: I copied some files for `xrp` in a separate commit to make it easier to see the difference between the two endpoints.

## Changes

Add `solana-balance` endpoint which given a list of addresses returns the balance for each individual address.

## Steps to Test

1. Unit/integration tests added.
2. Tested manually together with `proof-of-reserves` in another branch with a supporting change to `proof-of-reserves`.
3.
```
  "data": {
  "endpoint": "solana-balance",
  "addresses": [
    {
      "address": "7d73NFxuWQ2F248NA4XwxE95oFfbWZrc1sg4wcDJjzTq"
    }
  ]
}
}'
{
  "data": {
    "result": [
      {
        "address": "7d73NFxuWQ2F248NA4XwxE95oFfbWZrc1sg4wcDJjzTq",
        "balance": "8630400"
      }
    ],
    "decimals": 9
  },
  "statusCode": 200,
  "result": null,
  "timestamps": {
    "providerDataRequestedUnixMs": 1757956600397,
    "providerDataReceivedUnixMs": 1757956602611
  },
  "meta": {
    "adapterName": "TOKEN_BALANCE",
    "metrics": {
      "feedId": "{\"addresses\":[{\"address\":\"7d73nfxuwq2f248na4xwxe95offbwzrc1sg4wcdjjztq\"}]}"
    }
  }
}
```

## Quality Assurance


packages/sources/token-balance/src/transport/solana-balance.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+      },
+    )
+    return await Promise.all(addresses.map(getBalance))
+  }
+
+  async getTokenBalance(address: string): Promise<number> {
+    // getSolanaRpcUrl throws if the RPC URL is missing, in which case
+    // this.connection would not be initialized.
+    getSolanaRpcUrl(this.config)
```
mxiao-cll: This is confusing, can simply check if connection is null or not

dskloetc: Then I need an `if` and a `throw` while all of that is already there in the `getSolanaRpcUrl`.
Duplicating 6 lines instead of calling 1 function.
But if you really think it's confusing, maybe it's worth it to duplicate?
What do you think?

mxiao-cll: Can leave as is, or extract everything into a getConnection helper

Call it in endpoint to reject bad request early
No need to call in init
Here if connection is null call it as set it

dskloetc: I'm not sure about creating the connection every time just to validate.
But I've created a `getConnection()` on the transport which creates it if necessary.
What do you think?

PR #4031, opened 2025-09-17
Llama Guard EA

mxiao-cll:
## Description

New EA, it reads rawNav value from source EA and then bounds from on-chain contract, perform some calculations to determine what is the final nav value

## Quality Assurance


packages/composites/llama-guard/src/endpoint/nav.ts
```diff
@@ -0,0 +1,89 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+import { getEAUrl } from '../transport/ea'
+import { navTransport } from '../transport/transport'
+
+export const inputParameters = new InputParameters(
+  {
+    ea: {
```
dskloetc: "ea" is really short and jargonny.
I thought externally we call them "adapters".
In `implied-price` we call is "source".

packages/composites/llama-guard/src/transport/contract.ts
```diff
@@ -0,0 +1,73 @@
...
+      decimals: Number(decimals),
+    }
+  }
+
+  const [
+    { answer: lookbackNav, updatedAt: lookbackTime },
+    { answer: latestNav, updatedAt: latestTime },
+    decimals,
+  ] = await Promise.all([
+    registryContract.getLookbackData(asset),
```
dskloetc: Why not include this call on line 21 above? It looks like it doesn't depend on anything that isn't already available there.

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -0,0 +1,86 @@
...
+      adjustedNav: rawNavScaled.toString(),
+      riskFlag: false,
+      breachDirection: '',
+      isBounded: true,
+      ...results,
+    }
+  }
+}
+
+const mulBigInt = (value: bigint, multiplier: number, decimals: number) => {
```
dskloetc, line 91: What is this trying to accomplish?

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -0,0 +1,86 @@
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import { JsonRpcProvider, parseUnits } from 'ethers'
+import { getBounds } from './contract'
+import { getRawNav } from './ea'
+
+const SECONDS_IN_DAY = 24 * 60 * 60
+const BPS = 10000
+
+export const getNav = async (
```
dskloetc, line 12: There is a lot of logic and arithmetic here and it's not obvious why it's doing what it's doing.
Would it be possible to add some documenting comments to help understand what this is trying to do?
Because it looks quite random which makes it really hard to review if it's correct.

packages/composites/llama-guard/src/transport/contract.ts
```diff
@@ -0,0 +1,73 @@
+import { Contract, JsonRpcProvider } from 'ethers'
+import ABI from '../config/ABI.json'
+import EACAggregatorProxy from '../config/EACAggregatorProxy.json'
+
+export const getBounds = async (asset: string, registry: string, provider: JsonRpcProvider) => {
```
dskloetc: Let's use named parameters so the strings can't accidentally be switched.

packages/composites/llama-guard/src/endpoint/nav.ts
mxiao-cll: Renamed

packages/composites/llama-guard/src/transport/contract.ts
mxiao-cll: We don't need to make this call if `isUpperBoundEnabled` and `isLowerBoundEnabled` is false

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: Multiply a bigint with a float number into a bigint

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: Added some comments, there is also a doc provided by DP but not sure if appropriate to include

dskloetc: I meant documentation on the function itself explaining the purpose of the function.

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -31,6 +32,8 @@ export const getNav = async (
     const finalDiscount = scaledMaxDiscount * (1 - bounds.lower.lowerBoundTolerance / BPS)
     lowerBound = mulBigInt(bounds.lower.latestNav, finalDiscount, bounds.decimals)
```
dskloetc, line 46: How about checking the lower bound here and returning early if it's violated?
Then we don't need the special `-1` value of `lowerBound` and we don't need 2 separate if blocks for the lower bound logic.

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -39,6 +42,7 @@ export const getNav = async (
     upperBound = mulBigInt(bounds.upper.lookbackNav, finalBuffer, bounds.decimals)
```
dskloetc, line 65: and then check the upper bound here.

packages/composites/llama-guard/src/transport/nav.ts
dskloetc, line 46: Do I understand the intent correctly as follows?

* We allow the nav to go down by `maxDiscount` basis points per day since the previous data point.
* We always allow at least 1 day worth of movement even if the previous data point was less than a dag ago.
* Additionally to a per-day movement, we allow a fixed movement of `lowerBoundTolerance` basis points.

If so, something like this would be useful as documentation on the function itself.

And if I got that right, can we add `PerDay` to the `maxDiscount` variable name?

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -61,6 +66,7 @@ export const getNav = async (
       ...results,
     }
   } else if (upperBound > 0 && rawNavScaled > upperBound) {
+    // Breach Upper Bound
     return {
       adjustedNav: upperBound.toString(),
       riskFlag: false,
```
dskloetc: Should this not be `true` since the NAV is out of bounds?

packages/composites/llama-guard/src/transport/contract.ts
```diff
@@ -0,0 +1,76 @@
...
+    lower: {
+      isLowerBoundEnabled: Boolean(isLowerBoundEnabled),
+      latestNav: BigInt(latestNav),
+      latestTime: Number(latestTime),
+      maxDiscount: Number(maxDiscount),
+      lowerBoundTolerance: Number(lowerBoundTolerance),
+    },
+    upper: {
+      isUpperBoundEnabled: Boolean(isUpperBoundEnabled),
+      lookbackNav: BigInt(lookbackNav),
```
dskloetc, line 69: Why does the upper bound use a different "previous price" than the lower bound?

packages/composites/llama-guard/test/unit/nav.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        defaultParams.ea,
+        defaultParams.eaInput,
+        defaultParams.requester,
+        defaultParams.asset,
+        defaultParams.registry,
+        defaultParams.provider,
+      )
+
+      expect(result).toEqual({
+        rawNav: '100',
```
dskloetc: Why 100? Because 1 and 2 decimals?
Can you make it more obvious with variables and different numbers?
Maybe
```
const decimals = 6
const rawNavFromSource = 123
const expectedRawNav = '123000000'
```
or something

packages/composites/llama-guard/test/unit/nav.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        bases: {
+          lookback: { nav: '0', ts: 0 },
+          previous: { nav: '150', ts: now / 1000 - 86400 / 2 },
+        },
+        decimals: 2,
+        riskFlag: true,
+        breachDirection: 'lower',
+        isBounded: false,
+      })
+    })
```
dskloetc, line 128: Can you add separate tests to make it clear that the per-day offset is working and that the fixed offset is working as intended?

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: Updated

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: Is is specified by DP, we have a isBounded flag to indicate if out of bounds or not

packages/composites/llama-guard/src/transport/contract.ts
mxiao-cll: Specified by DP

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: I wanted to be able to report both bands even if one of them is violated but I guess that's fine.

mxiao-cll: You'd still need the -1 in order to return bounds for with-in bounds case though

packages/composites/llama-guard/test/unit/nav.test.ts
mxiao-cll: Yes because decimals

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: I am trying to see if it is ok to link the doc here, 

I don't want to modify `maxDiscount` because it is the name on-chain as well

packages/composites/llama-guard/test/unit/nav.test.ts
mxiao-cll: I do set a different maxDiscount and lowerBoundTolerance here, maybe add a test case with maxDiscount being 0?

PR #4041, opened 2025-09-22
change the EA to use custom transport

karen-stepanyan:
## Description

......

## Changes

- High level
- changes that
- you made

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance

### DIFF
--- a/packages/sources/nomia/src/config/index.ts
+++ b/packages/sources/nomia/src/config/index.ts
@@ -14,11 +14,16 @@ export const config = new AdapterConfig(
       required: true,
       sensitive: true,
     },
+    API_FREQUENCY: {
+      description: 'Frequency of API calls in minutes',
+      type: 'number',
+      default: 20,
+      required: false,
+    },
   },
   {
     envDefaultOverrides: {
       CACHE_MAX_AGE: 20 * 60 * 1000, // 20 minutes - max validated setting
-      BACKGROUND_EXECUTE_TIMEOUT: 180_000,
     },
   },
 )
--- a/packages/sources/nomia/src/transport/price.ts
+++ b/packages/sources/nomia/src/transport/price.ts
@@ -1,6 +1,11 @@
-import { HttpTransport } from '@chainlink/external-adapter-framework/transports'
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
+import { calculateHttpRequestKey } from '@chainlink/external-adapter-framework/cache'
+import { TransportDependencies } from '@chainlink/external-adapter-framework/transports'
+import { SubscriptionTransport } from '@chainlink/external-adapter-framework/transports/abstract/subscription'
 import { makeLogger } from '@chainlink/external-adapter-framework/util'
-import { BaseEndpointTypes } from '../endpoint/price'
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import schedule from 'node-schedule'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/price'
 
 const logger = makeLogger('nomia')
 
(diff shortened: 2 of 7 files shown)


packages/sources/nomia/src/config/index.ts
```diff
@@ -14,11 +14,16 @@ export const config = new AdapterConfig(
       required: true,
       sensitive: true,
     },
+    API_FREQUENCY: {
+      description: 'Frequency of API calls in minutes',
+      type: 'number',
+      default: 20,
```
cawthorne, line 20: Can we make the default 2, as we know that works

packages/sources/nomia/src/config/index.ts
```diff
@@ -14,11 +14,16 @@ export const config = new AdapterConfig(
       required: true,
       sensitive: true,
     },
+    API_FREQUENCY: {
+      description: 'Frequency of API calls in minutes',
```
cawthorne, line 18: should this be per minute?

packages/sources/nomia/src/transport/price.ts
```diff
@@ -13,68 +18,122 @@ export interface ResponseSchema {
...
+
+  async backgroundHandler(_: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
+    this.params = entries
+  }
+
+  runScheduler() {
+    schedule.scheduleJob(`0 */${this.settings.API_FREQUENCY} * * * *`, async () => {
+      logger.info(`Scheduled execution started at ${Date.now()}`)
+      for (let i = 0; i < this.params.length; i++) {
+        await this.executeRequest(this.params[i])
```
cawthorne, line 60: Does this call the subscriptions one after the other? Does it wait 60/API_FREQUENCY seconds?

packages/sources/nomia/src/transport/price.ts
```diff
@@ -13,68 +18,122 @@ export interface ResponseSchema {
...
-        request: {
-          baseURL: `${config.API_ENDPOINT}?${decodedQuery}`,
-        },
+
+  async backgroundHandler(_: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
+    this.params = entries
+  }
+
+  runScheduler() {
+    schedule.scheduleJob(`0 */${this.settings.API_FREQUENCY} * * * *`, async () => {
```
cawthorne, line 57: How often does this make it run the cron? The superstate EA uses a `const rule = new schedule.RecurrenceRule()` I would be in favour of doing the same.

packages/sources/nomia/src/transport/price.ts
karen-stepanyan: yes, so this is make a burst of requests (once the previous one is fetched)

packages/sources/nomia/src/config/index.ts
karen-stepanyan: since there is no background process limiting here, why not 20 minutes for 6 requests ? essentially 3.3 minute per request

packages/sources/nomia/src/config/index.ts
karen-stepanyan: this is for CRON, can be minutes or seconds I don't have preference

packages/sources/nomia/src/transport/price.ts
karen-stepanyan: this runs every this.settings.API_FREQUENCY minutes. superstate uses slightly different logic (runs at a specific time per day) so that's why it's different rule.

cawthorne: Karen explained, standard cron syntax

PR #4045, opened 2025-09-22
OPDATA-3674: Add optional address filter parameter to virtune endpoints of por-address-list

dskloetc:

## Description

The address list of Virtune's Avalanche reserves has both P-chain addresses and normal addresses.
The balances of these types of addresses are retrieved with different adapters.
So we need a way to get a list of just one type of address for `proof-of-reserves` to pass to the appropriate indexer.

## Changes

1. Add `addpressPattern` parameter to shared parameters of `virtune` and `virtune-token` endpoints`.
2. Add `filterAddresses` function which keeps the addresses matching pattern, or all addresses if no pattern is specified.
3. Pass the `filterAddresses` function to `createVirtuneTransportConfig`.

## Steps to Test

1. Unit tests added
2.
```
curl --silent -S -X POST http://localhost:8081 -H 'Content-Type: application/json' -d '{
  "data": {
  "endpoint": "virtune",
  "accountId": "VIRAVAX",
  "network": "avalanche",
  "chainId": "43114",
  "addressPattern": "^(?!P-)"
}
}'
{
  "result": null,
  "data": {
    "result": [
      {
        "address": "0xdF8e635E05dAfa055e219B34B37F3bD793C165Cd",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0xF7b2a630d5345d7a86889848ee92FDa4451c6BD4",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0x5fD5Df3e794f0140fE2930440c39098B158A000b",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0xbDa1ABDA39d1b672620688fA46B0B58449F7b4F9",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0x76D203FA71D0DF5cf7Df75b124A1cEeE105a374E",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0x0e43d3D4bFE82Cf706BAbb3fd5e077FE698Bdb0b",
        "network": "avalanche",
        "chainId": "43114"
      },
      {
        "address": "0x94bA68963620B53DEcf06e652eEbb8Bc65137c46",
        "network": "avalanche",
        "chainId": "43114"
      }
    ]
  },
  "timestamps": {
    "providerDataRequestedUnixMs": 1758550117246,
    "providerDataReceivedUnixMs": 1758550117987
  },
  "statusCode": 200,
  "meta": {
    "adapterName": "POR_ADDRESS_LIST",
    "metrics": {
      "feedId": "{\"accountId\":\"viravax\",\"network\":\"avalanche\",\"chainId\":\"43114\",\"addressPattern\":\"^(?!p-)\"}"
    }
  }
}
```

## Quality Assurance

### DIFF
--- a/packages/sources/por-address-list/src/transport/virtune-utils.ts
+++ b/packages/sources/por-address-list/src/transport/virtune-utils.ts
@@ -7,6 +7,19 @@ export const getUrl = (params: { accountId: string }): string => {
   return params.accountId
 }
 
+export const filterAddresses = ({
+  addresses,
+  params,
+}: {
+  addresses: string[]
+  params: { addressPattern?: string }
+}): string[] => {
+  const { addressPattern } = params
+  if (!addressPattern) return addresses
+  const re = new RegExp(addressPattern)
+  return addresses.filter((address) => re.test(address))
+}
+
 export interface ResponseSchema {
   accountName: string
   result: {
@@ -43,6 +56,16 @@ export const createVirtuneTransportConfig = <T extends VirtuneTransportGenerics>
   // because the compiler doesn't know getUrl satisfies the type until T
   // is instantiated.
   getUrlFromParams: (params: VirtuneParams<T>) => string,
+  // filterAddresses is always assigned to the exported filterAddresses.
+  // It's only a parameter because the compiler doesn't know
+  // filterAddresses satisfies the type until T is instantiated.
+  filterAddresses: ({
+    addresses,
+    params,
+  }: {
+    addresses: string[]
+    params: VirtuneParams<T>
+  }) => string[],
   getResultFromAddresses: (_: {
     params: VirtuneParams<T>
     addresses: string[]
@@ -88,8 +111,10 @@ export const createVirtuneTransportConfig = <T extends VirtuneTransportGenerics>
       ]
     }
 
+    const filteredAddresses = filterAddresses({ addresses, params: param })
+
     const result = getResultFromAddresses({
-      addresses,
+      addresses: filteredAddresses,
       params: param,
     })
 
--- a/.changeset/few-maps-cover.md
+++ b/.changeset/few-maps-cover.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/por-address-list-adapter': minor
+---
+
+Add optional address filter parameter to virtune endpoints
--- a/packages/sources/por-address-list/src/endpoint/virtune.ts
+++ b/packages/sources/por-address-list/src/endpoint/virtune.ts
@@ -21,6 +21,11 @@ export const sharedVirtuneInputParameters = {
     type: 'string',
     required: true,
   },
+  addressPattern: {
+    description: 'Return only addresses matching the given regular expression pattern.',
+    type: 'string',
+    required: false,
+  },
 } as const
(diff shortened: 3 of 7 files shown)


packages/sources/por-address-list/src/transport/virtune-utils.ts
```diff
@@ -7,6 +7,19 @@ export const getUrl = (params: { accountId: string }): string => {
...
+export const filterAddresses = ({
+  addresses,
+  params,
+}: {
+  addresses: string[]
+  params: { addressPattern?: string }
+}): string[] => {
+  const { addressPattern } = params
+  if (!addressPattern) return addresses
+  const re = new RegExp(addressPattern)
```
mmcallister-cll, line 19: How likely is it that we'll need anything other than P-, X-, or C-/0x, or none?

I agree regex as input param is very flexible, but I'm debating whether we should restrict this to known address types for simplicity, ie as an inputParam with options, since this is specific to virtune.

dskloetc: Currently we'd want `none`, `^P-` or `^(?!P-)` (i.e. all but `^P`). I guess it could be `0x` instead of `P-` but it would be less obvious that it's a partition.
I could make 3 options, such as `any`, `P-chain` and `all-but-P-chain` but it would be very specific to one network, while this endpoint of `por-address-list` is not specific to that network.
I could also add `X-` and `C-` options but they would be currently unused.

What do you think?

mxiao-cll: If the endpoint itself doesn't support certain address, it is better to filter it with-in the endpoint (vs creating a new param)

dskloetc: It has nothing to do with the endpoint of `por-address-list` supporting it.
The addresses that start with `P-` should be sent to `avalance-platform` to get balances and the addresses starting with `0x` should be sent to `eth-balance` to get the balances.
The Virtune API presents a single list of addresses and we need to split it before `proof-of-reserves` sends it off to the indexers and I believe `por-address-list` is the right place to do that.

mmcallister-cll: yeah my main point was regex may be confusing to look at in a job spec.
since addresses on avalanche are restricted to P-*, X-*, C-*, or 0x, what about something like `"addressType": "P"` input param in its place

dskloetc: That would make the parameter specific to Avalanche, while the endpoint is not specific to Avalanche. So the parameter would have to be called `avalancheAddressType` or something. I'm not sure I like adding an Avalanche specific parameter like that. We'd also have to invent a way to make sure the parameter is only used with Avalanche.

If you think the regex is too complicated, we could also change to `addressPrefix` and use `P-` vs. `0x`.
By only problem with that is that then it's less obvious that we are still including all addresses with those 2 prefixed.

Anyway, I've made my case. But if you tell me what to do, I'll do it.

PR #4026, opened 2025-09-16
OPDATA-4086: Build Blocksize Capital DEX State EA

paul-chain:
## Description
Adds a new WebSocket adapter for Blocksize Capital State API to provide real-time aggregated price feeds for DEX state data.

## Changes

- New adapter implementation for Blocksize Capital State API
- WebSocket transport with authentication and subscription management
- Out-of-order message detection and filtering
- Comprehensive test suite and integration tests
- Documentation and setup instructions

## Features

- **Real-time WebSocket streaming**: Live price feed updates via WebSocket connection
- **Provider timestamp preservation**: Maintains original timestamps for data traceability
- **Multiple quote currencies**: Support for USD and ETH quote currencies
- **Robust error handling**: Comprehensive error handling and logging
- **Rate limiting**: Built-in rate limiting configuration (60 requests/minute)

## Supported Assets

- `CBBTC` (Coinbase Wrapped BTC)
- `ALETH` (Alchemix ETH)
- `CBETH` (Coinbase Staked ETH)

### Automated Test
1. Navigate to the root of the external-adapters-js repo
2. Build the adapter: `yarn workspace @chainlink/blocksize-capital-state-adapter build`
3. Run the test: `yarn test packages/sources/blocksize-capital-state/test/integration/adapter.test.ts`

### Manual Testing
1. Set environment variables: `API_KEY="your-key-here"`
2. Build the adapter: `yarn workspace @chainlink/blocksize-capital-state-adapter build`
3. Start the adapter: `API_KEY="..." yarn start blocksize-capital-state`
4. Run test script: `bash packages/sources/blocksize-capital-state/test/test-ws.sh`
5. Verify WebSocket connections and real-time price updates work correctly

## Quality Assurance

- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `infra-k8s` configuration file.
- [ ] If a new adapter was made, or an existing one was modified so that its environment variables have changed, update the relevant `adapter-secrets` configuration file or update the [soak testing blacklist](/packages/scripts/src/get-changed-adapters/soakTestBlacklist.ts).
- [x] If a new adapter was made, or a new endpoint was added, update the `test-payload.json` file with relevant requests.
- [x] The branch naming follows git flow (`feature/x`, `chore/x`, `release/x`, `hotfix/x`, `fix/x`) or is created from Jira.
- [x] This is related to a maximum of one Jira story or GitHub issue.
[...]

PR #4058, opened 2025-09-26
Add Coinpaprika State adapter

paul-chain:
## Description
Adds a new Coinpaprika State external adapter that streams state_price via Server-Sent Events (SSE).
See package [README](https://github.com/smartcontractkit/external-adapters-js/blob/feat/add-coinpaprika-state-adapter/packages/sources/coinpaprika-state/README.md) for full details

## Changes

- New package: `@chainlink/coinpaprika-state-adapter`
- Endpoint: `coinpaprika-state` (alias: `state`)
- Transport: single-connection SSE with:
  - pair batching & dynamic reconnect on pair-set change
  - JSON parsing with malformed-event tolerance
  - graceful shutdown & reconnection backoff with jitter

## Steps to Test

1. Navigate to the root of the external-adapters-js repo
2. Build the adapter: `yarn workspace @chainlink/coinpaprika-state-adapter build`
3. Run tests:
- Integration test: `yarn test packages/sources/coinpaprika-state/test/integration/adapter.test.ts`
- Unit test: `yarn test packages/sources/coinpaprika-state/test/unit/sse.test.ts`

## Quality Assurance

### DIFF
--- a/packages/sources/coinpaprika-state/README.md
+++ b/packages/sources/coinpaprika-state/README.md
@@ -1,0 +1,199 @@
+# COINPAPRIKA_STATE
+
+The Coinpaprika State adapter streams **state_price** for supported pairs via **Server-Sent Events (SSE)** streaming.
+
+---
+
+## Environment Variables
+
+|        Variable         | Description                                          |  Type  | Required |                       Default                       |
+| :---------------------: | ---------------------------------------------------- | :----: | :------: | :-------------------------------------------------: |
+|        `API_KEY`        | Coinpaprika API key (sent as `Authorization` header) | string |    ✅    |                          —                          |
+|     `API_ENDPOINT`      | Coinpaprika streaming endpoint (POST)                | string |          | `https://chainlink-streaming.dexpaprika.com/stream` |
+| `BACKGROUND_EXECUTE_MS` | Background loop interval for pair set evaluation     | number |          |                       `3000`                        |
+
+---
+
+## Endpoint
+
+**Name:** `coinpaprika-state` (alias: `state`)
+
+### Input Parameters
+
+|    Name    | Required | Aliases        | Description        | Type   |            Example            |
+| :--------: | :------: | -------------- | ------------------ | ------ | :---------------------------: |
+|   `base`   |    ✅    | `coin`, `from` | Base asset symbol  | string |   `LUSD`, `ALETH`, `CBETH`    |
+|  `quote`   |    ✅    | `market`, `to` | Quote asset symbol | string |         `USD`, `ETH`          |
+| `endpoint` |          |                | Endpoint selector  | string | `coinpaprika-state` / `state` |
+
+### Example Requests
+
+**Using the primary endpoint `coinpaprika-state`**
+
+```bash
+curl -X POST http://localhost:8080 \
+  -H "Content-Type: application/json" \
+  -d '{"data":{"base":"LUSD","quote":"USD","endpoint":"coinpaprika-state"}}'
+```
+
+**Using the alias `state`**
+
+```bash
+curl -X POST http://localhost:8080 \
+    -H "Content-Type: application/json" \
+    -d '{"data": {"base": "LUSD", "quote": "USD", "endpoint": "state"}}'
+```
+
+**Without endpoint (uses default)**
+
+```bash
+curl -X POST http://localhost:8080 \
+  -H "Content-Type: application/json" \
+  -d '{"data": {"base": "LUSD", "quote": "USD"}}'
+```
+
+### Example Response
+
+```json
+{
+  "data": {
+    "result": 1.000979,
+    "timestamp": 1758888503
+  },
+  "statusCode": 200,
+  "result": 1.000979,
+  "timestamps": {
+    "providerDataRequestedUnixMs": 1758888508939,
+    "providerDataReceivedUnixMs": 1758888508939,
+    "providerIndicatedTimeUnixMs": 1758888503000
+  },
+  "meta": {
+    "adapterName": "COINPAPRIKA_STATE",
+    "metrics": {
+      "feedId": "{\"base\":\"lusd\",\"quote\":\"usd\"}"
+    }
+  }
+}
+```
+
+---
+
+## Streaming Data
+
+### Streaming Method Selection: Multiple Assets via POST (Sessionless)
+
+Per the “EA Requirements – Coinpaprika State” doc, Coinpaprika offers:
+
+1. **Single Asset via GET** — simple but requires one connection per pair
+2. **Multiple Assets via POST (Sessionless)** — ✅ **SELECTED APPROACH**
+3. **Multiple Assets via Session** — requires session management overhead
+
+### Streaming Data Response
+
+> Numeric values may arrive as **strings**; the adapter coerces them to numbers and skips malformed JSON frames.
+
+The adapter receives real-time data from Coinpaprika in the following format:
+
+```text
+data:{
+   "block_time":1758892511,
+   "send_timestamp":1758892514,
+   "base_token_symbol":"ALETH",
+   "quote_symbol":"USD",
+   "volume_7d_usd":190502.190952,
+   "market_depth_plus_1_usd":0.000000,
+   "market_depth_minus_1_usd":0.000000,
+   "state_price":3835.519084
+}
+event:t_s
+
+data:{
+   "block_time":1758892487,
+   "send_timestamp":1758892514,
+   "base_token_symbol":"EURA",
+   "quote_symbol":"USD",
+   "volume_7d_usd":36039.253665,
+   "market_depth_plus_1_usd":0.000000,
+   "market_depth_minus_1_usd":0.000000,
+   "state_price":1.158675
+}
+event:t_s
+
+data:{
+   "block_time":1758892511,
+   "send_timestamp":1758892515,
+   "base_token_symbol":"LUSD",
+   "quote_symbol":"USD",
+   "volume_7d_usd":23434.612065,
+   "market_depth_plus_1_usd":0.000000,
+   "market_depth_minus_1_usd":0.000000,
+   "state_price":1.000883
+}
+event:t_s
+```
+
+---
+
+## Architecture
+
+### Transport
+
+`CoinpaprikaStateTransport` (extends framework `SubscriptionTransport`)
+
+- **Single SSE connection** for all active pairs
+- **Dynamic pair batching**; reconnects when the pair set changes
+- **Malformed event tolerance** (skips bad JSON frames, keeps last good value)
+- **Numeric coercion** for `state_price` & `block_time` (string → number)
+- **Error mapping**: `401/400/429` passthrough, `500 → 502`, **no data → 504**
+- **Graceful shutdown** + reconnect backoff with jitter
+
+### Data Flow
+
+1. **Background Handler** monitors requested pairs and manages SSE connection
+2. **SSE Parser** handles chunked event streams (`src/transport/sse.ts`)
+3. **Processing** (`src/transport/coinpaprika-state.ts`) normalizes symbols, coerces numbers, validates payloads, and writes to the response cache
+4. **Serving requests** returns the latest cached tick for the requested pair
+
+### Error Handling
+
+| Provider Status | Adapter Status | Description                             |
+| :-------------: | :------------: | :-------------------------------------- |
+|      `401`      |     `401`      | Unauthorized                            |
+|      `400`      |     `400`      | Bad Request                             |
+|      `429`      |     `429`      | Rate Limited                            |
+|      `500`      |     `502`      | Bad Gateway - Provider error            |
+|   No data yet   |     `504`      | Foreground fallback when cache is empty |
+
+---
+
+## Development
+
+### Build
+
+```bash
+yarn workspace @chainlink/coinpaprika-state-adapter build
+```
+
+### Start
+
+```bash
+ API_KEY="your-api-key-here" yarn start coinpaprika-state
+```
+
+### Test
+
+```bash
+# Integration
+yarn test packages/sources/coinpaprika-state/test/integration/adapter.test.ts
+
+# Unit
+yarn test packages/sources/coinpaprika-state/test/unit/sse.test.ts
+```
+
+---
+
+## Notes
+
+- **First-tick latency**: Until the first valid tick is cached, foreground requests may return `504`.
+- **Symbol normalization**: `base`/`quote` are uppercased; stick to canonical symbols.
+- **Numeric coercion**: Provider may send numerics as strings; invalid numerics are dropped.
--- a/packages/sources/coinpaprika-state/src/endpoint/coinpaprika-state.ts
+++ b/packages/sources/coinpaprika-state/src/endpoint/coinpaprika-state.ts
@@ -1,0 +1,48 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+import overrides from '../config/overrides.json'
+import { CoinpaprikaSubscriptionTransport } from '../transport/coinpaprika-state'
+
+export const inputParameters = new InputParameters(
+  {
+    base: {
+      aliases: ['from', 'coin'],
+      required: true,
+      type: 'string',
+      description: 'The symbol of the currency to query',
+    },
+    quote: {
+      aliases: ['to', 'market'],
+      required: true,
+      type: 'string',
+      description: 'The symbol of the currency to convert to',
+    },
+  },
+  [
+    {
+      base: 'LUSD',
+      quote: 'USD',
+    },
+  ],
+)
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: {
+    Result: number
+    Data: {
+      result: number
+      timestamp: number
+    }
+  }
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'coinpaprika-state',
+  aliases: ['state'],
+  transport: CoinpaprikaSubscriptionTransport,
+  inputParameters,
+  overrides: overrides['coinpaprika-state'],
+})
--- a/packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
+++ b/packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
@@ -1,0 +1,368 @@
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
+import { ResponseCache } from '@chainlink/external-adapter-framework/cache/response'
+import { TransportDependencies } from '@chainlink/external-adapter-framework/transports'
+import { SubscriptionTransport } from '@chainlink/external-adapter-framework/transports/abstract/subscription'
+import { AdapterResponse, makeLogger, sleep } from '@chainlink/external-adapter-framework/util'
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import type { AxiosRequestConfig } from 'axios'
+import { Readable } from 'node:stream'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/coinpaprika-state'
+import { SSEParser } from './sse'
+
+const logger = makeLogger('CoinpaprikaStateTransport')
+
+type RequestParams = typeof inputParameters.validated
+
+// Coinpaprika’s OpenAPI doc says all numeric values are encoded as strings to preserve precision
+interface CoinpaprikaStreamData {
+  block_time: string | number
+  base_token_symbol: string
+  quote_symbol: string
+  volume_7d_usd: string | number
+  market_depth_plus_1_usd: string | number
+  market_depth_minus_1_usd: string | number
+  state_price: string | number
+}
+
+export type TransportTypes = BaseEndpointTypes & {
+  Provider: {
+    RequestBody: Array<{ base: string; quote: string }>
+    ResponseBody: CoinpaprikaStreamData
+  }
+}
+
+const norm = (s: string) => s.trim().toUpperCase()
+const normalizeParams = (p: RequestParams): RequestParams => ({
+  ...p,
+  base: norm(p.base),
+  quote: norm(p.quote),
+})
+
+/**
+ * Single-connection SSE transport that batches all pairs into one POST and streams state_price ticks into the cache.
+ *
+ * */
+export class CoinpaprikaStateTransport extends SubscriptionTransport<TransportTypes> {
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
+  requester!: Requester
+
+  // single SSE conn for all pairs
+  private activeConnection: AbortController | null = null
+  private activePairs: Map<string, RequestParams> = new Map()
+  // reconnection backoff base (ms); jitter added per attempt //TODO
+  private lastConnectionAttempt: number = 0
+  private reconnectDelay: number = 5000
+  // guard to prevent reconnects during/after shutdown
+  private isShuttingDown = false
+  private sseParser: SSEParser | null = null
+
+  async initialize(
+    dependencies: TransportDependencies<TransportTypes>,
+    adapterSettings: TransportTypes['Settings'],
+    endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    this.isShuttingDown = false
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+    this.requester = dependencies.requester
+  }
+
+  async backgroundHandler(context: EndpointContext<TransportTypes>, entries: RequestParams[]) {
+    if (this.isShuttingDown) return
+
+    // normalize params once at ingestion
+    const normalized = entries.map(normalizeParams)
+
+    // build current pairs map
+    const currentPairs = new Map<string, RequestParams>()
+    for (const e of normalized) currentPairs.set(`${e.base}/${e.quote}`, e)
+
+    // detect changes in the requested pair set (triggers reconnect if changed)
+    const pairsChanged = this.havePairsChanged(currentPairs)
+
+    if (pairsChanged || !this.activeConnection) {
+      logger.info(`Pairs changed or no connection. Updating stream with ${currentPairs.size} pairs`)
+      logger.debug(`Pairs: ${[...currentPairs.keys()].join(', ')}`)
+
+      await this.updateStream(context, currentPairs)
+    }
+    await sleep(context.adapterSettings.BACKGROUND_EXECUTE_MS)
+  }
+
+  private havePairsChanged(pairs: Map<string, RequestParams>): boolean {
+    if (pairs.size !== this.activePairs.size) return true
+    for (const [key] of pairs) {
+      if (!this.activePairs.has(key)) return true
+    }
+    return false
+  }
+
+  private async updateStream(
+    context: EndpointContext<TransportTypes>,
+    pairs: Map<string, RequestParams>,
+  ) {
+    if (this.isShuttingDown) return
+
+    // close existing connection if any
+    if (this.activeConnection) {
+      logger.info('Closing existing SSE connection')
+      this.activeConnection.abort()
+      this.activeConnection = null
+    }
+    // update active pairs
+    this.activePairs = new Map(pairs)
+
+    if (pairs.size === 0) {
+      logger.debug('No pairs to stream')
+      return
+    }
+
+    await this.createSSEConnection(context)
+  }
+
+  private buildSSERequest(
+    context: EndpointContext<TransportTypes>,
+    pairsArray: Array<{ base: string; quote: string }>,
+    signal: AbortSignal,
+  ): AxiosRequestConfig {
+    return {
+      url: context.adapterSettings.API_ENDPOINT,
+      method: 'POST',
+      headers: {
+        Authorization: context.adapterSettings.API_KEY,
+        'Content-Type': 'application/json',
+        Accept: 'text/event-stream',
+      },
+      data: pairsArray,
+      responseType: 'stream',
+      signal,
+      timeout: 60_000,
+      validateStatus: () => true,
+    }
+  }
+
+  private async createSSEConnection(context: EndpointContext<TransportTypes>) {
+    if (this.isShuttingDown) return
+
+    const pairsArray = Array.from(this.activePairs.values()).map((p) => ({
+      base: p.base,
+      quote: p.quote,
+    }))
+
+    logger.info(`Opening single SSE connection for ${pairsArray.length} pairs`)
+    logger.debug(`Pairs: ${pairsArray.map((p) => `${p.base}/${p.quote}`).join(', ')}`)
+
+    this.lastConnectionAttempt = Date.now()
+    const controller = new AbortController()
+    this.activeConnection = controller
+
+    try {
+      const req = this.buildSSERequest(context, pairsArray, controller.signal)
+      const key = `coinpaprika-state/stream:${[...this.activePairs.keys()].join(',')}`
+      const { response } = await this.requester.request<Readable>(key, req)
+      const axiosResp = response
+
+      if (axiosResp.status !== 200 || !axiosResp.data) {
+        // try to read error body if available
+        let errDetail = `HTTP error! status: ${axiosResp.status}`
+        let rawErrBody = ''
+        try {
+          const text = await new Promise<string>((resolve) => {
+            let buf = ''
+            ;(axiosResp.data as Readable)
+              .on('data', (c: Buffer) => (buf += c.toString('utf8')))
+              .on('end', () => resolve(buf))
+              .on('error', () => resolve(''))
+          })
+          rawErrBody = text
+          try {
+            const j = JSON.parse(text)
+            if (j?.error && j?.details) errDetail = `${j.error}: ${j.details}`
+          } catch {
+            logger.error(`Provider error body (not JSON): ${rawErrBody}`)
+          }
+        } catch {
+          // ignore parse error; keep default errDetail
+        }
+
+        const statusCode =
+          axiosResp.status === 500
+            ? 502
+            : axiosResp.status === 401
+            ? 401
+            : axiosResp.status === 400
+            ? 400
+            : axiosResp.status === 429
+            ? 429
+            : axiosResp.status
+
+        const errorResponse: AdapterResponse<TransportTypes['Response']> = {
+          statusCode,
+          errorMessage: errDetail,
+          timestamps: {
+            providerDataRequestedUnixMs: Date.now(),
+            providerDataReceivedUnixMs: Date.now(),
+            providerIndicatedTimeUnixMs: undefined,
+          },
+        }
+
+        for (const params of this.activePairs.values()) {
+          await this.responseCache.write(this.name, [{ params, response: errorResponse }])
+        }
+
+        if (this.activeConnection === controller) this.activeConnection = null
+        return
+      }
+
+      const stream: Readable = axiosResp.data as Readable
+      let aborted = false
+      this.sseParser = new SSEParser('t_s')
+
+      const handleParsedEvent = async (eventType: string, rawData: string) => {
+        if (this.isShuttingDown) return
+        if (eventType !== 't_s') {
+          logger.debug(`Skipping event type: ${eventType}`)
+          return
+        }
+
+        try {
+          const streamData: CoinpaprikaStreamData = JSON.parse(rawData)
+          const pairKey = `${norm(streamData.base_token_symbol)}/${norm(streamData.quote_symbol)}`
+          const params = this.activePairs.get(pairKey)
+          if (params) {
+            await this.handleStreamData(params, streamData)
+          } else {
+            logger.warn(`Received data for untracked pair: ${pairKey}`)
+          }
+        } catch (err) {
+          logger.debug(`Failed to parse SSE data: ${rawData} | Error: ${(err as Error).message}`)
+        }
+      }
+
+      const onData = async (chunk: Buffer) => {
+        if (this.isShuttingDown) return
+        const raw = chunk.toString('utf8')
+        logger.debug(`Raw price state update message received:\n${raw}`)
+
+        this.sseParser!.push(raw, (evt, data) => {
+          void handleParsedEvent(evt, data)
+        })
+      }
+
+      const onError = (err: Error) => {
+        if (err.name === 'CanceledError' || err.name === 'AbortError') {
+          aborted = true
+        } else {
+          logger.error(`Stream error: ${err}`)
+        }
+      }
+
+      const onEnd = async () => {
+        stream.off('data', onData)
+        stream.off('error', onError)
+        this.sseParser?.reset()
+
+        // mark closed immediately to prevent races
+        if (this.activeConnection === controller) this.activeConnection = null
+
+        if (!this.isShuttingDown && !aborted && this.activePairs.size > 0) {
+          const since = Date.now() - this.lastConnectionAttempt
+          const delay = Math.max(this.reconnectDelay - since, 0) + Math.floor(Math.random() * 1000)
+          logger.info(`SSE ended; reconnecting in ${delay} ms...`)
+          await sleep(delay)
+          if (!this.isShuttingDown) {
+            await this.createSSEConnection(context)
+          }
+        }
+      }
+
+      // wire events
+      stream.on('data', (chunk) => {
+        void onData(chunk as Buffer)
+      })
+      stream.once('error', onError)
+      stream.once('end', onEnd)
+    } catch (error) {
+      this.activeConnection = null
+      if (this.isShuttingDown) return
+      logger.error(`Failed to create SSE connection: ${error}`)
+      return
+    }
+  }
+
+  private async handleStreamData(param: RequestParams, streamData: CoinpaprikaStreamData) {
+    // coerce string -> number
+    const toNum = (value: string | number | undefined | null): number => {
+      if (typeof value === 'number') return value
+      if (typeof value === 'string' && value.trim() !== '') {
+        const n = Number(value)
+        return Number.isFinite(n) ? n : NaN
+      }
+      return NaN
+    }
+    const statePrice = toNum(streamData.state_price)
+    const blockTime = toNum(streamData.block_time)
+
+    // guard against bad payloads
+    if (!Number.isFinite(statePrice) || !Number.isFinite(blockTime) || blockTime <= 0) {
+      logger.warn(
+        `Bad numeric fields for ${param.base}/${param.quote}: ${JSON.stringify(streamData)}`,
+      )
+      return
+    }
+
+    const response: AdapterResponse<TransportTypes['Response']> = {
+      data: {
+        result: statePrice,
+        timestamp: blockTime,
+      },
+      statusCode: 200,
+      result: statePrice,
+      timestamps: {
+        providerDataRequestedUnixMs: Date.now(),
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: blockTime * 1000,
+      },
+    }
+    logger.debug(`tick ${param.base}/${param.quote}=${statePrice} t=${blockTime}`)
+    logger.trace(JSON.stringify(response))
+
+    await this.responseCache.write(this.name, [{ params: param, response }])
+  }
+
+  async _handleRequest(_: RequestParams): Promise<AdapterResponse<TransportTypes['Response']>> {
+    // This fallback shouldn't be called in normal operation - data should come from cache via backgroundHandler
+    logger.debug('Foreground request hit fallback: no cached data yet')
+    return {
+      statusCode: 504,
+      errorMessage:
+        'No cached data available. Streaming transport is initializing or waiting for data.',
+      timestamps: {
+        providerDataRequestedUnixMs: Date.now(),
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: undefined,
+      },
+    }
+  }
+
+  getSubscriptionTtlFromConfig(adapterSettings: TransportTypes['Settings']): number {
+    return adapterSettings.WARMUP_SUBSCRIPTION_TTL
+  }
+
+  async close(): Promise<void> {
+    logger.info('Closing SSE connection')
+    this.isShuttingDown = true
+
+    if (this.activeConnection) {
+      this.activeConnection.abort()
+      this.activeConnection = null
+    }
+    this.activePairs.clear()
+
+    if (this.sseParser) this.sseParser.reset()
+    this.sseParser = null
+  }
+}
+
+export const CoinpaprikaSubscriptionTransport = new CoinpaprikaStateTransport()
--- a/packages/sources/coinpaprika-state/src/transport/sse.ts
+++ b/packages/sources/coinpaprika-state/src/transport/sse.ts
@@ -1,0 +1,56 @@
+/**
+ * Minimal SSE parser that accumulates chunked lines and emits complete events.
+ * - Supports "event:" name (optional; defaults to 't_s')
+ * - Supports multi-line "data:" blocks
+ * - Ignores comment/heartbeat lines starting with ':'
+ */
+export class SSEParser {
+  private buffer = ''
+  private dataLines: string[] = []
+  private currentEvent: string | null = null
+  private readonly defaultEvent: string
+
+  constructor(defaultEvent = 't_s') {
+    this.defaultEvent = defaultEvent
+  }
+
+  /**
+   * Push a chunk and emit parsed events via callback.
+   * The callback is invoked once per complete event frame.
+   */
+  push(chunk: string, onEvent: (eventType: string, data: string) => void): void {
+    this.buffer += chunk
+    const lines = this.buffer.split('\n')
+    this.buffer = lines.pop() || ''
+
+    for (const rawLine of lines) {
+      const line = rawLine.replace(/\r$/, '')
+
+      if (line.startsWith(':')) continue
+
+      if (line.startsWith('event:')) {
+        this.currentEvent = line.slice(6).trim()
+        continue
+      }
+      if (line.startsWith('data:')) {
+        this.dataLines.push(line.slice(5).trim())
+        continue
+      }
+
+      // end of an event
+      if (line.trim() === '' && this.dataLines.length > 0) {
+        const rawData = this.dataLines.join('\n')
+        this.dataLines = []
+        const evt = this.currentEvent ?? this.defaultEvent
+        this.currentEvent = null
+        onEvent(evt, rawData)
+      }
+    }
+  }
+
+  reset(): void {
+    this.buffer = ''
+    this.dataLines = []
+    this.currentEvent = null
+  }
+}
--- a/packages/sources/coinpaprika-state/test/integration/adapter.test.ts
+++ b/packages/sources/coinpaprika-state/test/integration/adapter.test.ts
@@ -1,0 +1,338 @@
+import {
+  TestAdapter,
+  setEnvVariables,
+} from '@chainlink/external-adapter-framework/util/testing-utils'
+import nock from 'nock'
+import {
+  endSSEStream,
+  mockStreamPost,
+  mockStreamPostAnyBody,
+  mockStreamPostRawAnyBody,
+  mockStreamPostRawMatchingBody,
+  sseEventChunk,
+  waitFor,
+} from './fixtures'
+
+jest.setTimeout(10000)
+
+describe('coinpaprika-state adapter', () => {
+  let testAdapter: TestAdapter
+  let oldEnv: NodeJS.ProcessEnv
+  let outSpy: jest.SpyInstance, errSpy: jest.SpyInstance
+
+  beforeAll(async () => {
+    outSpy = jest.spyOn(process.stdout, 'write').mockImplementation(() => true as any)
+    errSpy = jest.spyOn(process.stderr, 'write').mockImplementation(() => true as any)
+
+    process.env.LOG_LEVEL = 'silent'
+    process.env.EA_LOG_LEVEL = 'silent'
+    process.env.PINO_LOG_LEVEL = 'silent'
+    process.env.METRICS_ENABLED = 'false'
+    process.env.RETRY = '0'
+
+    nock.disableNetConnect()
+    oldEnv = JSON.parse(JSON.stringify(process.env))
+    process.env.API_KEY = 'TEST-KEY'
+    process.env.API_ENDPOINT = 'http://localhost:1234/stream'
+    process.env.BACKGROUND_EXECUTE_MS = '100'
+
+    const adapter = (await import('../../src')).adapter
+    adapter.rateLimiting = undefined
+    testAdapter = await TestAdapter.startWithMockedCache(adapter, {
+      testAdapter: {} as TestAdapter<never>,
+    })
+    nock.enableNetConnect((host) => host.includes('127.0.0.1') || host.includes('localhost'))
+  })
+
+  afterAll(async () => {
+    await testAdapter.api.close()
+    await new Promise((r) => setTimeout(r, 400))
+    nock.abortPendingRequests()
+    nock.cleanAll()
+    nock.enableNetConnect()
+    setEnvVariables(oldEnv)
+    outSpy.mockRestore()
+    errSpy.mockRestore()
+  })
+
+  beforeEach(() => {
+    nock.cleanAll()
+  })
+
+  it('happy path: streams ticks and serves latest state_price for LUSD/USD', async () => {
+    const { scope, stream } = mockStreamPost({
+      apiBase: 'http://localhost:1234',
+      pairs: [{ base: 'LUSD', quote: 'USD' }],
+      events: [
+        {
+          block_time: 1756224311,
+          base_token_symbol: 'LUSD',
+          quote_symbol: 'USD',
+          state_price: 1.0005,
+          volume_7d_usd: 1234.56,
+          market_depth_plus_1_usd: 0,
+          market_depth_minus_1_usd: 0,
+        },
+        {
+          block_time: 1756224313,
+          base_token_symbol: 'LUSD',
+          quote_symbol: 'USD',
+          state_price: 1.0007,
+          volume_7d_usd: 2222,
+          market_depth_plus_1_usd: 0,
+          market_depth_minus_1_usd: 0,
+        },
+      ],
+    })
+
+    await waitFor(async () => {
+      const r = await testAdapter.request({ base: 'LUSD', quote: 'USD', endpoint: 'state' })
+      expect(r.statusCode).toBe(200)
+      expect(r.json().result).toBeCloseTo(1.0007, 3)
+      expect(r.json().data.timestamp).toBe(1756224313)
+    })
+
+    endSSEStream(stream)
+    scope.done()
+  })
+
+  it('returns 504 when no data available', async () => {
+    const scope = nock('http://localhost:1234')
+      .post('/stream')
+      .matchHeader('authorization', 'TEST-KEY')
+      .reply(200, () => ':heartbeat\n\n', { 'Content-Type': 'text/event-stream' })
+      .persist()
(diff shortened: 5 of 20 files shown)


packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+import type { AxiosRequestConfig } from 'axios'
+import { Readable } from 'node:stream'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/coinpaprika-state'
+import { SSEParser } from './sse'
+
+const logger = makeLogger('CoinpaprikaStateTransport')
+
+type RequestParams = typeof inputParameters.validated
+
+// Coinpaprika’s OpenAPI doc says all numeric values are encoded as strings to preserve precision
```
mxiao-cll: Then why is the type string or number

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+    ResponseBody: CoinpaprikaStreamData
+  }
+}
+
+const norm = (s: string) => s.trim().toUpperCase()
+const normalizeParams = (p: RequestParams): RequestParams => ({
+  ...p,
+  base: norm(p.base),
+  quote: norm(p.quote),
+})
```
mxiao-cll: You can actually do this in endpoint, search `requestTransforms`

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * Single-connection SSE transport that batches all pairs into one POST and streams state_price ticks into the cache.
+ *
+ * */
+export class CoinpaprikaStateTransport extends SubscriptionTransport<TransportTypes> {
```
mxiao-cll: Why not use `SseTransport` ?

karen-stepanyan:
any reason why this is not a new endpoint on existing coinpaprika EA ?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: I investigated SseTransport but found it's designed around the usual EventSource/GET model (and optional separate sub/unsub/keepalive calls). The current design of Coinpaprika State API instead opens a single SSE connection via POST with a JSON body containing all pairs and expects reconnects when the pair set changes. Given that shape, SubscriptionTransport is a cleaner fit for batching pairs and managing one shared connection + reconnection logic.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: The string | number type is intentional. From what I can see Coinpaprika's documentation states they encode numeric values as strings, so this type union is for defensive programming.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: updated

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: Then this comment is mis-leading

.changeset/witty-buses-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/coinpaprika-adapter': major
```
mxiao-cll: Major usually means breaking change

.gitignore
```diff
@@ -33,3 +33,4 @@ packages/k6/src/config/http.json
 # Secrets
 .envrc
 .env
+.yarn/cache/
```
mxiao-cll: Why?

packages/sources/coinpaprika/src/endpoint/state.ts
```diff
@@ -0,0 +1,53 @@
+import { AdapterEndpoint } from '@chainlink/external-adapter-framework/adapter'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { config } from '../config'
+// import overrides from '../config/overrides.json'
```
mxiao-cll: remove

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
...
+import type { AxiosRequestConfig } from 'axios'
+import { Readable } from 'node:stream'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/state'
+import { SSEParser } from './utils'
+
+const logger = makeLogger('CoinpaprikaStateTransport')
+
+type RequestParams = typeof inputParameters.validated
+
+// Coinpaprika's OpenAPI doc says all numeric values are encoded as strings to preserve precision
```
mxiao-cll: say that you are putting number as defensive

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
...
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+    this.requester = dependencies.requester
+  }
+
+  async backgroundHandler(context: EndpointContext<TransportTypes>, entries: RequestParams[]) {
+    if (this.isShuttingDown) return
+
+    // build current pairs map (params already normalized by endpoint's requestTransforms)
+    const currentPairs = new Map<string, RequestParams>()
+    for (const e of entries) currentPairs.set(`${e.base}/${e.quote}`, e)
```
mxiao-cll: Use { } everywhere

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
...
+      method: 'POST',
+      headers: {
+        Authorization: context.adapterSettings.API_KEY,
+        'Content-Type': 'application/json',
+        Accept: 'text/event-stream',
+      },
+      data: pairsArray,
+      responseType: 'stream',
+      signal,
+      timeout: 60_000,
```
mxiao-cll: move this to config or up

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
...
+
+        const statusCode =
+          axiosResp.status === 500
+            ? 502
+            : axiosResp.status === 401
+            ? 401
+            : axiosResp.status === 400
+            ? 400
+            : axiosResp.status === 429
+            ? 429
```
mxiao-cll: ???

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+            providerDataReceivedUnixMs: Date.now(),
+            providerIndicatedTimeUnixMs: undefined,
+          },
+        }
+
+        for (const params of this.activePairs.values()) {
+          await this.responseCache.write(this.name, [{ params, response: errorResponse }])
+        }
+
+        if (this.activeConnection === controller) this.activeConnection = null
```
mxiao-cll: What is this doing?

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
...
+
+        if (this.activeConnection === controller) this.activeConnection = null
+        return
+      }
+
+      const stream: Readable = axiosResp.data as Readable
+      let aborted = false
+      this.sseParser = new SSEParser('t_s')
+
+      const handleParsedEvent = async (eventType: string, rawData: string) => {
```
mxiao-cll: No function inside code

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+    endpointName: string,
+    transportName: string,
+  ): Promise<void> {
+    this.isShuttingDown = false
+    await super.initialize(dependencies, adapterSettings, endpointName, transportName)
+    this.requester = dependencies.requester
+  }
+
+  async backgroundHandler(context: EndpointContext<TransportTypes>, entries: RequestParams[]) {
+    if (this.isShuttingDown) return
```
mxiao-cll: Why do we need isShuttingDown

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
...
+              .on('end', () => resolve(buf))
+              .on('error', () => resolve(''))
+          })
+          rawErrBody = text
+          logger.error(`Provider error response (status ${axiosResp.status}): ${rawErrBody}`)
+          try {
+            const j = JSON.parse(text)
+            if (j?.error && j?.details) errDetail = `${j.error}: ${j.details}`
+            else if (j?.error) errDetail = `${j.error}`
+            else if (j?.message) errDetail = `${j.message}`
```
mxiao-cll: Why can't it just be the entire j

packages/sources/coinpaprika/src/transport/utils.ts
```diff
@@ -249,3 +249,60 @@ export const buildCryptoHttpTransport = (
...
+        continue
+      }
+
+      // end of an event
+      if (line.trim() === '' && this.dataLines.length > 0) {
+        const rawData = this.dataLines.join('\n')
+        this.dataLines = []
+        const evt = this.currentEvent ?? this.defaultEvent
+        this.currentEvent = null
+        onEvent(evt, rawData)
```
mxiao-cll: Can we just return something and call `onEvent` outside?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,368 @@
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
```
mxiao-cll, line 1: Prob need to break this file down further

.changeset/witty-buses-exist.md
mmcallister-cll: why isn't this a separate EA like blocksize-capital-state?

packages/sources/coinpaprika/src/transport/state.ts
```diff
@@ -0,0 +1,368 @@
+import { EndpointContext } from '@chainlink/external-adapter-framework/adapter'
+import { ResponseCache } from '@chainlink/external-adapter-framework/cache/response'
+import { TransportDependencies } from '@chainlink/external-adapter-framework/transports'
+import { SubscriptionTransport } from '@chainlink/external-adapter-framework/transports/abstract/subscription'
+import { AdapterResponse, makeLogger, sleep } from '@chainlink/external-adapter-framework/util'
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import type { AxiosRequestConfig } from 'axios'
```
mmcallister-cll: if possible remove the axios dependency

packages/sources/coinpaprika/test/integration/adapter.test.ts
```diff
@@ -125,4 +125,23 @@ describe('execute', () => {
...
+      mockStateStreamingResponse()
+
+      const data = {
+        endpoint: 'state',
+        base: 'LUSD',
+        quote: 'USD',
+      }
+
+      const response = await testAdapter.request(data)
+      expect([200, 504]).toContain(response.statusCode)
```
mmcallister-cll: ?

alejoberardino:
saw this pr randomly, there's an [sse transport in the framework](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/transports/sse.ts), what's the particular reason this adapter needs a custom one?

packages/sources/coinpaprika/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@chainlink/external-adapter-framework": "2.7.0",
+    "axios": "1.12.2",
```
mmcallister-cll: hey @paul-chain why is this needed?

.changeset/witty-buses-exist.md
mmcallister-cll: same context from https://github.com/smartcontractkit/external-adapters-js/pull/4026/files#r2402650936 applies here

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

.changeset/witty-buses-exist.md
paul-chain: after discussion, separated EA

packages/sources/coinpaprika/src/transport/state.ts
paul-chain: updated

packages/sources/coinpaprika/src/transport/state.ts
paul-chain: deleted

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika/package.json
paul-chain: removed axios

packages/sources/coinpaprika/src/transport/state.ts
paul-chain: done

packages/sources/coinpaprika/src/transport/state.ts
paul-chain: done

packages/sources/coinpaprika/src/transport/state.ts
paul-chain: done

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Without it, async callbacks from the SSE stream continue firing after close() completes, causing the adapter to process stale data and potentially reconnect. The framework doesn't provide connection lifecycle coordination

paul-chain: agreed, it looks like redundant check, removed

packages/sources/coinpaprika-state/README.md
```diff
@@ -0,0 +1,199 @@
...
+    "metrics": {
+      "feedId": "{\"base\":\"lusd\",\"quote\":\"usd\"}"
+    }
+  }
+}
+```
+
+---
+
+## Streaming Data
```
mxiao-cll: You can't write custom stuff here, it will be overwritten by workflow

Instead create a `known-issues.md` file

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+
+export type TransportTypes = BaseEndpointTypes & {
+  Provider: {
+    RequestBody: Array<{ base: string; quote: string }>
+    ResponseBody: CoinpaprikaStreamData
+  }
+}
+
+const norm = (s: string) => s.trim().toUpperCase()
+const normalizeParams = (p: RequestParams): RequestParams => ({
```
mxiao-cll: You can move this into endpoint via `requestTransforms`

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
+  requester!: Requester
+
+  // single SSE conn for all pairs
+  private activeConnection: AbortController | null = null
+  private activePairs: Map<string, RequestParams> = new Map()
+  // reconnection backoff base (ms); jitter added per attempt //TODO
+  private lastConnectionAttempt = 0
+  private reconnectDelay = 5000
```
mxiao-cll: Use value from config

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+ * */
+export class CoinpaprikaStateTransport extends SubscriptionTransport<TransportTypes> {
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
+  requester!: Requester
+
+  // single SSE conn for all pairs
+  private activeConnection: AbortController | null = null
+  private activePairs: Map<string, RequestParams> = new Map()
+  // reconnection backoff base (ms); jitter added per attempt //TODO
```
mxiao-cll: TODO what?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+    }
+    this.activePairs.clear()
+
+    if (this.sseParser) {
+      this.sseParser.reset()
+    }
+    this.sseParser = null
+  }
+
+  private mapHttpStatusCode(status: number): number {
```
mxiao-cll: Why do we need this?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+      } else {
+        logger.warn(`Received data for untracked pair: ${pairKey}`)
+      }
+    } catch (err) {
+      logger.debug(`Failed to parse SSE data: ${rawData} | Error: ${(err as Error).message}`)
+    }
+  }
+}
+
+export const CoinpaprikaSubscriptionTransport = new CoinpaprikaStateTransport()
```
mxiao-cll: variable lower case

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+ * Single-connection SSE transport that batches all pairs into one POST and streams state_price ticks into the cache.
+ *
+ * */
+export class CoinpaprikaStateTransport extends SubscriptionTransport<TransportTypes> {
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
+  requester!: Requester
+
+  // single SSE conn for all pairs
+  private activeConnection: AbortController | null = null
```
mxiao-cll: Why do we need an AbortController here? Looks like we can get away with just a boolean?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+
+        // mark closed immediately to prevent races
+        this.activeConnection = null
+
+        if (!aborted && this.activePairs.size > 0 && this.activeConnection) {
+          const since = Date.now() - this.lastConnectionAttempt
+          const delay = Math.max(this.reconnectDelay - since, 0) + Math.floor(Math.random() * 1000)
+          logger.info(`SSE ended; reconnecting in ${delay} ms...`)
+          await sleep(delay)
+          if (this.activeConnection) {
```
mxiao-cll: How can this be true?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+    // build current pairs map
+    const currentPairs = new Map<string, RequestParams>()
+    for (const e of normalized) {
+      currentPairs.set(`${e.base}/${e.quote}`, e)
+    }
+
+    // detect changes in the requested pair set (triggers reconnect if changed)
+    const pairsChanged = this.havePairsChanged(currentPairs)
+
+    if (pairsChanged || !this.activeConnection) {
```
mxiao-cll: Don't seem to need pairsChanged var

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+        return 400
+      case 429:
+        return 429
+      default:
+        return status
+    }
+  }
+
+  private async handleParsedSSEEvent(eventType: string, rawData: string): Promise<void> {
+    if (eventType !== 't_s') {
```
mxiao-cll: t_s needs to be a variable

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,384 @@
...
+      base: p.base,
+      quote: p.quote,
+    }))
+
+    logger.info(`Opening single SSE connection for ${pairsArray.length} pairs`)
+    logger.debug(`Pairs: ${pairsArray.map((p) => `${p.base}/${p.quote}`).join(', ')}`)
+
+    this.lastConnectionAttempt = Date.now()
+    const controller = new AbortController()
+    this.activeConnection = controller
```
mxiao-cll: Can be one line

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: This file is still too complex, there is callbacks all over and control logic here and there. See if it is possible to break it down into separate files, things like handling SSE connection into 1 and then business logic stays here

paul-chain:
## Sample Input and Output

### Sample Input
```json
{
  "base": "LUSD",
  "quote": "USD"
}
```

### Sample Output
```json
{
  "result": 1.0007,
  "statusCode": 200,
  "data": {
    "result": 1.0007,
    "timestamp": 1756224313
  },
  "timestamps": {
    "providerDataRequestedUnixMs": 1641035471111,
    "providerDataReceivedUnixMs": 1641035471111,
    "providerIndicatedTimeUnixMs": 1756224313000
  }
}
```

The adapter streams real-time state_price data via Server-Sent Events (SSE) and caches the latest values for each currency pair.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: correct, fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: This follows standard EA practice - mapping provider 500 errors to 502 (Bad Gateway) to distinguish provider issues from adapter issues. Multiple EAs use this same pattern (trueusd, the-network-firm, emgemx, etc.) for proper error attribution and monitoring.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: removed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: done

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Keeping normalizeParams in transport layer due to TypeScript complexity with requestTransforms - attempted migration caused type inference issues.

packages/sources/coinpaprika-state/README.md
paul-chain: done

packages/non-deployable/token-allocation/src/config/index.ts
```diff
@@ -35,7 +35,7 @@ export const adaptersV3: PriceAdapter<any>[] = [
   CoinGecko,
   CoinMarketCap,
   CoinMetrics,
-  CoinPaprika,
+  CoinPaprika as any, // Framework 2.7.1 compatibility
```
mxiao-cll: Needs to be fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: This does not seem to be fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,381 @@
...
+  name!: string
+  responseCache!: ResponseCache<TransportTypes>
+  requester!: Requester
+
+  // single SSE conn for all pairs
+  private activeConnection: AbortController | null = null
+  private activePairs: Map<string, RequestParams> = new Map()
+  // reconnection backoff base (ms); jitter added per attempt
+  private lastConnectionAttempt = 0
+  private reconnectDelay = 0
```
mxiao-cll: reconnectDelay!: number
lastConnectionAttempt!: number

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,381 @@
...
+              .on('error', () => resolve(''))
+          })
+          rawErrBody = text
+          try {
+            const j = JSON.parse(text)
+            errDetail = `Provider error: ${JSON.stringify(j)}`
+          } catch {
+            logger.error(`Provider error body (not JSON): ${rawErrBody}`)
+          }
+        } catch {
```
mxiao-cll: Just not try catch if you are not planning to do anything in catch

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: I get that part but why are we doing 401 -> 401 400 -> 400?

packages/sources/coinpaprika-state/src/transport/sse.ts
```diff
@@ -0,0 +1,56 @@
...
+ * - Supports multi-line "data:" blocks
+ * - Ignores comment/heartbeat lines starting with ':'
+ */
+export class SSEParser {
+  private buffer = ''
+  private dataLines: string[] = []
+  private currentEvent: string | null = null
+  private readonly defaultEvent: string
+
+  constructor(defaultEvent = 't_s') {
```
mxiao-cll: Use var

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    process.env.API_KEY = 'TEST-KEY'
+    process.env.API_ENDPOINT = 'http://localhost:1234/stream'
+    process.env.BACKGROUND_EXECUTE_MS = '100'
+
+    const adapter = (await import('../../src')).adapter
+    adapter.rateLimiting = undefined
+    testAdapter = await TestAdapter.startWithMockedCache(adapter, {
+      testAdapter: {} as TestAdapter<never>,
+    })
+    nock.enableNetConnect((host) => host.includes('127.0.0.1') || host.includes('localhost'))
```
mxiao-cll: Why do we need this?

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    process.env.EA_LOG_LEVEL = 'silent'
+    process.env.PINO_LOG_LEVEL = 'silent'
+    process.env.METRICS_ENABLED = 'false'
+    process.env.RETRY = '0'
+
+    nock.disableNetConnect()
+    oldEnv = JSON.parse(JSON.stringify(process.env))
+    process.env.API_KEY = 'TEST-KEY'
+    process.env.API_ENDPOINT = 'http://localhost:1234/stream'
+    process.env.BACKGROUND_EXECUTE_MS = '100'
```
mxiao-cll, line 36: You can set this to 0 in test

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: I don't quite get this

packages/sources/coinpaprika-state/src/endpoint/coinpaprika-state.ts
```diff
@@ -0,0 +1,46 @@
...
+      result: number
+      timestamp: number
+    }
+  }
+  Settings: typeof config.settings
+}
+
+export const endpoint = new AdapterEndpoint({
+  name: 'coinpaprika-state',
+  aliases: ['state'],
```
mmcallister-cll: this needs to have a price alias to work alongside other non-dex EAs properly -- can very likely use a `PriceEndpoint` with `priceEndpointInputParametersDefinition` input params and `SingleNumberResultResponse` response, which will put the timestamps object up one level

.changeset/beige-timers-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/coinpaprika-state-adapter': minor
```
mmcallister-cll:
```suggestion
'@chainlink/coinpaprika-state-adapter': major
```

paul-chain: fixed

packages/sources/coinpaprika-state/src/endpoint/coinpaprika-state.ts
paul-chain: fixed!

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
paul-chain: fixed

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
paul-chain: this is needed to allow localhost connections for TestAdapter internal communication

packages/sources/coinpaprika-state/src/transport/sse.ts
paul-chain: fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: fixed

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
mxiao-cll: None of our other tests have it though

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mxiao-cll: 429?

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    errSpy = jest.spyOn(process.stderr, 'write').mockImplementation(() => true)
+
+    process.env.LOG_LEVEL = 'silent'
+    process.env.EA_LOG_LEVEL = 'silent'
+    process.env.PINO_LOG_LEVEL = 'silent'
+    process.env.METRICS_ENABLED = 'false'
+    process.env.RETRY = '0'
+
+    nock.disableNetConnect()
+    nock.enableNetConnect((host) => host.includes('127.0.0.1') || host.includes('localhost'))
```
mxiao-cll: this is not addressed yet, none of our test requires enabling actual traffic like this

paul-chain: I see, removed and fixed

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
paul-chain: make sense, fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: updated again

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: ah updated

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: actually you are right, fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,366 @@
...
+      this.currentAbortController = null
+      this.isConnected = false
+      logger.error(`Failed to create SSE connection: ${error}`)
+      return
+    }
+  }
+
+  private async handleStreamData(param: RequestParams, streamData: CoinpaprikaStreamData) {
+    // coerce string -> number
+    const toNum = (value: string | number | undefined | null): number => {
```
mmcallister-cll: arguably this should be a separate testable function in a utils.ts file

paul-chain: alright fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+        providerIndicatedTimeUnixMs: blockTime * 1000,
+      },
+    }
+    logger.debug(`tick ${param.base}/${param.quote}=${statePrice} t=${blockTime}`)
+    logger.trace(JSON.stringify(response))
+
+    await this.responseCache.write(this.name, [{ params: param, response }])
+  }
+
+  async _handleRequest(
```
mmcallister-cll: can this ever be called?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+    RequestBody: Array<{ base: string; quote: string }>
+    ResponseBody: CoinpaprikaStreamData
+  }
+}
+
+/**
+ * Single-connection SSE transport that batches all pairs into one POST and streams state_price ticks into the cache.
+ *
+ * */
+export class CoinpaprikaStateTransport extends SubscriptionTransport<TransportTypes> {
```
mmcallister-cll: Just wondering, did you try extending `StreamingTransport` and implementing `streamHandler`?
At a glance it seems like that would manage subscriptions for you rather than rewriting `havePairsChanged`.

BTW: Coinmetrics is a different WS-based EA that switches URLs for every new subscription, not sure if there is any inspiration to pull from it: https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/sources/coinmetrics/src/transport/lwba.ts#L62-L75

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+        providerIndicatedTimeUnixMs: undefined,
+      },
+    }
+  }
+
+  getSubscriptionTtlFromConfig(adapterSettings: TransportTypes['Settings']): number {
+    return adapterSettings.WARMUP_SUBSCRIPTION_TTL
+  }
+
+  async close(): Promise<void> {
```
mmcallister-cll: is this used?

packages/sources/coinpaprika-state/src/transport/sse.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Minimal SSE parser that accumulates chunked lines and emits complete events.
+ * - Supports "event:" name (optional; defaults to 't_s')
+ * - Supports multi-line "data:" blocks
+ * - Ignores comment/heartbeat lines starting with ':'
+ */
+export class SSEParser {
```
mmcallister-cll, line 4: is there no library that can do some of this for us?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+      data: {
+        result: statePrice,
+      },
+      timestamps: {
+        providerDataRequestedUnixMs: Date.now(),
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: blockTime * 1000,
+      },
+    }
+    logger.debug(`tick ${param.base}/${param.quote}=${statePrice} t=${blockTime}`)
```
mmcallister-cll: are 2 logs needed or is 1 debug enough?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  private async createSSEConnection(context: EndpointContext<TransportTypes>) {
+    const pairsArray = Array.from(this.activePairs.values()).map((p) => ({
+      base: p.base,
+      quote: p.quote,
+    }))
+
+    logger.debug(`Opening single SSE connection for ${pairsArray.length} pairs`)
+    logger.debug(`Pairs: ${pairsArray.map((p) => `${p.base}/${p.quote}`).join(', ')}`)
```
mmcallister-cll: isn't this a duplicate log in `backgroundHandler`?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+    this.lastConnectionAttempt = Date.now()
+    this.currentAbortController = new AbortController()
+
+    try {
+      const req = this.buildSSERequest(context, pairsArray, this.currentAbortController.signal)
+      const key = `coinpaprika-state/stream:${[...this.activePairs.keys()].join(',')}`
+      const { response } = await this.requester.request<Readable>(key, req)
+      const httpResp = response
+
+      if (httpResp.status !== 200 || !httpResp.data) {
```
mmcallister-cll: maybe extract this section into an `sseConnectionErrorHandler` function for clarity since this is already a long function

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      .persist()
+
+    await new Promise((resolve) => setTimeout(resolve, 300))
+
+    const response = await testAdapter.request({
+      base: 'UNKNOWN_X',
+      quote: 'USD',
+    })
+
+    expect([504, 502]).toContain(response.statusCode)
```
mmcallister-cll: 504 or 502?

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      expect(r2.statusCode).toBe(200)
+      expect(r1.json().result).toBeCloseTo(1.03)
+      expect(r2.json().result).toBeCloseTo(1.02)
+    })
+
+    endSSEStream(stream)
+    scope.done()
+  })
+
+  it('pair-set change triggers reconnect immediately (first stream kept open)', async () => {
```
mmcallister-cll: is the (first stream kept open) part correct here?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+    const pairsArray = Array.from(this.activePairs.values()).map((p) => ({
+      base: p.base,
+      quote: p.quote,
+    }))
+
+    logger.debug(`Opening single SSE connection for ${pairsArray.length} pairs`)
+    logger.debug(`Pairs: ${pairsArray.map((p) => `${p.base}/${p.quote}`).join(', ')}`)
+
+    this.lastConnectionAttempt = Date.now()
+    this.currentAbortController = new AbortController()
```
mmcallister-cll: could investigate whether it's worth including this in a superclass over the requester (something like `AbortableRequester`). Might not be worth the effort but could save some of this complexity for setting/unsetting the AbortController

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,358 @@
...
+        // mark closed immediately to prevent races
+        this.currentAbortController = null
+        this.isConnected = false
+
+        if (!aborted && this.activePairs.size > 0) {
+          const since = Date.now() - this.lastConnectionAttempt
+          const delay = Math.max(this.reconnectDelay - since, 0) + Math.floor(Math.random() * 1000)
+          logger.info(`SSE ended; reconnecting in ${delay} ms...`)
+          await sleep(delay)
+          await this.createSSEConnection(context)
```
mmcallister-cll: should we be doing this here or just let the next backgroundHandler invocation start a new connection?

In fact, could also extract the management of timing these connections into some SSEConnectionManager class or similar

paul-chain: You're right. I make let backgroundHandler handle it.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Good idea but I think for now not priority. will put todo

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
paul-chain: Nope, we close it. Fixed title.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Yes - fallback when no cache yet. Returns 504.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Yes - app shutdown, Lambda cleanup.

packages/sources/coinpaprika-state/src/transport/sse.ts
paul-chain: Yeah, eventsource exists. Current impl works fine but added TODO.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - 1 debug only.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Yep, consolidated.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Done - extracted handleSSEConnectionError().

packages/sources/coinpaprika-state/test/integration/adapter.test.ts
paul-chain: Fixed - specific codes now.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Good point. Would be cleaner. Added TODO.

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
mmcallister-cll: are you sure the status will not be 200 but contain an error message?
asked another way: are we sure `error` and `data` fields are mutually exclusive?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: actually refactored to use StreamingTransport

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: removed this - unused

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: refactored and removed this part

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Updated error handling process to handle that error cases

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: ok split to connection manager and business logic one

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+    method: string
+    headers: Record<string, string>
+    data: PairRequest[]
+    responseType: 'stream'
+    signal: AbortSignal
+    timeout: number
+    validateStatus: () => boolean
+  } {
+    return {
+      url: `${config.apiEndpoint}/stream`,
```
mmcallister-cll, line 124: if you've just added `/stream` to this URL this you'll need to update the corresponding env var config in src/config/index.ts to remove it. Typically we don't override these env var defaults.

paul-chain: True, fixed!

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    // update active pairs
+    this.activePairs = new Map(pairs)
+
+    if (pairs.size === 0) {
+      logger.debug('No pairs, disconnecting')
+      await this.connectionManager.disconnect()
+      return
+    }
+
+    const pairRequests: PairRequest[] = Array.from(pairs.values()).map((p) => ({
```
alejoberardino: This starts from the params list above, then builds a map, then converts back to a list in three different places

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+interface CoinpaprikaStreamData {
+  block_time: string | number
+  base_token_symbol: string
+  quote_symbol: string
+  volume_7d_usd: string | number
+  market_depth_plus_1_usd: string | number
+  market_depth_minus_1_usd: string | number
+  state_price: string | number
+  // Error fields from API
+  error?: string
```
mmcallister-cll: re-looking at the API doc samples, this seems to be mixing up the stream data error event and connection response 400 error fields, right?

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+      const { response } = await this.requester.request<Readable>(key, request)
+
+      if (response.status !== 200) {
+        await callbacks.onConnectionError(response.status)
+        return
+      }
+
+      if (!response.data) {
+        logger.error('No data in HTTP 200 response')
+        await callbacks.onConnectionError(response.status)
```
mmcallister-cll, line 81: According to the API docs, I think this is where you might get a 400 json response with error and details

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    if (eventType !== COINPAPRIKA_STATE_EVENT_TYPE) {
+      logger.debug(`Skipping event type: ${eventType}`)
+      return
+    }
+
+    try {
+      const streamData: CoinpaprikaStreamData = JSON.parse(rawData)
+
+      // Check for API error structure
+      if ('error' in streamData || 'details' in streamData) {
```
mmcallister-cll: example event from API looks like 
```
event: error
data: {"message": "unsupported X-Y asset"}
```
and should already be handled in the above `if (eventType === 'error') {` case

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+      onReconnectNeeded: () => {
+        logger.info('SSE ended; will reconnect on next cycle')
+      },
+    }
+
+    await this.connectionManager.connect(pairRequests, config, callbacks)
+  }
+
+  private async handleConnectionError(status: number): Promise<void> {
+    const mappedStatus = status === 500 ? 502 : status
```
alejoberardino: This should be the status from the adapter and not the provider, so if the provider is sending an errored response most often that would mean a 502 directly unless we can actually determine directly what the problem is and forward accordingly (for example, if from the provider response we can tell that outbound request was improperly formatted due to a bad paramter in the inbound EA request, we can put a 400 here)

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+      await this.responseCache.write(this.name, [{ params, response: errorResponse }])
+    }
+  }
+
+  private async handleStreamData(param: RequestParams, streamData: CoinpaprikaStreamData) {
+    const statePrice = toNumber(streamData.state_price)
+    const blockTime = toNumber(streamData.block_time)
+
+    // guard against bad payloads
+    if (!Number.isFinite(statePrice) || !Number.isFinite(blockTime) || blockTime <= 0) {
```
alejoberardino: The util fn in this case makes the code harder to parse directly, means an extra hop when it can be replaced by builtins (Number()) and a couple checks

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+        result: statePrice,
+      },
+      timestamps: {
+        providerDataRequestedUnixMs: Date.now(),
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: blockTime * 1000,
+      },
+    }
+    logger.debug(`tick ${param.base}/${param.quote}=${statePrice} t=${blockTime}`)
+    logger.trace(JSON.stringify(response))
```
alejoberardino: These will stringify the response always regardless of whether the log is printed due to the selected level or not; if the stream is very heavy with updates, that's a lot of cpu time spent on this. We can either pass the object as-is and not stringify (the printing should be handled in the logger) or remove the log altogether

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+      statusCode: 502,
+      errorMessage: `SSE stream error: ${errorData}`,
+      timestamps: {
+        providerDataRequestedUnixMs: Date.now(),
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: undefined,
+      },
+    }
+
+    for (const params of this.activePairs.values()) {
```
mmcallister-cll: this causes every pair to get overwritten with the error message for a single stream?

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+  quote: string
+}
+
+/**
+ * SSE connection manager for Coinpaprika
+ */
+export class SSEConnectionManager {
+  private isConnected = false
+  private currentAbortController: AbortController | null = null
+  private sseParser: SSEParser | null = null
```
alejoberardino: Is there a case where we'd not have a parser?

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  private async updateConnection(
+    context: EndpointContext<TransportTypes>,
+    pairs: Map<string, RequestParams>,
+  ): Promise<void> {
+    logger.debug(`updateConnection called with ${pairs.size} pairs`)
+
+    // update active pairs
+    this.activePairs = new Map(pairs)
```
alejoberardino: Is this statefulness needed for something in particular? The key that we build later is simply stringifying the same base/quote parameters

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
```diff
@@ -0,0 +1,238 @@
...
+
+      // Check for API error structure
+      if ('error' in streamData || 'details' in streamData) {
+        logger.error(`API error in stream data: ${rawData}`)
+        await this.handleStreamError(rawData)
+        return
+      }
+
+      const pairKey = `${streamData.base_token_symbol.toUpperCase()}/${streamData.quote_symbol.toUpperCase()}`
+      const params = this.activePairs.get(pairKey)
```
alejoberardino: This should not be necessary, we can simply pass the `{base,quote}` here

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+  constructor(requester: Requester, defaultEventType: string) {
+    this.requester = requester
+    this.defaultEventType = defaultEventType
+  }
+
+  get connected(): boolean {
+    return this.isConnected
+  }
+
+  async connect(
```
alejoberardino, line 45: The reasoning behind not using the framework sse transport makes sense, EventSource does not support POST and the provider API doesn't handle multiple assets otherwise. These reconnection branches however might seem simple but in cases like the WebSocket transport they can be very difficult to polish, so I would recommend ample time soaking this EA, testing it with many different requests (good and bad) to make sure the connection is stable before we rely on it in production

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+  private async setupStream(stream: Readable, callbacks: SSEConnectionCallbacks): Promise<void> {
+    let aborted = false
+    this.sseParser = new SSEParser(this.defaultEventType)
+
+    const onData = (chunk: Buffer) => {
+      const raw = chunk.toString('utf8')
+      logger.debug(`Raw SSE message received:\n${raw}`)
+
+      if (this.sseParser) {
+        this.sseParser.push(raw, (eventType, data) => {
```
alejoberardino: Passing the event handler for each chunk feels weird, it will always be the same and not invoked per-chunk but at the end of a complete event. We should pass it in the parser constructor instead

packages/sources/coinpaprika-state/src/utils.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Utility functions for coinpaprika-state adapter
+ */
+
+/**
+ * Coerces a value to a number, handling string conversion and null/undefined cases
+ * @param value - The value to convert to a number
+ * @returns A finite number, or NaN if conversion fails
+ */
+export function toNumber(value: string | number | undefined | null): number {
```
alejoberardino: This should not be necessary, there's a comment above

alejoberardino (changes requested):

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - replaced custom toNumber()

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - separated CoinpaprikaStreamData interface

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Correct - removed redundant error field checks

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - stream errors now only log

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - proper 4xx/5xx mapping logic

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - removed stringify,pass objects directly

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
paul-chain: Fixed - moved event handler to constructor, get rid off null checks

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed - simplified from Map to array + lookup Map

packages/sources/coinpaprika-state/src/transport/coinpaprika-state.ts
paul-chain: Fixed!

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
paul-chain: Fixed - moved SSE event handler to constructor

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
paul-chain: Noted - added proper async error handling

packages/sources/coinpaprika-state/src/utils.ts
paul-chain: Fixed - removed utility function

packages/sources/coinpaprika-state/src/transport/sse-connection-manager.ts
paul-chain: Agreed - this is already properly handled in SSE connection manager's response validation.

PR #4059, opened 2025-09-26
Feature extend reserve endpoint for lax ripcording

fmunshi:
## Description

......

## Changes

- Updates the the reserve endpoint to take a laxRipcord option for data streams and always return ripcordAsInt also 

## Steps to Test

```
curl -s --header "Content-Type: application/json" --request POST --data '{ "data": {"endpoint": "reserve", "client": "backed-tslax-kyln3p"} }' localhost:8080  | jq 
{
  "result": 47327,
  "data": {
    "result": 47327,
    "ripcord": false,
    "ripcordAsInt": 0,
    "totalReserve": 47327,
    "totalToken": 46366.19985541
  },
  "timestamps": {
    "providerDataRequestedUnixMs": 1758915236519,
    "providerDataReceivedUnixMs": 1758915236960,
    "providerIndicatedTimeUnixMs": 1758913525602
  },
  "statusCode": 200,
  "meta": {
    "adapterName": "THE_NETWORK_FIRM",
    "metrics": {
      "feedId": "{\"client\":\"backed-tslax-kyln3p\",\"laxRipcord\":false}"
    }
  }
```

## Quality Assurance

### DIFF
--- a/packages/sources/the-network-firm/src/endpoint/reserve.ts
+++ b/packages/sources/the-network-firm/src/endpoint/reserve.ts
@@ -32,7 +32,11 @@ export const inputParameters = new InputParameters(
 
 export type BaseEndpointTypes = {
   Parameters: typeof inputParameters.definition
-  Response: PoRProviderResponse
+  Response: PoRProviderResponse & {
+    ripcordAsInt?: number
+    totalReserve?: number
+    totalToken?: number
+  }
   Settings: typeof config.settings
 }
 
--- a/packages/sources/the-network-firm/src/endpoint/reserve.ts
+++ b/packages/sources/the-network-firm/src/endpoint/reserve.ts
@@ -14,18 +14,18 @@ export const inputParameters = new InputParameters(
       required: true,
       description: 'The name of the TNF client to consume from',
     },
-    laxRipcord: {
+    noErrorOnRipcord: {
       type: 'boolean',
       required: false,
       default: false,
       description:
-        'Lax ripcord handling, return 200 on ripcord when laxRipcord is true, return 502 with ripcord details if laxRipcord is false or unset',
+        'Lax ripcord handling, return 200 on ripcord when noErrorOnRipcord is true, return 502 with ripcord details if noErrorOnRipcord is false or unset',
     },
   },
   [
     {
       client: 'acme',
-      laxRipcord: false,
+      noErrorOnRipcord: false,
     },
   ],
 )
--- a/packages/sources/the-network-firm/src/endpoint/reserve.ts
+++ b/packages/sources/the-network-firm/src/endpoint/reserve.ts
@@ -18,7 +18,8 @@ export const inputParameters = new InputParameters(
       type: 'boolean',
       required: false,
       default: false,
-      description: 'Lax ripcord handling, return 200 on ripcord when laxRipcord is true, return 502 with ripcord details if laxRipcord is false or unset',
+      description:
+        'Lax ripcord handling, return 200 on ripcord when laxRipcord is true, return 502 with ripcord details if laxRipcord is false or unset',
     },
   },
   [
--- a/packages/sources/the-network-firm/src/endpoint/reserve.ts
+++ b/packages/sources/the-network-firm/src/endpoint/reserve.ts
@@ -14,10 +14,17 @@ export const inputParameters = new InputParameters(
       required: true,
       description: 'The name of the TNF client to consume from',
     },
+    laxRipcord: {
+      type: 'boolean',
+      required: false,
+      default: false,
+      description: 'Lax ripcord handling, return 200 on ripcord when laxRipcord is true, return 502 with ripcord details if laxRipcord is false or unset',
+    },
   },
   [
     {
       client: 'acme',
+      laxRipcord: false,
     },
   ],
 )
--- a/packages/sources/the-network-firm/src/transport/reserve.ts
+++ b/packages/sources/the-network-firm/src/transport/reserve.ts
@@ -53,7 +53,7 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
   },
   parseResponse: (params, response) => {
     return params.map((param) => {
-      const laxRipcord = param.laxRipcord
+      const noErrorOnRipcord = param.noErrorOnRipcord
 
       const ripcord =
         response.data.ripcord || response.data.ripcord.toString().toLowerCase() === 'true'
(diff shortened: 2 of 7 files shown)


packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +54,52 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
         response: {
-          result: result.totalReserve,
+          result: totalReserve,
           data: {
-            result: result.totalReserve,
+            result: totalReserve,
             ripcord: response.data.ripcord,
+            ripcordAsInt,
+            totalReserve,
+            totalToken,
```
fmunshi, line 102: i added both of these as named responses to make it easier to understand whats going on but i could remove it - i dont think we'll need them - if you think i should remove this let me know

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -18,6 +19,8 @@ export type HttpTransportTypes = BaseEndpointTypes & {
   }
 }
 
+const logger = makeLogger('ReserveStreamsHTTPTransport')
```
mmcallister-cll: maybe don't name this streams specific

packages/sources/the-network-firm/src/endpoint/reserve.ts
```diff
@@ -14,10 +14,17 @@ export const inputParameters = new InputParameters(
       required: true,
       description: 'The name of the TNF client to consume from',
     },
+    laxRipcord: {
```
mxiao-cll: more like ignoreRipcord?

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
+
+        // If laxRipcord is false and ripcord is true return 502
+        if (!laxRipcord) {
+          const message = `Ripcord indicator true. Details: ${ripcordDetails}`
+          return {
+            params: param,
+            response: {
+              errorMessage: message,
+              ripcord: response.data.ripcord,
+              ripcordAsInt,
```
mxiao-cll, line 74: Why do we need this?

packages/sources/the-network-firm/src/transport/reserve.ts
fmunshi: woops i copied this over from when i had a separate endpoint here

packages/sources/the-network-firm/src/endpoint/reserve.ts
fmunshi: not quite - lax seems the most accurate to me 

since we dont ignore the ripcord status - we return it and let the client handle it however they choose to

packages/sources/the-network-firm/src/transport/reserve.ts
fmunshi: for consumption by data streams - we did this on other endpoints such as `uranium` as well

packages/sources/the-network-firm/src/endpoint/reserve.ts
mxiao-cll: errorOnRipcord?

fmunshi: @mxiao-cll - I wanted to keep the default behaviour to error - so this flag is to disable the error 

could word it - noErrorOnRipcord - but i think laxRipcord seems the most succinct way to describe that

mxiao-cll: noErrorOnRipcord is better IMO

fmunshi: @mxiao-cll updated

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
   },
   parseResponse: (params, response) => {
     return params.map((param) => {
+      const noErrorOnRipcord = param.noErrorOnRipcord
+
       const ripcord =
         response.data.ripcord || response.data.ripcord.toString().toLowerCase() === 'true'
+      const ripcordAsInt = ripcord ? 1 : 0
```
mmcallister-cll, line 60: i think this could be `Number(ripcord)`?

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
       return {
         params: param,
         response: {
-          result: result.totalReserve,
+          result,
           data: {
-            result: result.totalReserve,
+            result,
             ripcord: response.data.ripcord,
+            ripcordAsInt,
```
mmcallister-cll, line 98: the response type in the endpoint file might not be clear anymore 🤔

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
+              statusCode: 502,
+              timestamps: {
+                providerIndicatedTimeUnixMs: new Date(response.data.timestamp).getTime(),
+              },
             },
-          },
+          }
         }
+
+        logger.debug(`Ripcord indicator true. Details: ${ripcordDetails}`)
```
mmcallister-cll: won't this duplicate the error log when noErrorOnRipcord is false? maybe stick in an else instead?

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
+        const ripcordDetails = response.data.ripcordDetails.join(', ')
+
+        // If noErrorOnRipcord is false and ripcord is true return 502
+        if (!noErrorOnRipcord) {
+          const message = `Ripcord indicator true. Details: ${ripcordDetails}`
+          return {
+            params: param,
+            response: {
+              errorMessage: message,
+              ripcord: response.data.ripcord,
```
mmcallister-cll: we're already extracting ripcord above on L58, can we just use that here and in the successful response?

packages/sources/the-network-firm/src/transport/reserve.ts
```diff
@@ -50,34 +53,51 @@ export const httpTransport = new HttpTransport<HttpTransportTypes>({
...
-            errorMessage: message,
-            ripcord: response.data.ripcord,
-            ripcordDetails: response.data.ripcordDetails.join(', '),
-            statusCode: 502,
-            timestamps: {
-              providerIndicatedTimeUnixMs: new Date(response.data.timestamp).getTime(),
+        const ripcordDetails = response.data.ripcordDetails.join(', ')
+
+        // If noErrorOnRipcord is false and ripcord is true return 502
+        if (!noErrorOnRipcord) {
```
mmcallister-cll, line 68: i'd probably move this into a single `if (ripcord && !noErrorOnRipcord)` but this works too

mmcallister-cll (approved):

PR #4060, opened 2025-09-29
Llama support 18 decimals

mxiao-cll:
Previously we only support up-to 15 decimals due to use of numbers, here we try to use bigint more to support up to 18 decimals


packages/composites/llama-guard/src/transport/ea.ts
```diff
@@ -48,15 +48,19 @@ export const getEAUrl = (ea: string) => {
...
   if (typeof response === 'string') {
-    const num = Number(response)
-    if (!isNaN(num)) {
-      return num
+    if (!isNaN(Number(response))) {
+      if (response.indexOf('.') === -1) {
+        // Response may already be scaled, use BigInt to avoid overflow
+        return BigInt(response).toString()
+      } else {
+        return Number(response).toString()
```
dskloetc: Should we use Decimal here to support 18 decimals?

packages/composites/llama-guard/src/transport/nav.ts
```diff
@@ -93,6 +93,16 @@ export const getNav = async (
 
 const mulBigInt = (value: bigint, multiplier: number, decimals: number) => {
   const scale = 10n ** BigInt(decimals)
-  const scaledMultiplier = BigInt(Math.floor(multiplier * Number(scale)))
+  const scaledMultiplier = parseUnits(trimDecimals(multiplier.toString(), decimals), decimals)
```
dskloetc, line 96: Why do rounding before we reach the final result? Doesn't that just introduce unnecessary rounding errors?

packages/composites/llama-guard/test/unit/ea.test.ts
```diff
@@ -46,16 +46,24 @@ describe('ea.ts', () => {
       const requester = { request: jest.fn() } as any
 
       requester.request.mockResolvedValueOnce({ response: { data: { result: 10 } } })
-      await expect(getRawNav('ea', '{}', requester)).resolves.toEqual(10)
+      await expect(getRawNav('ea', '{}', requester)).resolves.toEqual('10')
+
+      requester.request.mockResolvedValueOnce({ response: { data: { result: '1.234456123123' } } })
+      await expect(getRawNav('ea', '{}', requester)).resolves.toEqual('1.234456123123')
```
dskloetc: Use a constant so it's easy to see that the value is the same as on the previous line.

packages/composites/llama-guard/src/transport/ea.ts
mxiao-cll: The output here is from another EA so we should already have this covered, in a sense that if the output is 18 decimals the EA should already output in bigint

packages/composites/llama-guard/src/transport/nav.ts
mxiao-cll: If we don't round here `parseUnits` will throw an error, it expects that the decimals does not exceed the decimal var.

For example, if you have 1.001 and decimals 2, then it will throw an error

packages/composites/llama-guard/src/transport/ea.ts
dskloetc: Should, yes. But this feels like leaving a trap that will one day we really hard do debug.
If it was outputting a number, then it would output it as a `number` rather than a `string`, wouldn't it?
So if it's a `string`, I think we should make sure not to lose precision.
Is it hard to use a `Decimal` here?

dskloetc: Or would it be possible to just throw if it's a string that contains a `'.'`?
That shouldn't happen anyway, right?

mxiao-cll: I guess my point is, there is no way to produce a number with 18 decimals correctly in the first place, if the EA returns a value like "1.12312312312312312321" then the number itself can not be accurate, thus there is no handling here that can make this right.

It is possible that we have an EA that returns "1.12" though, so doing a Number here is safe for that

dskloetc: How can you say it's not possible? There are other EAs that use big `Decimal`s.
Or it could use `bigint` and then insert a `'.'` somewhere, similar to how manipulate the number as a string in `trimDecimals`.

mxiao-cll: Yes, that make sense

dskloetc (approved):

PR #4061, opened 2025-09-30
creating new allium-state ea for OPDATA-4235

jordanbounds:
Closes #OPDATA-4235:
https://smartcontract-it.atlassian.net/browse/OPDATA-4235

Description
Creates allium-state EA that provides real-time state price data via WebSocket connection. The adapter subscribes to aggregated state price updates and returns normalized price data for supported cryptocurrency pairs.

Changes
New Adapter: Created allium-state external adapter with WebSocket transport
Price Endpoint: Implemented price endpoint supporting base/quote currency pairs e.g (CBBTC/USD, BUSD/USD)
Testing: Added integration tests with mocked WebSocket server

Steps to Test:
```
curl -X POST http://localhost:8080 \
  -d '{
    "id": "1",
    "data": {
      "base": "CBBTC",
      "quote": "USD"
    }
  }' | jq
```

`yarn test allium-state`


packages/sources/allium-state/src/transport/price.ts
```diff
@@ -0,0 +1,80 @@
...
+      return [
+        {
+          params: { base, quote },
+          response: {
+            data: {
+              result: state_price,
+            },
+            result: state_price,
+            timestamps: {
+              providerIndicatedTimeUnixMs: timestamp,
```
dskloetc: In the case where this is assigned `Date.now()` on line 44, it's not actually data provider-indicated.
It looks like this is allowed to be `undefined`:
https://github.com/smartcontractkit/ea-framework-js/blob/85d96c9dc2fcfd2c19af4428cc38b56183087e24/src/util/types.ts#L175

We can assign `Date.now()` to `providerDataReceivedUnixMs`.

packages/sources/allium-state/src/transport/price.ts
```diff
@@ -0,0 +1,80 @@
...
+  }),
+
+  handlers: {
+    message(message) {
+      if (message?.type !== 'aggregated_state_price_update' || !message?.data?.state_price) {
+        return []
+      }
+
+      const { base, quote, state_price } = message.data
+      const timestamp = message.timestamp ? new Date(message.timestamp).getTime() : Date.now()
```
dskloetc: Do we need this conditional?
Is the API response not consistent in whether `message.timestamp` is present or not?

packages/sources/allium-state/src/transport/price.ts
jordanbounds: fixed.

packages/sources/allium-state/src/transport/price.ts
jordanbounds: Good call out, there is no clear data from their docs or otherwise observed that they would not include a timestamp. fixed.

.changeset/thick-insects-rule.md
dskloetc, line 1: You only need 1 changeset per PR. Please remove these additional changesets.

dskloetc (approved):

.changeset/thick-insects-rule.md
jordanbounds: @dskloetc Gotcha this is my first EA build - should I just keep the original and delete this one and the next?

dskloetc: Yes, the original still describes the PR.

PR #4063, opened 2025-10-01
Call superstate repeatedly

mxiao-cll:
Currently we make a single call between 9:08 and 12:00
If the nav update comes after we call, we will miss this and wait till the next day to update.
We currently have to ask all the NOPs to bounce EA and bounce redis in order to pick it up on the same day.

Here, we make a change that we would run periodic calls between 9:08 and 12:00

### DIFF
--- a/packages/sources/superstate/test/unit/nav.test.ts
+++ b/packages/sources/superstate/test/unit/nav.test.ts
@@ -1,34 +1,25 @@
 import { LoggerFactoryProvider } from '@chainlink/external-adapter-framework/util'
 import { makeStub } from '@chainlink/external-adapter-framework/util/testing-utils'
 
-jest.mock('../../src/transport/utils', () => ({
-  ...jest.requireActual('../../src/transport/utils'),
-  isInTimeRange: jest.fn(() => true),
-}))
-
 import { NavTransport } from '../../src/transport/nav'
 
 LoggerFactoryProvider.set()
 
 describe('navTransport', () => {
-  beforeEach(() => {
-    jest.useFakeTimers()
-  })
-
   afterEach(() => {
     jest.useRealTimers()
   })
 
   it('runScheduler should trigger periodically', async () => {
+    jest.useFakeTimers({ now: new Date('2024-01-01T10:00:00.000-05:00') }) // 9:08 <-> 12:00
+
     const transport = new NavTransport()
 
     const mockExecute = jest.spyOn(transport, 'execute').mockResolvedValue(undefined)
-
     const mockDependencies = makeStub('dependencies', {
       responseCache: { write: jest.fn() },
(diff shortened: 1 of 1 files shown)


packages/sources/superstate/src/transport/nav.ts
```diff
@@ -80,20 +80,22 @@ class NavTransport implements Transport<BaseEndpointTypes> {
...
-    rule.minute = startTimeSegments[1]
-    rule.second = startTimeSegments[2]
-    rule.tz = TZ
-
-    schedule.scheduleJob(rule, () => {
-      logger.info(
-        `Scheduled execution started at ${Date.now()}. FundsMap - ${[...this.fundsMap].join(',')}`,
-      )
-      ;[...this.fundsMap].map(async (entry) => this.execute(entry[1][0], entry[1][1]))
+    schedule.scheduleJob('*/10 * * * *', () => {
```
dskloetc: We will continue to execute even after we've already found the new value for the day, correct?

.changeset/popular-pans-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/superstate-adapter': minor
+---
+
+Make regular calls to DP during open period. If you were ask by CLL to run this adapter using local cache, you can now safely switch back to redis.
```
dskloetc, line 5: I don't see how this change relates to using local cache vs. redis. Can you explain that to me?

packages/sources/superstate/src/transport/nav.ts
dskloetc, line 1: Please add a test that enforces the new behavior.

packages/sources/superstate/src/transport/nav.ts
karen-stepanyan: also makes sense to make this value of 10 configurable

mxiao-cll: Correct, moved to env var

.changeset/popular-pans-sleep.md
mxiao-cll: Previously to force an update I've asked NOP to move to local cache so that we can simply re-start EA to force an update

packages/sources/superstate/test/unit/nav.test.ts
```diff
@@ -0,0 +1,46 @@
...
+describe('navTransport', () => {
+  beforeEach(() => {
+    jest.useFakeTimers()
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
+  })
+
+  it('runScheduler should trigger periodically', async () => {
```
dskloetc, line 13: Also test that it does not execute outside trading hours?

dskloetc (approved):

PR #4064, opened 2025-10-02
Llama metrics

mxiao-cll:
## Description

Send a metric when results go out-of-bounds, this call should not impact normal flow.

## Quality Assurance

### DIFF
--- a/.changeset/rich-fishes-travel.md
+++ b/.changeset/rich-fishes-travel.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/asseto-finance-adapter': major
+---
+
+asseto-finance initial release
--- a/.changeset/selfish-apes-exist.md
+++ b/.changeset/selfish-apes-exist.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/allium-state-adapter': major
+---
+
+allium-state initial release
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -211,6 +211,10 @@ const RAW_RUNTIME_STATE =
       "reference": "workspace:packages/sources/aleno"\
     },\
     {\
+      "name": "@chainlink/allium-state-adapter",\
+      "reference": "workspace:packages/sources/allium-state"\
+    },\
+    {\
       "name": "@chainlink/alphachain-adapter",\
       "reference": "workspace:packages/sources/alphachain"\
     },\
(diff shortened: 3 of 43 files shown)


packages/composites/llama-guard/src/transport/metrics.ts
```diff
@@ -0,0 +1,51 @@
...
+  apiKey: string,
+  asset: string,
+  response: AdapterResponse<BaseEndpointTypes['Response']>,
+  requester: Requester,
+) => {
+  if (response.data?.isBounded === true) {
+    return
+  }
+  if (apiKey.length === 0) {
+    logger.warn('Missing LLAMA_RISK_API_KEY, skip sending metrics')
```
mmcallister-cll: qq, is the expectation that all NOPs send this or just an internal instance or llama themselves?

mxiao-cll: Actually good question I guess, we can sort of control this by sharing/not sharing API keys

mmcallister-cll: aha i figured that's why this check was added, nice side effect even if that wasn't the original intent. If we're going to use this as a feature flag though we should probably make this a debug log

packages/composites/llama-guard/src/config/index.ts
```diff
@@ -11,6 +11,17 @@ export const config = new AdapterConfig({
...
+    description: 'API endpoint for Llama Risk',
+    type: 'string',
+    default: 'https://api.llamarisk.com/protocols/horizon/alert/',
+  },
+  LLAMA_RISK_API_KEY: {
+    description: 'API key for Llama Risk',
+    type: 'string',
+    default: '',
+    sensitive: true,
+  },
```
zekiblue, line 24: We would want to keep this as a single sensitive variable and not to expose the default URL. Security crawlers can pick up the default and can exhaust our endpoint.

zekiblue (commented):
Thanks for this swift handling

packages/composites/llama-guard/src/config/index.ts
mxiao-cll: Updated, can you delete this comment as well to remove trace?

Side note: do you want all the NOPs to hit this endpoint?

zekiblue (commented):
Looks good to me, thanks

PR #4077, opened 2025-10-06
Refactors GM-token to support Botanix and dynamic fetching of tokens

mohamed-mehany:
## Description
This PR refactors the EA to enable support of Botanix specific tokens.
It also adds dynamic fetching of tokens, rather than static / hardcoded tokens.

......

## Changes

- Adding support to Botanix specific tokens
- Dynamic fetching of tokens using GMX APIs.

## Steps to Test

1. Unit tests
2. Integration tests
3. Running the EA locally

## Quality Assurance

### DIFF
--- a/packages/composites/gm-token/src/config/index.ts
+++ b/packages/composites/gm-token/src/config/index.ts
@@ -14,13 +14,36 @@ export const config = new AdapterConfig(
       required: true,
       default: 42161,
     },
+    BOTANIX_RPC_URL: {
+      description: 'RPC url of Arbitrum node',
+      type: 'string',
+      required: false,
+    },
+    BOTANIX_CHAIN_ID: {
+      description: 'The chain id to connect to',
+      type: 'number',
+      required: false,
+      default: 42161,
+    },
     DATASTORE_CONTRACT_ADDRESS: {
       description: 'Address of Data Store contract',
       type: 'string',
       required: true,
       default: '0xFD70de6b91282D8017aA4E741e9Ae325CAb992d8',
     },
+    BOTANIX_DATASTORE_CONTRACT_ADDRESS: {
+      description: 'Address of Data Store contract',
+      type: 'string',
+      required: true,
+      default: '0xFD70de6b91282D8017aA4E741e9Ae325CAb992d8',
+    },
     READER_CONTRACT_ADDRESS: {
+      description: 'Address of Reader contract',
+      type: 'string',
+      required: true,
+      default: '0xf60becbba223EEA9495Da3f606753867eC10d139',
+    },
+    BOTANIX_READER_CONTRACT_ADDRESS: {
       description: 'Address of Reader contract',
       type: 'string',
       required: true,
--- a/packages/composites/gm-token/src/config/index.ts
+++ b/packages/composites/gm-token/src/config/index.ts
@@ -23,7 +23,7 @@ export const config = new AdapterConfig(
       description: 'The chain id to connect to',
       type: 'number',
       required: false,
-      default: 42161,
+      default: 3637,
     },
     DATASTORE_CONTRACT_ADDRESS: {
       description: 'Address of Data Store contract',
@@ -35,7 +35,7 @@ export const config = new AdapterConfig(
       description: 'Address of Data Store contract',
       type: 'string',
       required: true,
-      default: '0xFD70de6b91282D8017aA4E741e9Ae325CAb992d8',
+      default: '0xA23B81a89Ab9D7D89fF8fc1b5d8508fB75Cc094d',
     },
     READER_CONTRACT_ADDRESS: {
       description: 'Address of Reader contract',
@@ -47,7 +47,25 @@ export const config = new AdapterConfig(
       description: 'Address of Reader contract',
       type: 'string',
       required: true,
-      default: '0xf60becbba223EEA9495Da3f606753867eC10d139',
+      default: '0xa254B60cbB85a92F6151B10E1233639F601f2F0F',
+    },
+    ARBITRUM_TOKENS_INFO_URL: {
+      description: 'URL to token meta data supported by GMX on Arbitrum',
+      type: 'string',
+      required: true,
+      default: 'https://arbitrum-api.gmxinfra.io/tokens',
+    },
+    BOTANIX_TOKENS_INFO_URL: {
+      description: 'URL to token meta data supported by GMX on Botanix',
+      type: 'string',
+      required: true,
+      default: 'https://botanix-api.gmxinfra.io/tokens',
+    },
+    GMX_TOKENS_CACHE_MS: {
+      description: 'TTL in milliseconds for GMX tokens cache',
+      type: 'number',
+      required: true,
+      default: 300_000,
     },
     PNL_FACTOR_TYPE: {
       description:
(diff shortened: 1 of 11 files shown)


packages/composites/gm-token/src/config/index.ts
```diff
@@ -14,18 +14,59 @@ export const config = new AdapterConfig(
       required: true,
       default: 42161,
     },
+    BOTANIX_RPC_URL: {
+      description: 'RPC url of Arbitrum node',
```
mxiao-cll: ?

packages/composites/gm-token/src/config/index.ts
```diff
@@ -14,18 +14,59 @@ export const config = new AdapterConfig(
...
+      type: 'string',
+      required: false,
+    },
+    BOTANIX_CHAIN_ID: {
+      description: 'The chain id to connect to',
+      type: 'number',
+      required: false,
```
mxiao-cll: I think we can remove this

packages/composites/gm-token/src/transport/price.ts
```diff
@@ -224,12 +237,15 @@ export class GmTokenTransport extends SubscriptionTransport<GmTokenTransportType
   }
 
   private unwrapAsset(asset: string) {
```
mxiao-cll: See if we can put this in another file, easier to find

mohamed-mehany: Can put this in the utils.ts file.

packages/composites/gm-token/src/endpoint/price.ts
```diff
@@ -55,26 +60,4 @@ export const endpoint = new AdapterEndpoint({
   name: 'price',
   transport: gmTokenTransport,
   inputParameters,
-  customInputValidation: (req): AdapterInputError | undefined => {
```
mxiao-cll, line 58: Is this no longer needed or we've moved it into transport, the reason here is preferred is if you do this in transport, we will keep looping on bad requests

packages/composites/gm-token/src/transport/token-resolver.ts
```diff
@@ -0,0 +1,48 @@
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
+import { GmTokenTransportTypes } from './price'
+
+export type ChainKey = 'arbitrum' | 'botanix'
```
mxiao-cll: I see this in a couple of places, see if you can use the type from the input param

packages/composites/gm-token/src/config/index.ts
mohamed-mehany: This one is to allow connecting to testnet for debugging / testing.

packages/composites/gm-token/src/transport/token-resolver.ts
```diff
@@ -0,0 +1,48 @@
+import { Requester } from '@chainlink/external-adapter-framework/util/requester'
```
mxiao-cll, line 1: Why are we refreshing this seperately from the transport>?

packages/composites/gm-token/src/config/index.ts
mohamed-mehany: Also already exists for Arbitrum so keeps parity.

mxiao-cll: I mean `required: false,`

mohamed-mehany: Ah, makes sense then.

packages/composites/gm-token/src/endpoint/price.ts
mohamed-mehany: This is difficult to check here because now we're fetching the addresses dynamically (not static as before).

packages/composites/gm-token/src/transport/token-resolver.ts
mohamed-mehany: Sorry didn't get what you mean here.
What this does is that it checks for token addresses from the GMX API, the reason why we need to refresh based on a TTL is that the token list gets updated, so we can't fetch only once because we've had an incident with GLV for example where a new token was added and the EA had to be restarted to cache the new token.

packages/composites/gm-token/src/transport/token-resolver.ts
mohamed-mehany: The input param is not typed currently. 
Do you mean to use that type to set the input param type?

packages/composites/gm-token/src/config/index.ts
mohamed-mehany: Copy pasta :) will fix.

packages/composites/gm-token/src/transport/token-resolver.ts
mxiao-cll: But we are using a subscription transport right? So we should be calling this API every round?

packages/composites/gm-token/src/transport/token-resolver.ts
mxiao-cll:
```
    chain: {
      description: 'Target chain for GM market',
      type: 'string',
      options: ['arbitrum', 'botanix'],
      default: 'arbitrum',
    },
```

If you set to type of this, it will be the same

mohamed-mehany: Fixed, thanks!

packages/composites/gm-token/src/transport/price.ts
mohamed-mehany: Fixed/

packages/composites/gm-token/src/config/index.ts
mohamed-mehany: Removed.

packages/composites/gm-token/src/transport/token-resolver.ts
mohamed-mehany: As discussed offline, we agreed to remove the Transport level caching as this is cached anyways at the requester level.

packages/composites/gm-token/src/transport/price.ts
```diff
@@ -32,10 +25,11 @@ export class GmTokenTransport extends SubscriptionTransport<GmTokenTransportType
   name!: string
   responseCache!: ResponseCache<GmTokenTransportTypes>
   requester!: Requester
-  provider!: ethers.providers.JsonRpcProvider
-  readerContract!: ethers.Contract
   abiEncoder!: utils.AbiCoder
   settings!: GmTokenTransportTypes['Settings']
+  tokenResolver!: TokenResolver
+  private providers: Partial<Record<ChainKey, ethers.providers.JsonRpcProvider>> = {}
+  private readers: Partial<Record<ChainKey, ethers.Contract>> = {}
```
mxiao-cll, line 32: Why are these partial?

mohamed-mehany: Because Botanix is optional for backward compatibility.

mxiao-cll: This makes no sense, if Botanix is optional then there would just be 1 entry in the map, we don't need Partial

mohamed-mehany: Aah I get what you mean. 
The ChainKey type was a union before the last round of feedback like: ChainKey = "arb" | "bot" etc..

In this case I think Record (without Partial) enforces a value for every key in the union, hence why Partial was required. When we changed this in the feedback, now not sure if this is still needed because it depends on how inputParameter.validated["chain"] is typed; if it's a union, I believe Partial would still be needed, if not then we can remove it. Will check and update.

PR #4079, opened 2025-10-06
OPDATA-4169, OPDATA-4237: Add anchor-data endpoint to solana-functions

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-4169
https://smartcontract-it.atlassian.net/browse/OPDATA-4237

## Description

Solana does not have a standard way to call view functions like Ethereum does.

In Solana, accounts have account data that can be read. But the meaning of this account data depends on the program that created it. There is not a single standard way to interpret the account data. But a common way is using an Anchor IDL which is created based on the program.

This PR adds an `anchor-data` endpoint to that uses Anchor IDL to read a specific field from a specific state account.

While there appear to be ways to store program IDL on the blockchain they don't seem to be universally used. So in this the IDL for the Frametric Liquid Restaking program is hard coded and used when the fragmetric program ID is recognized.

Other IDLs should be added to the map as needed.

## Changes

1. Added a new `anchor-data` endpoint that reads a field from a state account, given the account's address, the field's name and name of the account type the field belongs to.
2. Add Fragmetric program ID downloaded from https://explorer.solana.com/address/fragnAis7Bp6FTsMoa6YcH8UffhEw43Ph79qAiK3iF3/idl
3. Added the result of
```
ACCOUNT="3TK9fNePM4qdKC4dwvDe8Bamv14prDqdVfuANxPeiryb"

curl -X POST https://rpcs.cldev.sh/solana/mainnet -H 'Content-Type: application/json' -d '{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "getAccountInfo",
  "params": [
    "'$ACCOUNT'",
    {
      "encoding": "base64"
    }
  ]
}' | jq
```
as `packages/sources/solana-functions/test/fixtures/fragmetric-account-data-2025-10-06.json` to be used as test fixture.

## Steps to Test

1. Run unit tests: `yarn test packages/sources/solana-functions/test/unit/anchor-data.test.ts`
2. Run integration test: `yarn test packages/sources/solana-functions/test/integration/anchor-data.test.ts`
3. Test manually:
```
    "data": {
  "endpoint": "anchor-data",
  "stateAccountAddress": "3TK9fNePM4qdKC4dwvDe8Bamv14prDqdVfuANxPeiryb",
  "account": "FundAccount",
  "field": "one_receipt_token_as_sol"
}
  }'

{
  "data": {
    "result": "1079420728"
  },
  "statusCode": 200,
  "result": "1079420728",
  "timestamps": {
    "providerDataRequestedUnixMs": 1759754831251,
    "providerDataReceivedUnixMs": 1759754836326
  },
  "meta": {
    "adapterName": "SOLANA_FUNCTIONS",
    "metrics": {
      "feedId": "{\"stateAccountAddress\":\"3tk9fnepm4qdkc4dwvde8bamv14prdqdvfuanxpeiryb\",\"account\":\"fundaccount\",\"field\":\"one_receipt_token_as_sol\"}"
    }
  }
}
```

## Quality Assurance


packages/sources/solana-functions/src/transport/anchor-data.ts
```diff
@@ -0,0 +1,114 @@
...
+import { type Rpc, type SolanaRpcApi } from '@solana/rpc'
+import { BaseEndpointTypes, inputParameters } from '../endpoint/anchor-data'
+import * as fragmetricLiquidRestakingProgramIdl from '../idl/fragmetric_liquid_restaking.json'
+import { SolanaRpcFactory } from '../shared/solana-rpc-factory'
+
+const logger = makeLogger('AnchorDataTransport')
+
+const programToIdlMap: Record<string, Idl> = {
+  fragnAis7Bp6FTsMoa6YcH8UffhEw43Ph79qAiK3iF3: fragmetricLiquidRestakingProgramIdl as Idl,
+}
```
fmunshi, line 18: so every new account we want to query will require an addition to this map + adding the IDL to the package

packages/sources/solana-functions/src/transport/anchor-data.ts
```diff
@@ -0,0 +1,114 @@
...
+    const resp = await this.rpc
+      .getAccountInfo(params.stateAccountAddress as Address, { encoding })
+      .send()
+    const programAddress = resp.value?.owner
+    if (!programAddress) {
+      throw new AdapterInputError({
+        message: `No program address found for state account '${params.stateAccountAddress}'`,
+        statusCode: 500,
+      })
+    }
```
fmunshi, line 79: I think we could use `SolanaAccountReader` class here - as this seems to be re-implementing that logic

packages/sources/solana-functions/src/transport/anchor-data.ts
dskloetc: Not every account but every program.
If we want to read a new account of an already supported program, it should work without changes.

packages/sources/solana-functions/src/transport/anchor-data.ts
dskloetc: I looked at `SolanaAccountReader` but it seems to be specific for the case where the state account address is derived from the program address and seeds.
Did I see that wrong?

fmunshi: Theres two functions on the solana account reader
`fetchAccountInformationByAddressAndSeeds` & `fetchAccountInformation`
you can pass the address into `fetchAccountInformation`  without the seeds just like your PR does

packages/sources/solana-functions/src/transport/anchor-data.ts
fmunshi: ah yea - thats what i meant - sounds good

packages/sources/solana-functions/src/transport/anchor-data.ts
dskloetc: Unfortunately I can't use that either.
The IDL I use depends on the program, which is determined by the `owner` field on the account data.
So I can't pass in the IDL before fetching the data.

fmunshi: discussed offline

PR #4080, opened 2025-10-07
OPDATA-3087: Add sanctum-infinity endpoint to solana-functions

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-3087

## Description

There is not a single standard way to interpret the account data. A common way is using an `BufferLayout` specific to a program. The Sanctum Controller program and the standard Solana token program both use this method.

This PR adds a utility to read a field from an account that owned by such a program.
And it uses this utility to implement the `sanctum-infinity` endpoint which calculates the token price by dividing the total value by the total supply, which are both read using the introduced utility.

## Changes

1. 

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance


packages/sources/solana-functions/src/shared/buffer-layout-accounts.ts
```diff
@@ -0,0 +1,87 @@
...
+  is_rebalancing: number
+  padding: number[]
+  admin: Uint8Array
+  rebalance_authority: Uint8Array
+  protocol_fee_beneficiary: Uint8Array
+  pricing_program: Uint8Array
+  lp_token_mint: Uint8Array
+}
+
+const SanctumPoolStateLayout = BufferLayout.struct<SanctumPoolState>([
```
fmunshi, line 22: wouldnt this make more sense to live in the endpoint and be passed as an argument?

seems endpoint specific

dskloetc: I imagine there could also be an endpoint to get individual values from various programs.
Similar to the `anchor-data` endpoint.
And layouts would just be added to this map as needed.
The specific endpoint is for dividing the 2 values.
But this layout is not specific to that division.
Does this make sense?

PR #4818, opened 2026-04-09
One unified image and one workflow for JS and streams adapter

denis-chernov-smartcontract:
## Description
The problem is described here: https://docs.google.com/document/d/1bC1R1jkkr2Kkf59K1tK7fXthz-xhBWFN9yAVz_SGIG0/edit?tab=t.0#heading=h.bh7thiany39y

We are implementing Option 1: unified image solution. 
......

## Changes
- Updated Dockerfile with the `go-builder` section
- Removed Dockerfile.streams-adapter file
- `selector.sh` to dynamically run JS or Streams supervisor configuration based on the STREAMS_ADAPTER env variable  
- Removed all streams specific jobs from workflow files
- Updated Readme files  

## Steps to Test

- [x] Build and test both scenarios locally
- [ ] Merge this PR, initiate new release for streams adapters, deploy to Staging and verify


.github/workflows/upsert-release-pr.yml
```diff
@@ -122,45 +116,6 @@ jobs:
...
-            --jq ".[] | select(.title == \"Release $BUMPED_VERSION\") | .number")
-
-          if [[ -z "$PR_NUMBER" ]]; then
-            echo "Unable to find release PR for version $BUMPED_VERSION"
-            exit 1
-          fi
-
-          PR_BODY=$(gh pr view "$PR_NUMBER" --json body --jq '.body')
-
-          printf -v PR_BODY_PREFIX "This release was triggered by @%s.\n\n" "${{ github.actor }}"
```
dskloetc, line 143: This part is not specific to streams and we should keep it.
I added this in https://github.com/smartcontractkit/external-adapters-js/commit/ec12ee7c914abd7c9d6f0dc522b973d7e9db5298

start-supervisor.sh
dskloetc, line 1: This name is very generic but also doesn't really say what the script does.
It doesn't just select, it actually starts.
Would `start-supervisor.sh` be a better name or can you suggest a more descriptive name?

supervisord.conf
```diff
@@ -28,3 +40,27 @@ stdout_logfile_maxbytes=0
...
+startsecs=5
+
+[program:selector]
+command=/usr/local/bin/selector.sh
+autostart=true
+autorestart=false
+startsecs=0
+priority=1
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
```
dskloetc: Please don't remove the trailing newline. Maybe check your editor settings to make it stop doing this.

supervisord.conf
dskloetc, line 1: I'm not familiar with supervisord. Please have someone else review this file.

Dockerfile
```diff
@@ -3,18 +3,52 @@ ARG location
...
 COPY --from=builder /home/node/app/$location/package.json /home/node/app/$location/*test-payload.js* ./
 
-CMD ["yarn", "server"]
+# Copy Go binary and alias config
+COPY --from=go-builder /build/streams-adapter /usr/local/bin/streams-adapter
+COPY --from=builder /home/node/app/packages/streams-adapter/endpoint_aliases.json /home/node/app/endpoint_aliases.json
+COPY --from=builder /home/node/app/selector.sh /usr/local/bin/selector.sh
+
+# Make scripts executable
+RUN chmod +x /usr/local/bin/selector.sh /usr/local/bin/streams-adapter
```
dskloetc: If they are executable before copying, won't they also be executable after copying?

Dockerfile
```diff
@@ -3,18 +3,52 @@ ARG location
 ARG package
 WORKDIR /home/node/app
 COPY . .
-RUN yarn workspaces focus $package @chainlink/external-adapters-js @chainlink/ea-test-helpers @chainlink/ea-reference-data-reader @chainlink/ea-factories
+RUN yarn workspaces focus $package @chainlink/external-adapters-js @chainlink/ea-test-helpers @chainlink/ea-reference-data-reader @chainlink/ea-factories @chainlink/ea-scripts
 RUN yarn workspace $package build
+RUN METRICS_ENABLED=false yarn generate:endpoint-aliases
```
dskloetc: Setting `METRICS_ENABLED=false` is redundant because `generate:endpoint-aliases` already does that.
https://github.com/smartcontractkit/external-adapters-js/blob/c2f1b829cfcda6ae23e5f79b2f7ce5c12ea82eb3/package.json#L28

Dockerfile
```diff
@@ -3,18 +3,52 @@ ARG location
...
 FROM node:22-alpine
 ARG location
+ARG package
+ENV PACKAGE_NAME=$package
 
 EXPOSE 8080
 WORKDIR /home/node/app
 
+# Install supervisor
+RUN apk add --no-cache supervisor
```
dskloetc, line 31: Why are we avoiding the cache here?

Dockerfile
denis-chernov-smartcontract: Correct, it is not strictly required, but it allows us to avoid unexpected failures cause by GH runner config (e.g. core.fileMode false).

Dockerfile
denis-chernov-smartcontract: apk add --no-cache is there so the image stays smaller.
Without it, Alpine’s apk keeps fetched indexes and package files under /var/cache/apk/ and bloats the image for no runtime benefit.

.github/workflows/upsert-release-pr.yml
denis-chernov-smartcontract: Restored the "Annotate release PR description" job without the streams-adapter section.

start-supervisor.sh
denis-chernov-smartcontract: Updated to `start-supervisor.sh` everywhere.

Dockerfile
denis-chernov-smartcontract: Good catch, updated. Thanks.

supervisord.conf
denis-chernov-smartcontract: @brunotm could you please help with this?

Dockerfile
dskloetc: Thanks for explaining. That makes sense. Maybe worth adding a comment?

.github/workflows/upsert-release-pr.yml
```diff
@@ -125,36 +119,15 @@ jobs:
       - name: Annotate release PR description
         env:
           GITHUB_TOKEN: ${{ steps.setup-github-token.outputs.access-token }}
-        # We use gh pr list with exact title match instead of --search to avoid GitHub search index lag
```
dskloetc, line 128: I think this comment is still useful.

supervisord.conf
```diff
@@ -28,3 +40,27 @@ stdout_logfile_maxbytes=0
...
 startsecs=0
+
+[group:streams-adapter]
+programs=go-adapter,js-adapter
+
+[program:js-adapter-original]
+command=yarn server
+directory=/home/node/app
+autostart=false
+autorestart=unexpected
```
brunotm, line 51: do we also need `exitcodes=0` here ?

supervisord.conf
```diff
@@ -28,3 +40,27 @@ stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 startsecs=0
```
brunotm: startsecs=5

brunotm (changes requested):

brunotm (dismissed):

dskloetc (dismissed):

mxiao-cll (approved):

PR #4819, opened 2026-04-10
[Streams adapter] Forward JS mertics

denis-chernov-smartcontract:
## Description
Implements metrics forwarder to combine original JS and Go metrics.


packages/streams-adapter/config/config.go
```diff
@@ -45,6 +50,9 @@ func Load() *Config {
 		// Subscription
 		SubscriptionRetryDelaySeconds: getEnvAsInt("SUBSCRIPTION_RETRY_DELAY_SECONDS", 10),
 
+		// Metrics forwarding
+		MetricsForwardTimeoutSeconds: getEnvAsInt("METRICS_FORWARD_TIMEOUT_SECONDS", 2),
```
mxiao-cll, line 54: I am not sure if this is enough

mxiao-cll (approved):

ro-tex (approved):

packages/streams-adapter/config/config.go
denis-chernov-smartcontract: 2 seconds is more than enough. The actual response time will be under 10ms.

PR #4844, opened 2026-04-16
OPDATA-6440: iTick: Make region a param instead of having separate endpoints

dskloetc:
## Description

Originally we thought we needed to use different API keys for different regions.
To support this, each region had to be used through a different endpoint.
But it turns out we'll have a single API key that works for all the regions.
So now we only need to use different endpoints for quote vs. depth (because they use different message formats) and stock vs. indices (because they use different websocket endpoints).

## Changes

1. Remove the region from the endpoint construction and move it to the params.
2. Specify an optional region param to the message handler, to be used only by the rest transport. For the websocket transport, the response message contains the region but for the rest transport it needs to be derived from the request we made.

## Steps to Test

1. `yarn test packages/sources/itick`
2.
  ```
    "data": {
      "endpoint": "indices-quote",
      "transport": "ws",
      "symbol": "GER30",
      "region": "gb"
    }
  }'
  
  {
    "result": 24288.21,
    "data": {
      "symbol": "GER30",
      "lastPrice": 24288.21
    },
    "timestamps": {
      "providerDataStreamEstablishedUnixMs": 1776338391878,
      "providerDataReceivedUnixMs": 1776339785857,
      "providerIndicatedTimeUnixMs": 1776339785000
    },
    "statusCode": 200,
    "meta": {
      "adapterName": "ITICK",
      "metrics": {
        "feedId": "{\"base\":\"ger30\",\"region\":\"gb\"}"
      }
    }
  }
  ```

## Quality Assurance

[OPDATA-6440]: https://smartcontract-it.atlassian.net/browse/OPDATA-6440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/itick/test/integration/adapter-ws.test.ts
```diff
@@ -19,67 +19,32 @@ describe('websocket', () => {
   const requests = [
     {
       symbol: '100',
-      endpoint: 'hk-depth',
-      transport: 'ws',
-    },
-    {
-      symbol: '100',
-      endpoint: 'cn-depth',
```
mmcallister-cll, line 27: not worth keeping some of these additional tests? i suppose they are duplicates... 🤔

dskloetc: The fixture returns the same for them so they really are duplicates.

mmcallister-cll: all good

mmcallister-cll (approved):

PR #4857, opened 2026-04-17
OPDATA-6522: generic-api: Use SubscriptionTransport to avoid shared rate limit

dskloetc:
## Description

Currently, `generic-api` uses an `HttpTransport`. The `HttpTransport` does not do any rate limiting so when we use an `HttpTransport` we always apply rate limiting to the adapter, which applies to the single `Requester` used throughout the entire adapter.

This means that the more different requests we send to the adapter, the longer each individual request has to wait between updates.

Since with `generic-api` different requests probably go to different data providers, applying a single global rate limit is not a good solution. 

On the other hand, when we use a `SubscriptionTransport`, we don't apply rate limiting on the adapter level, but instead process every request once every fixed time interval. Typically 10 seconds.

## Changes

1. Add `BACKGROUND_EXECUTE_MS` to configure the time interval between rounds, as is usual with `SubscriptionTransport`.
2. Remove global rate limit.
3. Add `GenericApiSubscriptionTransport` base class used by both transports.
4. Adapt `prepareRequest` and `createResponse` and how they're used by the subscription transport.

## Steps to Test

```
yarn test packages/sources/generic-api
```

I also tested with all 8 existing `generic-api` feed requests.

## Quality Assurance

[OPDATA-6522]: https://smartcontract-it.atlassian.net/browse/OPDATA-6522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/generic-api/src/transport/utils.ts
```diff
@@ -150,36 +161,133 @@ const createResponse = (
...
-        response: {
-          statusCode,
-          errorMessage,
-          ...extraFields,
+      response = await this._handleRequest(context, param)
+    } catch (e: unknown) {
+      const errorMessage = e instanceof Error ? e.message : 'Unknown error occurred'
+      this.logger.error(e, errorMessage)
+      response = {
+        statusCode: (e as AdapterInputError)?.statusCode || 502,
```
mohamed-mehany: Why `AdapterInputError` and not `AdapterError` (more generic)?

packages/sources/generic-api/src/transport/multi-http.ts
```diff
@@ -1,33 +1,18 @@
...
-      apiResponse,
-      mapParam: (param) => param,
+// This was originally an HTTP transport and we wanted to keep the same
+// endpoint.
+export class MultiHttpTransport extends GenericApiTransport<BaseEndpointTypes> {
+  constructor() {
+    super({
+      logger,
+      mapParam: (p) => p,
       mapResponse: (adapterResponse) => adapterResponse,
```
mohamed-mehany, line 14: Since both http.ts and multi-http.ts now extend GenericApiTransport, should we move mapResponse as default identity function in the base class constructor?

packages/sources/generic-api/src/transport/utils.ts
```diff
@@ -150,36 +161,133 @@ const createResponse = (
...
 }: {
-  params: Params<EndpointTypes>[]
+  params: Params<EndpointTypes>
   apiResponse: { data: object | undefined }
   mapParam: (param: Params<EndpointTypes>) => MultiHttpParams
   mapResponse: (response: MultiHttpResponse) => Response<EndpointTypes>
-}): ProviderResult<EndpointTypes>[] => {
-  return params.map((param: Params<EndpointTypes>): ProviderResult<EndpointTypes> => {
+}): PartialAdapterResponse<EndpointTypes['Response']> => {
+  try {
```
mohamed-mehany: AI also mentioned this but kinda nitpick:
```
createResponse already catches all errors and returns an error-shaped partial response (lines 175–187). Then handleRequest wraps _handleRequest in another try/catch (lines 240–254). 
This means errors thrown inside createMultiResponse are silently swallowed by createResponse and the outer catch in handleRequest never fires for those. 
This isn't wrong per se, but it creates two places where errors are "handled" with slightly different shapes, which can be confusing.

In particular, _handleRequest (lines 258–288) spreads the createResponse result and adds timestamps unconditionally. 
If createResponse returns an error object (no timestamps key), 
the spread ...('timestamps' in response ? response.timestamps : {}) on line 283 correctly handles it, but the providerDataRequestedUnixMs and providerDataReceivedUnixMs will still be written — which is semantically odd for an error response (they'll have real timestamps, not zeros). 
This is inconsistent with the outer catch block which sets them to 0 (lines 248–251).

```

mohamed-mehany (dismissed):
A few comments but nothing functional

packages/sources/generic-api/src/transport/utils.ts
dskloetc: I just copied this from some other boiler plate. I'm not sure how we ended up with this in our code base since the generator doesn't have it:
https://github.com/smartcontractkit/ea-framework-js/blob/8635c16e5a4c841ca395f6d2c17138caf9272ad5/scripts/generator-adapter/generators/app/templates/src/transport/custombg.ts.ejs#L71

I've fixed this one here now.

packages/sources/generic-api/src/transport/multi-http.ts
dskloetc: The identity is only valid when `EndpointTypes` is the `BaseEndpointTypes` from `multi-http`. Otherwise it doesn't have the correct return type. So I don't think it's appropriate as default. It probably wouldn't even compile although I didn't check.

packages/sources/generic-api/src/transport/utils.ts
dskloetc: I wouldn't say the errors are silently swallowed. We actually return a response shaped like an error instead of a success response.

The first try/catch has a specific purpose of using a different code path for errors so that successful responses can all go through `mapResponse`. But I realized, I can just check whether the response is an error instead and that simplifies the code a lot.

packages/sources/generic-api/src/transport/multi-http.ts
mohamed-mehany: Aah, got it, makes sense.

mohamed-mehany (approved):

PR #4879, opened 2026-04-22
Refactor market-status

mxiao-cll:
## Closes #OPDATA-6860

Refactor how we structure hard coded market hours because we need to add more of these in the future

 - Route all calls to hard coded market hours via a single file `fallback.ts`
 - A map inside `fallback.ts` to map hard code name to the processing function
 - In the future, when adding a new hard coded market hour, we only need to create its file and then add to the map, 1 touch point

## Quality Assurance


dskloetc:
Can you provide more context? If it's a pure refactoring, why do you need to change tests?

mxiao-cll:
> Can you provide more context? If it's a pure refactoring, why do you need to change tests?

Yes pure refactoring, changing test since we now have all the calls routed through `fallback.ts` so testing at that level instead of the underlying file. 

Basically this
```
export const getStatus = (
  adapterName: HardcodedAdapterName,
  weekend?: string,
): MarketStatusResult => ({
  ...mapping[adapterName](weekend),
  source: adapterName,
})
```

packages/composites/market-status/src/adapter/fallback.ts
```diff
@@ -0,0 +1,18 @@
+import type { MarketStatusResult } from '../transport/base-market-status'
+import { getStatus as get245Status } from './245Adapter'
+
+const mapping = {
+  FALLBACK_245: get245Status,
+}
+
+export type HardcodedAdapterName = keyof typeof mapping
```
dskloetc: Rename "fallback" and "hardcoded" to "static" everywhere?

packages/composites/market-status/src/adapter/fallback.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const isFallbackAdapter = (name: string): name is HardcodedAdapterName => name in mapping
+
+export const getStatus = (
```
dskloetc: `getStatus` is a very generic name.

How about `getStatusFromStaticSchedule` or something like that?

packages/composites/market-status/src/adapter/adapters.ts
```diff
@@ -84,7 +87,7 @@ export const getMarketAdapters = (
     case '24/5':
       return {
         primary: 'TRADINGHOURS',
-        secondary: 'HARD_CODE_245',
+        secondary: 'FALLBACK_245',
```
dskloetc: Can we keep this the same so it's actually a pure refactoring?
If it needs to change, let's do it in a separate PR.

packages/composites/market-status/src/adapter/adapters.ts
```diff
@@ -1,9 +1,12 @@
...
 import { TypeFromDefinition } from '@chainlink/external-adapter-framework/validation/input-params'
 
-const adapterNames = ['NCFX', 'TRADINGHOURS', 'FINNHUB_SECONDARY', 'HARD_CODE_245'] as const
+import type { StaticAdapterName } from './static'
 
-export type AdapterName = (typeof adapterNames)[number]
+const REMOTE_ADAPTER_NAMES = ['NCFX', 'TRADINGHOURS', 'FINNHUB_SECONDARY'] as const
+export type RemoteAdapterName = (typeof REMOTE_ADAPTER_NAMES)[number]
+
+export type AdapterName = RemoteAdapterName | StaticAdapterName
```
dskloetc: Actually, why do we call these adapters? The static sources are not adapters, are they?
Shouldn't it be
```
export type MarketStatusSourceName = AdapterName | StaticSourceName
```
?

dskloetc (dismissed):

packages/composites/market-status/src/adapter/adapters.ts
mxiao-cll: Renamed adapter to source

dskloetc (approved):

PR #4890, opened 2026-04-27
[Streams Adapter] fixed race conditions in key mapping flow

denis-chernov-smartcontract:
### Problem
~30% of requests to the streams-adapter were returning data for the wrong symbol (e.g. a request for upepeusdrti returning uxautusdrti values).

The cache itself always contained correct data. The bug was a race condition that stored a corrupted `rawKey → transformedKey]` mapping in KeyMapper, causing correct cache entries to be looked up under the wrong key.

### Root cause:

When a cache miss occurs, the subscription goroutine fires two phases:

Phase 1 — immediately POSTs to the JS adapter for all ~300 symbols concurrently to minimise initial subscription time
Phase 2 — serialised per endpoint via SubscribeAndLearn, intercepts the ZADD the JS adapter emits, and learns the `rawCacheKey → transformedCacheKey` mapping
`pending` (the channel waiting for a ZADD notification in Phase 2) was keyed by canonical endpoint. Phase 1 fires subscribes for all symbols on the same endpoint concurrently. The JS adapter sends a ZADD on a redcon connection. When goroutine A (waiting for pepeusd in Phase 2) had `pending["cryptolwba"] = ch_A` , a late Phase-1 ZADD for xautusd from goroutine B could arrive and be delivered to ch_A, storing:

All future lookups for pepeusd would then return the cached xautusd observation.

### Fix
Added a `phase1InFlight` set in KeyMapper (keyed by endpoint → set of rawCacheKeys). Before each Phase-1 HTTP call, the symbol is registered; after Phase-1 returns, it is deregistered (before Phase-2 starts).

In NotifyZAdd, before delivering a ZADD to a waiting Phase-2 channel, the computed `transformedKey` is checked against `phase1InFlight. If it matches, the ZADD is a Phase-1 notification for a concurrent symbol on the same endpoint — it is dropped and deregistered atomically. This prevents it from being mis-attributed to whichever goroutine is currently waiting in SubscribeAndLearn.

Since NotifyZAdd deregisters on first ZADD arrival, the DeregisterPhase1InFlight call after subscribeToAsset returns is a no-op in the happy path, and a safety fallback for the case where the ZADD never arrives at all (JS error, dropped message).

### Additional change:
Added `Meta` info to match JS adapter response
```
    "meta": {
        "adapterName": "COINPAPRIKA",
        "metrics": {
            "feedId": "{\"exchange\":\"binance\",\"symbol\":\"btcusdt\",\"type\":\"top_of_book_spot\"}"
        }
    }
 ```

### CODE REVIEW

File: packages/streams-adapter/helpers/keymapper.go [hunk 1]
```diff
@@ -20,26 +20,58 @@ type KeyMapper struct {
...
-
-	epMu   sync.Mutex
+	pending   map[string]chan zaddNotification // canonical endpoint → channel
+
+	// phase1InFlight tracks the raw cache keys of subscriptions whose Phase-1
+	// HTTP request is still in-flight (sent but response not yet received).
+	// Keyed by canonical endpoint; values are sets of rawCacheKey strings.
+	// When a ZADD arrives and its computed transformedKey matches one of these
+	// keys, the ZADD originated from a concurrent Phase-1 request and must be
+	// dropped so it is not mistakenly attributed to a Phase-2 SubscribeAndLearn.
```
ro-tex, line 30: I highly appreciate these documenting comments. Thanks for adding them!

packages/streams-adapter/server/server.go
```diff
@@ -337,7 +326,15 @@ func (s *Server) adapterHandler(c *gin.Context) {
...
 			// adapter ASAP without waiting for the per-endpoint learning lock.
+			// Register as in-flight BEFORE the HTTP call so that the ZADD
+			// it triggers is excluded from Phase-2 learning of other goroutines.
+			s.keyMapper.RegisterPhase1InFlight(params["endpoint"], key)
 			s.subscribeToAsset(originalData)
+			// Deregister before Phase-2 so that Phase-2's own ZADD (same
+			// symbol, same transformedKey) is not mistakenly filtered out.
+			// Any late Phase-1 ZADD for this symbol reaching Phase-2's channel
+			// is harmless — it carries the correct params for this symbol.
+			s.keyMapper.DeregisterPhase1InFlight(params["endpoint"], key)
```
ro-tex, line 337: What happens if the endpoint is already registered for Phase1 from a previous call?

Specifically, I'm thinking about this case:
1. Call 1 registers endpoint e1.
2. Call 2 registers endpoint e1 (again).
3. Call 2's `subscribeToAsset` passes and it deregisters e1. Now e1 is not in the phase1 set.
4. Call 1's `subscribeToAsset` is still ongoing. If in this moment we hit `NotifyZAdd` and we check whether e1 is in phase1, we'll see that it's not, because it's not in the set. But it actually is - we added it twice and removed it only once.

If this is a plausible scenario, then replacing the `struct{}` with a counter and only removing it when the counter reaches zero, would eliminate the issue.

ro-tex: P.S.: If we want to test this, we can add a mock version of `subscribeToAsset` with a configureable delay, which blocks for a long time when first called and returns immediately on the second call.

ro-tex: Discussed offline, this scenario is not possible, we're good to go.

ro-tex (approved):

PR #4913, opened 2026-04-29
Fix race condition in key mapping: replace ZADD interception with feedId-based approach

denis-chernov-smartcontract:
### Problem

10-30% of requests were returning data for the wrong symbol. The root cause was a race condition in the `KeyMapper` which learned raw→transformed cache key mappings by intercepting `ZADD` writes from the JS adapter.

The ZADD-based approach had two fundamental flaws:

1. **Channel pollution** — the notification channel was per-endpoint, not per-symbol. Any concurrent subscription on the same endpoint (e.g. ~300 parallel Phase-1 subscribes on startup) could deliver the wrong ZADD to the waiting goroutine.
2. **Broken for param-transforming adapters** — cfbenchmarks converts client params (`from=AIXBT, to=USD`) into internal params (`index=u_aixbtusd_rti, adapterNameOverride=cfbenchmarks2`), so `rawCacheKey ≠ transformedCacheKey`. The Phase-1 in-flight filter that was meant to suppress stale ZADDs never fired, leaving the race fully open.

### Solution

Replace the ZADD-interception approach entirely. The JS adapter already returns the correct `meta.metrics.feedId` in its HTTP 200 response — use that as the source of truth for key mapping.

### Changes

**`common/types.go`**
- Added `CacheItemStatus` type with three states: `new`, `learned`, `active`
- Extended `CacheItem` with `Status` and `TransformedKey` fields; `Observation` and `OriginalAdapterKey` are only populated once the item is active

**`cache/cache.go`**
- Replaced `Set(params, obs, ...)` with three lifecycle methods:
  - `SetNew(rawKey) bool` — creates a placeholder item; return value acts as a subscription guard (replaces `subscriptionTracker sync.Map`)
  - `SetTransformedKey(rawKey, transformedKey)` — transitions to `learned`, registers `byTransformedKey` secondary index
  - `SetObservation(transformedKey, obs, ...)` — transitions to `active`; observations arriving before the mapping is learned are dropped
- Added `byTransformedKey` secondary index to route EVAL writes to the correct item
- Added `cache_items_total` and `cache_items_active` Prometheus gauges

**`redcon/redcon.go`**
- `handleEval` now computes `transformedKey` from the JS adapter Redis key params and calls `cache.SetObservation(transformedKey, ...)`
- Removed `keyMapper` field, `Config.KeyMapper`, and `NotifyZAdd` call from `handleZAdd`

**`server/server.go`**
- Removed `KeyMapper` and `subscriptionTracker` fields from `Server`
- `adapterHandler` checks `item.Status` directly: serves if `active`, returns 504 if `new`/`learned`, calls `cache.SetNew` on first miss
- On cache miss, a goroutine polls the JS adapter until it returns HTTP 200 with `meta.metrics.feedId`, then calls `cache.SetTransformedKey` — no ZADD involvement
- Replaced `keyMapperHandler` debug endpoint with enhanced `cacheHandler` that now exposes `status` and `transformedKey` per item
- `resubscribeAllAssets` skips non-active items

**`helpers/helpers.go`**
- Added `TransformedKeyFromFeedID(feedID, endpoint string) (string, error)` — parses the feedId JSON and produces the normalized transformed cache key via `BuildCacheKeyParams` + `CalculateCacheKey`

**`config/config.go`**
- Added `FeedIDPollIntervalSeconds` (env `FEEDID_POLL_INTERVAL_SECONDS`, default `2`)
- Added `FeedIDMaxRetries` (env `FEEDID_MAX_RETRIES`, default `90`)

**`main.go`**
- Removed `KeyMapper` initialization and passing to `server.New` and `redcon.New`

**Deleted**
- `helpers/keymapper.go` and `helpers/keymapper_test.go`


packages/streams-adapter/helpers/helpers.go
```diff
@@ -84,3 +85,24 @@ func CalculateCacheKey(params types.RequestParams) (string, error) {
...
 	return strings.Join(parts, ":"), nil
 }
+
+// TransformedKeyFromFeedID derives the internal transformed cache key from the
+// feedId string returned in the JS adapter's HTTP response under
+// meta.metrics.feedId (e.g. `{"index":"u_aixbtusd_rti","adapterNameOverride":"cfbenchmarks2"}`).
+//
+// The endpoint parameter is the canonical endpoint name (e.g. "cryptolwba") and
+// is injected into the params when the feedId JSON does not already contain it.
+func TransformedKeyFromFeedID(feedID, endpoint string) (string, error) {
```
ro-tex, line 95: The fact that this `feedID` is JSON is quite confusing. It might be normal within the context, though.

packages/streams-adapter/redcon/redcon.go
```diff
@@ -208,20 +205,26 @@ func (s *RedconServer) handleEval(conn redcon.Conn, cmd redcon.Command) {
 
 	// Extract data field
 	if data, hasData := rawJSON["data"]; hasData {
-		obs.Data = data
+		obs.Data = append(json.RawMessage(nil), data...)
```
ro-tex, line 208: Valid for all such appends. 
```suggestion
		obs.Data = append([]byte{}, data...)
```

packages/streams-adapter/server/server.go
```diff
@@ -316,53 +317,42 @@ func (s *Server) adapterHandler(c *gin.Context) {
...
+				for i := 0; i < maxRetries; i++ {
+					if i > 0 {
+						time.Sleep(retryInterval)
+					}
+					feedID, ok := s.queryAdapterForFeedID(originalData)
+					if ok {
+						transformedKey, err := helpers.TransformedKeyFromFeedID(feedID, endpoint)
+						if err != nil {
+							s.logger.Error("Failed to compute transformed key from feedId", "feedId", feedID, "error", err)
+							break
```
ro-tex, line 347: What do you think about removing the key from the cache at this point? This `break` stops the look without a successful outcome and we don't have other options for populating the value. If we keep seeing it, we will keep trying, potentially going around some invalid data?

packages/streams-adapter/server/server.go
```diff
@@ -418,6 +408,52 @@ func (s *Server) subscribeToAsset(data interface{}) {
...
+
+	resp, err := s.httpClient.Post(internalUrl, "application/json", bytes.NewBuffer(jsonBody))
+	if err != nil {
+		s.logger.Error("Failed to query JS adapter for feedId", "error", err, "url", internalUrl)
+		return "", false
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", false
```
ro-tex, line 438: Would it make sense to log here? For example if the code is 403 and we need to fix some permissions?
Feel free to ignore if this wouldn't make sense.

ro-tex (dismissed):
LGTM.

I left some comments but none of them are blocking.

packages/streams-adapter/helpers/helpers.go
denis-chernov-smartcontract: Agree, the naming is confusing, but this is what we get from JS adapter.
```
    "meta": {
        "adapterName": "NCFX",
        "metrics": {
            "feedId": "{\"base\":\"ldo\",\"quote\":\"usd\"}"
        }
    }
```

packages/streams-adapter/redcon/redcon.go
denis-chernov-smartcontract: same thing, but the intent is cleaned with `json.RawMessage(nil)`. I would keep it.

packages/streams-adapter/server/server.go
denis-chernov-smartcontract: This is expected that it returns 504 for the first request.
We can add waning log if it's not 200 or 504, but it's not informative without feed ID. At this stage we only have request payload.

packages/streams-adapter/server/server.go
denis-chernov-smartcontract: A have slightly refactored the cache items in this change. Sorry for not describing it fully in the PR description. Here is the logic:
- Each new asset pair request creates a cache item with status "new"
- When we learn the key mapping the status is updated to "learned"
- When we start receiving observations, status becomes "active"
This gives us more visibility on the adapter status, I have also added a couple metrics that cover this.
To answer your question - no, with this approach I don't want to delete the key. 
For invalid asset pair, the loop stops in 10 seconds and the cache item remains "new". Adapter will try to subscribe to it again if there is a request a node.

ro-tex (approved):

PR #4923, opened 2026-05-05
OPDATA-7103: proof-of-reserves-v2: Use decimals.js to handle decimal fractions

dskloetc:
## Description

The Bitcoin JSON RPC interface returns amounts as decimals fraction strings, such as `"76977.62420307"`.
Because there are at most `21e14` satoshis, the JavaScript `number` type can represent each possible number of satoshis exactly.
So one might think it's safe to handle Bitcoin amounts with the `number` type.
Unfortunately this is not true when representing amounts as decimal fractions.
For example:
```
> 76977.62420307 * 10 ** 8
7697762420306.999
```

## Changes

When handling a response that contains a decimal fraction string, use `decimal.js`.

## Steps to Test

```
yarn test packages/composites/proof-of-reserves-v2
```

## Quality Assurance

[OPDATA-7103]: https://smartcontract-it.atlassian.net/browse/OPDATA-7103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
+++ b/packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
@@ -1,3 +1,4 @@
+import Decimal from 'decimal.js'
 import * as objectPath from 'object-path'
 
 export type FixedPoint = {
@@ -5,13 +6,20 @@ export type FixedPoint = {
   decimals: number
 }
 
-type NumberType = FixedPoint | number
+type NumberType = FixedPoint | number | Decimal
 
 const isFixedPoint = (num: NumberType): num is FixedPoint => {
   return typeof num !== 'number' && 'amount' in num && 'decimals' in num
 }
 
 export const toFixedPointWithDecimals = (num: NumberType, decimals: number): FixedPoint => {
+  if (num instanceof Decimal) {
+    return {
+      amount: BigInt(num.mul(Decimal.pow(10, decimals)).trunc().toFixed(0)),
+      decimals,
+    }
+  }
+
   if (!isFixedPoint(num)) {
     return {
       amount: BigInt(Math.trunc(num * 10 ** decimals)),
@@ -86,5 +94,5 @@ export const getFixedPointFromResult = ({
       decimals: decimals,
     }
   }
-  return toFixedPointWithDecimals(Number(amount), defaultDecimals)
+  return toFixedPointWithDecimals(new Decimal(amount), defaultDecimals)
 }
--- a/packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
+++ b/packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
@@ -1,6 +1,8 @@
 import Decimal from 'decimal.js'
 import * as objectPath from 'object-path'
 
+Decimal.set({ precision: 50 })
+
 export type FixedPoint = {
   amount: bigint
   decimals: number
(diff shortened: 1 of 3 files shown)


packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
```diff
@@ -1,17 +1,25 @@
...
+  if (num instanceof Decimal) {
+    return {
+      amount: BigInt(num.mul(Decimal.pow(10, decimals)).trunc().toFixed(0)),
+      decimals,
+    }
+  }
+
   if (!isFixedPoint(num)) {
     return {
       amount: BigInt(Math.trunc(num * 10 ** decimals)),
```
mohamed-mehany: Worth using Decimal.js for this as well?

packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
```diff
@@ -1,17 +1,25 @@
...
   amount: bigint
   decimals: number
 }
 
-type NumberType = FixedPoint | number
+type NumberType = FixedPoint | number | Decimal
 
 const isFixedPoint = (num: NumberType): num is FixedPoint => {
   return typeof num !== 'number' && 'amount' in num && 'decimals' in num
 }
```
mohamed-mehany, line 18: AI raised this, kinda nit but worth noting:
> Since NumberType now includes Decimal, the guard checks typeof num !== 'number', a Decimal instance passes this check and then 'amount' in num / 'decimals' in num are checked on the Decimal object. 
Decimal objects happen to not have amount/decimals properties, so it works by accident. If Decimal ever adds those properties, this breaks silently. The guard should explicitly check for Decimal:
```
 const isFixedPoint = (num: NumberType): num is FixedPoint => {
    return !(num instanceof Decimal) && typeof num !== 'number' && 'amount' in num && 'decimals' in num
 }
```

packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
```diff
@@ -1,17 +1,25 @@
...
 
 export const toFixedPointWithDecimals = (num: NumberType, decimals: number): FixedPoint => {
+  if (num instanceof Decimal) {
+    return {
+      amount: BigInt(num.mul(Decimal.pow(10, decimals)).trunc().toFixed(0)),
```
mohamed-mehany, line 22: Default precision for Decimal.js is 20, is that safe for this? for example if `decimals` is 18 or higher could it cause issues?

mohamed-mehany (dismissed):
A few comments, nothing blocking

packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
dskloetc: Done

packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
dskloetc: I've set it to 50 now.

packages/composites/proof-of-reserves-v2/src/utils/fixed-point.ts
dskloetc: Done

mohamed-mehany (approved):
LGTM

PR #4924, opened 2026-05-05
Ondo Tokenized Equity Regular Hours Only EA

mxiao-cll:
## Closes #OPDATA-6833

 - Commit 1: Add a new smoother type "none" which skips smoothing logic
 - Commit 2: Allow providing 1 stream only, the EA will only fetch during that stream's open hours

## Quality Assurance

### DIFF
--- a/packages/composites/tokenized-equity/src/endpoint/common.ts
+++ b/packages/composites/tokenized-equity/src/endpoint/common.ts
@@ -47,33 +47,29 @@ export const inputDefinition = new InputParameters(
         'Ignore report if the streams report is older than current time by more than this value in seconds. If not provided, all reports will be processed regardless of age.',
     },
     sessionMarket: {
-      required: true,
       type: 'string',
       description:
         'The name of the market for session times, for example nyse. This is passed to the tradinghours adapter as the `market` parameter.',
     },
     sessionMarketType: {
-      required: true,
       type: 'string',
       description:
         'The type of the market for session times, for example 24/5. This is passed to the tradinghours adapter as the `type` parameter.',
     },
     sessionBoundaries: {
-      required: true,
       type: 'string',
       array: true,
       description:
         '(backup) A list of time where market trasition from 1 session to the next in the format of HH:MM. This is only used when the adapter is unable to fetch session times from the tradinghours EA',
     },
     sessionBoundariesTimeZone: {
-      required: true,
       type: 'string',
       description: 'ANA Time Zone Database format',
     },
     smoother: {
       type: 'string',
       description: 'Smoothing algorithm to apply to the price',
-      options: ['kalman', 'ema'],
+      options: ['kalman', 'ema', 'none'],
       default: 'kalman',
     },
     decimals: {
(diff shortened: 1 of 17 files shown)


packages/composites/tokenized-equity/src/endpoint/common.ts
```diff
@@ -87,7 +83,17 @@ export const inputDefinition = new InputParameters(
...
 
-export const validateSession = (sessionBoundaries: string[], sessionBoundariesTimeZone: string) => {
+export const validateSmoother = (
+  smoother: string,
+  sessionBoundaries: string[],
+  sessionBoundariesTimeZone?: string,
+  sessionMarket?: string,
+  sessionMarketType?: string,
+) => {
+  if (smoother === 'none') {
```
dskloetc, line 113: Would we ever want to specify `sessionBoundariesTimeZone`, `sessionBoundariesTimeZone`, `sessionMarket`, or `sessionMarketType` when `smoother === 'none'` or would it make sense to assert that those are `undefined` if `smoother === 'none'`?

packages/composites/tokenized-equity/src/transport/ondoTransport.ts
```diff
@@ -41,7 +41,7 @@ export class OndoTransport extends SubscriptionTransport<BaseEndpointTypes> {
   }
 
   async handleRequest(param: RequestParams) {
-    let responses: Record<Smoother, AdapterResponse<BaseEndpointTypes['Response']>>
+    let responses: Partial<Record<Smoother, AdapterResponse<BaseEndpointTypes['Response']>>>
```
dskloetc, line 44: A `Record` doesn't have any required fields so what does it mean to apply `Partial<>` to it?

packages/composites/tokenized-equity/src/endpoint/ondo.ts
```diff
@@ -50,6 +50,12 @@ export const endpoint = new AdapterEndpoint({
       })
     }
 
+    validateStreamIds(
```
dskloetc: If we destructure in `validateStreamIds`, then here we can just pass `req.requestContext.data`.

packages/composites/tokenized-equity/src/endpoint/common.ts
mxiao-cll: Added assertion

packages/composites/tokenized-equity/src/transport/ondoTransport.ts
mxiao-cll: Record means 
{
  ema: ...,
  kalman: ...,
  none: ...,
}

adding partial makes it

{
  ema?: ...,
  kalman?: ...,
  none?: ...,
}

dskloetc: Apparently `Record<Smoother` is fundamentally different from `Record<string`.
I didn't know. Thanks.

dskloetc (approved):

PR #2407, opened 2022-12-09
Reduce GHA workflow check workload, optimize dependency caching

ad0ll:
**Currently not ready for deployment until there's a mechanism to build composite adapters when a source adapter it depends on changes.**

This is a tweak to CI that cuts build times by about 3/4~4/5 by having certain workflows run against changed adapters instead of all adapters.

Affected workflows/steps:
* main.yml/run-documentation-check (yarn generate:readme)
* main.yml/build-adapters (build a limited set)
* generate-readmes.yml
* release.yml/publish-artifacts

Note, the main purpose of this PR is to save time on checks against develop, but it should also be safe to push to the production workflows, generate-readmes.yml and release.yml. See [this comment](https://github.com/smartcontractkit/external-adapters-js/pull/2407#issuecomment-1347422707) for details.

Here's what this PR does
* By default, the build matrix JS script (scripts/gha) will only present adapters that have changed in the source branch, based on filepaths from git diff. This affects all workflows using build matrixes (main and release).
* By default, changes generate:readme to only generate readmes for changed adapters, which is done through a bash script. This affects main and generate-readmes.
* If either of the above need to be bypassed, manual workflow runs have a BUILD_ALL parameter to bypass the filtering behavior. All of the affected workflows can now be run manually if they couldn't before.
* Regardless of whether the BUILD_ALL flag is passed or not, all workflows will run against all adapters if any changes to core/ or scripts/ appears in the diff.

How to test these changes:
1. Check out a new branch based on **this branch (feature/PDI-188)**  (calling the newly created branch `example` below)
2. In the example branch
        1. Make a change to any source, composite, non-deployable, example, or observable adapter
        2. Push that change
        3. Create a PR against this branch (feature/PDI-188)
        4.  Observe that Documentation Test and Test and Build steps only operate on your changed adapter
        5. Make any change to core or scripts
        6. Push that change
        7. Observe that Documentation Test and Test and Build run against all adapters

You can also test locally by creating a new branch based of this branch, changing files, and then running one or both of the following commands (alter UPSTREAM_BRANCH and BUILD_ALL to your liking):
* `source .github/get-changed-adapters-array.sh && ADAPTERS=$(getChangedAdapterArray) && UPSTREAM_BRANCH=develop getChangedAdapterArray`
* `BUILD_ALL=false yarn generate:gha:matrix`

See this draft commit to see this work in action (you can't see it in this branch because a change to packages/scripts will always build all): https://github.com/smartcontractkit/external-adapters-js/pull/2394

PR #4925, opened 2026-05-06
OPDATA-7274: Fix README generator test

dskloetc:
## Description

The [adapter README generation test](https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/scripts/src/generate-readme/test/integration/script.test.ts) is not working.

It’s supposed to test that it correctly generates the README for an adapter.

In reality it tries to generate the README, but says it can’t find the adapter, so it leaves the README unchanged. Then it checks that the README is as expected. This check passes because the README wasn’t changed because the script failed to do so.

The test seems to have been working correctly before https://github.com/smartcontractkit/external-adapters-js/pull/2407 which was merged over 3 years ago.

## Changes

1. Change `ReadmeGenerator` to take the `adapterPath` instead of the whole adapter object.
2. In `index.ts` when the path is passed to the options instead of the adapter name, pass the path straight to `ReadmeGenerator` instead of trying to find the adapter. This is necessary because `getWorkspacePackages` doesn't find the test adapter. We don't want it to find it because the test adapter is only for testing and shouldn't be treated as a real adapter for releases etc.
3. Move the code path when the adapter path is passed earlier in the code. It does not depend on the call to `getWorkspacePackages` and that call tries to find changed adapters which didn't work during the test. (see `UsageError: No ancestor could be found between any of HEAD and origin/main` [failed job](https://github.com/smartcontractkit/external-adapters-js/actions/runs/25436775551/job/74617323155))
4. Change the test to delete the README before generating it so we aren't accidentally doing nothing and passing.

### Drive-by improvement

Format the README file after generating it. This otherwise happens during pre-commit but it can be confusing to see formatting changes being generated which then disappear again when committing to the repo.

## Steps to Test

1. `yarn test packages/scripts/src/generate-readme/test/integration/`
2. Make a manual change to the README and check that the test still passes and restores the change.

## Quality Assurance

[OPDATA-7274]: https://smartcontract-it.atlassian.net/browse/OPDATA-7274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mxiao-cll (dismissed):

.changeset/heavy-streets-smash.md
```diff
@@ -0,0 +1,5 @@
+---
```
mxiao-cll: Is it possible to not create this?

dskloetc: I thought it was required but apparently it's not.

mxiao-cll (approved):

PR #4929, opened 2026-05-06
Upgrade yarn

mxiao-cll:
The current yarn version seems to break prettier on vscode

Manual changes:
 yarnPath: .yarn/releases/yarn-4.14.1.cjs
 "packageManager": "yarn@4.14.1",

Everything else is auto generated


.yarnrc.yml
```diff
@@ -1,7 +1,12 @@
+approvedGitRepositories:
+  - '**'
+
 compressionLevel: mixed
 
 enableGlobalCache: false
 
+enableScripts: true
```
dskloetc: What does this do?

mxiao-cll:
```
Define whether to run postinstall scripts or not.

If false (the default), Yarn will not execute the postinstall scripts from third-party packages when installing the project (workspaces will still see their postinstall scripts evaluated, as they're assumed to be safe if you're running an install within them).

Note that you also have the ability to disable scripts on a per-package basis using dependenciesMeta, or to re-enable a specific script by combining enableScripts and dependenciesMeta.
```

This was auto populated after I run yarn

dskloetc: This doesn't sound safe. If the default is false why was it set to true? Do we need it?

It seems most of the PR is auto generated. Can you describe in the PR description what you actually did?

mxiao-cll: done

dskloetc (approved):

PR #4939, opened 2026-05-08
enableCompositeTransport on Finage

mxiao-cll:
### OPDATA-6402

Pull result from both transport and return the result with latest timestamp

Uses https://github.com/smartcontractkit/ea-framework-js/pull/765


alejoberardino (approved):
we should test this out carefully, but lgtm

PR #4940, opened 2026-05-08
[Streams adapters] default endpoint support

denis-chernov-smartcontract:
## Description
- DefaultEndpoint for activeIndex
- If no endpoint is specified in the payload, use default endpoint from the index


ro-tex (approved):
LGTM!

There is an outdated dependency the CI is complaining about - `@chainlink/external-adapter-framework`. It would also be nice to bring `axios` to the latest version because of the recent CVEs. Neither of those is part of the responsibilies of this PR, though.

PR #4945, opened 2026-05-11
OPDATA-7107, OPDATA-7307: Make currency conversion optional in solana endpoint of token-balance

dskloetc:
## Description

The EA `proof-of-reserves-v2` has support for currency conversion built in. This means we no longer need to have currency conversion done in many separate endpoints.

When we migrate from `proof-of-reserves` to `proof-of-reserves-v2`, or create a new PoR feed, we want to be able to query balances on Solana while letting `proof-of-reserves-v2` do the conversion. And eventually when all feeds using the conversion have been migrated, we can remove conversion support.

## Changes

1. Make param `priceOracle` optional.
3. If the price oracle is not specified, use a price of 1.
4. Unit test added.

## Steps to Test

```
yarn test packages/sources/token-balance
```

## Quality Assurance

[OPDATA-7107]: https://smartcontract-it.atlassian.net/browse/OPDATA-7107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OPDATA-7307]: https://smartcontract-it.atlassian.net/browse/OPDATA-7307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sources/token-balance/src/transport/solana.ts
+++ b/packages/sources/token-balance/src/transport/solana.ts
@@ -72,10 +72,16 @@ export class SolanaTransport extends SubscriptionTransport<BaseEndpointTypes> {
     const providerDataRequestedUnixMs = Date.now()
 
     // 1. Fetch token price ONCE from oracle contract
-    const tokenPrice = await getTokenPrice({
-      priceOracleAddress: param.priceOracle.contractAddress,
-      priceOracleNetwork: param.priceOracle.network,
-    })
+    // Or use a price of 1 if no oracle provided.
+    const tokenPrice = param.priceOracle
+      ? await getTokenPrice({
+          priceOracleAddress: param.priceOracle.contractAddress,
+          priceOracleNetwork: param.priceOracle.network,
+        })
+      : {
+          value: 1n,
+          decimal: 0,
+        }
 
     // 2. Fetch balances for each Solana wallet and calculate their USD value using the SINGLE tokenPrice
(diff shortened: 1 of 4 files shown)


packages/sources/token-balance/src/transport/solana.ts
```diff
@@ -72,10 +72,16 @@ export class SolanaTransport extends SubscriptionTransport<BaseEndpointTypes> {
     const providerDataRequestedUnixMs = Date.now()
 
     // 1. Fetch token price ONCE from oracle contract
-    const tokenPrice = await getTokenPrice({
-      priceOracleAddress: param.priceOracle.contractAddress,
-      priceOracleNetwork: param.priceOracle.network,
-    })
+    // Or use a price of 1 if no oracle provided.
+    const tokenPrice = param.priceOracle
+      ? await getTokenPrice({
```
mohamed-mehany, line 77: Not sure ifgetTokenPrice can throw, but if that happens do we also want to default or fail the request?

dskloetc: Definitely fail the request.
If the price oracle is specified we have to use it. We can't suddenly say 1 BTC = 1 USD because we couldn't get the price of a BTC :).

mohamed-mehany (approved):

PR #4947, opened 2026-05-11
OPDATA-7109: Include decimals for individual balances of xrp response

dskloetc:
## Description

All balances returned by the `xrp` endpoint of the `token-balance` adapter are XRP balances, so they all have 6 decimals.
We indicate this with `decimas: 6` directly in the `data` part of the response.

But other endpoints provide decimals for each individual balance so doing the same for this endpoint, makes it more consistent and easier to process the response. In particular in `proof-of-reserves-v2`.

## Changes

1. Add `decimals: 6` to individual balances in the response.
2. Update tests.

## Steps to Test

```
yarn test packages/sources/token-balance
```

## Quality Assurance

[OPDATA-7109]: https://smartcontract-it.atlassian.net/browse/OPDATA-7109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sources/token-balance/src/endpoint/xrp.ts
+++ b/packages/sources/token-balance/src/endpoint/xrp.ts
@@ -34,6 +34,7 @@ export const inputParameters = new InputParameters(
 export type AddressWithBalance = {
   address: string
   balance: string
+  decimals: 6
 }
 
 export type BaseEndpointTypes = {
--- a/.changeset/tame-worlds-invite.md
+++ b/.changeset/tame-worlds-invite.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/token-balance-adapter': minor
+---
+
+Include decimals for individual balances for xrp response
--- a/packages/sources/token-balance/src/transport/xrp.ts
+++ b/packages/sources/token-balance/src/transport/xrp.ts
@@ -126,6 +126,7 @@ export class XrpTransport extends SubscriptionTransport<BaseEndpointTypes> {
         return {
           address,
           balance,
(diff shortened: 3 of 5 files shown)


packages/sources/token-balance/src/endpoint/xrp.ts
```diff
@@ -34,6 +34,7 @@ export const inputParameters = new InputParameters(
 export type AddressWithBalance = {
   address: string
   balance: string
+  decimals: 6
```
mxiao-cll, line 37: Can this reference RESULT_DECIMALS as well?

mxiao-cll (approved):

packages/sources/token-balance/src/endpoint/xrp.ts
dskloetc: I don't think so?
This is the type `6` but `RESULT_DECIMALS` is the value `6`.
```
packages/sources/token-balance/src/endpoint/xrp.ts:39:13 - error TS2749: 'RESULT_DECIMALS' refers to a value, but is being used as a type here. Did you mean 'typeof RESULT_DECIMALS'?

39   decimals: RESULT_DECIMALS
               ~~~~~~~~~~~~~~~
```

If I declare `RESULTS_DECIMALS` as
```
const RESULT_DECIMALS = 6 as const
```
then I could declare this as `decimals: typeof RESULT_DECIMALS` but I don't think that would make it at all obvious that it's `6` rather than `number`.

PR #4961, opened 2026-05-13
Implement price in six

mxiao-cll:
## Closes #OPDATA-6773

Implement the price endpoint with the following special cases:
- Six's ws server may appear to be working but stopped streaming data, this is detectable via ping/pong. We implemented a timer that if pong comes 2 seconds after ping, we re-start the connection. 
- Six only stream updates on data that has updates, for example, if there is only updates on bid it will only stream a message containing the bid information. We are expected to use the previous ask information to form the full bid/ask pair. We implemented a cache locally to store the latest bid and ask as they come-in. When generating response, we fetch data from cache.
- Six stream both lastPrice and bid/ask in one connection, thus we need to implement the stock & stock_quotes sharing a single ws connection
   - One endpoint with stock and stock_quotes alias with one ws transport
   - Inject the real endpoint name via framework into a hidden param
   - The transport will write data for stock or stock_quotes based on the messages streaming in
   - We will be able to match the response by using hidden param
   - The endpoint will return different data based on endpoint: stock vs endpoint: stock_quotes
   - We additionally dedupe the subscribe and un-subscribe message so only 1 is sent even when we request data on both endpoints
- Message over 300 seconds old is rejected

## Quality Assurance


packages/sources/six/src/endpoint/stock.ts
dskloetc, line 1: Why is the file called `price` while the endpoint is called `stock`?

packages/sources/six/src/endpoint/stock.ts
```diff
@@ -0,0 +1,60 @@
...
+    (req) => {
+      const [ticker, market] = req.requestContext.data.base.split('_')
+      if (!ticker || !market) {
+        /// Not using customInputValidation because we want to validate after overrides are applied
+        throw new AdapterInputError({
+          statusCode: 400,
+          message: 'base must be in the format of ${TICKER}_${MARKET}',
+        })
+      }
+      req.requestContext.data.rawEndpoint = req.requestContext.requestEndpointName
```
dskloetc: Do we need this? Is the `requestContext` not also available in the transport?

packages/sources/six/src/endpoint/stock.ts
```diff
@@ -0,0 +1,60 @@
...
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+import { wsTransport } from '../transport/price'
+
+export const inputParameters = new InputParameters(
+  {
+    ...stockEndpointInputParametersDefinition,
+    rawEndpoint: {
```
dskloetc: Why is this available in the params? Isn't it provided by the framework?

packages/sources/six/src/transport/stock-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.setBookSide(streamId, 'ask', ask)
+  }
+
+  private setBookSide(streamId: string, side: 'bid' | 'ask', msg?: PriceMessage) {
+    const cache = side === 'bid' ? this.bidCache : this.askCache
+    const price = msg?.value
+    const volume = msg?.size
+    const time = coverTimeToMs(msg?.unixTimestamp)
+
+    if (isMessageOld(time)) {
```
dskloetc: Does it matter how old it is? If there are no newer messages it's still the most recent data, isn't it?

packages/sources/six/src/transport/price-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+    const volume = msg?.size
+    const time = coverTimeToMs(msg?.unixTimestamp)
+
+    if (isMessageOld(time)) {
+      this.logger.warn(
+        `${side} message ${JSON.stringify(msg)} is more than ${MESSAGE_TTL_SECONDS}s old`,
+      )
+      return
+    }
+    if (price == undefined || Number.isNaN(price) || volume === undefined || Number.isNaN(volume)) {
```
dskloetc: Please use `===` instead of `==`.

packages/sources/six/src/transport/price-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.logger.debug(`Invalid or missing ${side} ${JSON.stringify(msg)}`)
```
dskloetc: If it's missing it just means we haven't received the data yet, right?
But if it's invalid that's a serious error?
Shouldn't we treat those cases separately?

packages/sources/six/src/transport/price-cache.ts
```diff
@@ -0,0 +1,115 @@
+import { makeLogger } from '@chainlink/external-adapter-framework/util'
+import { PriceMessage } from './price'
+
+const MESSAGE_TTL_SECONDS = 300
+const coverTimeToMs = (time?: number) => (time ? Math.floor(time * 1000) : undefined)
```
dskloetc: What do you mean by "cover"?

packages/sources/six/src/transport/price.ts
```diff
@@ -0,0 +1,145 @@
...
+    open() {
+      activeStreamIds.clear()
+    },
+    heartbeat(connection) {
+      connection.ping()
+      pingTime = Date.now()
+    },
+    pong(connection) {
+      const pongTime = Date.now() - pingTime
+      if (pongTime > PONG_TIMEOUT) {
```
dskloetc: What if the pong never comes or only after 10 minutes.
We shouldn't wait for that, right?
Isn't it better to set a timeout when the ping is sent and close the connection when the timeout is reached unless the pong arrives first?

packages/sources/six/src/transport/price.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      return message.data.startStream
+        .map((stream) => {
+          const [ticker, market] = stream.streamId.split('_')
+          if (!ticker || !market) {
+            logger.warn(`Unexpected streamId format ${stream.streamId}`)
+            return []
+          }
+          if (stream.type === 'ERROR') {
+            return ['stock', 'stock_quotes'].map((rawEndpoint) => ({
```
dskloetc: Can we call it `requestEndpoint` here as well instead of `rawEndpoint`?

packages/sources/six/src/index.ts
```diff
@@ -1,13 +1,13 @@
 import { expose, ServerInstance } from '@chainlink/external-adapter-framework'
 import { Adapter } from '@chainlink/external-adapter-framework/adapter'
 import { config } from './config'
-import { marketStatus } from './endpoint'
+import { marketStatus, price } from './endpoint'
 
 export const adapter = new Adapter({
-  defaultEndpoint: marketStatus.name,
+  defaultEndpoint: price.name,
```
dskloetc: This is a breaking change, right?

.changeset/true-sides-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@chainlink/six-adapter': patch
```
dskloetc, line 2: patch?

packages/sources/six/test/integration/adapter-stock.test.ts
```diff
@@ -0,0 +1,113 @@
+import { WebSocketClassProvider } from '@chainlink/external-adapter-framework/transports'
+import {
+  mockWebSocketProvider,
+  MockWebsocketServer,
+  setEnvVariables,
+  TestAdapter,
+} from '@chainlink/external-adapter-framework/util/testing-utils'
+import FakeTimers from '@sinonjs/fake-timers'
```
dskloetc, line 8: Why use sinon fake timers rather than stubbing `Date.now` or `jest.useFakeTimers` like we normally do?

packages/sources/six/src/endpoint/stock.ts
mxiao-cll: Renamed

packages/sources/six/src/endpoint/stock.ts
mxiao-cll: Framework strip endpoint out before passing it down, so we don't have access to that here.

packages/sources/six/src/endpoint/stock.ts
mxiao-cll: We need to inject the raw endpoint into the request itself because we need it to match the cache key when we generate the response from ws messages.

Another way to do it would be to modify the `base` itself to include the endpoint names


-----------

1 request comes in, it will be able to generate two cache entries, the two cache entry need to have different cache key, so we need something more than the request itself to distinguish (here adding a new param into request)

packages/sources/six/src/transport/stock-cache.ts
mxiao-cll: I think here we are preventing server from re-playing an old message, we don't have anything like this in our other ws. This comes from SA as
```
it is not a client or product’s requirement. we added it as a security measure to avoid writing stale data. if this is already managed elsewhere, feel free to remove that business logic from the adapter.
```

Can remove it if we think this is not important

packages/sources/six/src/transport/price-cache.ts
mxiao-cll: This is validating the ws message that server sents us, the data needs to come in pairs

Data need to have both price and volume and they both needs to be valid.

So those cases are the same, here we are basically ignoring bad ws messages.

However this should be a warn instead of debug, updated.

packages/sources/six/src/transport/price-cache.ts
mxiao-cll: convert*

packages/sources/six/src/index.ts
mxiao-cll: Yes but this adapter was never released

.changeset/true-sides-wait.md
mxiao-cll: Yes, the first PR is major and it was not release so when release it all squash into major

packages/sources/six/test/integration/adapter-stock.test.ts
mxiao-cll: FakeTimers also widely used in our codebase

packages/sources/six/src/endpoint/stock.ts
dskloetc: If this needs to be a param, I think we should give it a proper name.

What would we call this if it didn't have to be related to the endpoint?
Then in `requestTransforms` we add the comment explaining why this is also made available via the endpoint (i.e. in order to offer the same interface as other EA. Would be good to list this EAs as well).

Probably both `stock` and `stock_quotes` should be aliases so that there is also a way to pass this param without the endpoint overwriting it.

packages/sources/six/src/endpoint/stock.ts
dskloetc: This is pretty confusing and will be hard to remember in 3 months. Can you add a detailed comment explaining this here and also where the requests are injected into the cache?

packages/sources/six/src/transport/stock-cache.ts
dskloetc: I don't understand. The DP can send us whatever they want whenever they want. How does this add security? Can you push more on SA to understand their real concern and whether we need anything here?

If we're concerned that messages will come out of order, we should compare to the timestamp of the previous message. Not to some arbitrary cut-off.

packages/sources/six/src/transport/price.ts
mxiao-cll: Yes

packages/sources/six/src/endpoint/stock.ts
mxiao-cll: Re-named and added comments

packages/sources/six/src/transport/stock-cache.ts
mxiao-cll: Don't seem to be anything concrete, removed

dskloetc (approved):

PR #4962, opened 2026-05-14
Add SIX and BME to market status

mxiao-cll:
## Description

Add two new markets to six and tradinghours adapter

Also updated ws closure code because 1006 is reserved

## Quality Assurance


packages/sources/six/test/integration/market-status-fixtures.ts
```diff
@@ -12,7 +12,7 @@ export const mockOneMarket = (): nock.Scope =>
             {
               referenceData: {
                 marketBase: {
-                  bc: 1,
+                  bc: 4,
```
mmcallister-cll, line 15: are overrides not read in integration tests?

mxiao-cll: They are, in integration test the adapter input is six and the underlying request to DP is 4

packages/sources/six/src/config/overrides.json
mmcallister-cll, line 1: why overrides file vs in code like tradinghours?

mxiao-cll: I feel like overrides is easier to see and update

mmcallister-cll: do you think we should update tradinghours? can be a separate task if yes

packages/sources/six/test/integration/market-status-fixtures.ts
mmcallister-cll: sorry, was reading this as the ea request, all good

mmcallister-cll (dismissed):

packages/sources/six/src/config/overrides.json
mxiao-cll: I think we should refactor all these so that we don't need to change code to add new market

mmcallister-cll: ok let's discuss offline cause it's out of the scope of this ticket and would likely require a few reconfigs before we could roll it out. Going to resolve this so it can be merged

mmcallister-cll (approved):

PR #4965, opened 2026-05-15
[Streams adapters] Fix finalto and finnhub flatlines

denis-chernov-smartcontract:
## Description

Finalto and Finnhub data providers are case-sensitive. The resubscribe loop was re-creating the request payload from the adapter key, which resulted in all asset names being uppercased. This caused resubscribes to fail — the JS adapter invalidated the asset pair, and data stopped flowing until the Redcon cache expired.

The fix stores the original request payload in the cache item and uses it when resubscribing.

## Changes

- Cache cleanup interval from minutes to seconds to increase granilarity
- Added `originalRequestData` for cache item, updated `setNew`, and resubscribeAllAssets to use it
- Added overrides to the key items

## Steps to Test

Query Finalto and Finnhub EAs in staging and check that timestamps are no longer than few seconds.

### DIFF
--- a/packages/streams-adapter/README.md
+++ b/packages/streams-adapter/README.md
@@ -51,7 +51,7 @@ The Go runtime waits for the JS adapter `/health` endpoint before starting.
 - `REDCON_PORT` (default: `6379`): Redis-compatible ingestion server port.
 - `METRICS_PORT` (default: `9080`): Prometheus metrics server port.
 - `CACHE_TTL_MINUTES` (default: `5`): in-memory cache TTL.
-- `CACHE_CLEANUP_INTERVAL` (default: `60`): cache cleanup interval in seonds; also used as periodic resubscribe interval.
+- `CACHE_CLEANUP_INTERVAL` (default: `60`): cache cleanup interval in seconds; also used as periodic resubscribe interval.
 - `SUBSCRIPTION_RETRY_DELAY_SECONDS` (default: `10`): delay before allowing re-subscription after a miss.
 - `LOG_LEVEL` (default: `info`): use `debug` to enable Gin request logging and additional debug logs.
 - `PACKAGE_NAME` (required for alias mapping): adapter package name, for example `@chainlink/tiingo-adapter`.
--- a/packages/streams-adapter/README.md
+++ b/packages/streams-adapter/README.md
@@ -51,7 +51,7 @@ The Go runtime waits for the JS adapter `/health` endpoint before starting.
 - `REDCON_PORT` (default: `6379`): Redis-compatible ingestion server port.
 - `METRICS_PORT` (default: `9080`): Prometheus metrics server port.
 - `CACHE_TTL_MINUTES` (default: `5`): in-memory cache TTL.
-- `CACHE_CLEANUP_INTERVAL` (default: `1`): cache cleanup interval; also used as periodic resubscribe interval.
+- `CACHE_CLEANUP_INTERVAL` (default: `60`): cache cleanup interval in seonds; also used as periodic resubscribe interval.
 - `SUBSCRIPTION_RETRY_DELAY_SECONDS` (default: `10`): delay before allowing re-subscription after a miss.
 - `LOG_LEVEL` (default: `info`): use `debug` to enable Gin request logging and additional debug logs.
 - `PACKAGE_NAME` (required for alias mapping): adapter package name, for example `@chainlink/tiingo-adapter`.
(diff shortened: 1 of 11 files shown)


packages/streams-adapter/helpers/aliases.go
```diff
@@ -172,5 +172,36 @@ func BuildCacheKeyParams(data map[string]interface{}) (types.RequestParams, erro
 		out[strings.ToLower(k)] = strings.ToUpper(value)
 	}
 
+	// Apply adapter-specific symbol overrides so the cache key reflects the
+	// actual symbol the JS adapter will subscribe to. Without this, requests
+	// that map a symbol via overrides (e.g. CRCL → CRCL.xnys) would produce
+	// the same cache key as requests for the raw symbol, causing incorrect
+	// resubscription payloads to be stored and used.
```
ro-tex, line 179: Once again, I highly appreciate the level of detail and the "why" nature of these documentation comments. Great work!

packages/streams-adapter/helpers/aliases.go
```diff
@@ -172,5 +172,36 @@ func BuildCacheKeyParams(data map[string]interface{}) (types.RequestParams, erro
...
+						continue
+					}
+					for overrideKey, overrideVal := range adapterOverrides {
+						if strings.EqualFold(overrideKey, paramVal) {
+							out[paramKey] = strings.ToUpper(toString(overrideVal))
+							break
+						}
+					}
+				}
+				break
```
ro-tex, line 201: Are we sure there will be only one override?

packages/streams-adapter/README.md
```diff
@@ -51,7 +51,7 @@ The Go runtime waits for the JS adapter `/health` endpoint before starting.
 - `REDCON_PORT` (default: `6379`): Redis-compatible ingestion server port.
 - `METRICS_PORT` (default: `9080`): Prometheus metrics server port.
 - `CACHE_TTL_MINUTES` (default: `5`): in-memory cache TTL.
-- `CACHE_CLEANUP_INTERVAL` (default: `1`): cache cleanup interval; also used as periodic resubscribe interval.
+- `CACHE_CLEANUP_INTERVAL` (default: `60`): cache cleanup interval in seonds; also used as periodic resubscribe interval.
```
ro-tex: nit:
```suggestion
- `CACHE_CLEANUP_INTERVAL` (default: `60`): cache cleanup interval in seconds; also used as periodic resubscribe interval.
```

packages/streams-adapter/server/server_test.go
```diff
@@ -55,15 +55,15 @@ func TestMain(m *testing.M) {
 	}
 
 	cfg := &config.Config{
-		HTTPPort:             "0",
-		EAPort:               "0",
-		EAHost:               "localhost",
-		RedconPort:           "0",
-		GoMetricsPort:        "0",
-		CacheTTLMinutes:      5,
-		CacheCleanupInterval: 60,
```
ro-tex, line 64: We're going from an hour to a minute? And the TTL is still 5 minutes?
This might be fine, I just want to make sure.

packages/streams-adapter/config/config.go
```diff
@@ -19,7 +19,7 @@ type Config struct {
 
 	// Cache configuration
 	CacheTTLMinutes      uint // Cache TTL in minutes (0 = default 5 minutes)
-	CacheCleanupInterval uint // Cache cleanup interval in minutes (0 = default 1 minute)
+	CacheCleanupIntervalSeconds uint // Cache cleanup interval in seconds (0 = default 60 seconds)
```
ro-tex, line 22: This breaking change in the config is the only thing that worries me in this PR. If we don't have this deployed anywhere then we should be good to go. If we do - we'd better make sure those configs are updated. :)

packages/streams-adapter/helpers/aliases.go
denis-chernov-smartcontract: yes, there can only be one matching adapter key in the map.

packages/streams-adapter/server/server_test.go
denis-chernov-smartcontract: yes, TTL is still 5 minutes. No changes here

packages/streams-adapter/config/config.go
denis-chernov-smartcontract: ENV variable `CACHE_CLEANUP_INTERVAL` remains the same, and we (and external NOPs) are using default value, so I feel this is safe.

ro-tex (approved):

PR #4971, opened 2026-05-18
OPDATA-7129: Handle ripcord from balance source of proof-of-reserves-v2

dskloetc:
## Description

Some balance providers might indicate incomplete responses. In particular, `ethereum-cl-indexer` will include`isValid: false` in the response if it hasn't aggregated all the validators yet.

In this case we should not consider the returned total reserves as valid.

The old `proof-of-reserves` has custom code for `ethereum-cl-indexer` for this at https://github.com/smartcontractkit/external-adapters-js/blob/5224c0fc3795c06558d3f9235a91397aaa8efb44/packages/composites/proof-of-reserves/src/utils/reduce.ts#L77

In `proof-of-reserves-v2` we make this more generic by allowing to specify a ripcord `path` and `disabledValue` (that it should have under normal conditions), similar to `generic-api`.

## Changes

1. Add `ripcord` field to `balanceSource` config, which has a `path` and a `disabledValue`.
2. If the `ripcord` field is specified, include a `ripcord` field in the corresponding response component and in the top-level response.
3. If the ripcord is enabled, remove the result from the response and return status code 503.
4. If the ripcord is enabled, still try to process the response as normal, but catch any errors to account for the response being incomplete.

## Steps to Test

```
yarn test packages/composites/proof-of-reserves-v2
```

## Quality Assurance

[OPDATA-7129]: https://smartcontract-it.atlassian.net/browse/OPDATA-7129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/composites/proof-of-reserves-v2/src/endpoint/reserves.ts
```diff
@@ -93,6 +93,24 @@ export const inputParameters = new InputParameters(
           type: 'string',
           required: false,
         },
+        ripcord: {
```
mxiao-cll, line 96: Do you want to name this balancerRipcord?

packages/composites/proof-of-reserves-v2/src/transport/component.ts
```diff
@@ -157,6 +163,17 @@ export class ComponentRepo {
     )
   }
 
+  async getRipcord(): Promise<boolean | undefined> {
+    const ripcordValues = await Promise.all(this.components.map((c) => c.getRipcord()))
+    if (ripcordValues.some((value) => value === true)) {
```
mxiao-cll, line 168: ripcordValues.includes(true) ?

mxiao-cll (approved):

packages/composites/proof-of-reserves-v2/src/endpoint/reserves.ts
dskloetc: It's already inside the `balanceSource` field so I think that's redundant.

packages/composites/proof-of-reserves-v2/src/transport/component.ts
dskloetc: Good call. I'll do it in a follow-up PR to make reviewing easier.

dskloetc: https://github.com/smartcontractkit/external-adapters-js/pull/4974

PR #4991, opened 2026-05-25
solana-balance ripcord

mxiao-cll:
## Description

Only staking Solana address have balances, and it is only detectable when you actually try to fetch it. We throw a ripcord here so that PoR-v2 can pick it up

## Quality Assurance


packages/sources/token-balance/src/endpoint/solana-balance.ts
```diff
@@ -19,6 +19,11 @@ export const inputParameters = new InputParameters(
       array: true,
       description: 'List of addresses to read',
     },
+    throwOnRipcord: {
```
dskloetc: What is the context? Do we have a feed with an address list with non-existing staking account which are there on purpose and it's not a mistake?

dskloetc: I think this is a valid concern from Cursor:

> Naming diverges from the established convention in this repo. Existing ripcord-aware adapters use noErrorOnRipcord (negative, default false), e.g.:
>
> packages/sources/the-network-firm/src/endpoint/reserve.ts:17-23
> packages/sources/exchange-copter/src/endpoint/totalReserve.ts:16-22
> This PR introduces throwOnRipcord (positive, default true) at packages/sources/token-balance/src/endpoint/solana-balance.ts:22-26. The behavior is identical (default = strict), but now PoR-v2 operators have to remember a different param name per balance source. Why it matters: it doubles the set of conventions and is easy to misconfigure. Suggested fix: rename to noErrorOnRipcord for parity, or, if you prefer positive naming, raise it as a repo-wide refactor.

packages/sources/token-balance/src/transport/solana-balance.ts
```diff
@@ -99,10 +124,7 @@ export class SolanaBalanceTransport extends SubscriptionTransport<BaseEndpointTy
   async getTokenBalance(address: string): Promise<number> {
     const result = await this.getConnection().getAccountInfo(new PublicKey(address))
     if (!result) {
-      throw new AdapterInputError({
-        statusCode: 400,
```
dskloetc, line 103: From Cursor, is this intentional?

> Status code silently regresses from 400 to 502 for the default path. Previously getTokenBalance threw AdapterInputError({ statusCode: 400 }). Now it throws RipcordError, which has no statusCode, so the catch's (e as AdapterInputError)?.statusCode || 502 falls through to 502 (packages/sources/token-balance/src/transport/solana-balance.ts:47 on main vs the diff). Direct (non-PoR-v2) consumers that distinguish 400 (bad address) from 502 (upstream issue) will see a change. Either give RipcordError a statusCode (400, or 503 if you want to align with PoR-v2's "ripcord = 503" convention in packages/composites/proof-of-reserves-v2/src/transport/reserves.ts:128), or call this out in the changeset.

packages/sources/token-balance/src/endpoint/solana-balance.ts
mxiao-cll: Currently you throw an error when there is an non-existing staking account, in order for this to work with ripcord setting in PoR-v2, you need this EA to not throw an error.

However, if you don't throw error, existing feed will under-report, so we need this EA to throw error default and not throw error for PoR-v2 only

mxiao-cll: Will rename

packages/sources/token-balance/src/transport/solana-balance.ts
mxiao-cll: I can switch to 503

packages/sources/token-balance/src/endpoint/solana-balance.ts
dskloetc: I still don't understand. I see 2 possibilities:

1. The address list contains an address that's wrong. The wrong address happens to not exist. By throwing an error, we make sure we are alerted to the fact that the address is wrong.
2. The address list contain an non-existing address on purpose. Maybe the address will be used in the future. We need the balance of the non-existing address to be reported as zero so the feed can continue working until the non-existing address starts to exist.

Are you adding this functionality because you know a feed for which option 2 is the case? Or is there another possibility that I'm missing?

I don't see what any of this has to do with `proof-of-reserves-v2`. Both cases above could also happen with `proof-of-reserves` (v1), right?

mxiao-cll: I am only considering #1 here

For PoR v1, you need to throw an error to stall the feed
For PoR v2, both throwing an error and not throwing an error can stall the feed, however, you must not throw an error in order to raise the ripcord field properly

dskloetc (approved):

PR #4999, opened 2026-05-27
Handle address ripcord in por-v2

mxiao-cll:
## Description

Addresser can now throw ripcord as well, handling it the same way as we handle balancer ripcord

## Quality Assurance


packages/composites/proof-of-reserves-v2/src/transport/address.ts
```diff
@@ -2,14 +2,18 @@ import objectPath from 'object-path'
...
+import { getRipcord } from './utils'
 
 type AddressListConfig = RequestParams['addressLists'][number]
 
 class AddressList {
   fetchFromProvider: Fetcher
   shortJsonForError: Stringifier
 
-  addressArray: Promise<unknown[]>
+  addressArray: Promise<{
```
dskloetc: An address array holding an address array is a bit confusing. Shall we call the outer one addressListResult?

Maybe also declare an `AddressListResult` type?

packages/composites/proof-of-reserves-v2/src/transport/utils.ts
```diff
@@ -52,3 +53,17 @@ export const shortJsonForError = (maxLen: number, obj: unknown): string => {
   }
   return longString.length > maxLen ? `${longString.slice(0, maxLen)}...` : longString
 }
+
+export const getRipcord = (
+  responseData: Record<string, unknown>,
+  config?: { path: string; disabledValue: string },
+): boolean | undefined => {
+  let ripcord: boolean | undefined = undefined
```
dskloetc: You don't need this mutable variable if you change the structure:
```
  if (config === undefined) {
    return undefined
  }
  const ripcordValue = String(objectPath.get(responseData, config.path))
  const expectedValue = config.disabledValue
  return ripcordValue !== expectedValue
```

packages/composites/proof-of-reserves-v2/src/transport/component.ts
```diff
@@ -67,7 +70,7 @@ export class Component {
         originalCurrency: component.currency,
         totalBalanceInOriginalCurrency: totalBalance,
         addressCount,
-        ripcord,
+        ripcord: addressRipcord || balanceRipcord,
```
dskloetc: Is there a way to see in the output where the ripcord came from?
Would it make sense to output `addressRipcord` and `balanceRipcord` in addition to `ripcord`?

dskloetc (approved):

PR #5020, opened 2026-06-02
Feat/OPDATA-6934 nobi ea

mmcallister-cll:
## Description
Adds Nobi state price EA

## Changes
- Added standard plumbing & WsTransport implementation
- Interesting part: NobiTransportManager in `src/endpoint/transportManager` that controls request routing based on a max number of WS subscriptions per connection and max number of connections per key (see ticket for details)

## Steps to Test
1. `yarn test nobi`
2. with payloads
```json
{
    "data": {
        "endpoint": "price",
        "base": "BTC",
        "quote": "USD"
    }
}
```

Set `MAX_SUBSCRIPTIONS_PER_TRANSPORT` sufficiently low (eg: 2) and request other common crypto pairs (eg: ETH/USD, SOL/USD) to see transport rollover

## Quality Assurance

[OPDATA-6934]: https://smartcontract-it.atlassian.net/browse/OPDATA-6934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sources/nobi/docs/custom.md
```diff
@@ -0,0 +1,6 @@
+## Additional Env Variables
+
+These env vars are available to configure the amount of WebSocket transports and subscriptions per transport that the adapter will manage. The defaults should be sufficient for most use cases:
+
+- `MAX_TRANSPORTS`: The maximum number of WebSocket transports to manage. Default is 10.
+- `MAX_SUBSCRIPTIONS_PER_TRANSPORT`: The maximum number of currency pairs to route through a single transport before routing to the next one. Default is 100.
```
mxiao-cll: Why are these not in `packages/sources/nobi/src/config/index.ts`

mmcallister-cll: I think they could be, but then we'd have to lazy load the transport manager as a singleton

packages/sources/nobi/src/transport/price.ts
```diff
@@ -0,0 +1,112 @@
...
+                timestamps: {
+                  providerIndicatedTimeUnixMs: dateMs.getTime(),
+                },
+              },
+            },
+          ]
+        },
+      },
+      builders: {
+        // Nobi's WS API requires sending the full list of pairs to subscribe, keep track of activePairs
```
alejoberardino: should use [this](https://github.com/smartcontractkit/ea-framework-js/pull/801/changes#diff-8516aabfa53f8fa1cddae5decb8a8c695d361244ace4cf85b5ffc9567cce6b26R551-R556)

packages/sources/nobi/src/endpoint/transportManager.ts
```diff
@@ -0,0 +1,94 @@
...
+  return indexToLetters(Math.floor(i / 26) - 1) + String.fromCharCode(97 + (i % 26))
+}
+
+type NobiTransportTracker = {
+  transport: NobiWsTransport
+  name: string
+  requestCount: number
+}
+
+export class NobiTransportManager {
```
alejoberardino: I get why but it's way too much imo, I don't believe having this be configurable is a good thing. I'd still loop and make 10 transports (the documented max connections) and have max subs as a proper config (which you could since the only place we'd need it is in the customRouter I think)

If we do see this as a repeated pattern (haven't so far afaik) and want to build some more complex abstracted logic we should put this in the fw

alejoberardino (dismissed):
PR with suggested approach: https://github.com/smartcontractkit/external-adapters-js/pull/5021

packages/sources/nobi/docs/custom.md
mmcallister-cll: Moved `MAX_TRANSPORTS` to be hardcoded in endpoint file and MAX_SUBSCRIPTIONS_PER_TRANSPORT to be in the config. Removed this custom doc and regenerated expected readme

packages/sources/nobi/src/endpoint/transportManager.ts
mmcallister-cll: cool, used most of your suggestions and got rid of the TransportManager in favour of the custom builder ⭐

packages/sources/nobi/src/transport/price.ts
mmcallister-cll: tyty, used most of it

alejoberardino (approved):

PR #5021, opened 2026-06-02
Nobi suggested changes

alejoberardino:
did not update tests, + depends on unmerged EA fw version with changes from Eric


mmcallister-cll:
closed out nobi, took a lot of inspiration from this one, but ultimately didn't fold it in.

PR #5022, opened 2026-06-02
DF-24762: Migrates blocksize-capital to use batch transport builders

cl-efornaciari:
## Description
- blocksize-capital sends 1006 when a burst of subscriptions occur at once
- We leverage the API's support of subscribing to multiple tickers in a single subscription to avoid this
- Apply to all transports

## Changes

- bump framework version to 2.15.0
- update all transports to use batch variants of the subscribe/unsubscribe builder methods

### What changed

Blocksize WebSocket subscribe messages treat the `tickers` list as the **full active subscription set** (replace), not “add these on top.” The initial batch implementation only sent **new** pairs each background-execute cycle, so earlier pairs were dropped as more feeds registered. This PR fixes that with two changes:

1. **Full-set subscribe** — batch subscribe sends already-synced pairs plus new ones (`localSubscriptions` + delta).
2. **Skip idle resubscribe** — no subscribe message when there are no new pairs (matches prod’s per-ticker behavior on idle loops).

---

#### 1. Unit tests (automated, CI)

**File:** `packages/sources/blocksize-capital/test/unit/transport-utils.test.ts`

**Covers:**

- Ticker building and message shape
- Batch subscribe sends **local state + new pairs**, not delta only
- Batch subscribe is **skipped** when there are no new pairs
- Batch unsubscribe still sends the stale delta only
- All three WS transports (price, vwap, crypto-lwba) use batch builders

**Why it matters:** Locks in the Blocksize replace-semantics contract; CI catches accidental reverts.

#### 2. DF1 mainnet payload regression (manual, local vs prod)

**Scope:** Every distinct request body used by **live DF1 mainnet feeds** that list `blocksize-capital` as a data provider.

| | |
|--|--|
| Feed instances | 693 live on mainnet |
| Unique payloads tested | **250** |
| Endpoint | `crypto` (price) |
| Mix | 213 default-path, 37 explicit `transport=ws` |
| Pairs | Full production set, including RDD overrides where applicable |

**Method:** POST the same JSON each node would send to:

- **Prod EA** — current deployed adapter (`adapters.main.prod.cldev.sh/blocksize-capital`)
- **Local EA** — this branch (`localhost:8080`)

**Pass criteria:** HTTP 200, EA `statusCode` 200, numeric `result` present (identical bar for local and prod).

**Execution:** Sequential requests, 180s timeout, retry on 504 with backoff (local is slower on cold WS subscribe; prod typically succeeds on first attempt).

#### 3. Before / after on local (proves the fix)

| Stage | Local result | Notes |
|-------|--------------|-------|
| Before batch fix | **187 / 250** | Delta-only subscribe dropped pairs under load |
| After full-set merge only | **213 / 250** | Better; idle resubscribe still caused churn |
| **After both fixes** | **250 / 250** | Matches prod |

Prod was **250 / 250** at every stage for the same payloads. Failures were local-only and resolved by this PR.

---

### Why we expect no regression on release

1. **Production-real inputs** — The 250 bodies are the actual distinct requests from live DF1 RDD entries using blocksize-capital, not a synthetic subset.

2. **Local branch matches prod on every payload**

   | | Prod (today) | Local (this branch) |
   |--|--------------|---------------------|
   | EA success | **250 / 250** | **250 / 250** |
   | Fail locally, pass on prod | — | **0** |
   | `http/default` | 213 / 213 | 213 / 213 |
   | `ws` | 37 / 37 | 37 / 37 |

3. **Prod baseline validates provider coverage** — Prod already serves all 250 payloads today (per-ticker subscribe). Local 250/250 after the fix shows the **batch path** achieves the same outcomes.

4. **Unit tests guard the two bugs that caused failures** — Full-set merge and skip-idle-resubscribe are covered in CI.

5. **Retries on local are timing, not correctness** — 143/250 local payloads needed retries (cold single-process EA registering ~250 WS pairs); all recovered. Prod needed 0 retries (warm deployment). No pair returned wrong data or a provider error.

---

### Summary

Unit tests enforce batch subscribe/unsubscribe behavior. **250 live DF1 payloads** were run against prod (**250/250**) and this branch locally (**250/250**) with the same pass criteria. Pre-fix local failures (187 → 213 → 250) map directly to the subscribe bugs fixed here; **zero payloads fail locally that succeed on prod.**

## Quality Assurance

### DIFF
--- a/.changeset/short-knives-dress.md
+++ b/.changeset/short-knives-dress.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/blocksize-capital-adapter': minor
+---
+
+Adjusts blocksize-capital EA to use batch builder on transport
--- a/.pnp.cjs
+++ b/.pnp.cjs
@@ -5208,7 +5208,7 @@ const RAW_RUNTIME_STATE =
         "packageLocation": "./packages/sources/blocksize-capital/",\
         "packageDependencies": [\
           ["@chainlink/blocksize-capital-adapter", "workspace:packages/sources/blocksize-capital"],\
-          ["@chainlink/external-adapter-framework", "npm:2.14.6"],\
+          ["@chainlink/external-adapter-framework", "npm:2.15.0"],\
           ["@sinonjs/fake-timers", "npm:9.1.2"],\
           ["@types/jest", "npm:29.5.14"],\
           ["@types/node", "npm:22.14.1"],\
@@ -6340,7 +6340,7 @@ const RAW_RUNTIME_STATE =
         "packageDependencies": [\
           ["@chainlink/external-adapter-framework", "npm:2.14.5"],\
           ["@date-fns/tz", "npm:1.5.0"],\
-          ["@solana/kit", "virtual:e164854d300ea35b7fa3cd6a156e91619b18483c375f0e19e840c2407b9c2b3cb28559dd82b52bd774a11ef6eda89d1c73e1df84aa088be71b59fe686bdc8d15#npm:6.9.0"],\
+          ["@solana/kit", "virtual:fde7122715fe4674cd559fa84a90d4b4c123710c30ab850b0d20be61e32f48e3c7ea9ec823835399ad8a457c7ec3c620c6fce914a30913ec8e978849be19288a#npm:6.9.0"],\
           ["ajv", "npm:8.20.0"],\
           ["axios", "npm:1.16.1"],\
           ["ethers", "npm:6.16.0"],\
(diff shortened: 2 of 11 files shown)


alejoberardino (commented):
lol ok, I assumed blocksize had a batch sub for new subs, not the whole list. Funnily enough this is also what @mmcallister-cll needed for the nobi EA, so we should actually roll back the batch builders if no one will use them and implement something that will serve both (that will send the entire view of subs: new, stale, desired)

PR #5047, opened 2026-06-10
DF-25487 fix request key bug in asseto-finance EA

karen-stepanyan:
Fixes incorrect NAV/reserve responses when multiple fundId values are subscribed concurrently on ea-framework 2.16.1+.

The HTTP request coalesce key was built from requestConfig.data, but Asseto Finance puts fundId in the URL path, not the body. All requests shared DEFAULT_CACHE_KEY, so the Requester returned the same in-flight response for different funds. This was masked on older framework versions by a race in request coalescing; [#805](https://github.com/smartcontractkit/ea-framework-js/pull/805) made coalescing reliable and exposed the bug.

Pass { fundId } into calculateHttpRequestKey so each fund gets a distinct coalesce key.

### DIFF
--- a/.changeset/thick-geese-rush.md
+++ b/.changeset/thick-geese-rush.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/asseto-finance-adapter': patch
+---
+
+Bugfix. Fix request coalescing bug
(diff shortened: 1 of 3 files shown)


mmcallister-cll (approved):

augustbleeds:
thanks, makes sense. had never heard of http coalescing before this

augustbleeds (approved):

PR #5057, opened 2026-06-12
Remove six from market-status

mxiao-cll:
Revert the six changes from: https://github.com/smartcontractkit/external-adapters-js/commit/c0c0ecee59b8bbf4c429e048b1558a0b8cbbfa47

Because six is not deployed yet


karen-stepanyan:
I think there is one test that explicitly checks for `six`

karen-stepanyan (approved):

PR #5095, opened 2026-06-17
OPDATA-7578 Add stSLX exchange rate endpoint

magiodev-cll:
## Closes

OPDATA-7578

## Description

Adds a `stslx-exchange-rate` endpoint to the existing `solana-functions` adapter for the Solstice stSLX-SLX exchange-rate feed.

## Changes

- Register a new `stslx-exchange-rate` endpoint in `solana-functions`.
- Derive the GLAM vault PDA and canonical SLX base-asset ATA for stSLX.
- Read raw SPL account data for SLX mint, stSLX mint, and the SLX base-asset ATA.
- Compute the normalized fixed-point exchange rate using bigint math and return 18-decimal integer strings.
- Add unit and integration coverage plus a changeset.

## Steps to Test

1. `yarn prettier --check "packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts" "packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts" "packages/sources/solana-functions/test/unit/stslx-exchange-rate.test.ts" "packages/sources/solana-functions/test/integration/stslx-exchange-rate.test.ts"`
2. `yarn workspace @chainlink/solana-functions-adapter build`
3. `yarn eslint "packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts" "packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts" "packages/sources/solana-functions/test/unit/stslx-exchange-rate.test.ts" "packages/sources/solana-functions/test/integration/stslx-exchange-rate.test.ts"`
4. `yarn tsc -p "packages/sources/solana-functions/tsconfig.test.json"`
5. `yarn test "packages/sources/solana-functions/test" --runInBand`

## Quality Assurance

- [x] No environment variables changed; no `infra-k8s` update needed.
- [x] No environment variables changed; no `adapter-secrets` update needed.
- [x] No `test-payload.json` exists under `packages/sources/solana-functions`; unit and integration tests were added instead.
- [x] The branch naming follows Jira/git flow: `feat/OPDATA-7578-stslx-exchange-rate`.
- [x] This is related to one Jira story: OPDATA-7578.
- [x] Types avoid unsafe `any`/disable patterns.
- [x] Unit and integration tests cover the new endpoint behavior.

## Automated EA Review and Live Evidence

- Full redacted output from the experimental `ea-review` static + dynamic harness was posted in this PR comment: https://github.com/smartcontractkit/external-adapters-js/pull/5095#issuecomment-4768258577
- Live endpoint test was run against `https://api.mainnet-beta.solana.com`.
- Live `stslx-exchange-rate` response returned `statusCode: 200`, `result: 1009061601987286399`, `computedResult: 1009061601987286399`, and `boundsApplied: false`.


packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,32 @@
...
+import { stslxExchangeRateTransport } from '../transport/stslx-exchange-rate'
+
+// TODO: Add optional address inputs if future stSLX-like feeds need configurable accounts.
+export const inputParameters = new InputParameters({}, [{}])
+
+export type BaseEndpointTypes = {
+  Parameters: typeof inputParameters.definition
+  Response: {
+    Result: string
+    Data: {
```
jcly99, line 75: @magiodev-cll  We will need to add "result" (string) to the data object. Or replace the 'rate: string' with 
`result: string`

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (stslxMint.supply === 0n) {
+      throw new AdapterInputError({
+        message: `stSLX mint '${STSLX_MINT_ADDRESS}' has zero supply`,
+        statusCode: 500,
+      })
+    }
+
+    // Rate is returned as an integer with 18 decimals:
+    //   SLX balance / stSLX supply, normalized by each mint's native decimals.
+    const rate =
```
jcly99: @magiodev-cll Wish-list: To add a max_rate_increase_bps parameter where we can input a BPS as configurable param. Max rate BPS check will ensure the new computed rate does not exceed X% of the current rate, else it assumes the current rate will be the new rate.

This is an additional safeguard rate calculations like the following example to prevent unexpected spikes or dips to the calculation (from a deployed contract, please re-adapt): 
```
 /// @notice Updates the active exchange rate. Caller must have RATE_UPDATER_ROLE.
/// @dev Rate can only increase by at most 5 bps per update.

    function setExchangeRate(uint256 newRate) external onlyRole(RATE_UPDATER_ROLE) {
        if (newRate == 0) revert RateMustBePositive();

        uint256 oldRate = exchangeRate;
        if (newRate <= oldRate) revert RateMustIncrease();

        uint256 increase = newRate - oldRate;
        uint256 maxIncrease = oldRate * MAX_RATE_INCREASE_BPS / BPS_DENOMINATOR;
        if (increase > maxIncrease) revert RateIncreaseTooLarge(increase, maxIncrease);

        exchangeRate = newRate;
        lastUpdatedAt = block.timestamp;

        emit ExchangeRateUpdated(oldRate, newRate, block.timestamp);
    }
```

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,286 @@
...
+import { BaseEndpointTypes, inputParameters } from '../endpoint/stslx-exchange-rate'
+import { SolanaRpcFactory } from '../shared/solana-rpc-factory'
+
+const logger = makeLogger('StslxExchangeRateTransport')
+
+const RESULT_DECIMALS = 18
+
+// stSLX-specific accounts. The endpoint is intentionally hardcoded for this feed today;
+// optional input plumbing can be added later if we need to reuse the same logic elsewhere.
+const GLAM_STATE_ADDRESS = '5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB'
```
jcly99: might want make this as an input param

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
magiodev-cll: agreed. I’ll replace data.rate with data.result.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
magiodev-cll: I’m going to leave this out of this PR for now. The Solidity example has a canonical stored `exchangeRate` to compare against, but this EA is stateless per request and currently only computes the live Solana rate.

I don’t think local adapter `responseCache` should be used as the “current rate” source because it is node-local, can reset on restart, and can diverge across DON nodes. If we want this guard in the EA, I think we need a canonical comparison value supplied explicitly, e.g. `currentRate` + `maxRateIncreaseBps` as request params from the jobspec, or this should live in feed/OCR config instead. Happy to add it once the source of the current accepted rate is confirmed.

magiodev-cll: Follow-up: implemented this as stateless jobspec-configured bounds with required `minRate` and `maxRate` request params, rather than using a cached/current-rate BPS comparison.

The endpoint now:
- Computes `computedResult` from the deterministic GLAM SLX `base_asset_ata` and stSLX mint supply.
- Validates `minRate` / `maxRate` as positive base-10 integer strings and enforces `minRate <= maxRate`.
- Clamps the computed rate into `[minRate, maxRate]`.
- Returns `data.result` as the final bounded value, plus `data.computedResult`, `data.minRate`, `data.maxRate`, and `data.boundsApplied` for observability.

This gives us explicit upper/lower guardrails against transient RPC/read-skew spikes or dips without relying on adapter `responseCache`, which is node-local, reset-prone, and not a canonical source of the last accepted rate across DON nodes.

I also added unit/integration coverage for bounds and required input validation, and a live smoke test with the final required params returned `statusCode: 200` with `boundsApplied: false`.

magiodev-cll:
Spawning the local server:
```
➜  external-adapters-js git:(feat/OPDATA-7578-stslx-exchange-rate) ✗ RPC_URL=https://api.mainnet-beta.solana.com/ \
EA_HOST=127.0.0.1 \
EA_PORT=18080 \
METRICS_ENABLED=false \
yarn workspace @chainlink/solana-functions-adapter server:dist

{"level":"warn","time":1781695665025,"isoTime":"2026-06-17T11:27:45.025Z","pid":85088,"hostname":"MB-JYCTKQD9RJ","layer":"Adapter","msg":"METRICS_ENABLED has been set to false. Metrics should not be disabled in a production environment."}
(node:85088) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
{"level":"info","time":1781695665025,"isoTime":"2026-06-17T11:27:45.025Z","pid":85088,"hostname":"MB-JYCTKQD9RJ","layer":"CacheFactory","msg":"Using \"local\" cache."}
{"level":"info","time":1781695665058,"isoTime":"2026-06-17T11:27:45.058Z","pid":85088,"hostname":"MB-JYCTKQD9RJ","layer":"Main","msg":"Starting background execution loop"}
{"level":"info","time":1781695665062,"isoTime":"2026-06-17T11:27:45.062Z","pid":85088,"hostname":"MB-JYCTKQD9RJ","layer":"Main","msg":"Listening on port 18080"}
```

Testing it:
```
➜  external-adapters-js git:(feat/OPDATA-7578-stslx-exchange-rate) ✗ curl -sS -X POST http://127.0.0.1:18080/ \
  -H 'Content-Type: application/json' \
  --data '{
    "data": {
      "endpoint": "stslx-exchange-rate",
      "glamStateAddress": "5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB",
    }
  }' | jq
{
  "data": {
    "result": "1006963459819162228",
    "computedResult": "1006963459819162228",
    "decimals": 18,
    "maxRate": "1050000000000000000",
    "boundsApplied": false
  },
  "statusCode": 200,
  "result": "1006963459819162228",
  "timestamps": {
    "providerDataRequestedUnixMs": 1781695670121,
    "providerDataReceivedUnixMs": 1781695670318
  }
}
➜  external-adapters-js git:(feat/OPDATA-7578-stslx-exchange-rate) ✗
```

magiodev-cll:
## Automated EA Review Harness Output + Live API Evidence

PR: #5095
Branch: `feat/OPDATA-7578-stslx-exchange-rate`
Commit tested: `9d0ad80a3`
Adapter: `@chainlink/solana-functions-adapter`
Endpoint: `stslx-exchange-rate`

### Static Harness Command

```bash
EA_REVIEW_NO_TELEMETRY=1 node <ea-review>/dist/cli.js scan \
  -a solana-functions \
  -r <external-adapters-js> \
  -p ci-fast \
  --severity critical,high \
  -f markdown
```

### Static Harness Output

## EA Review Results

**Profile:** ci-fast | **Adapters:** solana-functions | **Duration:** 89ms

### Summary

| Metric | Count |
|--------|-------|
| Total | 70 |
| Passed | 28 |
| Failed | 42 |
| Critical | 5 |
| High | 37 |

### solana-functions

<details>
<summary>🔴 <b>DATA-002</b> NaN/Infinity Result `src/transport/eusx-price.ts:90`</summary>

**Severity:** critical
**Message:** Division operation without NaN/Infinity guard; add isFinite() check or guard before computing
**Why:** NaN or Infinity propagating as a price result causes the Chainlink node to submit invalid data on-chain, potentially triggering incorrect liquidations or contract behavior
**Fix:** Add Number.isNaN() and Number.isFinite() checks after numeric operations before returning results

```typescript
numerator / denominator
```

</details>

<details>
<summary>🔴 <b>DATA-002</b> NaN/Infinity Result `src/transport/sanctum-infinity.ts:76`</summary>

**Severity:** critical
**Message:** Division operation without NaN/Infinity guard; add isFinite() check or guard before computing
**Why:** NaN or Infinity propagating as a price result causes the Chainlink node to submit invalid data on-chain, potentially triggering incorrect liquidations or contract behavior
**Fix:** Add Number.isNaN() and Number.isFinite() checks after numeric operations before returning results

```typescript
Number(totalPoolValueString.result) / Number(totalPoolTokenSupplyString.result)
```

</details>

<details>
<summary>🔴 <b>DATA-002</b> NaN/Infinity Result `src/transport/sanctum-infinity.ts:76`</summary>

**Severity:** critical
**Message:** Number() call without NaN validation; add Number.isNaN() check or return null/default if invalid
**Why:** NaN or Infinity propagating as a price result causes the Chainlink node to submit invalid data on-chain, potentially triggering incorrect liquidations or contract behavior
**Fix:** Add Number.isNaN() and Number.isFinite() checks after numeric operations before returning results

```typescript
Number(totalPoolValueString.result)
```

</details>

<details>
<summary>🔴 <b>DATA-002</b> NaN/Infinity Result `src/transport/sanctum-infinity.ts:76`</summary>

**Severity:** critical
**Message:** Number() call without NaN validation; add Number.isNaN() check or return null/default if invalid
**Why:** NaN or Infinity propagating as a price result causes the Chainlink node to submit invalid data on-chain, potentially triggering incorrect liquidations or contract behavior
**Fix:** Add Number.isNaN() and Number.isFinite() checks after numeric operations before returning results

```typescript
Number(totalPoolTokenSupplyString.result)
```

</details>

<details>
<summary>🔴 <b>DATA-002</b> NaN/Infinity Result `src/transport/stslx-exchange-rate.ts:218`</summary>

**Severity:** critical
**Message:** Division operation without NaN/Infinity guard; add isFinite() check or guard before computing
**Why:** NaN or Infinity propagating as a price result causes the Chainlink node to submit invalid data on-chain, potentially triggering incorrect liquidations or contract behavior
**Fix:** Add Number.isNaN() and Number.isFinite() checks after numeric operations before returning results

```typescript
(slxBalance * 10n ** BigInt(RESULT_DECIMALS + stslxMint.decimals)) /
      (stslxMint.supply * 10n ** BigInt(slxMint.decimals))
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/shared/account-reader.ts:16`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
  ): Promise<T> {
    const [pda] = await getProgramDerivedAddress({
      programAddress,
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/eusx-price.ts:99`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
    const programAddress = params.address as Address
    const [vestingSchedule, yieldPool] = await Promise.all([
      accountReader.fetchAccountInformationByAddressAndSeeds<VestingSchedule>(
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/extension.ts:54`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
    const extensionData = buffer.slice(offset, offset + extensionLength)
    extensionDataByType[extensionType] = extensionData
    offset += extensionLength
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/extension.ts:119`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
      for (const baseField of params.baseFields) {
        resultData[baseField.name] = getFieldFromBuffer(baseField, data)
      }
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/extension.ts:127`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
      for (const extensionField of params.extensionFields) {
        resultData[extensionField.name] = getExtensionField(extensionField, extensionDataByType)
      }
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/extension.ts:132`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
    if (params.resultDecimals !== undefined) {
      resultData['decimals'] = params.resultDecimals
    }
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/sanctum-infinity.ts:62`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript

    const [totalPoolValueString, totalPoolTokenSupplyString] = await Promise.all([
      fetchFieldFromBufferLayoutStateAccount({
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/stslx-exchange-rate.ts:128`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
  // GLAM stores token assets in a vault PDA derived from the state account and protocol program.
  const [vaultAddress] = PublicKey.findProgramAddressSync(
    [Buffer.from('vault'), new PublicKey(glamStateAddress).toBuffer()],
```

</details>

<details>
<summary>🟠 <b>SEC-007</b> Prototype Pollution `src/transport/stslx-exchange-rate.ts:202`</summary>

**Severity:** high
**Message:** Detects potential prototype pollution via dynamic property assignment
**Why:** Prototype pollution can corrupt shared JavaScript objects, potentially altering price data across all adapter responses or bypassing validation logic
**Fix:** Validate property keys against a whitelist before dynamic assignment; use Object.create(null) for dictionaries

```typescript
    // divided by stSLX mint supply, normalized by each mint's native decimals.
    const [slxMint, stslxMint, slxBalance] = await Promise.all([
      this.fetchMintInfo(SLX_MINT_ADDRESS, 'SLX mint'),
```

</details>

<details>
<summary>🟠 <b>CFG-001</b> Missing Rate Limiting `src/index.ts:1`</summary>

**Severity:** high
**Message:** Missing rate limiting for http transport; add rateLimiting config to prevent API quota exhaustion
**Why:** Missing rate limiting on HTTP/REST transports can exhaust API quotas, causing cascading failures. Streaming transports (WebSocket, batch) don't hit per-request API quotas and are exempt.
**Fix:** Add rateLimiting configuration to HTTP/REST adapters; WebSocket/batch adapters can safely omit this

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/anchor-data.ts:39`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/buffer-layout.ts:29`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/buffer-layout.ts:59`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript

    const result = await fetchFieldFromBufferLayoutStateAccount({
      stateAccountAddress: params.stateAccountAddress,
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/eusx-price.ts:61`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  ) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/eusx-price.ts:99`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
    const programAddress = params.address as Address
    const [vestingSchedule, yieldPool] = await Promise.all([
      accountReader.fetchAccountInformationByAddressAndSeeds<VestingSchedule>(
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/extension.ts:80`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/pool-token-rate.ts:31`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/pool-token-rate.ts:61`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript

    const { programAddress, data } = await fetchDataFromBufferLayoutStateAccount({
      stateAccountAddress: params.stakePoolAccountAddress,
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/sanctum-infinity.ts:32`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/sanctum-infinity.ts:62`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript

    const [totalPoolValueString, totalPoolTokenSupplyString] = await Promise.all([
      fetchFieldFromBufferLayoutStateAccount({
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/stslx-exchange-rate.ts:158`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
  async backgroundHandler(context: EndpointContext<BaseEndpointTypes>, entries: RequestParams[]) {
```

</details>

<details>
<summary>🟠 <b>DATA-001</b> Missing Error Response `src/transport/stslx-exchange-rate.ts:202`</summary>

**Severity:** high
**Message:** Network/API call without try-catch error handling
**Why:** Unhandled API errors crash the adapter or return undefined results to the Chainlink node, causing failed oracle reports and potential node penalties
**Fix:** Wrap all network/API calls in try-catch blocks and return proper error responses

```typescript
    // divided by stSLX mint supply, normalized by each mint's native decimals.
    const [slxMint, stslxMint, slxBalance] = await Promise.all([
      this.fetchMintInfo(SLX_MINT_ADDRESS, 'SLX mint'),
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/anchor-data.ts:56`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/anchor-data.ts:117`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/buffer-layout.ts:46`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/buffer-layout.ts:76`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/eusx-price.ts:78`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/eusx-price.ts:165`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/extension.ts:97`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/extension.ts:142`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/pool-token-rate.ts:48`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/pool-token-rate.ts:92`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/sanctum-infinity.ts:49`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/sanctum-infinity.ts:87`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/stslx-exchange-rate.ts:175`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>ERR-002</b> Zero/Null Timestamp in Error Response `src/transport/stslx-exchange-rate.ts:241`</summary>

**Severity:** high
**Message:** Timestamp set to undefined in response — can corrupt cache TTL
**Why:** Generic error handling without distinguishing error types (network vs parsing vs validation) prevents proper recovery strategies and makes debugging harder
**Fix:** Always return Date.now() as timestamps in error responses to prevent cache TTL corruption

```typescript
providerIndicatedTimeUnixMs: undefined
```

</details>

<details>
<summary>🟠 <b>CHAIN-001</b> RPC Timeout `src/transport/stslx-exchange-rate.ts:1`</summary>

**Severity:** high
**Message:** RPC provider configured without explicit timeout — calls can hang indefinitely
**Why:** RPC calls without timeout can hang indefinitely when a node is unresponsive, blocking the adapter.
**Fix:** Add explicit timeout to JsonRpcProvider or ethers provider configuration

</details>

### Dynamic Harness Command

```bash
EA_REVIEW_NO_TELEMETRY=1 node <ea-review>/dist/cli.js scan \
  -a solana-functions \
  -r <external-adapters-js> \
  -p dynamic-only \
  --dynamic \
  --dynamic-port 18295 \
  --dynamic-env RPC_URL=https://api.mainnet-beta.solana.com \
  -f markdown
```

### Dynamic Harness Output

## EA Review Results

**Profile:** dynamic-only | **Adapters:** solana-functions | **Duration:** 5.5s

### Summary

| Metric | Count |
|--------|-------|
| Total | 12 |
| Passed | 8 |
| Failed | 4 |
| High | 3 |
| Medium | 1 |

### solana-functions

<details>
<summary>🟠 <b>DYN-002</b> Runtime Price Validation</summary>

**Severity:** high
**Message:** Expected status 200, got 404
**Why:** Price feeds are the core function of an EA — a request that returns non-numeric or non-positive data will poison oracle reports
**Fix:** Verify the adapter's default endpoint handles standard base/quote pairs and returns a numeric result > 0

</details>

<details>
<summary>🟠 <b>DYN-006</b> Data Quality Bounds</summary>

**Severity:** high
**Message:** Non-200 response: 404
**Why:** Feeding NaN, Infinity, or extreme values into an oracle report can corrupt on-chain price feeds and trigger liquidations or arbitrage exploits
**Fix:** Add result validation that rejects non-finite, non-positive, or unreasonably large numbers before returning

</details>

<details>
<summary>🟠 <b>DYN-012</b> Error Recovery</summary>

**Severity:** high
**Message:** Adapter returned 404 after recovery attempt
**Why:** If a single bad request leaves the adapter in a broken state, all subsequent requests from every node will fail until the process is restarted
**Fix:** Ensure error paths do not corrupt shared state — use try/catch around request handling and reset any per-request state on failure

</details>

<details>
<summary>🟡 <b>DYN-009</b> Response Consistency</summary>

**Severity:** medium
**Message:** Non-numeric results: undefined, undefined
**Why:** Large swings between identical requests may indicate race conditions, stale cache issues, or unstable upstream data
**Fix:** Investigate why consecutive identical requests return significantly different results — check caching and data source stability

</details>

### Live Endpoint Runtime Command

```bash
EA_PORT=18205 \
RPC_URL=https://api.mainnet-beta.solana.com \
METRICS_ENABLED=false \
yarn node -e "require('./dist/index.js').server()"
```

### Live Endpoint Request/Response

```json
{
  "rpcUrl": "https://api.mainnet-beta.solana.com",
  "request": {
    "data": {
      "endpoint": "stslx-exchange-rate",
      "glamStateAddress": "5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB",
    }
  },
  "response": {
    "statusCode": 200,
    "elapsedMs": 1008,
    "body": {
      "data": {
        "result": "1009079081993244827",
        "computedResult": "1009079081993244827",
        "decimals": 18,
        "maxRate": "1050000000000000000",
        "boundsApplied": false
      },
      "statusCode": 200,
      "result": "1009079081993244827",
      "timestamps": {
        "providerDataRequestedUnixMs": 1782130917209,
        "providerDataReceivedUnixMs": 1782130917432
      }
    }
  }
}
```

magiodev-cll:
## Follow-up: EA Review Finding Scope

The automated `ea-review` harness scans the full existing `solana-functions` adapter, while this PR only adds/registers the new `stslx-exchange-rate` endpoint. I split the posted harness findings by whether they are in code introduced by this PR versus pre-existing adapter files.

### Static Harness Scope

Static harness failures reported: `42`

| Classification | Count | Notes |
| --- | ---: | --- |
| Findings in the new stSLX transport | 8 | These are the only findings directly in newly added endpoint code. |
| Adapter entrypoint heuristic | 1 | `src/index.ts` was touched to register the endpoint, but the finding is adapter-wide. |
| Pre-existing `solana-functions` files | 33 | These are outside this PR's endpoint implementation. |

Findings directly in new stSLX code:

| Rule | Location |
| --- | --- |
| `DATA-002` | `src/transport/stslx-exchange-rate.ts:218` |
| `SEC-007` | `src/transport/stslx-exchange-rate.ts:128` |
| `SEC-007` | `src/transport/stslx-exchange-rate.ts:202` |
| `DATA-001` | `src/transport/stslx-exchange-rate.ts:158` |
| `DATA-001` | `src/transport/stslx-exchange-rate.ts:202` |
| `ERR-002` | `src/transport/stslx-exchange-rate.ts:175` |
| `ERR-002` | `src/transport/stslx-exchange-rate.ts:241` |
| `CHAIN-001` | `src/transport/stslx-exchange-rate.ts:1` |

Touched but not specific to the new endpoint:

| Rule | Location | Why it is adapter-wide |
| --- | --- | --- |
| `CFG-001` | `src/index.ts:1` | This flags adapter-level rate limiting. This PR only touches `src/index.ts` to register the endpoint. |

Pre-existing files with findings:

- `src/transport/eusx-price.ts`
- `src/transport/sanctum-infinity.ts`
- `src/shared/account-reader.ts`
- `src/transport/extension.ts`
- `src/transport/anchor-data.ts`
- `src/transport/buffer-layout.ts`
- `src/transport/pool-token-rate.ts`

### Dynamic Harness Scope

Dynamic harness failures reported:

- `DYN-002`
- `DYN-006`
- `DYN-009`
- `DYN-012`

These are not clearly caused by the new stSLX endpoint. The experimental dynamic harness sends generic/default adapter requests and received `404` / non-numeric responses for value-oriented scenarios. The endpoint-specific live request in the previous comment is the stronger runtime evidence for this PR and returned `statusCode: 200` with `result == computedResult` and `boundsApplied: false`.

### Summary

Most harness findings are from pre-existing `solana-functions` code. The PR-relevant static findings are limited to the new `stslx-exchange-rate` transport plus one adapter-wide entrypoint heuristic. The live endpoint test confirms the new endpoint returns a successful response against the actual Solana RPC path.

magiodev-cll:
## Follow-up: Why the PR-local EA review findings are no-fix

This is a concise rationale for the findings that landed in files introduced or touched by this PR. I am not treating all findings as generic false positives; some are real-ish but adapter-wide or cross-cutting. I do not see a required endpoint code change for this PR.

| Finding | Scope | Decision | Rationale |
| --- | --- | --- | --- |
| `DATA-002` at `src/transport/stslx-exchange-rate.ts:218` | New stSLX transport | No fix | The calculation uses `bigint`, not JS `number`, so it cannot produce `NaN` or `Infinity`. The relevant denominator, `stslxMint.supply`, is checked for `0n` before division. Adding `Number.isFinite()` would be wrong for this type. |
| `SEC-007` at `src/transport/stslx-exchange-rate.ts:128` and `:202` | New stSLX transport | No fix | These are tuple/array destructuring patterns from `PublicKey.findProgramAddressSync` and `Promise.all`. Prototype pollution requires dynamic object-key writes such as `obj[userInput] = value`; that is not happening here. |
| `DATA-001` at `src/transport/stslx-exchange-rate.ts:158` and `:202` | New stSLX transport | No fix | The scanner sees async/RPC calls, but this transport follows the existing `solana-functions` pattern: `backgroundHandler` calls `handleRequest`, and `handleRequest` wraps `_handleRequest` in `try/catch` and writes an adapter error response. |
| `ERR-002` at `src/transport/stslx-exchange-rate.ts:175` and `:241` | New stSLX transport | No fix | This endpoint has no provider-supplied timestamp. Setting `providerIndicatedTimeUnixMs` to local wall-clock time would be misleading. The existing adapter uses this same pattern for on-chain reads/errors. |
| `CHAIN-001` at `src/transport/stslx-exchange-rate.ts:1` | Shared RPC behavior surfaced in new file | No fix in this PR | This is the one finding I would consider real engineering debt, but it belongs in shared `SolanaRpcFactory`/adapter config work because every Solana endpoint uses the same RPC factory. A one-off endpoint patch would be inconsistent. |
| `CFG-001` at `src/index.ts:1` | Adapter entrypoint touched for endpoint registration | No fix in this PR | The finding is adapter-wide rate-limiting heuristics. This PR only touches `src/index.ts` to register `stslxExchangeRate`; it does not introduce a new external HTTP provider quota path. |

Dynamic harness note: the `DYN-*` failures came from generic/default adapter requests returning `404` or non-numeric results. The endpoint-specific live request for `stslx-exchange-rate` returned HTTP `200` with `result == computedResult` and `boundsApplied: false`, so I do not treat those generic dynamic failures as endpoint bugs.

Suggested follow-up outside this PR: if the team wants to address the most legitimate scanner concern, create a separate shared task for explicit Solana RPC timeout support in `SolanaRpcFactory`.

mohamed-mehany (changes requested):
## Review summary

Thanks @magiodev-cll  for putting this together. 
To be upfront: I went through the logic and I believe while it maybe functional and matches the methodology in [OPDATA-7578](https://smartcontract-it.atlassian.net/browse/OPDATA-7578) (vault SLX balance / stSLX supply, bigint fixed-point, min/max clamp). 
The problem is the **state of the code**, it's quite hard to read and review, and it reads as largely LLM-generated. That's fine as a starting point, but for this to meet Chainlink standards it needs to follow our best practices, reuse the abstractions we already have, and be reliable/maintainable enough for the next person to review and own. Right now it forks its own parallel idiom instead of building on the adapter.

Could you please refactor this and, when you do, **explicitly instruct the model to follow best practices, prefer existing abstractions, and not reinvent what the adapter already provides**? Also, for large PRs like this one, please use a stronger model (ideally **Opus 4.8** or equivalent) to avoid the code smells below. There should be **no magic numbers or arbitrary hardcoded values** left in the final version; every constant should be named and traceable to a source (IDL, on-chain address, or config).

## Correctness issues

- **Non-atomic reads cause the exact read-skew the ticket warns about.** The SLX balance and stSLX supply are fetched as three independent `getAccountInfo` calls via `Promise.all`, so they can resolve at different slots and produce a transiently wrong rate. Use a single `getMultipleAccounts` call to read them atomically (same slot). Note the strcUSX PR already does this correctly, please mirror that.
- **Confirm the ATA assumption with Solstice/GLAM.** We derive the canonical associated token account for the vault's SLX, whereas the ticket's reference code enumerates `getTokenAccountsByOwner`. If GLAM ever holds SLX in a non-ATA account, this endpoint breaks. It fails safe today (mint/owner are validated), but the assumption should be confirmed and documented.

## Code quality / structure

- **Reuse existing abstractions.** The adapter already has `SolanaAccountReader` and `shared/buffer-layout-accounts.ts` (typed `MintLayout`/`AccountLayout` decode + owner/length checks). This PR re-implements raw RPC reads and buffer slicing with hand-rolled response types (`AccountInfoRpcResponse`, `as` casts) instead of extending those. Please build on the shared readers.
- **De-duplicate shared helpers.** `parseRateBound`, `getAccountDataBuffer`, `assertOwnerProgram`/`assertTokenProgramOwner`, the bounds-clamp, and the `AccountInfo`/`MintInfo` types are duplicated verbatim with the strcUSX PR. Extract them into `src/shared/` so both endpoints share one implementation.
- **Stick to one Solana stack.** New code pulls in legacy `@solana/web3.js` `PublicKey` and `@solana/spl-token` helpers, while the rest of the adapter uses the modern `@solana/addresses` / `@solana/rpc` stack (`getProgramDerivedAddress`). Please align with the existing stack rather than mixing both.
- **No hardcoded values / magic numbers.** Pull the bounds-clamp out into a single named helper instead of an inline nested ternary, and ensure any constant (decimals, addresses) is a named, documented constant.
- **`boundsApplied`** is computed by string comparison (`result !== computedResult`); compare the bigint values instead.

mohamed-mehany:
## Suggestion: extract a shared-components PR and stack these on top

Stepping back from the per-PR comments, both PRs (#5095 (stSLX) and #5097 (strcUSX)) and endpoints duplicate a substantial amount of the same code (`parseRateBound`, `getAccountDataBuffer`, `assertOwnerProgram`/`assertTokenProgramOwner`, the min/max bounds-clamp, the
shared `AccountInfo`/`MintInfo` types, and the raw account-read plumbing). 

Rather than refactoring each PR independently and reconciling the two copies later, it
would likely be cleaner to:

1. **Open a third PR first** that lands the shared building blocks in
   `src/shared/`, e.g. rate-bound parsing/validation, the bounds-clamp helper,
   account-buffer extraction + owner/length assertions, and (ideally) IDL/typed
   `buffer-layout` decoding built on the existing `SolanaAccountReader` /
   `buffer-layout-accounts.ts`. This PR has no new endpoint and is small and easy
   to review on its own.
2. **Rebase/stack #5095 (stSLX) and #5097 (strcUSX) on top of it**, so each
   endpoint PR shrinks to just its endpoint definition, its decode layout, and its
   rate formula, no duplicated scaffolding.

Benefits:
- One canonical implementation of the shared logic instead of two copies that can
  drift.
- Each endpoint PR becomes much smaller and genuinely reviewable.
- The shared module can be unit-tested once, in isolation.
- Future Solana exchange-rate endpoints get the same foundation for free.

mohamed-mehany:
Two follow-ups:

1. **vaultPDA derivation.** The vault PDA (and SLX ATA) are derived per-request,
   but for this feed they're constant. Also note the endpoint is half-parameterized:
   `glamStateAddress` is a request param while the mints/program are hardcoded, so
   it's not actually generic. Either hardcode the resolved vault/ATA as named
   constants with a unit test asserting the derivation matches, or make it fully
   generic (mints + program as params too). Not the current half-and-half.

2. **Reuse.** This is essentially "token-account balance / mint supply." The adapter
   already has `shared/buffer-layout-accounts.ts` (decodes `AccountLayout`/`MintLayout`).
   Please reuse it instead of the bespoke `fetchMintInfo`/`fetchBaseAssetBalance`.

danwilliams-cll:
Code Review: PR #5095 — stSLX exchange rate: use new shared helpers

Verdict: APPROVE — tight, focused refactoring. Three minor points, none blocking.

---
What Changed (+ lines only)

Adopts assertTokenProgramOwner, derivePda, and parseSolanaAddress from PR #5128. Adds deriveVaultAddress and deriveSlxTokenAccountAddress as exported helpers. Exports intermediate addresses (GLAM_STATE_ADDRESS, GLAM_PROTOCOL_PROGRAM_ADDRESS, ASSOCIATED_TOKEN_PROGRAM_ADDRESS) that were previously implicit.

---
Findings

stslx-exchange-rate.ts:L59: [LOW] correctness: parseSolanaAddress(LEGACY_TOKEN_PROGRAM_ADDRESS, 'tokenProgramAddress') validates a known-good constant (TOKEN_PROGRAM_ID.toBase58()). This call can never throw — redundant guard. Pass it directly or cast with address().

stslx-exchange-rate.ts:L51: [LOW] shrink: vaultAddress.toString() — derivePda returns Address which already extends string. No-op. Return vaultAddress directly.

stslx-exchange-rate.ts:L44-61: [LOW] yagni: deriveVaultAddress and deriveSlxTokenAccountAddress are exported but _handleRequest still uses the hardcoded GLAM_VAULT_ADDRESS and SLX_TOKEN_ACCOUNT_ADDRESS constants. If these only exist to verify the hardcoded values in tests, a single test assertion is enough — no need to export them from the transport module.

jcly99 (approved):

danwilliams-cll (approved):

packages/sources/solana-functions/README.md
```diff
@@ -239,4 +239,31 @@ Request:
...
+
+Request:
+
+```json
+{
+  "data": {
+    "endpoint": "stslx-exchange-rate",
+    "minRate": "950000000000000000",
+    "maxRate": "1050000000000000000"
+  }
```
jcly99, line 271: consider: 
1. The ability to enable/disable rateBound. For ie: applyBounds(boolean)

mohamed-mehany:
1. Reuse miss: inline assertAddressMatches. The two mint/owner checks are hand-written if (slxToken.x !== Y) throw, but #5128 now exports assertAddressMatches. Swap them, it's the same helper #5097 already uses, and trims the transport.
2. Those two checks are also largely redundant. A token account's mint and owner are immutable after init, and the derivation test already proves SLX_TOKEN_ACCOUNT_ADDRESS is the canonical ATA of the vault for the SLX mint. So at runtime the decoded mint/owner can't differ (a missing account is already caught by getAccountDataBuffer). If they're not needed, let's remove them.

3. Status codes: upstream conditions ("zero supply", missing account) throw AdapterInputError with 500. Consistent with the EA, but AdapterDataProviderError would label provider failures correctly in metrics.

4. Minor response over-echo: minRate/maxRate are echoed back in Data (caller already supplied them), no need to return inputs in the output (unless they're needed somehow).

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,163 @@
...
+  getAccountDataBuffer,
+} from '../shared/solana-account-utils'
+import { SolanaRpcFactory } from '../shared/solana-rpc-factory'
+
+const logger = makeLogger('StslxExchangeRateTransport')
+
+export const SLX_MINT_ADDRESS = 'SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq'
+export const STSLX_MINT_ADDRESS = 'GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq'
+export const GLAM_VAULT_ADDRESS = 'GMwdh2jTdTrrhA7dMR7Cc2zC6gV38UePzAXeoFHrXnfH'
+export const SLX_TOKEN_ACCOUNT_ADDRESS = '7CssRFNePpnDiCzjRC5kPRDpEJn87JMeDG7s6Gww9CTf'
```
mohamed-mehany: On the hardcoded addresses in the transport (`SLX_MINT_ADDRESS`, `STSLX_MINT_ADDRESS`, `GLAM_VAULT_ADDRESS`, `SLX_TOKEN_ACCOUNT_ADDRESS`): it works and the derivation test is a good mitigation, but I'd reconsider the approach. Three points:

**1. It diverges from this EA's convention.** The other endpoints parameterize their addresses via `inputParameters` with a default rather than hardcoding them in the transport, e.g. `eusx-price` takes `address` as a required param with a default, and `pool-token-rate`/`anchor-data` take their account addresses as params. Parameterizing with defaults gives the same out-of-the-box behavior plus: the addresses get documented in the generated README, they're visible at the job-spec layer, and they're overridable without a code change. And there's no determinism cost, since all NOPs share the same defaults so the computed rate is identical across operators.

**2. Hardcoding makes the feed un-reconfigurable without a release.** If Solstice migrates the GLAM vault, redeploys, or the mint changes, this needs a code change + EA release + redeploy instead of a config/job-spec update. That's a slow path for a live price feed.

**3. Two of the four constants are derived values and shouldn't be hardcoded at all.** `GLAM_VAULT_ADDRESS` and `SLX_TOKEN_ACCOUNT_ADDRESS` are deterministic functions of the primary addresses:
- vault = PDA of `["vault", glamState]` under the GLAM program
- SLX ATA = ATA of `(vault, SLX_MINT)` under the legacy token program

Hardcoding a derived value adds stale-risk for no benefit, and the derivation test is already computing them to prove the constants, so the code derives them twice. Cleaner to derive vault + ATA once from the source addresses via the shared `derivePda` (`@solana/addresses`, no web3.js, no I/O), which removes the constants-to-keep-in-sync and the need for the derivation test.

One caveat on that test: it only proves the constants are internally consistent (given the test's glamState/program/mint, the vault/ATA derive correctly). It does not prove the primary addresses are the correct real-world ones, so real-world correctness still rests on the mainnet smoke test.

**Suggested shape:** expose the primary addresses (or just `glamState`) as input params with defaults, and derive vault + ATA at runtime from them via `derivePda`. If you'd rather keep it locked/hardcoded for a single feed (also fine), then at least only hardcode the source-of-truth addresses (program, mints, glamState) and derive vault + ATA from them rather than hardcoding derived values.

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,209 @@
...
+  )
+
+const asProviderError = <T>(callback: () => T) => {
+  try {
+    return callback()
+  } catch (e: unknown) {
+    throw providerError(e instanceof Error ? e.message : 'Unknown provider error')
+  }
+}
+
```
mohamed-mehany: **`providerError` is duplicated with #5097.** Extract it to `src/shared/` rather than defining it in both transports.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,209 @@
...
+        boundsApplied,
+      },
+      statusCode: 200,
+      result,
+      timestamps: {
+        providerDataRequestedUnixMs,
+        providerDataReceivedUnixMs: Date.now(),
+        providerIndicatedTimeUnixMs: undefined,
+      },
+    }
```
mohamed-mehany, line 166: Please add `slxBalance` + `stslxSupply` so a wrong rate is debuggable from the response and the two endpoints are consistent.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,209 @@
...
+    addressEncoder.encode(parseSolanaAddress(glamStateAddress, 'glamStateAddress')),
+  ])
+
+const deriveSlxTokenAccountAddress = (vaultAddress: string, slxMintAddress: string) =>
+  derivePda(ASSOCIATED_TOKEN_PROGRAM_ADDRESS, [
+    addressEncoder.encode(parseSolanaAddress(vaultAddress, 'vaultAddress')),
+    addressEncoder.encode(address(LEGACY_TOKEN_PROGRAM_ADDRESS)),
+    addressEncoder.encode(parseSolanaAddress(slxMintAddress, 'slxMintAddress')),
+  ])
+
```
mohamed-mehany, line 48: **Redundant address re-parsing.** `parseSolanaAddress` runs in `_handleRequest`, then again inside `deriveVaultAddress`/`deriveSlxTokenAccountAddress`, then again inside `derivePda`. Idempotent but wasteful/confusing; consider threading the parsed `Address` through.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,187 @@
...
+type RequestParams = typeof inputParameters.validated
+
+const asProviderError = <T>(callback: () => T) => {
+  try {
+    return callback()
+  } catch (e: unknown) {
+    throw providerError(e instanceof Error ? e.message : 'Unknown provider error')
+  }
+}
```
mohamed-mehany: **Delete `asProviderError` once #5128 fixes the error taxonomy.** It exists only to convert the shared helpers' `AdapterInputError` 500 into a 502. Once the shared account/decode helpers throw `AdapterDataProviderError` directly (see #5128), the wrapper and all its call sites (lines 117, 118, 121, 130, 136, 142) collapse to plain calls. This also removes the current inconsistency where strcUSX does not wrap and emits 500 for the same failures.

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,88 @@
...
+import { type AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+import { parseRateBounds } from '../shared/exchange-rate-utils'
+import { stslxExchangeRateTransport } from '../transport/stslx-exchange-rate'
+
+export const DEFAULT_SLX_MINT_ADDRESS = 'SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq'
+export const DEFAULT_STSLX_MINT_ADDRESS = 'GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq'
+export const DEFAULT_GLAM_STATE_ADDRESS = '5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB'
+export const DEFAULT_GLAM_PROTOCOL_PROGRAM_ADDRESS = 'GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz'
+
```
mohamed-mehany, line 21: I might be confused, but why did we revert to hardcode all addresses as opposed to deriving them? Didn't we already address this in this comment: https://github.com/smartcontractkit/external-adapters-js/pull/5095#discussion_r3473745512

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,88 @@
...
+import { InputParameters } from '@chainlink/external-adapter-framework/validation'
+import { type AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
+import { parseRateBounds } from '../shared/exchange-rate-utils'
+import { stslxExchangeRateTransport } from '../transport/stslx-exchange-rate'
+
+export const DEFAULT_SLX_MINT_ADDRESS = 'SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq'
+export const DEFAULT_STSLX_MINT_ADDRESS = 'GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq'
+export const DEFAULT_GLAM_STATE_ADDRESS = '5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB'
+export const DEFAULT_GLAM_PROTOCOL_PROGRAM_ADDRESS = 'GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz'
```
mohamed-mehany, line 20: Also for any hardcoded address/constant, please document how we set this and where did we source it from and what happens if it changes.

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
magiodev-cll: Also addressed in https://github.com/smartcontractkit/external-adapters-js/pull/5095/commits/a81dc7eb80d5a649eaf8950557d35fb6431580d2

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
magiodev-cll: yeah i see where the confusion comes from. these are default input parameter fallbacks, not hardcoded transport addresses. The GLAM vault and SLX ATA are still derived at runtime in stslx-exchange-rate.ts from the source addresses

btw I added source/change documentation for the default address params, and also replaced the associated token program hardcode with the @solana/spl-token exported constant.

please check the new commit: https://github.com/smartcontractkit/external-adapters-js/pull/5095/commits/a81dc7eb80d5a649eaf8950557d35fb6431580d2

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
mohamed-mehany: @magiodev-cll The explorer links are redundant, can we please remove them? by commenting the source I meant an external official source where we get those addresses from, if the only source is the ticket or via comms with the Solstice team, then we can mention that.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    assertTokenProgramOwner(slxMintAccount, `SLX mint '${slxMintAddress}'`)
+    assertTokenProgramOwner(stslxMintAccount, `stSLX mint '${stslxMintAddress}'`)
+    assertOwnerProgram(
+      slxTokenAccount,
+      `SLX token account '${slxTokenAccountAddress}'`,
+      [LEGACY_TOKEN_PROGRAM_ADDRESS],
+      'the legacy SPL Token program',
+    )
```
mohamed-mehany: slxTokenAccount is derived so this assert is also redundant and can be removed.

mohamed-mehany (approved):
A few comments, please resolve before merging into the #5128 PR/branch

PR #5128, opened 2026-06-24
OPDATA-7578: Refactor Solana exchange-rate shared utilities

magiodev-cll:
## Summary

This is the entry-point PR for OPDATA-7578 Solana exchange-rate support. It now includes the shared Solana account/rate utilities from this branch plus the two endpoint PRs that were merged into it:

- #5095: stSLX-SLX exchange-rate endpoint
- #5097: strcUSX tranche exchange-rate endpoint

The requirement is to support three Solstice exchange-rate streams through the `solana-functions` EA:

| Stream | Endpoint | Network used for local smoke test | Notes |
| --- | --- | --- | --- |
| stSLX-SLX | `stslx-exchange-rate` | Solana mainnet | Uses Solstice-provided production feed/config defaults from OPDATA-7578 and Solstice comms. |
| strcUSX junior | `strcusx-exchange-rate` | Solana devnet | Current Solstice strcUSX deployment is on devnet. |
| strcUSX senior | `strcusx-exchange-rate` | Solana devnet | Same deployed devnet program/config as junior, with `tranche: "senior"`. |

`RPC_URL` remains a required adapter env var. Set it explicitly to the Solana cluster RPC for the target feed; the local smoke tests below supplied mainnet/devnet RPC URLs in the server environment.

## Closes

OPDATA-7578

## Solution

This PR adds shared utilities and endpoint implementations needed to read the required Solana account state deterministically and return normalized 18-decimal exchange rates.

### Shared Solana/rate utilities

- Adds shared PDA derivation via `@solana/addresses`.
- Adds shared Solana account helpers for address validation, multi-account reads, base64 data extraction, provider-style errors, and owner/data assertions.
- Adds shared Clock sysvar decoding via `@solana/sysvars`, used for deterministic on-chain time in vesting calculations.
- Adds SPL mint/token-account prefix decoding that works for legacy SPL Token and Token-2022 mint/account data with extension bytes.
- Adds shared exchange-rate helpers for 18-decimal normalization, optional min/max rate bounds, and vesting/unvested-asset calculations.
- Splits rate-bound validation from conversion: endpoints validate in `customInputValidation`, transports only convert validated strings to `bigint`.
- Moves provider-style account/RPC/decode failures to 502s via shared account utilities, while keeping genuine request input validation as 400s.
- Refactors the existing Anchor account reader to reuse the shared `derivePda` path and tightens PDA seeds from `any[]` to `PdaSeed[]`.
- Updates generated README docs for the new endpoints and request params.

### Bounds and safeguards

- `minRate` and `maxRate` are stateless request/job-spec guardrails for all three feeds.
- The final response `result` is clamped into `[minRate, maxRate]` when bounds are supplied.
- `computedResult` remains the unclamped on-chain computed rate for observability.
- `boundsApplied` indicates whether clamping happened.
- A warning log is emitted when bounds are applied.

### `stslx-exchange-rate`

- Reads the GLAM vault's canonical SLX ATA balance as the SLX backing balance.
- Reads stSLX mint supply directly.
- Supports the current Solstice-provided default addresses as request/job-spec fallbacks:

```json
{
  "endpoint": "stslx-exchange-rate",
  "slxMintAddress": "SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq",
  "stslxMintAddress": "GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq",
  "glamStateAddress": "5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB",
  "glamProtocolProgramAddress": "GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz",
}
```

- Derives the GLAM vault PDA from the GLAM protocol program, configured GLAM state address, and `"vault"` seed.
- Derives the SLX ATA using the legacy SPL Token program assumption for SLX. stSLX is read directly and can be owned by legacy SPL Token or Token-2022.
- Returns the computed result, final bounded result, decimals, bounds status, SLX balance, and stSLX supply.

### `strcusx-exchange-rate`

- Supports `tranche: "junior"` and `tranche: "senior"` from a shared `TRANCHES` tuple/type source of truth.
- Requires Solstice strategy params for the current devnet deployment:

```json
{
  "endpoint": "strcusx-exchange-rate",
  "programAddress": "7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe",
  "strategyName": "STRC-USX-1",
  "tranche": "junior",
}
```

```json
{
  "endpoint": "strcusx-exchange-rate",
  "programAddress": "7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe",
  "strategyName": "STRC-USX-1",
  "tranche": "senior",
}
```

- Derives controller, strategy, and accounting PDAs from the Solstice yield-strategy program and strategy seed.
- Decodes Solstice Anchor account data via the committed strcUSX IDL.
- Reads Solana Clock sysvar and removes unvested assets before computing tranche book values.
- Prices senior as vested senior assets divided by senior shares.
- Prices junior as residual vested assets divided by junior shares.
- Returns the computed result, final bounded result, tranche, decimals, bounds status, tranche assets, and tranche shares.

## Linked PRs

- #5095: OPDATA-7578 Add stSLX exchange-rate endpoint
- #5097: OPDATA-7578 Add strcUSX tranche exchange-rate endpoint

## Release / Generated Artifacts

| File | Purpose |
| --- | --- |
| `.changeset/quiet-ducks-compare.md` | Patch changeset for shared Solana exchange-rate utilities. |
| `.changeset/sharp-lions-serve.md` | Minor changeset for the stSLX exchange-rate endpoint. |
| `.changeset/fresh-tigers-pay.md` | Minor changeset for the strcUSX junior/senior exchange-rate endpoint. |
| `packages/sources/solana-functions/README.md` | Regenerated endpoint/input-param docs for `stslx-exchange-rate` and `strcusx-exchange-rate`. |
| `packages/sources/solana-functions/package.json` / `yarn.lock` | Adds `@solana/sysvars@3.0.2`; keeps `@chainlink/external-adapter-framework@2.17.1`. |

## Local Validation

Latest local validation after the final review fixes:

| Command | Result |
| --- | --- |
| `yarn test packages/sources/solana-functions/test/unit` | Passed: 11 suites, 85 tests. |
| `yarn test packages/sources/solana-functions/test/integration --runInBand` | Passed: 8 suites, 17 tests, 10 snapshots. |
| `yarn tsc -p packages/sources/solana-functions/tsconfig.test.json` | Passed. |
| `yarn workspace @chainlink/solana-functions-adapter build` | Passed. |
| `yarn prettier --check packages/sources/solana-functions` | Passed. |
| `yarn eslint --ext ts packages/sources/solana-functions/src packages/sources/solana-functions/test/unit/exchange-rate-utils.test.ts packages/sources/solana-functions/test/unit/solana-account-utils.test.ts packages/sources/solana-functions/test/unit/stslx-exchange-rate.test.ts packages/sources/solana-functions/test/unit/strcusx-exchange-rate.test.ts` | Passed with 0 errors; existing warnings only. |
| `git diff --check` | Passed. |

Review-driven test coverage included in the above suites:

- Shared utilities: rate-bound validation/conversion, clamp behavior, 18-decimal normalization, vesting math, Solana address/PDA helpers, account data extraction, owner/data assertions, Clock sysvar decoding, and multi-account fetch behavior.
- Existing Anchor account reader: PDA derivation now routes through the shared helper with typed `PdaSeed[]` seeds.
- Buffer-layout helpers: SPL mint/token-account prefix decoders and supported token-program owner checks are covered, including Token-2022 extension bytes.
- stSLX: normalized rate calculation, min/max clamping, warning logs when bounds apply, zero stSLX supply error, missing derived SLX ATA error, and integration coverage for default params and invalid bounds.
- strcUSX: junior and senior tranche rates, recorded real account fixture decoding, vesting/unvested asset handling, rounding/flooring behavior, Token-2022 mint accounts with extension bytes, `getMultipleAccounts().send()` rejection producing cached 502 responses, min/max clamping, warning logs when bounds apply, inconsistent account-state error, and integration coverage for both tranche values and invalid bounds.

## Local Endpoint Curl Testing

Local server commands used:

```bash
EA_PORT=18081 RPC_URL="https://api.mainnet-beta.solana.com" METRICS_ENABLED=false yarn workspace @chainlink/solana-functions-adapter start
EA_PORT=18081 RPC_URL="https://api.devnet.solana.com" METRICS_ENABLED=false yarn workspace @chainlink/solana-functions-adapter start
```

Coverage summary:

| Endpoint | Network | Coverage | Result |
| --- | --- | --- | --- |
| `stslx-exchange-rate` | Mainnet | All 64 present/omitted combinations of optional endpoint fields: `slxMintAddress`, `stslxMintAddress`, `glamStateAddress`, `glamProtocolProgramAddress`, `minRate`, `maxRate`. Production default values were used when optional fields were present. | All returned `statusCode: 200`, `result: "1012435281927010494"`, `boundsApplied: false`. |
| `strcusx-exchange-rate` | Devnet | Both tranche values across all 4 present/omitted combinations of optional bounds: bounds omitted, `minRate` only, `maxRate` only, both bounds. | All returned `statusCode: 200`. Junior result: `"1000000002222222222"`. Senior result: `"1000000000000000000"`. `boundsApplied: false`. |

### stSLX-SLX mainnet curl

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"stslx-exchange-rate","slxMintAddress":"SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq","stslxMintAddress":"GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq","glamStateAddress":"5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB","glamProtocolProgramAddress":"GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1012435281927010494","computedResult":"1012435281927010494","decimals":18,"boundsApplied":false,"slxBalance":"3741442753241","stslxSupply":"3695488314196"},"statusCode":200,"result":"1012435281927010494","timestamps":{"providerDataRequestedUnixMs":1782824099796,"providerDataReceivedUnixMs":1782824099855}}
```

### strcUSX junior devnet curl

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"strcusx-exchange-rate","programAddress":"7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe","strategyName":"STRC-USX-1","tranche":"junior","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1000000002222222222","computedResult":"1000000002222222222","tranche":"junior","decimals":18,"boundsApplied":false,"trancheAssets":"450000001","trancheShares":"450000000"},"statusCode":200,"result":"1000000002222222222","timestamps":{"providerDataRequestedUnixMs":1782823786378,"providerDataReceivedUnixMs":1782823786604,"providerIndicatedTimeUnixMs":1782823785000}}
```

### strcUSX senior devnet curl

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"strcusx-exchange-rate","programAddress":"7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe","strategyName":"STRC-USX-1","tranche":"senior","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1000000000000000000","computedResult":"1000000000000000000","tranche":"senior","decimals":18,"boundsApplied":false,"trancheAssets":"200000000","trancheShares":"200000000"},"statusCode":200,"result":"1000000000000000000","timestamps":{"providerDataRequestedUnixMs":1782823995257,"providerDataReceivedUnixMs":1782823995472,"providerIndicatedTimeUnixMs":1782823994000}}
```

## Operational Context

- `stslx-exchange-rate` was tested against mainnet because the stSLX-SLX accounts are live on mainnet.
- `strcusx-exchange-rate` was tested against devnet because the current Solstice strcUSX deployment is on devnet.
- The endpoint request defaults are limited to Solstice-provided feed/config values where applicable.
- Set `RPC_URL` to the Solana cluster RPC for the target feed.
- Input-validation failures return 400; Solana RPC/account/decode/provider failures return provider-style 502 errors.
- stSLX default addresses are sourced from OPDATA-7578 and Solstice comms. If Solstice migrates the feed config, these should be overridden at the request/job-spec layer first and only changed in code when production defaults change.
- Further staging validation is expected after this PR is merged and deployed.


mohamed-mehany:
Thank you @magiodev-cll for the refactor! 

A few gaps to close before #5095/#5097 stack on this, mostly in the **on-chain access 
layer**
Note: apologies if by the time I post this some of these are already addressed. 


**1. Extract `derivePda` and put it on `@solana/addresses`, not web3.js.**

Both endpoint PRs hand-roll PDA derivation (#5097 has 5 near-identical
`findProgramAddressSync` wrappers, #5095 has the vault + ATA derivation). None of that
is abstracted here, so the duplication survives the rebase. Just as important: those
wrappers use legacy `@solana/web3.js` `PublicKey.findProgramAddressSync`, but on `main`
`solana-functions` uses `@solana/addresses` exclusively (`getProgramDerivedAddress` in
`account-reader.ts`), web3.js isn't imported anywhere in this EA's `src`. Please add a
shared `derivePda(programAddress, seeds)` built on `@solana/addresses` so the whole EA
stays on one stack.

**2. Extract a shared Clock-sysvar reader.**

The vesting *math* (`calculateUnvestedAssets`) was extracted, but the on-chain
timestamp read that feeds it (#5097's `CLOCK_SYSVAR_ADDRESS` / offset /
`decodeClockUnixTimestamp`) was left inline. Reading the on-chain clock is the core
determinism primitive and is reused by strcUSX (and a future eUSX retrofit), it belongs
in shared next to the vesting math.

**3. Fold in the remaining duplicated helpers.**

`assertTokenProgramOwner` was duplicated verbatim in both PRs, the constant moved here
but not the wrapper. `parsePublicKey` (input validation) and `assertNameMatches` /
`assertAddressMatches` are generic and were left behind too. Either extract them or
consciously drop them.

**4. Manual byte readers**

`assertDiscriminator` was extracted but the readers that accompany it
(`readU128LE`/`readU64LE`/`readPublicKey`/`readPaddedString`) were not, so half the
manual-decode toolkit is shared and half isn't. Preferred direction is still to decode
strcUSX via the program IDL (or a named `buffer-layout` struct) and commit the IDL,
which makes most of these readers unnecessary. 

**5. Token-2022 + the exact-span `getLayout` map (correctness).**

Adding `token-2022 -> [AccountLayout, MintLayout]` to `programToBufferLayoutMap` is
fragile: `getLayout` matches on exact `layout.span === dataLength`, but Token-2022
accounts can carry extension bytes (and the stSLX mint is Token-2022), so that path will
throw. The new `decodeMintInfo`/`decodeTokenAccountInfo` are fine (they assert `>= span`
and decode the prefix), so prefer those and either drop the token-2022 entry from the
exact-span map or make `getLayout` tolerate `>= span` for token programs.

**6. the hand-rolled `AccountInfo` type + `as MultipleAccountsRpcResponse` cast
diverges slightly from how the existing readers lean on the `@solana/rpc` SDK types, are the custom/handrolled types needed?

magiodev-cll:
> Thank you @magiodev-cll for the refactor!
> 
> A few gaps to close before #5095/#5097 stack on this, mostly in the **on-chain access layer** Note: apologies if by the time I post this some of these are already addressed.
> 
> **1. Extract `derivePda` and put it on `@solana/addresses`, not web3.js.**
> 
> Both endpoint PRs hand-roll PDA derivation (#5097 has 5 near-identical `findProgramAddressSync` wrappers, #5095 has the vault + ATA derivation). None of that is abstracted here, so the duplication survives the rebase. Just as important: those wrappers use legacy `@solana/web3.js` `PublicKey.findProgramAddressSync`, but on `main` `solana-functions` uses `@solana/addresses` exclusively (`getProgramDerivedAddress` in `account-reader.ts`), web3.js isn't imported anywhere in this EA's `src`. Please add a shared `derivePda(programAddress, seeds)` built on `@solana/addresses` so the whole EA stays on one stack.
> 
> **2. Extract a shared Clock-sysvar reader.**
> 
> The vesting _math_ (`calculateUnvestedAssets`) was extracted, but the on-chain timestamp read that feeds it (#5097's `CLOCK_SYSVAR_ADDRESS` / offset / `decodeClockUnixTimestamp`) was left inline. Reading the on-chain clock is the core determinism primitive and is reused by strcUSX (and a future eUSX retrofit), it belongs in shared next to the vesting math.
> 
> **3. Fold in the remaining duplicated helpers.**
> 
> `assertTokenProgramOwner` was duplicated verbatim in both PRs, the constant moved here but not the wrapper. `parsePublicKey` (input validation) and `assertNameMatches` / `assertAddressMatches` are generic and were left behind too. Either extract them or consciously drop them.
> 
> **4. Manual byte readers**
> 
> `assertDiscriminator` was extracted but the readers that accompany it (`readU128LE`/`readU64LE`/`readPublicKey`/`readPaddedString`) were not, so half the manual-decode toolkit is shared and half isn't. Preferred direction is still to decode strcUSX via the program IDL (or a named `buffer-layout` struct) and commit the IDL, which makes most of these readers unnecessary.
> 
> **5. Token-2022 + the exact-span `getLayout` map (correctness).**
> 
> Adding `token-2022 -> [AccountLayout, MintLayout]` to `programToBufferLayoutMap` is fragile: `getLayout` matches on exact `layout.span === dataLength`, but Token-2022 accounts can carry extension bytes (and the stSLX mint is Token-2022), so that path will throw. The new `decodeMintInfo`/`decodeTokenAccountInfo` are fine (they assert `>= span` and decode the prefix), so prefer those and either drop the token-2022 entry from the exact-span map or make `getLayout` tolerate `>= span` for token programs.
> 
> **6. the hand-rolled `AccountInfo` type + `as MultipleAccountsRpcResponse` cast diverges slightly from how the existing readers lean on the `@solana/rpc` SDK types, are the custom/handrolled types needed?

Addressed in `[bd429a7a0](https://github.com/smartcontractkit/external-adapters-js/pull/5128/changes/bd429a7a0ebca2b58b618a9fbb52249a4a646cdc)`.

For point 5: removed Token-2022 from the exact-span `programToBufferLayoutMap`, so generic `getLayout` no longer tries to match Token-2022 accounts by exact length. The prefix-safe `decodeMintInfo` / `decodeTokenAccountInfo` helpers remain the intended path for SPL/Token-2022 mint/account decoding and still assert `>= span`.

For point 6: removed the hand-rolled `MultipleAccountsRpcResponse` cast, and changed `AccountInfo` to be derived from the `@solana/rpc` `getMultipleAccounts(...).send()` return type instead of maintaining our own response shape. `fetchMultipleAccounts` now validates addresses through `@solana/addresses` before calling the SDK.

Also added the shared access-layer pieces from the earlier points: `derivePda` on `@solana/addresses`, shared Clock sysvar decoding, shared token-program owner assertion, and generic address/name assertions.

danwilliams-cll:
---
Code Review: PR #5128 — "refactor: expand Solana account helpers"

**Verdict: APPROVE** — Focused, clean refactoring. Stronger typing from SDK derivation, address validation added to the RPC path, clock sysvar decoding centralised. Two minor points worth noting before merge.

---
What Changed (actual + lines)

- solana-account-utils.ts — AccountInfo replaced with a type derived from the actual getMultipleAccounts RPC return type (strictly stronger than the hand-rolled version). Added parseSolanaAddress, derivePda, assertNameMatches, assertAddressMatches, decodeClockUnixTimestamp. fetchMultipleAccounts now validates addresses before the RPC call.
- account-reader.ts — seeds: any[] tightened to PdaSeed[]. Inline PDA derivation replaced with the new derivePda util.
- buffer-layout-accounts.ts — Token 2022 removed from programToBufferLayoutMap. assertTokenProgramOwner wrapper added.

---
Findings

buffer-layout-accounts.ts:80-84: [LOW] yagni: assertTokenProgramOwner wraps assertOwnerProgram with two args baked in — one caller in this PR. Inline at the call site; export when a second caller exists.

solana-account-utils.ts:7-8: [LOW] correctness: CLOCK_ACCOUNT_LENGTH = 40 and CLOCK_UNIX_TIMESTAMP_OFFSET = 32 are correct but unattributed to the Solana clock sysvar layout. A comment or link makes decodeClockUnixTimestamp self-auditable without knowing the spec.

danwilliams-cll (dismissed):

packages/sources/solana-functions/src/shared/exchange-rate-utils.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const RESULT_DECIMALS = 18
+const POSITIVE_INTEGER_PATTERN = /^[1-9]\d*$/
+
+export const parseRateBound = (value: string, name: string) => {
+  if (!POSITIVE_INTEGER_PATTERN.test(value)) {
+    throw new AdapterInputError({
+      message: `${name} must be a positive base-10 integer string`,
+      statusCode: 400,
+    })
```
mohamed-mehany, line 11: Please consider validation using `customInputValidation`, [example](https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/tokenized-equity/src/endpoint/price.ts#L24)

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export const derivePda = async (programAddress: string, seeds: PdaSeed[]) => {
+  const [pda] = await getProgramDerivedAddress({
+    programAddress: parseSolanaAddress(programAddress, 'programAddress'),
+    seeds,
+  })
+
+  return pda
+}
```
mohamed-mehany, line 51: Now that `derivePda` lives here (and `PdaSeed` is exported just above), `account-reader.ts` should consume it rather than keeping its own derivation path. Today `SolanaAccountReader.fetchAccountInformationByAddressAndSeeds` still does:

```ts
async fetchAccountInformationByAddressAndSeeds<T>(
  rpc, programAddress: Address, seeds: any[], accountName, idl,
): Promise<T> {
  const [pda] = await getProgramDerivedAddress({ programAddress, seeds })
}
```
which is the same `getProgramDerivedAddress` call `derivePda` wraps. So after this PR there are two derivation paths in the EA: the new shared one (used by the strcUSX/stSLX transports) and this inline one (used by eusx-price/anchor-data via the reader).

Two asks:

- Drop the any. The seeds: any[] has a stale comment ("typed as any due to type not being exported by @solana/addresses"), but that's now obsolete since this PR exports PdaSeed. So seeds: any[] -> seeds: PdaSeed[] removes an any with zero behavioral change.
- Route the reader through `derivePda` so there's a single derivation function `(const pda = await derivePda(programAddress, seeds))`. One impedance note: `derivePda` takes a string and re-validates via `parseSolanaAddress`, while the reader already has a typed Address, passing the Address through works (it's a branded string, re-validation is a no-op), or derivePda could accept Address | string.

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export const assertAddressMatches = (actual: string, expected: string, description: string) => {
+  if (actual !== expected) {
+    throw new AdapterInputError({
+      message: `Expected ${description} to be '${expected}', found '${actual}'`,
+      statusCode: 500,
+    })
+  }
+}
```
mohamed-mehany: Remove dead exports.** `assertNameMatches` and `assertAddressMatches` are no longer used by any endpoint (strcUSX dropped the redundant equality checks). Remove them and their unit tests.

packages/sources/solana-functions/src/shared/exchange-rate-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+      message,
+      statusCode: 502,
+    },
+    {
+      providerDataRequestedUnixMs: 0,
+      providerDataReceivedUnixMs: 0,
+      providerIndicatedTimeUnixMs: undefined,
+    },
+  )
+
```
mohamed-mehany: **Relocate `providerError`.** It currently lives in `exchange-rate-utils.ts` but it's an error/account concern, not rate math. Move it into `solana-account-utils.ts` (or a small `errors.ts`) so `exchange-rate-utils.ts` is purely rate math + bounds.

packages/sources/solana-functions/src/shared/solana-account-utils.ts
mohamed-mehany, line 79: **Fix the error taxonomy in `solana-account-utils.ts`.** The account/decode helpers throw `AdapterInputError` with `statusCode: 500`, but these are upstream/data failures, not input errors. They should throw `AdapterDataProviderError` (502) via the `providerError` factory:
   - `getAccountDataBuffer` 
   - `assertOwnerProgram` 
   - `assertDataLength` 
   - `assertDiscriminator` 
   - `fetchMultipleAccounts`
   Keep `parseSolanaAddress` on `AdapterInputError` 400 since that's genuine input validation. This fix lets the stSLX PR delete its `asProviderError` wrapper and makes both endpoints emit consistent 502s for the same failures (today strcUSX emits 500 for these).

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/shared/exchange-rate-utils.ts
mohamed-mehany: @magiodev-cll I think this one was missed, please let me know if it's unclear.

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -0,0 +1,127 @@
+import {
+  AdapterDataProviderError,
+  AdapterInputError,
+} from '@chainlink/external-adapter-framework/validation/error'
+import { address, getProgramDerivedAddress } from '@solana/addresses'
+import { type Rpc, type SolanaRpcApi } from '@solana/rpc'
+
+export const CLOCK_SYSVAR_ADDRESS = 'SysvarC1ock11111111111111111111111111111111'
```
mohamed-mehany: Can we please document how we set those constants and where did we source them from?

mohamed-mehany (changes requested):

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -5,9 +5,12 @@ import {
 import { address, getProgramDerivedAddress } from '@solana/addresses'
 import { type Rpc, type SolanaRpcApi } from '@solana/rpc'
 
+// Sources:
+// - address: @solana/web3.js `SYSVAR_CLOCK_PUBKEY`
+//   https://github.com/solana-foundation/solana-web3.js/blob/maintenance/v1.x/src/sysvar.ts
```
mohamed-mehany: @magiodev-cll Any reason not to simply import it from there rather than hardcode it? since we already depend on the solana libraries

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -5,9 +5,12 @@ import {
...
+// - layout: Solana SDK `Clock` has five 8-byte fields, with `unix_timestamp` fifth
+//   https://github.com/anza-xyz/solana-sdk/blob/master/clock/src/lib.rs
 export const CLOCK_SYSVAR_ADDRESS = 'SysvarC1ock11111111111111111111111111111111'
-
-// Solana Clock sysvar layout stores unix_timestamp as an i64 at byte offset 32.
 const CLOCK_ACCOUNT_LENGTH = 40
 const CLOCK_UNIX_TIMESTAMP_OFFSET = 32
```
mohamed-mehany: @magiodev-cll Same applies to these constants as well, import if possible, otherwise please document.

packages/sources/solana-functions/src/shared/solana-account-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+export const assertDiscriminator = (data: Buffer, description: string, discriminator: Buffer) => {
+  if (!data.subarray(0, discriminator.length).equals(discriminator)) {
+    throw providerError(
+      `Expected ${description} discriminator to be ${discriminator.toString('hex')}, found ${data
+        .subarray(0, discriminator.length)
+        .toString('hex')}`,
+    )
+  }
+}
+
```
mohamed-mehany: This seems to be dead code, please remove

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,174 @@
...
+const deriveVaultAddress = (glamStateAddress: Address, glamProtocolProgramAddress: Address) =>
+  derivePda(glamProtocolProgramAddress, [GLAM_VAULT_SEED, addressEncoder.encode(glamStateAddress)])
+
+const deriveSlxTokenAccountAddress = (vaultAddress: Address, slxMintAddress: Address) =>
+  derivePda(ASSOCIATED_TOKEN_PROGRAM_ADDRESS, [
+    addressEncoder.encode(vaultAddress),
+    addressEncoder.encode(address(LEGACY_TOKEN_PROGRAM_ADDRESS)),
+    addressEncoder.encode(slxMintAddress),
+  ])
+
```
mohamed-mehany, line 52: This part deserves a bit of documentation.
Where did we source the seed, the methodology for deriving the vault address and the token account address.

packages/sources/solana-functions/src/endpoint/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// Defaults come from Solstice's OPDATA-7578 production stSLX feed config. They
+// are request/job-spec fallbacks; if Solstice migrates, override these params
+// first and update defaults only when the production feed defaults change. The
+// GLAM vault and SLX ATA are derived from these source addresses in transport.
+// Explorer checks:
+// - https://explorer.solana.com/address/SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq
+// - https://explorer.solana.com/address/GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq
+// - https://explorer.solana.com/address/5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB
+// - https://explorer.solana.com/address/GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz
```
mohamed-mehany: I think [this one](https://github.com/smartcontractkit/external-adapters-js/pull/5095/changes#r3495096956) was missed from the previous PR, can we please address it here?

packages/sources/solana-functions/src/config/index.ts
```diff
@@ -4,7 +4,7 @@ export const config = new AdapterConfig({
   RPC_URL: {
     description: 'The RPC URL for the Solana cluster',
     type: 'string',
-    required: true,
+    default: 'https://api.mainnet-beta.solana.com',
```
mohamed-mehany: Let's please remove this change.
We would like to avoid to use the public APIs accidentally by any user, so not having a default there and having it required is intentional

packages/sources/solana-functions/src/shared/exchange-rate-utils.ts
```diff
@@ -0,0 +1,75 @@
+import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+
+export const RESULT_DECIMALS = 18
+const POSITIVE_INTEGER_PATTERN = /^[1-9]\d*$/
+
+export const parseRateBound = (value: string, name: string) => {
```
mohamed-mehany: Can we please rename this to validateRateBound and use it accordingly only in the customInput validation, and everywhere else simply convert so that we don't double validate everywhere?

packages/sources/solana-functions/src/transport/strcusx-exchange-rate.ts
```diff
@@ -0,0 +1,196 @@
...
+    )
+    const strategy = decodeStrategyState(
+      getAccountDataBuffer(strategyAccount, `Strategy account '${strategyAddress}'`),
+    )
+    const accounting = decodeAccountingState(
+      getAccountDataBuffer(accountingAccount, `Accounting account '${accountingAddress}'`),
+    )
+
+    const trancheMintAddress =
+      tranche === 'junior' ? strategy.juniorMintAddress : strategy.seniorMintAddress
```
mohamed-mehany: Once types are added, please use it to check instead of hardcoded strings.

packages/sources/solana-functions/src/endpoint/strcusx-exchange-rate.ts
```diff
@@ -0,0 +1,84 @@
...
+      description:
+        'Solstice strcUSX strategy/accounting PDA seed from the current deployment/feed config',
+      type: 'string',
+      options: ['STRC-USX-1'],
+      required: true,
+    },
+    tranche: {
+      description: 'The tranche to price: junior or senior',
+      type: 'string',
+      options: ['junior', 'senior'],
```
mohamed-mehany: For both parameters, it's cleaner to define the options in a different way.
For example: `tranche` is currently declared three ways: this inline `options` literal, the hand-written `type Tranche = 'junior' | 'senior'` in `transport/strcusx-accounts.ts`, and the `as Tranche` cast in the transport. They can drift, and the cast is actually unnecessary — the framework already narrows the validated param to the union of `options` (`TypeFromOptionsOrTypeString` + the `const` type param on `InputParameters`).

Let's collapse to the established `gmx-tokens` convention (`CHAIN_OPTIONS` / `ChainKey` in `transport/shared/chain.ts`): one `as const` tuple as the single source of truth, with the type derived from it.

```ts
// transport/strcusx-accounts.ts
export const TRANCHES = ['junior', 'senior'] as const
export type Tranche = (typeof TRANCHES)[number]
// endpoint/strcusx-exchange-rate.ts
import { TRANCHES } from '../transport/strcusx-accounts'
tranche: {
  description: 'The tranche to price: junior or senior',
  type: 'string',
  options: [...TRANCHES],
  required: true,
},
```
Then the cast in the transport can go (see other comment). 
Also we can tighten Response.Data.tranche from string to Tranche too.
For an example, see: https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/gmx-tokens/src/endpoint/glv-price.ts#L18, https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/gmx-tokens/src/transport/glv-price.ts#L21

packages/sources/solana-functions/src/transport/strcusx-exchange-rate.ts
```diff
@@ -0,0 +1,196 @@
...
+    await this.responseCache.write(this.name, [{ params: param, response }])
+  }
+
+  async _handleRequest(
+    params: RequestParams,
+  ): Promise<AdapterResponse<BaseEndpointTypes['Response']>> {
+    const providerDataRequestedUnixMs = Date.now()
+    const programAddress = parseSolanaAddress(params.programAddress, 'programAddress').toString()
+    const strategyName = parseStrategyName(params.strategyName)
+    const tranche = params.tranche as Tranche
```
mohamed-mehany: With the `TRANCHES`-derived type (see endpoint comment), `params.tranche` is already typed `'junior' | 'senior'`, so this cast is a no-op, and a cast would mask a wrong value rather than catch it (the framework's runtime `options` validation is what actually guarantees it). Drop it:

```ts
const tranche = params.tranche
```
type Tranche in strcusx-accounts.ts currently has no other consumer, so this is the only change needed besides the tuple refactor.

packages/sources/solana-functions/src/transport/stslx-exchange-rate.ts
```diff
@@ -0,0 +1,174 @@
...
+const deriveVaultAddress = (glamStateAddress: Address, glamProtocolProgramAddress: Address) =>
+  derivePda(glamProtocolProgramAddress, [GLAM_VAULT_SEED, addressEncoder.encode(glamStateAddress)])
+
+const deriveSlxTokenAccountAddress = (vaultAddress: Address, slxMintAddress: Address) =>
+  derivePda(ASSOCIATED_TOKEN_PROGRAM_ADDRESS, [
+    addressEncoder.encode(vaultAddress),
+    addressEncoder.encode(address(LEGACY_TOKEN_PROGRAM_ADDRESS)),
+    addressEncoder.encode(slxMintAddress),
+  ])
```
mohamed-mehany, line 51: The ATA seed hardcodes the legacy SPL Token program, so this derivation is only correct if SLX is a legacy SPL mint; a Token-2022 SLX would derive a different address. 
That's fine for the current mint, but the assumption is implicit. 
Could we add a one-line comment noting SLX is assumed legacy-SPL (and that stSLX can be either program, since its mint is read directly with no ATA derived)? Keeps the assumption discoverable without re-introducing an assert.

mohamed-mehany (changes requested):
Left a few comments.
Also the PR description is stale (since we merged the other 2 PRs) so can we please rewrite it to reflect the real scope since this would be the entrypoint PR as a reference? 

We can add a high level description of the requirement and scope and link the other 2 PRs.
Can we also add the final tests that was run locally for all endpoints with their results?

magiodev-cll:
## Local endpoint curl testing

Tested the new endpoints locally with explicit `RPC_URL` values. `RPC_URL` remains required in the adapter config; these URLs were supplied only in the local server environment for smoke testing.

Server commands used:

```bash
EA_PORT=18081 RPC_URL="https://api.mainnet-beta.solana.com" METRICS_ENABLED=false yarn workspace @chainlink/solana-functions-adapter start
EA_PORT=18081 RPC_URL="https://api.devnet.solana.com" METRICS_ENABLED=false yarn workspace @chainlink/solana-functions-adapter start
```

Coverage summary:

- `stslx-exchange-rate` on mainnet: ran all 64 present/omitted combinations of optional endpoint fields (`slxMintAddress`, `stslxMintAddress`, `glamStateAddress`, `glamProtocolProgramAddress`, `minRate`, `maxRate`) using the production default values when present. All returned `statusCode: 200`, `result: "1012435281927010494"`, `boundsApplied: false`.
- `strcusx-exchange-rate` on devnet: ran both tranche values (`junior`, `senior`) across all 4 present/omitted combinations of optional bounds (`minRate`, `maxRate`). All returned `statusCode: 200`; junior result was `"1000000002222222222"`, senior result was `"1000000000000000000"`, and `boundsApplied: false`.

### stSLX-SLX mainnet

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"stslx-exchange-rate","slxMintAddress":"SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq","stslxMintAddress":"GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq","glamStateAddress":"5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB","glamProtocolProgramAddress":"GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1012435281927010494","computedResult":"1012435281927010494","decimals":18,"boundsApplied":false,"slxBalance":"3741442753241","stslxSupply":"3695488314196"},"statusCode":200,"result":"1012435281927010494","timestamps":{"providerDataRequestedUnixMs":1782824099796,"providerDataReceivedUnixMs":1782824099855}}
```

EA params:

```json
{
  "endpoint": "stslx-exchange-rate",
  "slxMintAddress": "SLXdx4BUt2v9uJQNzWqSfzTJ9UKLUDsvxHFMEEdrfgq",
  "stslxMintAddress": "GxHksENo754dKj6kv5d2z7ey9KwE7YSRYgRCtoFYd2yq",
  "glamStateAddress": "5E2scHi8LyZAqZeVHnXLeFhwoePxD2CTdSruWmjgVEoB",
  "glamProtocolProgramAddress": "GLAMpaME8wdTEzxtiYEAa5yD8fZbxZiz2hNtV58RZiEz",
}
```

### strcUSX junior devnet

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"strcusx-exchange-rate","programAddress":"7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe","strategyName":"STRC-USX-1","tranche":"junior","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1000000002222222222","computedResult":"1000000002222222222","tranche":"junior","decimals":18,"boundsApplied":false,"trancheAssets":"450000001","trancheShares":"450000000"},"statusCode":200,"result":"1000000002222222222","timestamps":{"providerDataRequestedUnixMs":1782823786378,"providerDataReceivedUnixMs":1782823786604,"providerIndicatedTimeUnixMs":1782823785000}}
```

EA params:

```json
{
  "endpoint": "strcusx-exchange-rate",
  "programAddress": "7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe",
  "strategyName": "STRC-USX-1",
  "tranche": "junior",
}
```

### strcUSX senior devnet

```bash
curl -sS -X POST "http://localhost:18081" \
  --data '{"data":{"endpoint":"strcusx-exchange-rate","programAddress":"7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe","strategyName":"STRC-USX-1","tranche":"senior","minRate":"950000000000000000","maxRate":"1050000000000000000"}}'
```

Result:

```json
{"data":{"result":"1000000000000000000","computedResult":"1000000000000000000","tranche":"senior","decimals":18,"boundsApplied":false,"trancheAssets":"200000000","trancheShares":"200000000"},"statusCode":200,"result":"1000000000000000000","timestamps":{"providerDataRequestedUnixMs":1782823995257,"providerDataReceivedUnixMs":1782823995472,"providerIndicatedTimeUnixMs":1782823994000}}
```

EA params:

```json
{
  "endpoint": "strcusx-exchange-rate",
  "programAddress": "7iNvMc3x5VvwNmYomAAg86CpWeEw7QfDF2z5GgtDzHXe",
  "strategyName": "STRC-USX-1",
  "tranche": "senior",
}
```

mohamed-mehany (approved):

PR #5140, opened 2026-06-28
fix(coinmarketcap): scope crypto batch requests to their own quote currency

tyrion70:
Symptom: coinmarketcap crypto feeds intermittently return 502 "Data for quote "<QUOTE>" was not found in response"; the failing set is non-deterministic, reshuffles on restart, and is independent of API key/plan.

Cause: In crypto.ts one HTTP request is built per unique quote (convert=<quote>), but each is tagged with params: list — the entire param set across all quotes. parseResponse then looks up data.quote[p.quote] for every attached param, so any param whose quote ≠ the request's convert 502s.

Fix: Filter each request's params down to the params sharing its quote, so each param is resolved only by the request that actually requested its convert. One line; no behavioural change for single-quote batches.

### DIFF
--- a/.changeset/coinmarketcap-crypto-quote-filter.md
+++ b/.changeset/coinmarketcap-crypto-quote-filter.md
@@ -1,0 +1,5 @@
+---
+'@chainlink/coinmarketcap-adapter': patch
+---
+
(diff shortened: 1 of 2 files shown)


dskloetc:
The fix was merged today: https://github.com/smartcontractkit/external-adapters-js/pull/5141

PR #5228, opened 2026-07-24
Add stock_quotes endpoint to ticketlayer

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-8220

## Description

This PR adds the `stock_quotes` endpoint to the `ticketlayer` adapter.

## Changes

1. Copy `stock` endpoint files in a separate commit.
2. Make adjustments for `stock_quotes`.

## Steps to Test

Only works during US extended trading hours:
```
  "data": {
    "endpoint": "stock_quotes",
    "symbol": "us:aapl"
  }
}'

{
  "result": null,
  "data": {
    "mid_price": 321.755,
    "bid_price": 321.7,
    "ask_price": 321.81,
    "bid_volume": 360,
    "ask_volume": 80
  },
  "timestamps": {
    "providerDataStreamEstablishedUnixMs": 1784813472673,
    "providerDataReceivedUnixMs": 1784813478519,
    "providerIndicatedTimeUnixMs": 1784813478194
  },
  "statusCode": 200,
  "meta": {
    "adapterName": "TICKERLAYER",
    "transportName": "default_single_transport",
    "metrics": {
      "feedId": "{\"base\":\"us:aapl\"}"
    }
  }
}
```

## Quality Assurance


packages/sources/tickerlayer/src/transport/stock_quotes.ts
```diff
@@ -0,0 +1,103 @@
...
+            {
+              params: { base: message.symbol },
+              response: {
+                result: null,
+                data: {
+                  mid_price,
+                  bid_price,
+                  ask_price,
+                  bid_volume: Number(message.bid_size),
+                  ask_volume: Number(message.ask_size),
```
mohamed-mehany: can we reuse the computed `bid_volume` and `ask_volume`>

packages/sources/tickerlayer/test/unit/stock_quotes.test.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    expect(responseCache.write).not.toHaveBeenCalled()
+    expect(log).toHaveBeenCalledWith({
+      msg: 'Ignoring unexpected message',
+      ignoredMessage: malformedMessage,
+    })
+    log.mockClear()
+  })
+
+  it('should ignore messages with zero ask volume', async () => {
```
mohamed-mehany, line 234: Is that one intentional? i.e if we have 0 ask_volume or bid_volume (but not both) we should drop the message?

packages/sources/tickerlayer/src/transport/stock_quotes.ts
```diff
@@ -0,0 +1,103 @@
...
+        message(message) {
+          if (message.type === 'system') {
+            logger.debug({ msg: 'Ignoring system message', ignoredMessage: message })
+            return
+          }
+          const bid_price = toNumber(message.bid)
+          const ask_price = toNumber(message.ask)
+          const bid_volume = toNumber(message.bid_size)
+          const ask_volume = toNumber(message.ask_size)
+          const providerIndicatedTimeUnixMs = toNumber(message.ts)
```
mohamed-mehany, line 49: nit: redundant conversion since ts is already number according to the `WSResponse` interface?

packages/sources/tickerlayer/src/transport/stock_quotes.ts
```diff
@@ -0,0 +1,103 @@
...
+          if (
+            message.type !== 'quote' ||
+            message.channel !== 'stocks.quotes' ||
```
mohamed-mehany: Can we have constants for those static strings since they're used in multiple places in the code?

packages/sources/tickerlayer/test/unit/stock_quotes.test.ts
```diff
@@ -0,0 +1,314 @@
...
+      base: symbol,
+    })
+    subscriptionSet.getAll.mockReturnValue([params])
+
+    const context = makeStub('context', {
+      adapterSettings,
+      endpointName,
+    } as EndpointContext<WsTransportTypes>)
+
+    await runAllUntilSettled(clock, transport.backgroundExecute(context))
```
mohamed-mehany: nit: This part is repeated many times, might be useful to define sth like: `setupSubscription(symbol?)` and reuse.

packages/sources/tickerlayer/src/transport/stock_quotes.ts
```diff
@@ -0,0 +1,103 @@
...
+            message.asset !== 'stocks' ||
+            !message.symbol ||
+            !bid_price ||
+            !ask_price ||
+            !bid_volume ||
+            !ask_volume ||
+            !providerIndicatedTimeUnixMs
```
mohamed-mehany, line 59: nit: This condition interleaves structural checks and numeric validation checks making it a bit too long, I would split it into two checks with different warnings that are more specific but I can also understand if you prefer the single condition.

mohamed-mehany (changes requested):
A few minor comments otherwise looks good.

packages/sources/tickerlayer/src/transport/stock_quotes.ts
dskloetc: A TypeScript type doesn't enforce anything about messages coming over the wire.

packages/sources/tickerlayer/src/transport/stock_quotes.ts
dskloetc: Done

packages/sources/tickerlayer/src/transport/stock_quotes.ts
dskloetc: lol, I had already done this but I think Claude reverted that change. Done (again).

packages/sources/tickerlayer/test/unit/stock_quotes.test.ts
dskloetc: Done

packages/sources/tickerlayer/test/unit/stock_quotes.test.ts
dskloetc: This is intentional. I consider this a logical consequence of the following from the ticket:

> If bid or ask (or both) are null, don’t calculate a new mid, return nothing so that the cached price is returned by the EA

packages/sources/tickerlayer/src/transport/stock_quotes.ts
dskloetc: If I split the `if` here, then it makes the function longer, which I think will hurt readability.
But if I move the condition of the `if` to a separate function, then the compiler no longer realizes that all the fields are defined, which changes the return type.
So I prefer to keep it like this.

mohamed-mehany (approved):

PR #5229, opened 2026-07-24
Tradinghours: Remove duplication in markets list

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-8204

## Description

`markets` and `marketToFinId` list (and should list) exactly the same markets. They should not both be hard coded.

## Changes

1. Derive `markets` from `marketToFinId`.
2. Move comments.
3. Add tests.

## Steps to Test

```
yarn test packages/sources/tradinghours
```

## Quality Assurance


packages/sources/tradinghours/src/transport/utils.ts
```diff
@@ -37,16 +15,18 @@ const marketToFinId: Record<Market, string> = {
...
+  jpx: 'JP.JPX', // Japan Exchange Group
+  sse: 'CN.SSE', // Shanghai Stock Exchange
+  szse: 'CN.SZSE', // Shenzhen Stock Exchange
   nymex: 'US.CHNLNK.WTI',
   ice_europe_energy: 'US.ICE.ENERGY.GROUP3',
   bme: 'ES.BME',
-}
+} as const
+
+export const markets = Object.keys(marketToFinId)
```
mohamed-mehany: nit: Obligatory AI recommendation :D 
```
1. Oject.keys() returns string[] in TypeScript, not Market[]. 
The original was a precise readonly ['forex', 'metals', ...] tuple. 

This widens the type in consumers:
- market-status.ts:18: options: [...markets, ...] loses its precise union type
- market-status.ts:31: aliases: markets.map(...) returns string[] instead of template-literal types
2. markets is now mutable — Nit
The original was as const (immutable). Object.keys() returns a mutable array. If immutability matters, consider as readonly Market[] or Object.freeze().

Recommendation:
export const markets: Market[] = Object.keys(marketToFinId) as readonly Market[]
```

mohamed-mehany (dismissed):
Just 1 AI nit but non blocking

packages/sources/tradinghours/src/transport/utils.ts
```diff
@@ -59,4 +39,5 @@ export const getFinId = (market: Market, type: MarketType) => {
   return marketToFinId[market]
 }
 
-export const isMarket = (v: string): v is Market => markets.includes(v as Market)
+export const isMarket = (v: string): v is Market =>
+  Object.prototype.hasOwnProperty.call(marketToFinId, v)
```
mohamed-mehany, line 43: nit: Also apparently Object.hasOwn can be used since it's more readable (I didn't know it existed tbh :D), so sth like this:
`export const isMarket = (v: string): v is Market => Object.hasOwn(marketToFinId, v)`

packages/sources/tradinghours/src/transport/utils.ts
dskloetc: Done

packages/sources/tradinghours/src/transport/utils.ts
dskloetc: Nice! Unfortunately it seems to be too new for us:
```
packages/sources/tradinghours/src/transport/utils.ts:45:10 - error TS2550: Property 'hasOwn' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2022' or later.

45   Object.hasOwn(marketToFinId, v)
            ~~~~~~
```

mohamed-mehany: Ah well that sucks :D sorry and thanks for checking!

mohamed-mehany (approved):

PR #5254, opened 2026-07-31
[market-status] Disable missing env var flaky test

denis-chernov-smartcontract:
## Description

......

## Changes

- High level
- changes that
- you made

## Steps to Test

1. Steps
2. to
3. test

## Quality Assurance

### DIFF
--- a/packages/composites/market-status/test/integration/adapter.test.ts
+++ b/packages/composites/market-status/test/integration/adapter.test.ts
@@ -523,7 +523,7 @@ describe('execute', () => {
       })
     })
   })
-// This is working locally but for some reason failing on pipeline
+  // This is working locally but for some reason failing on pipeline
   // describe('missing env var', () => {
   //   it('throws error', async () => {
   //     let response
--- a/packages/composites/market-status/test/integration/adapter.test.ts
+++ b/packages/composites/market-status/test/integration/adapter.test.ts
@@ -523,18 +523,18 @@ describe('execute', () => {
       })
     })
   })
-
-  describe('missing env var', () => {
-    it('throws error', async () => {
-      let response
-      for (let i = 0; i < 10; i++) {
-        response = await testAdapter.request({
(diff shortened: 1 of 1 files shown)


mxiao-cll (dismissed):

mxiao-cll:
### Potential fix later

What failed: market-status integration test execute › missing env var › throws error at packages/composites/market-status/test/integration/adapter.test.ts:535. It expected errorMessage: "Missing required environment variable: SIX_ADAPTER_URL" but got undefined.

Why it's a race: The test deliberately leaves SIX_ADAPTER_URL unset and fires 10 requests in a loop, hoping that by the last one the transport's background executor has run, hit the throw from PROVIDER_ADAPTER_URL.get('SIX') (src/transport/base-market-status.ts:105), and written a 502 with that errorMessage to the cache. If the background executor hasn't landed its cache write within those 10 iterations, the request instead exhausts cache polling and returns the framework's AdapterTimeoutError — which serializes as {status, statusCode, error: {message}} with no top-level errorMessage field. That's exactly the Received: undefined in the log. The 10-iteration loop never advances the fake clock or calls waitForCache(), so it's pure scheduling luck.

Why CI but not locally: the GitHub runner is a slow 2-core box running 31 test suites across parallel Jest workers (plus CI is on Node 22 vs your Node 24 — event-loop scheduling differs across versions). Either is enough to flip the race the wrong way.

Why it surfaced now: the test hasn't changed since July 14 (#5192 restored SIX). It only entered this release PR's test set because the gRPC streams changeset (#5250, merged July 30) bumps market-status — the July 29 run of this same branch didn't include it.

Suggested fix (in a follow-up PR, since this one is bot-managed): make the test deterministic — after the first request, do await testAdapter.waitForCache() (which advances the fake clock until the cache is populated, like the other tests do), then make one more request and assert on errorMessage.

mxiao-cll (approved):

PR #5296, opened 2026-08-13
Support additional asset types in tickerlayer

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-8339

## Description

Currently `tickerlayer` only supports stock prices and stock quotes.
We want to support prices and quotes for other asset types as well. In particular for ETFs.
We can get this over the same websocket connection and the API is almost the same.
We just need to specify a different `channel` in the subscription and take into account that while stock messages have numbers as strings, some other asset types have numbers as numbers.

## Changes

1. Add an `assetType` param to the `stock` and `stock_quotes` endpoints, defaulting to `stocks` for backwards compatilibilty.
2. Add `price` and `quotes` aliases to the endpoints to avoid confusion about using "stock" endpoints with a different asset type.
3. Use the `assetType` to construct the subscription messages.
4. Remove the restriction on `message.channel` and `message.asset` as they now depend on the asset type.
5. Update and add unit tests.

## Steps to Test

When Asian markets are open:
```
  "data": {
    "endpoint": "price",
    "assetType": "etfs",
    "symbol": "JP:1329"
  }
}'
```

## Quality Assurance

### DIFF
--- a/packages/sources/tickerlayer/src/endpoint/stock_quotes.ts
+++ b/packages/sources/tickerlayer/src/endpoint/stock_quotes.ts
@@ -4,11 +4,25 @@ import { InputParameters } from '@chainlink/external-adapter-framework/validatio
 import { config } from '../config'
 import { wsTransport } from '../transport/stock_quotes'
 
-export const inputParameters = new InputParameters(stockEndpointInputParametersDefinition, [
+export const inputParameters = new InputParameters(
   {
-    base: 'US:AAPL',
+    ...stockEndpointInputParametersDefinition,
+    assetType: {
+      description:
+        'The asset type is used to determine what channel to subscribe to. The Tickerlayer API supports "stocks", "fores", "crypto", "indices", "etfs", and "commodities".',
+      type: 'string',
+      default: 'stocks',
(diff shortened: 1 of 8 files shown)


packages/sources/tickerlayer/src/endpoint/stock_quotes.ts
```diff
@@ -4,11 +4,25 @@ import { InputParameters } from '@chainlink/external-adapter-framework/validatio
 import { config } from '../config'
 import { wsTransport } from '../transport/stock_quotes'
 
-export const inputParameters = new InputParameters(stockEndpointInputParametersDefinition, [
+export const inputParameters = new InputParameters(
```
mxiao-cll: This seems to be duplicated across two endpoints  ( as well as validators)

mxiao-cll (dismissed):

packages/sources/tickerlayer/src/endpoint/stock_quotes.ts
dskloetc: Done

mxiao-cll (approved):

PR #5306, opened 2026-08-21
Add tradinghours schedule generator script

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-8204

## Description

Add a script that can be used to generate simplified schedule JSON from the official TradingHours to be used as configuration for the `static-market-hours` adapter.

## Changes

1. Add the script
2. Add a script entry in `package.json`.
3. Add `README.md` describing how to use it.

## Steps to Test

1. Generate JSON (see README).
2. Add the JSON as env var `FOO_REGULAR_SCHEDULE` to `static-market-hours`.
3.
    ```
      "data": {
        "market": "foo",
      }
    }'
    ```

I also verified that the generated schedule matches the official data for the 15 markets from the ticket using a python script that I will add in a follow-up PR. 

## Quality Assurance


mxiao-cll (commented):
Do we want to put this inside the static EA instead?

dskloetc:
> Do we want to put this inside the static EA instead?

I thought about it but I wanted to keep the EA independent from tradinghours (you could have schedules made by hand or form other sources) and I wasn't sure if it would be confusing to have a script in an EA directory. But I'm not against moving it if you think it's better.

mxiao-cll:
> > Do we want to put this inside the static EA instead?
> 
> I thought about it but I wanted to keep the EA independent from tradinghours (you could have schedules made by hand or form other sources) and I wasn't sure if it would be confusing to have a script in an EA directory. But I'm not against moving it if you think it's better.

I think in the EA is better, you can have mutiple scripts in that EA

mxiao-cll (approved):

dskloetc:
> I think in the EA is better, you can have mutiple scripts in that EA

Sounds good. I'll move it in the follow-up PR.

PR #5314, opened 2026-08-26
OPDATA-8530 securitize asset pubkeys

mmcallister-cll:
## Description
Use variable env var framework feature to replace process.env lookup for `*_PUBKEYS` env vars

## Changes
- Replace `*_PUBKEYS` lookup with `ASSET_PUBKEYS` variable env var
- Update places where used & tests

## Steps to Test
1. yarn test securitize/
2. with payload
```json
{
    "data": {
        "endpoint": "nav",
        "assetId": "17eab62a-5347-484f-b26f-35b25c087b07",
        "envVarPrefix": "acred"
    }
}
```

## Quality Assurance

[OPDATA-8530]: https://smartcontract-it.atlassian.net/browse/OPDATA-8530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sources/securitize/src/transport/nav.ts
+++ b/packages/sources/securitize/src/transport/nav.ts
@@ -2,7 +2,6 @@ import {
   HttpTransport,
   HttpTransportConfig,
 } from '@chainlink/external-adapter-framework/transports'
-import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
 import { config } from '../config'
 import { BaseEndpointTypes } from '../endpoint/nav'
 import { validateResponseSignature } from './sigutils'
@@ -38,14 +37,7 @@ export type HttpTransportTypes = BaseEndpointTypes & {
 }
 
 export function getPubKeys(assetEnvVarPrefix: string, settings: typeof config.settings): string[] {
-  const envVarName = `${assetEnvVarPrefix.toUpperCase()}_PUBKEYS`
   const pubkeys = settings.ASSET_PUBKEYS.get(assetEnvVarPrefix)
-  if (!pubkeys) {
-    throw new AdapterInputError({
-      message: `Missing env var ${envVarName}`,
-      statusCode: 400,
-    })
-  }
   return pubkeys.split(',').map((s) => s.trim()) ?? []
 }
 
--- a/packages/sources/securitize/src/transport/nav.ts
+++ b/packages/sources/securitize/src/transport/nav.ts
@@ -3,6 +3,7 @@ import {
   HttpTransportConfig,
 } from '@chainlink/external-adapter-framework/transports'
 import { AdapterInputError } from '@chainlink/external-adapter-framework/validation/error'
+import { config } from '../config'
 import { BaseEndpointTypes } from '../endpoint/nav'
 import { validateResponseSignature } from './sigutils'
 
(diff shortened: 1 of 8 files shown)


packages/sources/securitize/src/transport/nav.ts
```diff
@@ -36,9 +37,9 @@ export type HttpTransportTypes = BaseEndpointTypes & {
   }
 }
 
-export function getPubKeys(assetEnvVarPrefix: string): string[] {
+export function getPubKeys(assetEnvVarPrefix: string, settings: typeof config.settings): string[] {
   const envVarName = `${assetEnvVarPrefix.toUpperCase()}_PUBKEYS`
-  const pubkeys = process.env[envVarName]
+  const pubkeys = settings.ASSET_PUBKEYS.get(assetEnvVarPrefix)
   if (!pubkeys) {
```
dskloetc: Make `ASSET_PUBKEYS` required and remove this check.

packages/sources/securitize/test/test-utils/signature-helper.ts
```diff
@@ -10,15 +10,14 @@ export function generateSignature(
   recordDate: Date,
   previousSignature: string,
   previousHash: string,
+  keyPair: nacl.SignKeyPair = nacl.sign.keyPair(),
```
dskloetc: This is not equivalent. Now the default key pair will be created only once and reused across calls that don't specify a key pair. Is this intended?

packages/sources/securitize/test/unit/nav.test.ts
```diff
@@ -49,28 +50,33 @@ describe('nav transport', () => {
...
-      const actualPubKeys = getPubKeys('single')
+      const actualPubKeys = getPubKeys('single', settings)
       expect(actualPubKeys).toEqual([keyA])
     })
 
-    it('should handle empty pubkey list', async () => {
-      process.env.EMPTY_PUBKEY = ''
-      expect(() => getPubKeys('empty')).toThrow()
+    it('should handle empty pubkey', async () => {
+      expect(() => getPubKeys('empty', settings)).toThrow()
```
dskloetc: You didn't define empty keys on line 55 so this throws because the keys don't exist. Not because they are empty.
So this test is now equivalent to the test on line 69.

But if you follow my suggestion to make the setting required, then the throwing logic will be inside the function that you are mocking here so then you can't test it here. You would just be relying on the existing code (which is tested in the framework) working as intended.

dskloetc (approved):

packages/sources/securitize/src/transport/nav.ts
mmcallister-cll: done, thx

packages/sources/securitize/test/unit/nav.test.ts
mmcallister-cll: removed, thx

packages/sources/securitize/test/test-utils/signature-helper.ts
mmcallister-cll: moved back

PR #5375, opened 2026-09-09
Add 21x EA

dskloetc:
https://smartcontract-it.atlassian.net/browse/OPDATA-7298

## Description

Adds a new adapter to be used with schema V11 Streams.
Makes 2 requests, to `tradeinfo` and `orderbook` to return the required data.

## Changes

1. Add the result of `yarn new` in a separate commit for easy reviewing.
2. Implement the EA.

## Steps to Test

```
  "data": {
    "id": "09befe9e-c95d-4856-ab4c-c811202a9cfb"
  }
}'

{
  "data": {
    "last_price": "0.9990000",
    "mid_price": "0.99999995",
    "bid_price": "0.9999999",
    "bid_volume": "11000.0000000",
    "ask_price": "1.0000000",
    "ask_volume": "136500.0000000",
    "market_status": 2,
    "trading_status_string": "CONTINUOUS_TRADING"
  },
  "statusCode": 200,
  "result": "0.9990000",
  "timestamps": {
    "providerDataRequestedUnixMs": 1788945861168,
    "providerDataReceivedUnixMs": 1788945861244
  },
  "meta": {
    "adapterName": "21X",
    "transportName": "default_single_transport",
    "metrics": {
      "feedId": "{\"base\":\"09befe9e-c95d-4856-ab4c-c811202a9cfb\"}"
    }
  }
}
```

## Quality Assurance


packages/sources/21x/src/transport/price.ts
```diff
@@ -0,0 +1,173 @@
...
+    const requestConfig = {
+      method: 'GET',
+      baseURL: this.config.API_ENDPOINT,
+      url: `${id}/${endpoint}`,
+    }
+
+    const requestKey = requestConfig.url
+    const response = await this.requester.request<T>(requestKey, requestConfig)
+
+    console.log('dskloet', endpoint, response.response.data)
```
mohamed-mehany: Leftover debug log

packages/sources/21x/src/config/index.ts
```diff
@@ -0,0 +1,18 @@
+import { AdapterConfig } from '@chainlink/external-adapter-framework/config'
+
+export const config = new AdapterConfig({
+  API_ENDPOINT: {
+    description: 'An API endpoint for Data Provider',
+    type: 'string',
+    default: 'https://trading.21x.eu/api/v1/tradingpairs/',
```
mohamed-mehany, line 7: Since this seems to be a public endpoint, have we checked for rate limits? Given the background_execute_ms, we seem to do 2 req / 10s per pair. Not high but just double checking

packages/sources/21x/src/transport/price.ts
```diff
@@ -0,0 +1,173 @@
...
+      this.getTradeInfo(params.base),
+      this.getOrderBook(params.base),
+    ])
+
+    const last_price = tradeInfo.lastPrice
+
+    const bid_price = orderBook.buy[0]?.limit
+    const bid_volume = orderBook.buy[0]?.quantity
+    const ask_price = orderBook.sell[0]?.limit
+    const ask_volume = orderBook.sell[0]?.quantity
```
mohamed-mehany, line 116: Can we please add a comment here that we're guaranteed top of book as the first item?

packages/sources/21x/src/transport/price.ts
```diff
@@ -0,0 +1,173 @@
...
+  switch (status) {
+    case 'CREATED':
+    case 'DISABLED':
+    case 'PERMANENTLY_DELETED':
+      return 0
+    case 'CONTINUOUS_TRADING':
+      return 2
+    case 'OUT_OF_TRADING':
+    case 'AUTOMATIC_TRADING_HALT':
+    case 'MANUAL_TRADING_HALT':
```
mohamed-mehany, line 51: Is there a public reference or some updated source that we can document in case those statuses change in the future (for example if they add a new status) and it needs to be added here too?

packages/sources/21x/src/transport/price.ts
```diff
@@ -0,0 +1,173 @@
...
+      return 5
+    default:
+      throw new Error(`Unknown trading status: ${status}`)
```
mohamed-mehany: Should we default to 0 in this case? The [public docs](https://docs.chain.link/data-streams/reference/report-schema-v11) refers to 0 as `Market status cannot be determined`, should we check with product if we should default to that as well? They might prefer to have reports still produced with unkown market status rather than failing the report.

mohamed-mehany (changes requested):

packages/sources/21x/src/config/index.ts
dskloetc: Especially with a public endpoint I'd be surprised if there's a rate limit but I've added a question in the ticket: https://smartcontract-it.atlassian.net/browse/OPDATA-7298?focusedCommentId=720288

Do you think this should be blocking or can we follow up if it turns out there is rate limiting?

packages/sources/21x/src/transport/price.ts
dskloetc: This is the mapping from the [requirements doc](https://docs.google.com/document/d/1nqds8WAyOBYIG0ITkzDtW8b5nqNeZVa1ZRIFuvV1N8Y/edit?tab=t.0#heading=h.1fdj37k47goi):

Image

The possible values are [documented here](https://docs.21x.eu/api-reference-v1.0/public-market-data#get-tradingpairs-id-tradeinfo) and only include these 7.

packages/sources/21x/src/transport/price.ts
dskloetc: Sounds good. Done.

packages/sources/21x/src/transport/price.ts
dskloetc: Done

packages/sources/21x/src/transport/price.ts
dskloetc: Oops. Thanks. Removed.

packages/sources/21x/src/config/index.ts
dskloetc: Answer: The 21X API gateway implements a limit of 3000 requests per IP within 5 minutes enforced by 21X firewall.

mohamed-mehany: Yeah the reason I ask is that I've found public endpoints to always need some form of protection against DDoS and that's usually sth like Cloudflare or sth like that. Usually those can have annoying rate limits.
So if my LLM's math is not wrong, we're good up until around 45-50 assets? For future proofing, might be worth it to add an EA rateLimit so as to self throttle. If it ends up too complicated, I think it's also not blocking, and we can simply document it somewhere that we have this limitation.

packages/sources/21x/src/transport/price.ts
mohamed-mehany: Nice, I think the second list/link would be great to be added as a comment here for quick access in case something changes.

mohamed-mehany (dismissed):
Looks good, 2 minor comments non-blocking

packages/sources/21x/src/config/index.ts
dskloetc: I'm not aware of plans to do more than 1 asset. I've added at rate limit of half the documented limit.

packages/sources/21x/src/transport/price.ts
dskloetc: Done

packages/sources/21x/src/config/index.ts
mohamed-mehany: Thanks!

mohamed-mehany (approved):

PR #5384, opened 2026-09-11
static-market-hours: 24/5 schedules

calvwang9:
## Description

Add support for generating and validating 24/5 schedules to the static-market-hours tradinghours scripts.

script-only change, so doesnt require a release/deploy

......

## Changes

- High level
- changes that
- you made

## Steps to Test

Added unit tests to packages/sources/static-market-hours/test/unit/schedule_generator.test.ts.

Manually verified accuracy of script output, including the check-market-status.py script:
```
=== waiting for all ===
ALL DONE
JP.JPX :: checking 527041 timestamps :: No mismatches found after checking 527041 timestamps
KR.KRX :: checking 527041 timestamps :: No mismatches found after checking 527041 timestamps
US.CHNLNK.FX :: checking 527041 timestamps :: No mismatches found after checking 527041 timestamps
US.CHNLNK.METAL :: checking 527041 timestamps :: No mismatches found after checking 527041 timestamps
US.CHNLNK.NYSE :: checking 525165 timestamps :: No mismatches found after checking 525165 timestamps
US.CHNLNK.WTI :: checking 527041 timestamps :: No mismatches found after checking 527041 timestamps
US.COMEX.METALS.COPPER :: checking 162296 timestamps :: No mismatches found after checking 162296 timestamps
US.NYMEX.ENERGY.CRUDEOIL1 :: checking 162296 timestamps :: No mismatches found after checking 162296 timestamps
US.NYMEX.ENERGY.NATURALGAS1 :: checking 162296 timestamps :: No mismatches found after checking 162296 timestamps
```

## Quality Assurance

### DIFF
--- a/packages/sources/static-market-hours/src/scripts/schedule_generator.ts
+++ b/packages/sources/static-market-hours/src/scripts/schedule_generator.ts
@@ -88,7 +88,17 @@ export type Session = {
   start: string // "HH:mm:ss"
   end: string // "HH:mm:ss"
   endDateOffset: 0 | 1
+  // The market status during the session. When absent, a status of "OPEN" is
+  // assumed (which is the case for regular markets, where only sessions of
+  // open phases are included).
+  status?: string
 }
+
+// The market status type determines which statuses are used in the generated
+// schedule. "regular" schedules use "OPEN"/"CLOSED" statuses. "24/5" schedules
+// use the TwentyfourFiveMarketStatus statuses of the static-market-hours
+// adapter: "REGULAR", "PRE_MARKET", "POST_MARKET", "OVERNIGHT" and "WEEKEND".
+export type MarketStatusType = 'regular' | '24/5'
 
 const FILE_MARKETS = 'markets.csv'
 const FILE_PHASES = 'phases.csv'
@@ -101,6 +111,7 @@ const COLUMN_PHASE_STATUS = 'Status'
 const COLUMN_FIN_ID = 'FinID'
 const COLUMN_SCHEDULE_GROUP = 'Schedule Group'
 const COLUMN_PHASE_TYPE = 'Phase Type'
+const COLUMN_SCHEDULE_PHASE_NAME = 'Phase Name'
 const COLUMN_IN_FORCE_START = 'In Force Start Date'
 const COLUMN_IN_FORCE_END = 'In Force End Date'
 const COLUMN_DAYS_OF_WEEK = 'Days'
@@ -112,6 +123,17 @@ const COLUMN_SCHEDULE = 'Schedule'
 
 const SCHEDULE_GROUP_REGULAR = 'Regular'
 const STATUS_OPEN = 'Open'
+
+const STATUS_OPEN_SCHEDULE = 'OPEN'
+const STATUS_CLOSED_SCHEDULE = 'CLOSED'
+const STATUS_WEEKEND = 'WEEKEND'
+const STATUS_REGULAR_245 = 'REGULAR'
+const STATUS_PRE_MARKET = 'PRE_MARKET'
+const STATUS_POST_MARKET = 'POST_MARKET'
+const STATUS_OVERNIGHT = 'OVERNIGHT'
+
+// The phase name used by TradingHours for overnight sessions of 24/5 markets.
+const PHASE_NAME_OVERNIGHT = 'Overnight'
 
 const CSV_DAY_NAMES = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'] as const
 type CsvDayOfWeek = (typeof CSV_DAY_NAMES)[number]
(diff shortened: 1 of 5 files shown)


.changeset/four-icons-carry.md
dskloetc, line 1: What's this?

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
```diff
@@ -101,6 +111,7 @@ const COLUMN_PHASE_STATUS = 'Status'
 const COLUMN_FIN_ID = 'FinID'
 const COLUMN_SCHEDULE_GROUP = 'Schedule Group'
 const COLUMN_PHASE_TYPE = 'Phase Type'
+const COLUMN_SCHEDULE_PHASE_NAME = 'Phase Name'
```
dskloetc, line 114: Should this be called`COLUMN_PHASE_NAME` instead of `COLUMN_SCHEDULE_PHASE_NAME` (to be consistent with `COLUMN_PHASE_TYPE`)?

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
```diff
@@ -264,20 +295,58 @@ export const getExceptionsFromSessionDifference = (
   return exceptions
 }
 
+// Exceptions may be overlapping or adjacent because sessions of consecutive
+// days can span into each other (e.g. overnight sessions). This sorts the
+// exceptions, clips overlaps, and merges adjacent exceptions with the same
+// status.
+export const mergeExceptions = (exceptions: Schedule['exceptions']): Schedule['exceptions'] => {
```
dskloetc, line 302: It's my understanding that with the way we compute exceptions, this is not necessary.
Can you give an example of where this is needed?

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
```diff
@@ -410,13 +479,13 @@ export class ScheduleGenerator {
     })
   }
 
-  getOpenScheduleRows(scheduleGroup: string): Row[] {
-    this.initOpenRowsBySchedule(scheduleGroup)
-    return this.openRowsBySchedule.get(scheduleGroup)!
+  getSessionRows(scheduleGroup: string): Row[] {
```
dskloetc, line 482: I called them schedule rows because they are rows from `schedules.csv`. I don't see a `sessions.csv` so I'm not sure what is meant by "SessionsRows".

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
```diff
@@ -426,12 +495,41 @@ export class ScheduleGenerator {
...
+    if (this.marketStatusType === 'regular') {
+      return this.getPhaseTypeStatus(row) === STATUS_OPEN ? STATUS_OPEN_SCHEDULE : null
+    }
+
+    // 24/5 markets distinguish multiple session types, which map onto the
+    // TwentyfourFiveMarketStatus statuses of the static-market-hours adapter.
+    // Time not covered by any session gets the WEEKEND status.
+    const phaseName = rowGet(row, COLUMN_SCHEDULE_PHASE_NAME)
+    if (phaseName === PHASE_NAME_OVERNIGHT) {
+      return STATUS_OVERNIGHT
```
dskloetc: I see some rows in schedules.csv where Phase Name is "Overnight" but Phase Type is "Pre-Trading Session". The script would treat this as OVERNIGHT but it's not obvious to me that it shouldn't be STATUS_PRE_MARKET.
If we're currently not dealing with any such markets, it would be good to make some assertion on the value of Phase Type in the case where we return OVERNIGHT solely based on Phase Name.

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
```diff
@@ -426,12 +495,44 @@ export class ScheduleGenerator {
...
```
dskloetc, line 520: I see some rows in schedules.csv where Phase Name is "Overnight" but Phase Type is "Pre-Trading Session". The script would treat this as `OVERNIGHT` but it's not obvious to me that it shouldn't be `STATUS_PRE_MARKET`.
If we're currently not dealing with any such markets, it would be good to make some assertion on the value of Phase Type in the case where we return `OVERNIGHT` solely based on Phase Name.

dskloetc (commented):
Where does the
```
=== waiting for all ===output come from?
```
I don't see that string in the check script. Do you have an additional check script.
I think it would be really useful to include enough in the PR description to reproduce your validation.

.changeset/four-icons-carry.md
calvwang9: empty changeset, as required by CI. no actual release or deploy is required for this change so I figured we can skip the changeset

dskloetc: I'm a bit worried that such a confusing changeset might break the release process. Can you just add a patch version bump on the EA? You can just not do a release until there are more changes for the EA.

calvwang9: this is the output of `yarn changeset add --empty` so it should be recognized, but i added a patch changeset anyway 👍

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
calvwang9: changed to COLUMN_PHASE_NAME for consistency 👍

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
calvwang9: Have updated the code comment for more clarity, but i have seen back to back exceptions being generated across 24/5 schedules that are being compressed via this function - stubbing it out and running the current tests returns 8 distinct exceptions rather than 2

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
calvwang9: updated naming to keep 'schedule' for clarity. 'session' replaces 'open'

packages/sources/static-market-hours/src/scripts/schedule_generator.ts
calvwang9: good catch, it seems that phase type should take precendence over phase name - have updated the code accordingly

dskloetc (approved):

PR #5403, opened 2026-09-15
Drop coinpaprika markprice WS messages with quote_fallback set

cl-efornaciari:
Coinpaprika sets `quote_fallback` on top-of-book messages when quote currency normalization fails, meaning the payload is quoted in the default currency instead of the requested one. These messages are now dropped (with a warning log) so the EA serves a 504 rather than a mis-quoted price.

Includes an integration test verifying a `quote_fallback` message produces no cache entry and requests return 504.


packages/sources/coinpaprika/src/transport/markprice.ts
```diff
@@ -63,6 +67,15 @@ export const wsTransport = new WebSocketTransport<WsTransportTypes>({
         return
       }
 
+      if (message.data.quote_fallback) {
+        logger.warn(
```
alecgard, line 72: Can we also expose a metric? Then we can see against our internal EA's how often this occurs

packages/sources/coinpaprika/src/transport/markprice.ts
```diff
@@ -63,6 +67,15 @@ export const wsTransport = new WebSocketTransport<WsTransportTypes>({
         return
       }
```
alecgard, line 70: Question: Should we consider filtering on `quote` as well?

cl-efornaciari: quote is always `USD` from what I can tell. Detecting a fallback to a non-USD quoted prices should be sufficient in the above check

packages/sources/coinpaprika/src/transport/markprice.ts
cl-efornaciari: In chatting with @alejoberardino , EA specific metrics seem like an anti-pattern. We should be able to glean the info we need from this warn and/or create an alert off of this. WDYT?

alecgard (dismissed):

packages/sources/coinpaprika/src/transport/markprice.ts
alecgard: I do wonder if there is a 'category' metric we could introduce so that we could still capture this without being EA specific - maybe something like 'response doesn't match request/expected'

But no need to block this PR on that - warn log is enough for now

cl-efornaciari: no generic metric exists. going to omit, we can detect through logql if we need to

alejoberardino (dismissed):

alejoberardino (dismissed):

alejoberardino (dismissed):

alejoberardino (dismissed):

alejoberardino (approved):

PR #5404, opened 2026-09-16
DXFeed subscription issue poc

alejoberardino:


cl-efornaciari:
## PR Review

### Problem

The framework's `StreamingTransport.backgroundHandler` tracks subscriptions by JSON-stringifying the **entire** validated params object. For `stock_quotes`, params like `{base:'TSLA', requireVolume:false}` and `{base:'TSLA', requireVolume:true}` are distinct subscriptions, but `formatTicker` only looks at `base`, so both collapse to the same DxFeed ticker (`Quote: ['TSLA']`). When one combo expires, the old `unsubscribeMessage` builder would blindly unsubscribe the shared ticker, killing the feed for the still-active combo.

### Solution Assessment

The fix replaces the per-subscription `subscribeMessage`/`unsubscribeMessage` builders with a single `customSubscriptionMessages` builder that receives the full `SubscriptionDeltas` (desired, new, stale). It dedupes by ticker key and only unsubscribes tickers not present in the `desired` set. **This is the correct approach** and aligns with the framework's `customSubscriptionMessages` API design.

### Verified Against Framework

Reviewed the framework source (`ea-framework-js`). Key confirmations:

- **`customSubscriptionMessages` returns `unknown[]`** where each element is sent as a separate WS frame via `sendMessages` → `wsConnection.send(JSON.stringify(element))`. The PR returns `[removeBatchArray, addBatchArray]`, so DxFeed receives 2 frames max (one for all removes, one for all adds). This is consistent with DxFeed's Bayeux-style protocol which expects JSON arrays of message objects.
- **Removes sent before adds** — matches the default builder's ordering (`[...unsubscribeMessages, ...subscribeMessages]`).
- **Empty deltas** — `messages` is `[]`, framework checks `messages?.length > 0` and skips. Correct.
- **`SubscriptionDeltas` import** — correctly sourced from `@chainlink/external-adapter-framework/transports/abstract/streaming`.

### Impact on `price-ws.ts` (second consumer)

`buildWsTransport` is shared by `price-ws.ts`, whose `formatTicker` can return **two** tickers for after-hours symbols (`[{Trade: [ticker]}, {TradeETH: [ticker]}]`). The dedup logic handles this correctly — each ticker has a distinct JSON key, so both are added/removed independently. The param-collision bug doesn't affect `price` (its only param is `base`), so the change is a no-op semantically for that endpoint, just a batching change (fewer WS frames).

### Issues & Suggestions

**1. Test coverage is thin (Medium)**

Only one scenario is tested — "shared ticker not unsubscribed when still desired." Missing cases:
- Truly stale ticker (not in `desired`) **is** unsubscribed — verifies the positive path
- New subscriptions are correctly added (and deduped when multiple `new` entries map to the same ticker)
- Mixed add+remove in the same delta cycle
- Empty deltas produce no messages
- `price-ws` path with multi-ticker `formatTicker` (`Trade` + `TradeETH`)

**2. No integration/transport-level test (Low)**

The test calls `subscriptionMessageBuilder` directly, which validates the builder logic but not the actual WS frame serialization. The message structure (`Array<Array<message>>` → each inner array serialized as a JSON array frame) differs from the old per-subscription framing. A test verifying `sendMessages` receives the expected serialized strings would add confidence.

**3. Duplicate `add` for already-active tickers (Low / not a regression)**

When a `new` subscription maps to a ticker that is already subscribed (via a different still-desired param combo), `tickersToAdd` will include it and send a duplicate `add`. This is not a regression (old code did the same), and DxFeed likely handles it idempotently. Could be guarded by checking `desiredTickerKeys` minus `new` entries, but it is not critical.

**4. `messages: unknown[]` typing (Nit)**

Could be typed more specifically (e.g., as the message object shape) for readability, but matches the framework's loose `unknown[]` API.

### Verdict

**Approve with suggestions.** The core fix is correct, well-reasoned, and properly uses the framework's `customSubscriptionMessages` API. The changeset is appropriate. Recommend adding a few more test cases before merge to cover the positive unsubscribe path and the add/dedup path.

cl-efornaciari (dismissed):

cl-efornaciari (approved):

